Issue #233, opened 2014-09-19
App Engine dev tries and fails to reach Datastore


stephenplusplus:
## Updated 12/12/14

I'm running https://github.com/stephenplusplus/sillycloud/tree/gae with an app.yaml and Dockerfile copied from the [appengine quickstart project](https://github.com/googlecloudplatform/appengine-nodejs-quickstart). I only made small modifications to match my structure.

When I run `gcloud preview app run .` I see that my routes are correctly configured. I have a handler for GET requests to /api/users: https://github.com/stephenplusplus/sillycloud/blob/master/actions.js#L10. The first api request, `runQuery`, is where I start experiencing errors.

**NOTE: I'm using my branch from #328 to test against**

See https://github.com/GoogleCloudPlatform/gcloud-node/issues/233#issuecomment-57473169 for the error I'm experiencing.

jgeewax:
Are we talking about the App Engine dev server? or the GCD dev server (https://cloud.google.com/datastore/docs/tools/devserver) ?

ryanseys:
https://cloud.google.com/datastore/docs/tools/devserver

stephenplusplus:
I think this will be resolved by #513. "Managed VMs default service account doesn't include scopes to access datastore, pubsub, etc."

Issue #104, opened 2014-08-08
Support App Engine local server


rakyll:
If user is running this client library from GAE context locally, we should turn authorization off and use local server for Datastore operations.

jgeewax:
Possibly related to https://github.com/GoogleCloudPlatform/gcloud-meta/issues/4

mstade:
Out of curiosity, how would this work? Would you have to run `gcloud preview app run .`, meaning you can't easily make use of tooling such as [nodemon]? Is there a way to run a local dev datastore instance by itself, without having to buy into the (a little bit too prescriptive) `gcloud preview app run .` run thing?

[nodemon]: https://github.com/remy/nodemon

ryanseys:
"Should be able to talk to a Datastore local development server."

stephenplusplus:
Custom API endpoints for Datastore can now be configured with `options.apiEndpoint`: https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.13.2/datastore/dataset

It's not done automatically, however. Should we follow this plan: https://github.com/GoogleCloudPlatform/gcloud-node/issues/104#issuecomment-53913532 and default to a custom api endpoint of "localhost:8080"? That host and port can be changed per machine, though those are the defaults for a local server.

stephenplusplus:
Ping @jgeewax - I'm stuck.

From `gcloud preview app run .`, how are requests to Google API endpoints authenticated?

\- Possible answer: They're not. You use a different endpoint on a local server that handles that for you.
\- Response: What is that endpoint? Here are the environment variables I have access to from a GAE dev-spun Node.js instance: https://gist.github.com/stephenplusplus/9851606fde29685af8ef

\- Possible answer: You ping a metadata server.
\- Response: Where do I find that? `http://metadata/` doesn't exist.

jgeewax:
With the App Engine SDK (`gcloud preview app run .`) version of Datastore, I don't know if it "speaks" the Cloud Datastore API. When I said "local development server" I meant https://cloud.google.com/datastore/docs/tools/

That server, once running, shouldn't require any special auth -- all we should need is to update some magical environment variables.. So to summarize:

1. Turn on GCD local server (gcd.sh thing...)
1. Set env variables saying "talk to the datastore service at this host and this port"
1. Run my script that uses `require('gcloud')` and it should hit the local dev server

stephenplusplus:
[...]

PR #328, opened 2014-12-12
integrate support for GAE dev

stephenplusplus:
## Do not merge

RE: #104
Blocked by: #233
Sucessor to: #154 

This isn't immediately useful, but is a modernized adaptation of #154.

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -344,7 +344,21 @@ function makeWritableStream(dup, options, onComplete) {
 module.exports.makeWritableStream = makeWritableStream;
 
 function makeAuthorizedRequest(config) {
-  var GAE_OR_GCE = !config || (!config.credentials && !config.keyFile);
+  var ENV = {
+    GAE_DEV: process.env.GAE_LONG_APP_ID && !process.env.GAE_VM,
+    NO_CREDENTIALS: !config || (!config.credentials && !config.keyFile)
+  };
+
+  var authorize;
+
+  if (ENV.GAE_DEV && ENV.NO_CREDENTIALS) {
+    // Google App Engine Development mode doesn't require authorization.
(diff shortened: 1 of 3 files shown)


stephenplusplus:
To do: use the environment's project ID - https://github.com/GoogleCloudPlatform/gcloud-node/issues/365#issuecomment-72556291

jgeewax:
(Added label based on the heading...)

jgeewax:
Just out of curiosity -- If I'm in App Engine, and want to talk to *another* project's datastore, and I *do* have the keyfile, can I still do that?

To put it another way, in App Engine, if I provide nothing to the code and just "hope that it figures it out", I'd want it to write to my App Engine app's datastore. I still want the ability to provide a specific set of credentials and a datastore/dataset ID (right now, this is equivalent to a project ID) and have my code talk outside my current app's dataset. Is that doable?

ryanseys:
Just quickly looking at the PR here, it doesn't look like it gives us that option. If it detects that we are on GAE, it will automatically use those credentials to authorize the calls. It should definitely provide the option, i.e. if we provide configuration options, it will use those ALWAYS before defaulting to attempting GAE detection.

jgeewax:
:+1: to @ryanseys : I think we really need to make sure that code does the same thing across platforms.

If I manually configure my connection with a specific key, it should do the same thing on GCE as it does on GAE. It sounds like that isn't the case here.

@stephenplusplus : Can you take a look?

stephenplusplus:
If a user provides configuration details, regardless of environment, that should definitely be honored (good catch!). It was overlooked in my original implementation, but I updated to support custom connection configuration never being precedented by magic.

stephenplusplus:
#458 will squash this.

stephenplusplus:
@coveralls you're so weird.

PR #342, opened 2015-01-06
Add prefixes to results from bucket.getFiles()

ghost:


mchmielarski:
Fixed

regression/storage.js
```diff
@@ -42,12 +42,12 @@ var files = {
 };
 
 function deleteFiles(bucket, callback) {
-  bucket.getFiles(function(err, files) {
+  bucket.getFiles(function (err, files) {
```
ryanseys: Are these single-space changes a result of your editor? They are unnecessary as far as our coding styles are concerned and should be reverted for consistency.

regression/storage.js
```diff
@@ -61,200 +61,200 @@ function setHash(obj, file, done) {
...
         bucket.acl.add({
           scope: USER_ACCOUNT,
           role: storage.acl.OWNER_ROLE
-        }, function(err, accessControl) {
+        }, function (err, accessControl) {
           assert.ifError(err);
           assert.equal(accessControl.role, storage.acl.OWNER_ROLE);
 
-          bucket.acl.get({ scope: USER_ACCOUNT }, function(err, accessControl) {
+          bucket.acl.get({scope: USER_ACCOUNT}, function (err, accessControl) {
```
ryanseys: Similarly, the spaces between the `{ }` helps for readability. Please revert these spacing changes.

lib/storage/bucket.js
```diff
@@ -252,16 +274,18 @@ Bucket.prototype.getFiles = function(query, callback) {
...
       file.metadata = item;
       return file;
     });
+    var prefixes = resp.prefixes || null;
     var nextQuery = null;
     if (resp.nextPageToken) {
       nextQuery = extend({}, query, { pageToken: resp.nextPageToken });
     }
-    callback(null, files, nextQuery);
+    callback(null, files, nextQuery, prefixes);
```
ryanseys: My thoughts are we should provide the prefixes before `nextQuery` i.e. `nextQuery` should always be the last object however this will result in a breaking change so I understand why it was put here for now. Thinking ahead, are there any other possibilities of different objects that can be returned such that it may be better to return in an object rather than polluting the callback parameters will a large chain of different values? Looking at [the docs](https://cloud.google.com/storage/docs/json_api/v1/objects/list) it appears not but typically I try to stick to a maximum of 3 parameters per function (callback or otherwise)... after that, it's hard to remember the order of parameters.

Pinging @stephenplusplus for additional thoughts here.

regression/storage.js
```diff
@@ -288,7 +288,7 @@ describe('storage', function() {
       var metadata = {
         website: {
           mainPageSuffix: 'http://fakeuri',
-          notFoundPage: 'http://fakeuri/404.html'
+          notFoundPage:   'http://fakeuri/404.html'
```
ryanseys: These alignments are unnecessary too.

lib/storage/bucket.js
mchmielarski: For me the best solution is to return object with files, prefixes instead files but I'm afraid that it's too big change, so I proposed it in this way.

silvolu:
@mchmielarski However it goes, could you squash the 'Fix *' commits at some point?

stephenplusplus:
Sorry for my novice question, but could you provide a practical use of prefixes, delimiters, and handling the result returned?

CONTRIBUTING.md
```diff
@@ -5,6 +5,14 @@
 3. Add your (or your organization's) name to the AUTHORS and CONTRIBUTORS files.
 4. Send a pull request.
 
+## Code quality
+
+To check code use:
+
+``` sh
+$ npm run lint
+```
```
stephenplusplus: :+1: thanks for adding this!

ryanseys:
Re: use case for prefixes is as I understand it a simple filter for files e.g. file search

stephenplusplus:
I'm interested in terms of how one would use this library after merging this PR. I figured instead of guessing by the code, it would be best to let the suggester (PR-sender) help me understand :)

mchmielarski:
with prefixes and delimiters you can use buckets as tree structure in easy way. I think that comment added to `bucket.getList()` explains it.

stephenplusplus:
Would you mind demonstrating with a practical use?

sindresorhus:
Would be nice with some practical use-cases in the docs. As a normal users I would have no idea what I should be using it for.

ryanseys:
Not considering the added complexity of `nextQuery` traversal, a simple practical example for this would be the following:

### Get folders in bucket
``` js
function getFolders(bucket, callback) {
  var query = { delimiter: '/' };
  bucket.getFiles(query, function(err, files, nextQuery, prefixes) {
    callback(err, prefixes);
  });
}
```

stephenplusplus:
Thank you for that. I know the functionality is in the API, so it makes sense to expose it. But before I can figure out what a good API looks like for it, I need to see how it's meant to be used. Are there other use case anyone can think of? The complexity of prefixes + delimiters still seems quite expensive for such simple results -- assuming the results are "folder names".

ryanseys:
I'm only using it with delimiter = `/` right now but you can combine that with the prefix to list folders in folders. Again, a very simple use case.

``` js
function getFoldersInFolder(bucket, folder, callback) {
  var query = { delimiter: '/', prefix: folder };
  bucket.getFiles(query, function(err, files, nextQuery, folders) {
    callback(err, folders);
  });
}

getFoldersInFolder(storage.bucket('myBucket'), 'my-photos/', cb);
```

ryanseys:
Using delimiter as something other than `/` does not seem to have a useful use case unless you're mimicking a directory structure with flat files where the delimiter is a well-known value in the file names e.g. `var__usr__bin__file1.png` and `var__usr__bin__file2.png` where the delimiter would be `__` in this case. The prefixes value returned in the response is definitely useful when used in combination with the delimiter value.

As for real-world applications and why-the-heck-would-anyone-do-that-anyway questions, it's beyond me.

stephenplusplus:
As a library that isn't about being a straight API bridge, maybe we can abstract what we're interested in with delimiters and prefixes, and expose them in a more usable way. So far, it sounds like getting files from a specified folder is the benefit. We should be able to make this API work to accomplish that:

```js
var query = {
  folder: 'images'
};

myBucket.getFiles(query, function (err, files, nextQuery) {
  // files = only files in a directory "images"
  // delimiter defaults to "/" when `directory` option is given
});
```

And if we think this is important, we can go so far as to create a class for folders:

```js
myBucket.getFolders(function(err, folders, nextQuery) {})
myFolder.file("image1.jpg")
```

ryanseys:
Thought experiment: Is it possible to upload a file with a filename that contains the character `"/"` but isn't a folder? If this is true, then this approach may not work for all cases. The only response value that isn't currently exposed is the prefixes value. Another thought would be to expose the raw response object for "advanced" users that need to do special-cases such as this.

stephenplusplus:
I'm confused on this as well, and if I remember correctly, even the Dev Console UI was confused about it. However, as far as I know, `/` is a reserved character that is invalid in filenames, so a file by the name of "photo/1.jpg" wouldn't be valid, instead, we should all assume that means folder "photo", file "1.jpg" ("all" being: our library, the API, the console UI, the user)

sindresorhus:
> However, as far as I know, / is a reserved character that is invalid in filenames

Not on OS X. Creating a folder named `foo/bar` works fine.

mchmielarski:
Check it: https://github.com/evenemento/gcloud-node/tree/v2
and examples/example-1.js

ryanseys:
@sindresorhus You're right, you can name the folder with a `/` in OS X strangely enough, but it's actually stored as a `:` instead. You can see this by a simple `ls` or by attempting to `cd` into the directory. Also, when you upload this folder to google cloud storage, it's stored as a `:` as well. 

screen shot 2015-01-08 at 2 30 30 pm

At least, this is the case that occurs when using the "Upload Folder" feature in the Google Cloud Developer Console. Might not have anything to worry about in this case then.

sindresorhus:
@ryanseys Yeah, weird, you're right.

stephenplusplus:
I think supporting a Folder object / folder-specific methods is unnecessary. If we can trust `/` as a directory separator, it should be straightforward for the user to get a file in a folder with: `bucket.file('file/in/folder.jpg')`. I do think exposing a method to get only files from a folder is important, though: `bucket.getFiles({ folder: 'folder/folder' }, function (err, files) {})` where we can internally use the API's prefix/delimiter options.

ryanseys:
Yes! 100% agree.

mchmielarski:
it's ok, but one question. how can I get list of folders?

stephenplusplus:
Anyone know if that is possible?

I think if we can support `bucket.getFiles({ folder: 'folder/folder' }, function (err, files) {})`, that's the main thing we want to support.

ryanseys:
Yes the folder option is good in your above snippet. But how are we going to return prefixes value to the developers if they want it? (e.g. for listing folders)

stephenplusplus:
We haven't made much progress with this one. Is there anything left to discuss here, or did we slowly decide against this?

I think `bucket.getFiles({ folder: 'folder/folder' }, function (err, files) {})` would be handy, but not necessary. 

Is it crucial to provide devs access to specifying prefixes and delimiters?

mchmielarski:
without prefixes and delimiters devs can't get folders list.

javorosas:
Any update on how to get folders along with files given a prefix?

ryanseys:
We now return the apiResponse body as the last parameter of nearly every call, so you can now use this object to get the prefixes. We realize this isn't the most ideal solution however it's effectively doing the same thing as this PR, however giving you more flexibility in the future for these corner cases. We haven't yet pushed a new version to npm however you can try it out from master right now.

Let us know how it goes.

PR #353, opened 2015-01-17
Storage: File: add bucket check

coderaiser:
Added `bucket` check same as `name` has.

### DIFF
--- a/lib/storage/bucket.js
+++ b/lib/storage/bucket.js
@@ -198,6 +198,10 @@ Bucket.prototype.delete = function(callback) {
  * var file = bucket.file('my-existing-file.png');
  */
 Bucket.prototype.file = function(name) {
+  if (!name) {
+    throw Error('A file name must be specified.');
+  }
+  
   return new File(this, name);
 };
 
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -64,10 +64,6 @@ var STORAGE_UPLOAD_BASE_URL = 'https://www.googleapis.com/upload/storage/v1/b';
  * @constructor
  */
 function File(bucket, name, metadata) {
-  if (!name) {
-    throw Error('A file name must be specified.');
-  }
-
   this.bucket = bucket;
   this.makeReq_ = bucket.makeReq_.bind(bucket);
   this.metadata = metadata || {};
--- a/test/storage/bucket.js
+++ b/test/storage/bucket.js
@@ -118,15 +118,19 @@ describe('Bucket', function() {
     var file;
     var metadata = { a: 'b' };
 
-    beforeEach(function() {
-      file = bucket.file(FILE_NAME, metadata);
+    it('should throw if no name was provided', function() {
+      assert.throws(function() {
+        file = bucket.file();
+      }, /A file name must be specified./);
     });
 
     it('should return a File object', function() {
+      file = bucket.file(FILE_NAME, metadata);
       assert(file instanceof FakeFile);
     });
 
     it('should pass filename to File object', function() {
+      file = bucket.file(FILE_NAME, metadata);
       assert.equal(file.name, FILE_NAME);
     });
   });
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -125,12 +125,6 @@ describe('File', function() {
   });
 
   describe('initialization', function() {
-    it('should throw if no name is provided', function() {
-      assert.throws(function() {
-        new File(bucket);
-      }, /A file name must be specified/);
-    });
-
     it('should assign file name', function() {
       assert.equal(file.name, FILE_NAME);
     });


stephenplusplus:
That error is actually only there because of this: https://github.com/coderaiser/gcloud-node/blob/patch-1/lib/storage/bucket.js#L201

A File instance is only created by our library. A user does not directly instantiate a File object without going through our methods that internally create them, always providing the first argument (bucket) for the user. Did you run into a problem where an error should have been thrown?

coderaiser:
No, but I think that all necessary arguments should be checked.
Anyway maybe arguments check should be in `Bucket.prototype.file`,
in function users interacting with.

stephenplusplus:
:+1: to checking and throwing in Bucket#file. I don't believe file objects can be created by name in any other part of the library.

If you'd like to change this PR to just move the error catching from the File constructor to the Bucket#file method, that'd be great. And even better if you can update/add tests to suit the new structure.

ryanseys:
I'm okay with Bucket#file too but I don't think putting it in File is wrong because when creating a File object, you DEFINITELY needs to go through File constructor but should not necessarily need to go through Bucket (even though our library only does that). Depends how much defensive programming you want to do. I'll be content with either but to be honest, think closer to File the better.

stephenplusplus:
I'm fine with that as well. Let's let @coderaiser decide :)

ryanseys:
In either case, whatever you decide @coderaiser, a test would be a great addition to this PR! :smile:

coderaiser:
`name` check moved to `Bucket.prototype.file` from `File`.
Tests updated.

test/storage/bucket.js
```diff
@@ -118,15 +118,19 @@ describe('Bucket', function() {
     var file;
     var metadata = { a: 'b' };
 
-    beforeEach(function() {
-      file = bucket.file(FILE_NAME, metadata);
+    it('should throw if no name was provided', function() {
+      assert.throws(function() {
+        file = bucket.file();
+      }, /A file name must be specified./);
```
stephenplusplus: You don't need to assign this to a variable. Leave the beforeEach.

test/storage/bucket.js
```diff
@@ -118,15 +118,19 @@ describe('Bucket', function() {
     var file;
     var metadata = { a: 'b' };
 
-    beforeEach(function() {
-      file = bucket.file(FILE_NAME, metadata);
+    it('should throw if no name was provided', function() {
+      assert.throws(function() {
+        file = bucket.file();
```
ryanseys: This test would fail under the current state of the project. Not sure what is meant to be checked here.

lib/storage/file.js
```diff
@@ -64,10 +64,6 @@ var STORAGE_UPLOAD_BASE_URL = 'https://www.googleapis.com/upload/storage/v1/b';
  * @constructor
  */
 function File(bucket, name, metadata) {
-  if (!name) {
```
ryanseys: Don't remove this check. I thought this PR was to be checking the bucket value?

lib/storage/bucket.js
```diff
@@ -198,6 +198,10 @@ Bucket.prototype.delete = function(callback) {
  * var file = bucket.file('my-existing-file.png');
  */
 Bucket.prototype.file = function(name) {
+  if (!name) {
```
ryanseys: If we are executing Bucket#file then we for sure have a Bucket object. And when you create a Bucket object, you must specify a name so I think this is not necessary if we are moving the check into Bucket#file.

ryanseys:
So I think that either the bucket check gets added to the File constructor like originally was done or the check doesn't get added at all because by checking for Bucket object in a Bucket method is silly.

lib/storage/bucket.js
stephenplusplus: He's just moving the error handling to the method that can actually cause it. A File is only directly instantiated by a user through this method, so it makes sense to me to throw the error here. I'm fine with it staying where it is as well. I thought that's what we were letting him decide. Checking for a bucket isn't necessary, since there will always be a bucket in order to create a file.

ryanseys: The original diff as I remember it just checked if `bucket` param in File constructor was not false-y because @coderaiser was using the File constructor directly and not passing in a bucket parameter. Therefore they added the check for the bucket parameter. The current change now just moves the check of the filename (not the bucket like this PR title seems to suggest) from the File constructor into Bucket#file. This is actually worse I think because now File doesn't check its own params and Bucket#file is now responsible. The current change actually does nothing to help their original problem of not checking the bucket parameter.

stephenplusplus: Checking for the presence of a Bucket is out since my first reply in the issue. It's not needed, since the only code that would be caught by that error is our own code (not a user's), and we can better prevent that error from happening by our test suite, rather than a user seeing an error that makes no sense to them.

I said above it makes sense to have the error handling done in Bucket#file since, if we're being modular, in order for that method to successfully execute, it needs a name, because we, as the authors of this library, know that our File object needs a name. We can throw in either place, it really doesn't matter, and since we each thought differently, I thought @coderaiser could save us the trouble and decide for us. If you feel more strongly that everything is fine the way it is now, feel free to close this out.

(Either way, thanks @coderaiser for helping us out and your patience!)

ryanseys: Sorry for the confusion, just returned from 3-day conference and catching up. I wanted to make sure that @coderaiser knows that this PR doesn't really do anything for his original issue and just now catches the original error one function call sooner. It's helpful to have that test in place because I think it was missed. I'm fine with merging this as long as its understood that File is NEVER to be used independently of Bucket.

As an aside, maybe naming the File filename something like `bucket.file.js` to show the hard hierarchical relationship between `bucket.js` and `file.js`.

stephenplusplus:
We'll just leave things the way they are. Thanks @coderaiser :)

Issue #350, opened 2015-01-15
Add an example of "how to make a file public" in the docs


jgeewax:
I've heard from a lot of people that this is the most common request, so it'd be really cool if we had an example in our docs.

ie:

---

To make an object publicly readable, you might do:

```nodejs
var storage = gcloud.storage();
var albums = storage.bucket('albums');

albums.acl.add({
  scope: 'allUsers',
  permission: Storage.acl.READER_ROLE
}, function(err, aclObject) {});
```

ryanseys:
Yes :+1: absolutely. This is implemented as a simple check-box in the google developer console. What about even a dedicated function for this such as:

``` js
albums.acl.makePublic(function() {}); // set to allUsers reader
albums.acl.makePrivate(function() {}); // set to no permissions (default)
```

jgeewax:
Somewhat dependent on what we expose per #351

stephenplusplus:
Integrating `makePublic` and `makePrivate` over in #351. Merged #357 to address the lacking docs issue.

PR #357, opened 2015-01-20
add examples of making an object publicly-readable

stephenplusplus:
RE: #350

### DIFF
--- a/lib/storage/index.js
+++ b/lib/storage/index.js
@@ -132,9 +132,17 @@ function Storage(config) {
  * var albums = storage.bucket('albums');
  *
  * //-
- * // Make a bucket's contents publicly readable.
+ * // Make all of the files currently in a bucket publicly readable.
  * //-
  * albums.acl.add({
+ *   scope: 'allUsers',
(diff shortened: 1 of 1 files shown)


lib/storage/index.js
```diff
@@ -131,6 +131,18 @@ function Storage(config) {
  * var storage = gcloud.storage();
  * var albums = storage.bucket('albums');
  *
+ * //-
+ * // Make a bucket's contents publicly readable.
```
ryanseys: Can you elaborate here to say that new items added to the bucket will be made public by default?

stephenplusplus: Good catch. I think this snippet would only apply to the objects already in the bucket. Maybe I should use `albums.acl.default.add` instead?

ryanseys: Oops, good catch on my good catch! I didn't actually realize they were different. You should definitely clarify whether it makes new objects public or not which I think is the standard case here. The use case that I'm imagining is a website asset storage bucket.

stephenplusplus:
@ryanseys added a new example. PTAL!

PR #358, opened 2015-01-20
docs: storage/acl - elaborate on `generation`

### LINKED ISSUE #349: Clarify what "generation" does in the File.acl method
The docs mention that I *can* specify a generation property when adding an ACL, but doesn't give much background on how that works.

Could we add a real life example of when generation is used ? And a link to "learn more about object versioning" ?

References: 
- https://github.com/GoogleCloudPlatform/gcloud-node/blob/6904e1181c585e3e5042d124df6bc295da3e5c42/lib/storage/acl.js#L94
- https://cloud.google.com/storage/docs/object-versioning

stephenplusplus:
Fixes #349 (maybe? :))

### DIFF
--- a/lib/storage/acl.js
+++ b/lib/storage/acl.js
@@ -92,6 +92,8 @@ function Acl(options) {
  *
  * //-
  * // For file ACL operations, you can also specify a `generation` property.
+ * // Here is how you would grant ownership permissions to a user on a specific
+ * // revision of a file.
  * //-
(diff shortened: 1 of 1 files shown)


ryanseys:
Calling on @jgeewax for this. Looks good to me, but not sure if he wanted an example or something different than this so I'm holding out on a merge for more information.

jgeewax:
:+1:

PR #359, opened 2015-01-20
fix storage acl roles documentation

ryanseys:
Using `storage` (the instance) instead of `Storage` the class in many places in storage documentation. Also sometimes a storage instance is instantiated in-line and that's a pattern I don't want to emerge, so I've split up:

``` js
gcloud.storage().bucket('my-bucket');
```

into 

``` js
var storage = gcloud.storage(); 
storage.bucket('mybucket')
```

Also, sometimes `gcloud.storage.bucket('my-bucket')` was used, which is just wrong so I fixed that too.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -96,7 +96,8 @@ var util = require('./common/util.js');
  * //     projectId: 'my-project'
  * //   });
  *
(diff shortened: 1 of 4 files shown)


stephenplusplus:
Cool. Thanks!

PR #360, opened 2015-01-21
storage: acl: refactor

### LINKED ISSUE #351: Can we look into making ACL operations really nice for users?
** revised proposal: https://github.com/GoogleCloudPlatform/gcloud-node/issues/351#issuecomment-70695351 **

Hi guys,

In gcloud-node, the ACL methods are very closely tied to the API specification, for example:

```javascript
myBucket.acl.add({
  scope: 'user-useremail@example.com',
  permission: Storage.acl.OWNER_ROLE
}, function(err, aclObject) {});
```

There are definitely times where I want to do low level operations, but other times I want to first-class citizen methods for the different options... Additionally, I now have to know a few things about this API: 

1. the name of the `scope` parameter
1. the name of the `permission` parameter
1. the fact that user's need to be prefixed with `user-`
1. the constant for the name of the role (`OWNER_ROLE`... is it `READER_ROLE` or `READ_ROLE` ?) and where to look for those constants when I inevitably forget them...

Can take a stab at making this more friendly? For example, it might be cool to have:

```javascript
myBucket.acl.owners.addUser('email@example.com', function(err, aclObject) {});
myBucket.acl.readers.removeDomain('example.com', function(err, aclObject) {});
myBucket.acl.writers.addAllUsers(function(err, aclObject) {});
myBucket.acl.readers.addAllAuthenticatedUsers(function(err, aclObject) {});
```

Or we could go the same route that gcloud-python went with the `grant_*` and `revoke_*` directives acting as the "commit" operation, so the code would look like:

```javascript
myBucket.acl.user('email@example.com').grantOwner(function(err, aclObject) {})
myBucket.acl.domain('example.com').revokeRead(function(err, aclObject) {})
myBucket.acl.allUsers().grantWrite(function(err, aclObject) {})
myBucket.acl.allAuthenticatedUsers().grantRead(function(err, aclObject) {})
```

Ideally you could chain this stuff together:

```javascript
myBucket.acl.user('email@example.com').user('other@example.com').grantOwner(function(err, aclObject) {})
```

Thoughts?

/cc @ryanseys @stephenplusplus

stephenplusplus:
Fixes #351 

This introduces support for a nicer API to ACL manipulations. The new methods are broken into groups representing a permission level, owners, readers, and writers. Each group has access methods to add or delete various entities: allAuthenticatedUsers, allUsers, domains, groups, projects, and users.

The full list of methods:

``` js
var acl = (myFile || myBucket).(acl || acl.default).(owners || readers || writers);

acl.addAllAuthenticatedUsers(function(err, aclObject) {});
acl.deleteAllAuthenticatedUsers(function(err) {});

acl.addAllUsers(function(err, aclObject) {});
acl.deleteAllUsers(function(err) {});

acl.addDomain(function(err, aclObject) {});
acl.deleteDomain(function(err) {});

acl.addGroup(function(err, aclObject) {});
acl.deleteGroup(function(err) {});

acl.addProject(function(err, aclObject) {});
acl.deleteProject(function(err) {});

acl.addUser(function(err, aclObject) {});
acl.deleteUser(function(err) {});
```

Additionally, a couple methods were added to simplify common operations; making a bucket or file publicly readable or the inverse.

``` js
(myFile || myBucket).makePrivate(function(err) {});
(myFile || myBucket).makePublic(function(err) {});
```

This will run `setMetadata` with a `predefinedAcl` value set to `private` or `publicRead`.

### DIFF
--- a/lib/storage/acl.js
+++ b/lib/storage/acl.js
@@ -139,7 +139,7 @@ Acl.prototype.owners = {};
  *
  * @example
  * //-
- * // Add a user as an owner of a file.
+ * // Add a user as a reader of a file.
  * //-
  * myFile.acl.readers.addUser('email@example.com', function(err, aclObject) {});
  *
@@ -178,7 +178,7 @@ Acl.prototype.readers = {};
  *
  * @example
  * //-
- * // Add a user as an owner of a file.
+ * // Add a user as a writer of a file.
  * //-
  * myFile.acl.writers.addUser('email@example.com', function(err, aclObject) {});
  *
--- a/docs/site/components/docs/docs.html
+++ b/docs/site/components/docs/docs.html
@@ -114,6 +114,20 @@ var pubsub = gcloud.pubsub({
       </p>
       <div hljs>
 var bucket = gcloud.storage.bucket('my-bucket');</div>
+      <h4>ACLs</h4>
+      <p>
+        Google Cloud Storage uses access control lists (ACLs) to manage object and bucket access. ACLs are the mechanism you use to share files with other users and allow other users to access your buckets and files.
+      </p>
+      <p>
+        Convenience methods are provided to perform common operations, such as the following.
+      </p>
+      <div hljs>
+// Allow a user to read files from a bucket.
+bucket.acl.readers.addUser('email@example.com', function(err, aclObject) {});
+
+// Revoke file ownership permissions from a group.
+var myFile = bucket.file('my-file.txt');
+myFile.acl.owners.deleteGroup('group-id', function(err, aclObject) {});</div>
       <p>
         See examples below for more on how to access your bucket to upload a file, read its files, create signed URLs, and more.
       </p>


lib/storage/acl.js
```diff
@@ -295,13 +418,55 @@ Acl.prototype.update = function(options, callback) {
...
+
+/**
  * Transform API responses to a consistent object format.
  *
  * @private
  */
 Acl.prototype.makeAclObject_ = function(accessControlObject) {
   var obj = {
-    scope: accessControlObject.scope,
+    scope: accessControlObject.entity,
```
stephenplusplus: Not sure if the API changed, this was always a bug, or I'm crazy. I noticed the API was responding with an `entity` property and not `scope`. I changed this to support the API.

lib/storage/acl.js
```diff
@@ -295,13 +418,55 @@ Acl.prototype.update = function(options, callback) {
...
+ * @alias acl.makePrivate
+ *
+ * @example
+ * myBucket.acl.makePrivate(function(err, aclObject) {});
+ * myFile.acl.makePrivate(function(err, aclObject) {});
+ */
+Acl.prototype.makePrivate = function(callback) {
+  this.delete({
+    scope: 'allUsers'
+  }, callback);
```
stephenplusplus: :us: Flagging this for attention. Not sure this is the intended effect making a bucket/object private is meant to have.

lib/storage/acl.js
```diff
@@ -295,13 +418,55 @@ Acl.prototype.update = function(options, callback) {
...
+ * @example
+ * myBucket.acl.makePublic(function(err, aclObject) {});
+ * myFile.acl.makePublic(function(err, aclObject) {});
+ */
+Acl.prototype.makePublic = function(callback) {
+  this.add({
+    scope: 'allUsers',
+    role: 'READER'
+  }, callback);
+};
```
stephenplusplus: (same as above)

:us: Flagging this for attention. Not sure this is the intended effect making a bucket/object public is meant to have.

ryanseys:
hmm... do you like `delete` or `remove` more? I'm feeling `remove` more because `acl.writers.removeUser` sounds more realistic than `acl.writers.deleteUser` but because we already have `delete` it might make sense to keep it that way... tough call.

lib/storage/acl.js
ryanseys: And the tests didn't catch it? :(

stephenplusplus: I believe we had regression tests checking the part of the response that was consistent, `role`. Our new tests check for `scope` as well.

lib/storage/acl.js
ryanseys: Hmm... yeah if this is a bucket, should it be `acl.default.add()`? go go gadget @jgeewax! (because you suggested the methods)

stephenplusplus:
I prefer `delete` for consistency across our library. Files, buckets, datasets, etc. are all deleted, and as you pointed out, we have `acl.delete` as well.

lib/storage/acl.js
ryanseys: This one seems reasonable for files. But if it's a bucket, again, might need to do defaults as well.

ryanseys:
Works for me! I've been writing too much ruby recently.

lib/storage/acl.js
stephenplusplus: `makePublic` can be done on a bucket via `myBucket.acl.makePublic()` and also on the default object as well: `myBucket.acl.default.makePublic()`.

stephenplusplus:
ha_ha()

ryanseys:
> `ha_ha()`

No brackets necessary.

stephenplusplus:
I think we have PR review inception here.

lib/storage/acl.js
ryanseys: What about:

``` js
if(this.isFile()) {
  this.acl.add({
    scope: 'allUsers',
    role: 'READER'
  }, callback);
} 
else if(this.isBucket()) {
  this.acl.add({
   scope: 'allUsers',
   role: 'READER'
 }, function(err, resp) {
    if(!err) {
      this.acl.default.add({
        scope: 'allUsers',
        role: 'READER'
      }, callback);
    } else {
      // uh oh
    }
  });
}
else {
  // nope chuck testa
}

Comment (stephenplusplus) [inline lib/storage/acl.js]: Basically, if it's a file, make it public. If it's a bucket, make the default acl public as well as the current files in the bucket?

I could see that being confusing since we have the `myBucket.acl.default` object as well, which would have its own `makePublic`. And in the case where a user wants to only make their current OR future files public, they would have to do the long-hand method.

File: lib/storage/acl.js [hunk 4]
```diff
@@ -328,3 +493,90 @@ Acl.prototype.makeReq_ = function(method, path, query, body, callback) {
...
+          // If the scope is not a prefix, it is a special entity group that
+          // does not require further details. The accessor methods only accept
+          // a callback.
+          scope = entity;
+          callback = entityId;
+        }
+
+        that[accessMethod]({
+          scope: scope,
+          role: role
```
ryanseys: No requirement to be uppercase e.g. OWNER, READER, WRITER?

lib/storage/acl.js
ryanseys: Yep just realized that and came back to say that we should handle both separately. :+1:

lib/storage/acl.js
stephenplusplus: Judging from the API working in the regression tests, it's not required to be uppercase. Nonetheless, it's an oversight that I didn't. Fixed: https://github.com/GoogleCloudPlatform/gcloud-node/commit/aefdd45899ee4e6ec2bd012ff3e195bf9ebbdcf3

regression/storage.js
```diff
@@ -149,6 +149,18 @@ describe('storage', function() {
...
+      it('should make a bucket publicly readable', function(done) {
+        bucket.acl.makePublic(function(err, accessControl) {
+          assert.equal(accessControl.role, storage.acl.READER_ROLE);
+          assert.equal(accessControl.scope, 'allUsers');
+          done();
+        });
+      });
+
+      it('should make a bucket private', function(done) {
+        bucket.acl.makePrivate(done);
```
ryanseys: assert role and scope like above test.

ryanseys: Oops this be regression tests.

ryanseys: Wait, how do we know this will be executed in order? Shouldn't this be done inside the other test?

stephenplusplus: Deleting comes back with an empty response body.

stephenplusplus: Updating any docs that didn't list the arguments to the callback correctly.

stephenplusplus: Tests merged: https://github.com/stephenplusplus/gcloud-node/commit/3a70ef724dcd18b19ee083225cea9809d61f342d

ryanseys:
Looks good to me :)

lib/storage/acl.js
```diff
@@ -295,13 +418,55 @@ Acl.prototype.update = function(options, callback) {
...
+ *
+ * This is a short-hand method which will grant "reader" permissions to the
+ * "allUsers" scope.
+ *
+ * @param {function} callback - The callback function.
+ *
+ * @alias acl.makePublic
+ *
+ * @example
+ * myBucket.acl.makePublic(function(err, aclObject) {});
```
ryanseys: `aclObject` --> `acl` ?

lib/storage/acl.js
```diff
@@ -67,11 +69,132 @@ var util = require('../common/util.js');
...
+ * //-
+ * myFile.acl.writers.addUser('email@example.com', function(err, aclObject) {});
+ *
+ * //-
+ * // For reference, the above command is the same as running the following.
+ * //-
+ * myFile.acl.add({
+ *   scope: 'user-email@example.com',
+ *   role: storage.acl.WRITER_ROLE
+ * }, function(err, aclObject) {});
```
ryanseys: `aclObject` --> `acl`?

lib/storage/acl.js
stephenplusplus: The examples have used `aclObject` since we added acl support. I tried to make the name overly descriptive just to help. Do you think we should switch them all to `acl`?

ryanseys: Hmm... if you are keeping them `aclObject` then can you change [this](https://github.com/stephenplusplus/gcloud-node/commit/6db8b3a64ebefd7cc91ad3870568e0d2dfcbba44#diff-712237668475919bc2aceb8288701749R508) to `aclObject` ?

stephenplusplus: I used `acl` for short-hand, as using `aclObject` would force a new line. That's our private, developer-documentation. I don't think we'll be confused :)

ryanseys: Oh, well I was confused.

stephenplusplus: What confused you, that it was private documentation? It's marked `@private` to keep it out of the published docs on the site. I could add the `/*! Developer Documentation` heading, though that's typically for broader documentation purposes, such as "The methods that follow are for...". This is just the documentation for this class.

If the confusion was `acl` vs `aclObject`, were you confused about the object itself, or that `acl` in this example wouldn't be the same object as `aclObject` from another example?

ryanseys: I just wasn't sure about the consistency of the variable names. The fact that the 80-char line limit trumps code consistency upsets me but I'm over it. I didn't notice the `@private` so I'll be more careful about that next time.

ryanseys:
I think this is acceptable for merge. Should probably wait for sign-off by @jgeewax for the makePrivate and makePublic however.

stephenplusplus:
Updated `makePublic()` and `makePrivate()`: https://github.com/stephenplusplus/gcloud-node/commit/fd35f2746b1620ca180961be811ca6d2c520fc10

They are now on the File and Bucket object themselves.

ryanseys:
Is this ready for another look?

lib/storage/acl.js
jgeewax: So do we want to create a ticket to change ours from `scope` to `entity` ?

stephenplusplus: We could do that with the benefit of being consistent with the upstream API, but I think we can still use `scope` if we think the term "scope" is clearer to the user. I'm happy with either option.

stephenplusplus:
Yes, and I should have explained the new `makePublic()` and `makePrivate()` methods use a shorthand "predefinedAcl" value that Ryan found. When a user makes an object public or private, it updates the "predefinedAcl" value as a metadata request to `publicRead` or `private`.

lib/storage/acl.js
ryanseys: The API never changed, it was always `entity` ([see here](https://github.com/google/google-api-nodejs-client/tree/master/apis/storage)) so I'm not sure where `scope` came from, may have been a simplification of the terminology which makes sense to me. For upstream support it makes sense to change to `entity` now that we have more reasonable methods to interacting with ACLs, people who want to do it manually can stick to the docs more closely.

test/storage/acl.js
```diff
@@ -20,7 +20,8 @@
 
 var Acl = require('../../lib/storage/acl.js');
 var assert = require('assert');
-var storage = require('../../lib/storage/index.js');
+var async = require('async');
+var Storage = require('../../lib/storage/index.js');
```
jgeewax: Why'd we change this to uppercase Storage? Just curious.

ryanseys: Capitalization suggests it's not an instance, which it isn't.

lib/storage/acl.js
jgeewax: I'm +1 on entity then.

test/storage/acl.js
jgeewax: Gotcha - thanks.

lib/storage/bucket.js
```diff
@@ -288,8 +306,51 @@ Bucket.prototype.getMetadata = function(callback) {
...
+ * @param {function=} callback - The callback function.
+ *
+ * @example
+ * bucket.makePublic(function(err) {});
+ */
+Bucket.prototype.makePublic = function(callback) {
+  callback = callback || util.noop;
+
+  this.setMetadata({
+    predefinedAcl: 'publicRead'
```
jgeewax: Does this make all the files in the bucket public as well? ie,

- create bucket
- create file in bucket
- `file.makePrivate()`
- `bucket.makePublic()`

Is file public? (I argue it should be.)

ryanseys: From the [docs](https://cloud.google.com/storage/docs/accesscontrol#predefined-acl):

> Gives the bucket or object owner OWNER permission and gives all users, both authenticated and anonymous, READER permission. When you apply this to an object, anyone on the Internet can read the object without authenticating. When you apply this to a bucket, anyone on the Internet can list objects without authenticating. 

If by "list objects", they mean read the objects contents, then yes.

ryanseys: Should we consider using `Cache-Control:private, max-age=0, no-transform` on the request to make the changes propagate immediately?

stephenplusplus: I tested @jgeewax's series of events, and the file remained private after making the bucket public. I'm not sure the URL I would visit to see a directory listing of my bucket to confirm whether it is included in that list.

jgeewax: +1 to making changes propagate immediately.

Re "list objects" : I don't believe that it means objects are readable... List means GET /bucket will tell you the list of keys, but GET /bucket/objects/object is what we're looking for.

I think `bucket.makePublic` needs to do more than just mark the bucket as public. It should also (depending on arguments):

- set the defaultACL to 'public' (so all future items are made public) (ie, `if (future) { /* set default ACL to public */ }`)
- set all existing items to 'public' (so all existing items are public (ie `if (recursive) { /* set everything to public */ }`)

ryanseys: > I'm not sure the URL I would visit to see a directory listing of my bucket to confirm whether it is included in that list.

http://storage.googleapis.com/my-bucket-name/ should work.

ryanseys:
@stephenplusplus For now, you should just remove `makePublic` and `makePrivate` and we can land these changes. Those methods are going to be more troublesome and are special cases and thus should be treated separately.

stephenplusplus:
Sounds good. :+1:

jgeewax:
Cool - can we create a ticket to make sure we don't forget those two methods?

ryanseys:
@jgeewax Done: #364

stephenplusplus:
This is now updated to remove makePublic() and makePrivate() as well as change `scope` -> `entity`.

lib/storage/acl.js
```diff
@@ -67,17 +69,138 @@ var util = require('../common/util.js');
...
  */
 function Acl(options) {
+  AclRoleAccessorMethods.call(this);
+
   this.makeReq = options.makeReq;
   this.pathPrefix = options.pathPrefix;
 }
 
 /**
+ * A convenience object of methods to add or delete ACL permissions from an
```
ryanseys: I'm assuming this is public documentation so perhaps we should clarify the role of these methods is for owners. Example:  "A convenience object of methods to add and delete **owner** ACL permissions from different entities."

lib/storage/acl.js
```diff
@@ -67,17 +69,138 @@ var util = require('../common/util.js');
...
+ * //-
+ * myFile.acl.add({
+ *   entity: 'user-email@example.com',
+ *   role: storage.acl.OWNER_ROLE
+ * }, function(err, aclObject) {});
+ */
+Acl.prototype.owners = {};
+
+/**
+ * A convenience object of methods to add or delete ACL permissions from an
```
ryanseys: Similarily, we might wish to specify this deals with the reader role. Example: "A convenience object of methods to add and delete **reader** ACL permissions from different entities."

lib/storage/acl.js
```diff
@@ -67,17 +69,138 @@ var util = require('../common/util.js');
...
+ * //-
+ * myFile.acl.add({
+ *   entity: 'user-email@example.com',
+ *   role: storage.acl.READER_ROLE
+ * }, function(err, aclObject) {});
+ */
+Acl.prototype.readers = {};
+
+/**
+ * A convenience object of methods to add or delete ACL permissions from an
```
ryanseys: And finally writer role. Example: "A convenience object of methods to add and delete **writer** ACL permissions from different entities."

lib/storage/acl.js
```diff
@@ -328,3 +451,90 @@ Acl.prototype.makeReq_ = function(method, path, query, body, callback) {
...
+
+AclRoleAccessorMethods.prototype._assignAccessMethods = function(role) {
+  var that = this;
+
+  var accessMethods = AclRoleAccessorMethods.accessMethods;
+  var entities = AclRoleAccessorMethods.entities;
+  var roleGroup = role.toLowerCase() + 's';
+
+  this[roleGroup] = entities.reduce(function(acc, entity) {
+    var isPrefix = entity.substr(-1) === '-';
```
ryanseys: It's more readable to use `entity.charAt(entity.length-1)` (and it's [faster](http://jsperf.com/splice-1-vs-charat-lenght-1-vs-index)!)

ryanseys:
Overall LGTM, just lightly commented on a couple things.

lib/storage/acl.js
```diff
@@ -67,17 +69,138 @@ var util = require('../common/util.js');
...
 function Acl(options) {
+  AclRoleAccessorMethods.call(this);
+
   this.makeReq = options.makeReq;
   this.pathPrefix = options.pathPrefix;
 }
 
 /**
+ * An object of convenience methods to add or delete owner ACL permissions for a
+ * given entity.
```
stephenplusplus: @ryanseys - the first of the three re-writes suggested.

stephenplusplus:
Addressed those last issues and squashed down to one commit.

Docs:
- http://stephenplusplus.github.io/gcloud-node/#/docs/master/storage/bucket
- http://stephenplusplus.github.io/gcloud-node/#/docs/master/storage/file

jgeewax:
For the docs, can we add a "suggested example" at the beginning of the page? something so that people don't have to scroll past the ".add()" confusion?

stephenplusplus:
I added some examples to the top of Storage pages: http://stephenplusplus.github.io/gcloud-node/#/docs/master/storage

Let me know if you prefer other examples to showcase.

jgeewax:
:+1:

PR #361, opened 2015-01-23
Specify folder parameter for folder-based filter

ryanseys:
Add folder-based filtering to `Bucket#getFiles` with a `folder` query parameter. 

Also implemented weird idea for returning prefixes for #342. I'm not sure how bad an idea it is, but basically return the `prefixes` as a property of the `files` Array. Yes, a property on an Array... How bad of an idea is that? :baby:

### DIFF
--- a/lib/storage/bucket.js
+++ b/lib/storage/bucket.js
@@ -226,6 +226,8 @@ Bucket.prototype.file = function(name) {
  *     return.
  * @param {string} query.pageToken - A previously-returned page token
  *     representing part of the larger set of results to view.
+ * @param {string} query.folder - Filter results to objects that are contained
+ *     within the given folder. Results will also include the folder itself.
  * @param {function} callback - The callback function.
  *
  * @example
@@ -234,6 +236,11 @@ Bucket.prototype.file = function(name) {
  *     // nextQuery will be non-null if there are more results.
  *     bucket.getFiles(nextQuery, function(err, files, nextQuery) {});
  *   }
+ *
+ *   // The prefixes property contains any prefixes returned in the response.
+ *   // This will only be populated when your request contains the 'folder' or
+ *   // 'delimiter' property.
+ *   files.prefixes;
  *
  *   // The `metadata` property is populated for you with the metadata at the
  *   // time of fetching.
@@ -250,12 +257,25 @@ Bucket.prototype.file = function(name) {
  * bucket.getFiles({
  *   maxResults: 5
  * }, function(err, files, nextQuery) {});
+ *
+ * //-
+ * // Fetch results within the folder 'hello/world'.
+ * // This is the same as specifying delimiter: '/', prefix: 'hello/world/'.
+ * //-
+ * bucket.getFiles({
+ *   folder: 'hello/world'
+ * }, function(err, files, nextQuery) {});
  */
 Bucket.prototype.getFiles = function(query, callback) {
   var that = this;
   if (!callback) {
     callback = query;
     query = {};
+  }
+  if (query.folder) {
+    query.delimiter = '/';
+    query.prefix = query.folder + '/';
+    delete query.folder;
   }
   this.makeReq_('GET', '/o', query, true, function(err, resp) {
     if (err) {
@@ -267,6 +287,9 @@ Bucket.prototype.getFiles = function(query, callback) {
       file.metadata = item;
       return file;
     });
+    if (resp.prefixes) {
+      files.prefixes = resp.prefixes; // bad idea?
+    }
     var nextQuery = null;
     if (resp.nextPageToken) {
       nextQuery = extend({}, query, { pageToken: resp.nextPageToken });
--- a/test/storage/bucket.js
+++ b/test/storage/bucket.js
@@ -152,6 +152,26 @@ describe('Bucket', function() {
       bucket.getFiles({ maxResults: 5, pageToken: token }, util.noop);
     });
 
+    it('should get files based on folder', function(done) {
+      var folder = 'hello/world';
+      bucket.makeReq_ = function(method, path, query, body, callback) {
+        assert.deepEqual(query, { delimiter: '/', prefix: 'hello/world/' });
+        callback(null, { items: [] });
+      };
+      bucket.getFiles({ folder: folder }, done);
+    });
+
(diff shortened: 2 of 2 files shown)


ryanseys:
Agh, tags. What ones do I use?? I just added a boat-load of them.

lib/storage/bucket.js
```diff
@@ -235,6 +237,11 @@ Bucket.prototype.file = function(name) {
  *     bucket.getFiles(nextQuery, function(err, files, nextQuery) {});
  *   }
  *
+ *   // The prefixes property contains any prefixes returned in the response.
```
parkr: Silly: can you put quotes around `prefixes`? I was reading too quickly and didn't realize it was a property until I realized that there was a verb-noun disagreement with "prefixes" and "contains". It seems you have quotes around the properties below, so it'd be helpful to have them here too. :sparkles:

lib/storage/bucket.js
```diff
@@ -257,6 +272,11 @@ Bucket.prototype.getFiles = function(query, callback) {
     callback = query;
     query = {};
   }
+  if (query.folder) {
+    query.delimiter = '/';
+    query.prefix = query.folder + '/';
```
parkr: Is the prefix handler smart enough to handle `hello/world//` (i.e. multiple trailing slashes)? What about a leading slash?

lib/storage/bucket.js
```diff
@@ -267,6 +287,9 @@ Bucket.prototype.getFiles = function(query, callback) {
       file.metadata = item;
       return file;
     });
+    if (resp.prefixes) {
+      files.prefixes = resp.prefixes; // bad idea?
```
parkr: What's the motivation for this? If the motivation is :+1: then I'd say it's a good idea :smile:

stephenplusplus:
> Agh, tags. What ones do I use?? I just added a boat-load of them.

Lol

What does the API response (`resp.prefixes`) look like? I don't feel like giving the dev a `prefixes` property on the files array is too important or useful. If they're using the `folder` property on a query passed to `getFiles`, they may not even realize we're using "prefixes" and "delimiters" under the hood, and wouldn't know what to do with that.

stephenplusplus:
`in progress` and `ready for review` are relics from the days of waffle.io. `opinions wanted` and `enhancement` (and really, anything except for the service a PR is affecting) make more sense to me to be used on issues. But, I won't remove them again if you would like them back :)

ryanseys:
> What does the API response (resp.prefixes) look like?

Looks something like `[ 'test/goodbye/', 'test/hello/' ]` when folder is set to `test` so basically, it's the folders in a given folder. What if you used a new query flag that the developer can specify and then we will return them in the `prefixes` property. Like `includePrefixes: true` or something?

lib/storage/bucket.js
ryanseys: Will do! :+1:

lib/storage/bucket.js
ryanseys: Nope, it's very dumb. This is largely a POC so not sure how we should handle those cases.

stephenplusplus:
What if on the File object itself, it had a "folder" property that was it's prefix? I don't think an array of prefixes is too useful.

ryanseys:
Yeah, thought of that too! A complaint I have with that with doing that is that suddenly everyone's code will need to handle the case where the file isn't actually a file and has no metadata or id or anything, and just contains a folder property instead.

lib/storage/bucket.js
ryanseys: HAHA that's what I'm trying to figure out too. It's a returned value from the API that we're trying to figure what it actually does and how important it is. This is the simplest way I could think of returning them without breaking too much or complicating the API beyond what's necessary.

stephenplusplus:
How would that happen? I'm envisioning:

```js
bucket.getFiles(function(err, files) {
  files[0].name = 'iamin/a/folder.jpg'
  files[0].folder = 'iamin/a'

  files[1].name = 'iamnotinafolder.jpg'
  files[0].folder = undefined
});
```

Though, chasing this thought down, I'm still not sure that's a helpful bit of knowledge to have. If I really wanted to, I could use the native `path` module to parse out the folder/extension from a file path. The most important thing should be `bucket.getFiles({ folder: 'a/folder' })` to get only the files from a folder, and `bucket.getFolders()` (which I don't think we know how to do yet / haven't implemented).

ryanseys:
The problem with your example is that the files might not exist but the folders would. Also, `bucket.getFiles({ folder: 'test' })` returns the files in the `test` folder (and the folder itself), and `prefixes` returns the folders in that folder.

ryanseys:
Because, the `test` folder is returned the results, we could add the prefixes as a `subfolders` property on that object in the files list? This won't work for the root folder (`/`) because it's not a folder, and does not exist. In that case, getting the subfolders in the root would be impossible which isn't ideal.

stephenplusplus:
My head is spinning a bit, because I still don't see the value to a user. There's clear value to `bucket.getFiles({ folder: 'a/folder' })`, and `bucket.getFolders()` would be nice as well, though not as vital. Even for a power-user, getting a File object that retains knowledge of the folder it's in or sub-folder, or prefix, etc, isn't something I can see being useful. And if I'm wrong and the user does need that information, they can solve the problem simply by putting files that were retrieved from a folder into a data structure that's labeled as such:

```js
var folders = { dir1: [], dir2: [] };

Object.keys(folders).forEach(function(folder) {
  bucket.getFiles({ folder: folder }, function (err, files) {
    folders[folder] = files;
  });
});
```

Hope I'm not overlooking something obvious. As is evident from my history on this project, it sometimes just takes a while for me to understand the Google APIs :)

ryanseys:
Yeah my head is spinning too. In this case I think it's best if we just find the simplest way to return the parameter rather than potentially messing up some folder-related methods that we don't fully understand.

ryanseys:
Ok, no `folder` value, how about just returning prefixes so that this value can be used to allow developers to create their own similar functionality if required (e.g. like in #342)

stephenplusplus:
I'll let you make the call on this.

ryanseys:
Yeah, I've decided this idea sucks. Will reconsider #342 or something similar.

PR #371, opened 2015-02-02
Clarify bucket metadata example

### LINKED ISSUE #370: Docstring for bucket.setMetadata has invalid argument
From [the code][1]:

```javascript
bucket.setMetadata({
  contentType: 'application/x-font-ttf',
  metadata: {
    my: 'custom',
    properties: 'go here'
  }
}, function(err, metadata) {});
```

[1]: https://github.com/GoogleCloudPlatform/gcloud-node/blob/fcc6123ddb0678f29d9ccd606e81318f988b4fa8/lib/storage/bucket.js#L308

ryanseys:
Fixes #370

### DIFF
--- a/lib/storage/bucket.js
+++ b/lib/storage/bucket.js
@@ -304,11 +304,13 @@ Bucket.prototype.getMetadata = function(callback) {
  * @param {function=} callback - The callback function.
  *
  * @example
+ * //-
+ * // Set website metadata field on the bucket.
+ * //-
  * bucket.setMetadata({
- *   contentType: 'application/x-font-ttf',
- *   metadata: {
- *     my: 'custom',
- *     properties: 'go here'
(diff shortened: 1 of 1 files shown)


dhermes:
Shouldn't the key just be `metadata`?

ryanseys:
According to [this](https://cloud.google.com/storage/docs/json_api/v1/buckets/patch) no such key exists.

stephenplusplus:
LGTM - this is the way our regression test does it: https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/regression/storage.js#L289

dhermes:
Sorry I confused myself (was looking on my phone) both about the value being passed and the relevant noun (I thought we were talking about objects, not buckets).

:+1:

PR #373, opened 2015-02-02
storage: pass metadata during stream upload

### LINKED ISSUE #368: Allow setting of custom metadata
I understand from the documents for cloud storage that it is possible to set custom metadata with the `x-goog-meta-` prefix as specified here: https://cloud.google.com/storage/docs/reference-headers#xgoogmeta

I've tried to take this approach to set custom metadata on objects created using `gcloud-node` as follows:

```javascript
    var stream = file.createWriteStream({
      metadata: {
        'x-goog-meta-entry-type': entry.type,
      }
    });

    stream.once("finish", resolve);
    stream.once("error", reject);
    stream.end(buffer);
```

However, no objects created using that approach have the custom metadata reflected when requested via `file.getMetadata()` nor when viewed via the online storage browser.

How can one set custom metadata using `gcloud-node`?

stephenplusplus:
Fixes #368

Looks like there are two concepts of metadata in an Object response ([seen here](https://cloud.google.com/storage/docs/json_api/v1/objects#resource)),
1. metadata, such as `etag`, `selfLink`, `contentType`, `updated`, `kind`, `id`, etc.
2. custom metadata, embedded within `metadata.metadata`

This PR fixes the way in which the insert API receives the request body from us , so that custom properties can be set as [we say they are](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/storage/bucket.js#L400) during a file upload.

However, that would stop a user from specifying other, non-custom, metadata properties. Maybe we should just change the docs to say "custom metadata properties go inside a `metadata` object", e.g.

``` js
var options = {
  metadata: {
    contentType: 'content-type',
    metadata: {
      custom: 'metadata'
    }
  }
}

bucket.upload("/file/path.txt", options, //...

bucket.file("file.txt").createWriteStream(options)
```

I'm talking myself out of thinking this PR is a good idea. Halp!

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -304,7 +304,9 @@ function makeWritableStream(dup, options, onComplete) {
       // Write the metadata to the request.
       stream.write('--' + boundary + '\n');
       stream.write('Content-Type: application/json\n\n');
-      stream.write(JSON.stringify(options.metadata));
+      stream.write(JSON.stringify({
+        metadata: options.metadata
+      }));
       stream.write('\n\n');
       stream.write('--' + boundary + '\n');
       stream.write('Content-Type: ' + streamType + '\n\n');
--- a/regression/storage.js
+++ b/regression/storage.js
@@ -376,7 +376,10 @@ describe('storage', function() {
 
       it('should write metadata', function(done) {
         var options = {
-          metadata: { contentType: 'image/png' },
+          metadata: {
+            contentType: 'image/png',
+            property: 'value'
+          },
           resumable: false
         };
 
@@ -386,6 +389,7 @@ describe('storage', function() {
           file.getMetadata(function(err, metadata) {
             assert.ifError(err);
             assert.equal(metadata.contentType, options.metadata.contentType);
+            assert.equal(metadata.metadata.property, options.metadata.property);
             file.delete(done);
           });
         });
--- a/test/common/util.js
+++ b/test/common/util.js
@@ -210,7 +210,9 @@ describe('common/util', function() {
           assert(endFirstBoundaryIdx > startFirstBoundaryIdx);
           assert(endBoundaryIdx > -1);
 
-          assert(written.indexOf(JSON.stringify(metadata)) > -1);
+          assert(written.indexOf(JSON.stringify({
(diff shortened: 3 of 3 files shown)


ryanseys:
`metadata.metadata` is annoying but that's how it's documented :(

> Maybe we should just change the docs to say "custom metadata properties go inside a metadata object"

Yes. They also can know this from just reading the docs. The outer "metadata" is actually just request body in this case so maybe we can clarify this when we change the docs. One is "User-provided metadata" and the other is "Google provided metadata" (???) I think that might help, but yeah.

dhermes:
I'm quite interested in what you decide. We (`gcloud-python`) currently have a property for each key in the dictionary and I am debating just throwing it out and using the dictionary as-is.

The `metadata.metadata` concept makes me sad all day, but the API is "set in stone" at this point.

ryanseys:
So wait, what other properties go on the same level as the outer metadata object as seen at https://github.com/GoogleCloudPlatform/gcloud-node/issues/368#issuecomment-72519661 ?

stephenplusplus:
I believe we should keep things as-is (close this PR). `metadata.metadata` is a bit silly, but if that's what the API wants, I don't think we can get around it.

ryanseys:
Agreed. Is that what this PR aims to accomplish right now?

stephenplusplus:
Yeah, it just unboxed the `metadata = { metadata: {} }` object to send just the custom part. I thought it was a bug I was fixing, but it's just a quirk.

wong2:
ha, it wasted me hours to figure out the silly `file.setMetadata({metadata: {}}, ...)` api, good job

jgeewax:
Hey @wong2 : Can you help us understand what we could've done different to avoid wasting time?

wong2:
@jgeewax sorry for my previous impolite comment, you've documented it, it's my fault for not noticing it.

jgeewax:
@wong2 No problem at all -- just wanted to make sure make sure it wasn't something we could make better ! :)

wong2:
@jgeewax maybe you can add a more obvious notice aboout this in the document?

PR #381, opened 2015-02-04
storage: add download() method

### LINKED ISSUE #367: add download() method
Hi, 
I spent few hours looking into the code, but didn't find the solution. 
What am i suppose to do in order to load the content of a file from a createReadStream into a variable ?
In the example there is only a read-to-pipe-to-write example.

I just want to do something like 

```javascript
var streamPointer = cloudstorage.createReadStream(myTextFile)
    streamPointer.on('complete', function(res) {
            myVar = res
        })
        
```

stephenplusplus:
Fixes #367

Adds a download method to a File object:

``` js
// Getting the contents of a file.
file.download(function(err, contents) {})
file.download({ start: 0, end: 100 }, function(err, contents) {});

// Writing the file to a local destination
file.download({ destination: '/Users/stephen/file.txt' }, function(err) {});
file.download({ start: 0, end: 100, destination: '/Users/stephen/file.txt' }, function(err) {});
```

This method just blankets `file.createReadStream(options)`, so the options passed to `download` are passed up to `createReadStream`.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -25,6 +25,8 @@ var ConfigStore = require('configstore');
 var crc = require('fast-crc32c');
 var crypto = require('crypto');
 var duplexify = require('duplexify');
+var fs = require('fs');
+var once = require('once');
 var request = require('request');
 var streamEvents = require('stream-events');
 var through = require('through2');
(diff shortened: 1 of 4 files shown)


lib/storage/file.js
```diff
@@ -620,6 +621,59 @@ File.prototype.delete = function(callback) {
...
+  var destination = options.destination;
+  delete options.destination;
+
+  var fileStream = this.createReadStream(options);
+
+  if (destination) {
+    fileStream
+      .on('error', callback)
+      .pipe(fs.createWriteStream(destination))
+      .on('error', callback)
```
ryanseys: Will this callback be potentially called twice?

test/storage/file.js
```diff
@@ -741,6 +743,127 @@ describe('File', function() {
...
+        return fileReadStream;
+      };
+    });
+
+    it('should accept just a callback', function(done) {
+      fileReadStream._read = function() {
+        done();
+      };
+
+      file.download(assert.ifError);
```
ryanseys: Elegant :)

test/storage/file.js
```diff
@@ -741,6 +743,127 @@ describe('File', function() {
...
+
+        file.download(function(err, remoteFileContents) {
+          assert.ifError(err);
+
+          assert.equal(fileContents, remoteFileContents);
+          done();
+        });
+      });
+
+      it('execute callback with error', function(done) {
```
ryanseys: should execute callback with error

regression/storage.js
```diff
@@ -362,6 +362,20 @@ describe('storage', function() {
...
     });
 
+    it('should download a file to memory', function(done) {
+      var fileContents = fs.readFileSync(files.big.path);
+
+      bucket.upload(files.big.path, function(err, file) {
+        assert.ifError(err);
+
+        file.download(function(err, remoteContents) {
+          assert.ifError();
```
chuckjaz: This looks like it should be:
``` js
    assert.ifError(err);
```

lib/storage/file.js
```diff
@@ -620,6 +621,59 @@ File.prototype.delete = function(callback) {
...
+ *
+ * @param {object=} options - Optional configuration. The arguments match those
+ *     passed to {module:storage/file#createReadStream}.
+ * @param {string} options.destination - Local file path to write the file's
+ *     contents to.
+ * @param {function} callback - The callback function.
+ *
+ * @example
+ * //-
+ * // Download a file into memory.
```
ryanseys: Clarify that it will be available in the `contents` variable.

regression/storage.js
ryanseys: Good catch!

test/storage/file.js
stephenplusplus: There I go with my anti-should streak again!

lib/storage/file.js
stephenplusplus: Good question!

stephenplusplus: Looks like yes: https://gist.github.com/stephenplusplus/fdd874987fc090b43e16

ryanseys: Hmm we should just call it once I'd think. As a developer, I would assume it only gets called once.

stephenplusplus: That makes sense to me. LMKWYT: https://github.com/GoogleCloudPlatform/gcloud-node/commit/47b69b36e364a015579390f8a8d7e390d73b3b4b

ryanseys:
Will require a rebase

stephenplusplus:
Using the new `once` module + rebased.

ryanseys:
LGTM! :) :ship:

PR #383, opened 2015-02-06
Update package.json dependencies

ryanseys:


package.json
```diff
@@ -17,6 +17,10 @@
       "email": "pcostell@google.com"
     },
     {
+      "name": "Ryan Seys",
+      "email": "ryan@ryanseys.com"
+    },
```
stephenplusplus: Who's this guy?!

stephenplusplus:
The only thing that will actually be changed was `extend` having a major bump from 1 to 2.0. The change was minimal though, according to their changelog. Merge city, here we come.

package.json
ryanseys: Am I worthy yet??

ryanseys:
Ah, true dat. Didn't think of this. Just used `npm-check-updates` to bump up the jam all over.

package.json
stephenplusplus: Of course! :dancer: :lollipop:

PR #386, opened 2015-02-07
Add Node v0.12.0 and io.js to .travis.yml

### LINKED ISSUE #380: Is gcloud-node going to work with iojs?
Maybe we should add some tests to make sure our build passes using the iojs.org runtime? Not sure if it's worthwhile, but figured I'd toss this out.

ryanseys:
Seems to be no changes required to support Node.js v0.12.0 so adding it to Travis.

Closes #380

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 node_js:
 - 0.10
 - 0.12
+- iojs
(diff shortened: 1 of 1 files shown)


ryanseys:
Added io.js here too. We can remove support if we run into large issues down the line but for now it should work okay.

Issue #384, opened 2015-02-06
Expose something finer than just `.save()` through the datastore abstraction


chilts:
One thing I'd like is more features than just `.save()`. At the moment, I'm inserting a kind=User with name=chilts. However if I were to have another user ask for the name 'chilts', then `.save()` will say "Oh, the key is complete, so I'll do an update". I would like to make it try an insert and fail because that name already exists (ie. the primary key) ... is that currently possible via your api (from what I can tell it isn't)?

I love how all of the thorny details are abstracted away, but possibly exposing more of the API through here would be awesome. i.e. `.insert()`, `.update()` as well as `.save()`. Sometimes the key logic in `.save()` isn't needed.

Thanks again for a great library.

dhermes:
@chilts The [relevant][1] verbs are:

- `update` - Must exist
- `insert` - Must not exist
- `upsert` - Can be brand new or exist already
- `insert_auto_id` - Incomplete path

I am also interested in

1. The demand for this from end users
1. The solution to this demand

[1]: https://github.com/GoogleCloudPlatform/google-cloud-datastore/blob/21e029b2657ebb15fef5429c90c4315326855fd1/proto/datastore_v1.proto#L388-L400

dhermes:
ISTM after a quick scan that the [relevant bits of code][1] only use `upsert` and `insert_auto_id` (as is done in `gcloud-python`).

It just build an accumulator which starts out as

```javascript
    {
      upsert: [],
      insert_auto_id: []
    }
```

and is translated to the relevant protobuf via `DatastoreRequest.prototype.makeReq_`.

You could build your own accumulator with `update` and `insert` keys and then execute

```
DatastoreRequest().makeReq_('commit', req, onCommitCallback);
```

[1]: https://github.com/GoogleCloudPlatform/gcloud-node/blob/d73ce95ee552bf4e0d0edd711e4200cfa70cb5a4/lib/datastore/request.js#L255-L260

ryanseys:
Agreed, we should support `insert` functionality on its own as it's pretty standard for the "attempting to add new users" use-case that @chilts described above.

ryanseys:
@chilts How does #387 fill this need? Basically you'll explicitly set a `method` for your entity like so:

``` js
var gcloud = require('gcloud');
var datastore = gcloud.datastore;
var dataset = datastore.dataset();

dataset.save([
  {
    key: dataset.key({ path: ['User', 'chilts'] }),
[...]

PR #387, opened 2015-02-08
Allow explicit set of method on Dataset#save

ryanseys:
Should satisfy the needs of #384

### DIFF
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -166,6 +166,8 @@ DatastoreRequest.prototype.get = function(keys, callback) {
  *
  * @param {object|object[]} entities - Datastore key object(s).
  * @param {Key} entities.key - Datastore key object.
+ * @param {string=} entities.method - Optional method to explicity use for save.
+ *     The choices include 'insert', 'update', 'upsert' and 'auto_insert_id'.
  * @param {object|object[]} entities.data - Data to save with the provided key.
  *     If you provide an array of objects, you must use the explicit syntax:
  *     `name` for the name of the property and `value` for its value. You may
@@ -209,7 +211,7 @@ DatastoreRequest.prototype.get = function(keys, callback) {
  * //-
  * var companyKey = dataset.key(['Company', 123]);
  * var productKey = dataset.key(['Product', 'Computer']);
- * 
+ *
  * dataset.save([
  *   {
  *     key: companyKey,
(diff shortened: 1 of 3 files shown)


stephenplusplus:
The way you implemented this is quite brilliant in its simplicity!

Do you think we should branch out the API to add these methods in? `dataset.insert/upsert/etc` that would just wrap the longer `method: "insert/upsert/etc"` call?

ryanseys:
I wasn't sure what was preferred here, so your suggestions are very welcome! This is one less-obvious technique, however if you think more value could come from separate methods, let's do that instead.

stephenplusplus:
My only thought is `save` is magic, so `insert/upsert/etc` is declarative. I just think it would be expected that we have those methods. Your implementation is still perfect, we can just add the convenience methods (which we're growing quite a collection of!)

ryanseys:
Should we merge this and then we can open an issue to add the declarative methods?

test/datastore/request.js
```diff
@@ -205,6 +205,25 @@ describe('Request', function() {
...
+    it('should save with specific method', function(done) {
+      request.makeReq_ = function(method, req, callback) {
+        assert.equal(method, 'commit');
+        assert.equal(req.mutation.insert.length, 1);
+        assert.equal(req.mutation.update.length, 1);
+        assert.equal(req.mutation.insert[0].property[0].name, 'k');
+        assert.equal(
+          req.mutation.insert[0].property[0].value.string_value, 'v');
+        assert.equal(req.mutation.update[0].property[0].name, 'k2');
+          assert.equal(
```
stephenplusplus: Extra indent.

lib/datastore/request.js
```diff
@@ -252,15 +271,35 @@ DatastoreRequest.prototype.save = function(entities, callback) {
 
       ent.key = entity.keyToKeyProto(entityObject.key);
 
-      if (entity.isKeyComplete(entityObject.key)) {
-        acc.upsert.push(ent);
+      if (method) {
+        switch(method) {
```
stephenplusplus: `switch (method)`

regression/datastore.js
```diff
@@ -119,6 +119,38 @@ describe('datastore', function() {
...
+      var postKey = ds.key('Post');
+
+      ds.save({ key: postKey, data: post }, function(err) {
+        assert.ifError(err);
+
+        // The key's path should now be complete.
+        assert(postKey.path[1]);
+
+        ds.save({ key: postKey, method: 'insert', data: post }, function(err) {
+          assert.notEqual(err, null); // should fail insert
```
stephenplusplus: I like!

stephenplusplus:
SGTM. Minor formatting comments, then self-merging is cool with me.

lib/datastore/request.js
```diff
@@ -224,6 +226,21 @@ DatastoreRequest.prototype.get = function(keys, callback) {
...
+ * //-
+ * var userKey = dataset.key(['User', 'chilts']);
+ *
+ * dataset.save([
+ *   {
+ *     key: userKey,
+ *     method: 'insert', // force the method to 'insert'
+ *     data: {
+ *       fullName: 'Andrew Chilton'
+ *     }
```
chilts: I don't mind if you change this to another (made up) name. :)

stephenplusplus: No way, you're part of this library forever now. https://github.com/GoogleCloudPlatform/gcloud-node/commit/d3fe788b6a4c59269d17dad246f108d8160a7b3b

Seriously though, I just used your name there to be funny. If you want me to remove the honorable mentions now, I'll be happy to. :+1:

PR #388, opened 2015-02-09
Implement publishBatch and pullBatch

### LINKED ISSUE #377: Looks like the build is broken?
Either that or the badge is broken... Can someone take a look today?

### LINKED ISSUE #379: Implement publishBatch and pullBatch methods
RE: https://github.com/GoogleCloudPlatform/gcloud-node/issues/377#issuecomment-72664423

API specs:
- https://cloud.google.com/pubsub/v1beta1/topics/publishBatch
- https://cloud.google.com/pubsub/v1beta1/subscriptions/pullBatch

Ideally, `publishBatch` could accept messages in a raw or simple format.

stephenplusplus:
Fixes #379
Fixes #377

Topics have a new `topic.publishBatch` method that takes an array of messages in any format (string, object, or raw [like what would be sent to `topic.publishRaw`).

Subscriptions accept a new parameter, `maxCount`, to `subscription.pull` which triggers a batched `pull` for messages, up to the count specified.

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -92,9 +92,14 @@ module.exports.extendGlobalConfig = extendGlobalConfig;
  * // [ 'Hi' ]
  */
 function arrayize(input) {
+  if (input === null || input === undefined) {
+    return [];
+  }
+
   if (!Array.isArray(input)) {
     return [input];
   }
+
   return input;
 }
 
--- a/lib/pubsub/subscription.js
+++ b/lib/pubsub/subscription.js
@@ -146,16 +146,24 @@ Subscription.formatName_ = function(projectId, name) {
  * @private
  */
 Subscription.formatMessage_ = function(msg) {
+  var event = msg.pubsubEvent;
+
   var message = {
-    id: msg.ackId
+    ackId: msg.ackId
   };
-  var evt = msg.pubsubEvent;
-  if (evt && evt.message && evt.message.data) {
-    message.data = new Buffer(evt.message.data, 'base64').toString('utf-8');
-    try {
-      message.data = JSON.parse(message.data);
-    } catch(e) {}
+
+  if (event && event.message) {
+    message.id = event.message.messageId;
+
+    if (event.message.data) {
+      message.data = new Buffer(event.message.data, 'base64').toString('utf-8');
+
+      try {
+        message.data = JSON.parse(message.data);
+      } catch(e) {}
+    }
   }
+
   return message;
 };
 
@@ -216,6 +224,7 @@ Subscription.prototype.startPulling_ = function() {
     return;
   }
   this.pull({
+    maxResults: 1,
     returnImmediately: false
   }, function(err, message) {
     if (err) {
@@ -243,7 +252,7 @@ Subscription.prototype.startPulling_ = function() {
 Subscription.prototype.ack = function(ids, callback) {
   if (!ids || ids.length === 0) {
     throw new Error(
-        'At least one ID must be specified before it can be acknowledged');
+        'At least one ID must be specified before it can be acknowledged.');
   }
   ids = util.arrayize(ids);
   var body = {
@@ -278,50 +287,86 @@ Subscription.prototype.delete = function(callback) {
 
 /**
  * Pull messages from the subscribed topic. If messages were found, your
- * callback is executed with the message object.
+ * callback is executed with an array of message objects.
  *
  * Note that messages are pulled automatically once you register your first
  * event listener to the subscription, thus the call to `pull` is handled for
  * you. If you don't want to start pulling, simply don't register a
  * `subscription.on('message', function() {})` event handler.
  *
+ * @todo Should not be racing with other pull.
+ * @todo Fix API to return a list of messages.
+ *
  * @param {object=} options - Configuration object.
- * @param {boolean=} options.returnImmediately - If set, the system will respond
+ * @param {boolean} options.returnImmediately - If set, the system will respond
  *     immediately. Otherwise, wait until new messages are available. Returns if
  *     timeout is reached.
+ * @param {number} options.maxResults - Limit the amount of messages pulled.
  * @param {function} callback - The callback function.
  *
  * @example
- * subscription.pull(function(err, message) {
- *   // message.id = ID used to acknowledge its receival.
- *   // message.data = Contents of the message.
+ * //-
+ * // Pull all available messages.
+ * //-
+ * subscription.pull(function(err, messages) {
+ *   // messages = [
+ *   //   {
+ *   //     ackId: '', // ID used to acknowledge its receival.
+ *   //     id: '',    // Unique message ID.
+ *   //     data: ''   // Contents of the message.
+ *   //   },
+ *   //   // ...
+ *   // ]
  * });
+ *
+ * //-
+ * // Pull a single message.
+ * //-
+ * var opts = {
+ *   maxResults: 1
+ * };
+ *
+ * subscription.pull(opts, function(err, messages) {});
  */
 Subscription.prototype.pull = function(options, callback) {
   var that = this;
-  // TODO(jbd): Should not be racing with other pull.
+  var MAX_EVENTS_LIMIT = 1000;
+  var apiEndpoint = 'subscriptions/pullBatch';
+
   if (!callback) {
     callback = options;
     options = {};
   }
+
+  if (!util.is(options.maxResults, 'number')) {
+    options.maxResults = MAX_EVENTS_LIMIT;
+  }
+
   var body = {
     subscription: this.name,
-    returnImmediately: !!options.returnImmediately
+    returnImmediately: !!options.returnImmediately,
+    maxEvents: options.maxResults
   };
-  this.makeReq_(
-      'POST', 'subscriptions/pull', null, body, function(err, message) {
-    // TODO(jbd): Fix API to return a list of messages.
+
+  this.makeReq_('POST', apiEndpoint, null, body, function(err, response) {
     if (err) {
       callback(err);
       return;
     }
-    message = Subscription.formatMessage_(message);
+
+    var messages = response.pullResponses || [response];
+    messages = messages.map(Subscription.formatMessage_);
+
     if (that.autoAck) {
-      that.ack(message.id, function(err) {
-        callback(err, message);
+      var ackIds = messages.map(function(message) {
+        return message.ackId;
+      });
+
+      that.ack(ackIds, function(err) {
+        callback(err, messages);
       });
     } else {
-      callback(null, message);
+      callback(null, messages);
     }
   });
 };
--- a/lib/pubsub/topic.js
+++ b/lib/pubsub/topic.js
@@ -63,6 +63,23 @@ function Topic(pubsub, options) {
 }
 
 /**
+ * Format a message object as the upstream API expects it.
+ *
+ * @private
+ *
+ * @return {object}
+ */
+Topic.formatMessage_ = function(message) {
+  if (!util.is(message.data, 'buffer')) {
+    message.data = new Buffer(JSON.stringify(message.data));
+  }
+
+  message.data = message.data.toString('base64');
+
+  return message;
+};
+
+/**
  * Format the name of a topic. A Topic's full name is in the format of
  * /topics/{projectId}/{name}.
  *
@@ -83,59 +100,55 @@ Topic.formatName_ = function(projectId, name) {
  *
  * @throws {Error} If no message is provided.
  *
- * @param {*} message - The message to publish.
+ * @param {object|object[]} message - The message(s) to publish.
+ * @param {*} message.data - The contents of the message.
+ * @param {array=} message.labels - Labels to apply to the message.
  * @param {function=} callback - The callback function.
  *
  * @example
- * topic.publish('New message!', function(err) {});
+ * var registrationMessage = {
+ *   data: {
+ *     userId: 3,
+ *     name: 'Stephen',
+ *     event: 'new user'
+ *   },
+ *   labels: [
+ *     'registration'
+ *   ]
+ * };
+ * topic.publish(registrationMessage, function(err) {});
  *
- * topic.publish({
- *   user_id: 3,
- *   name: 'Stephen',
- *   message: 'Hello from me!'
- * }, function(err) {});
+ * //-
+ * // You can publish a batch of messages at once by supplying an array.
+ * //-
+ * var purchaseMessage = {
+ *   data: {
+ *     userId: 3,
+ *     product: 'computer',
+ *     event: 'purchase'
+ *   }
+ * };
+ *
+ * topic.publish([
+ *   registrationMessage,
+ *   purchaseMessage
+ * ], function(err) {});
  */
-Topic.prototype.publish = function(message, callback) {
-  if (!message) {
-    throw new Error('Cannot publish an empty message.');
-  }
-  callback = callback || util.noop;
-  if (!util.is(message, 'string') && !util.is(message, 'buffer')) {
-    message = JSON.stringify(message);
-  }
-  this.publishRaw({
-    data: new Buffer(message).toString('base64')
-  }, callback);
-};
+Topic.prototype.publish = function(messages, callback) {
+  messages = util.arrayize(messages);
 
-/**
- * Publish a raw message.
- *
- * @throws {Error} If no message is provided.
- *
- * @param {object} message - Raw message to publish.
- * @param {array=} message.label - List of labels for the message.
- * @param {string} message.data - The base64-encoded contents of the message.
- * @param {function=} callback - The callback function.
- *
- * @example
- * topic.publishRaw({
- *   data: new Buffer('New message!').toString('base64')
- * }, function(err) {});
- */
-Topic.prototype.publishRaw = function(message, callback) {
-  if (!message) {
-    throw new Error('Cannot publish an empty message.');
+  if (messages.length === 0) {
+    throw new Error('Cannot publish without a message.');
   }
+
   callback = callback || util.noop;
-  if (!util.is(message.data, 'string') && !util.is(message.data, 'buffer')) {
-    message.data = new Buffer(JSON.stringify(message.data)).toString('base64');
-  }
+
   var body = {
-    message: message,
-    topic: this.name
+    topic: this.name,
+    messages: messages.map(Topic.formatMessage_)
   };
-  this.makeReq_('POST', 'topics/publish', null, body, callback);
+
+  this.makeReq_('POST', 'topics/publishBatch', null, body, callback);
 };
 
 /**
--- a/regression/pubsub.js
+++ b/regression/pubsub.js
@@ -81,7 +81,7 @@ describe('pubsub', function() {
 
     it('should publish a message', function(done) {
       pubsub.topic(topicNames[0])
-        .publish('message from me', done);
+        .publish({ data: 'message from me' }, done);
     });
 
     it('should be deleted', function(done) {
@@ -166,6 +166,7 @@ describe('pubsub', function() {
 
     it('should error when using a non-existent subscription', function(done) {
       var subscription = topic.subscription('non-existent-subscription');
+
       subscription.pull(function(err) {
         assert.equal(err.code, 404);
         done();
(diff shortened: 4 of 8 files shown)


lib/pubsub/subscription.js
```diff
@@ -296,32 +300,53 @@ Subscription.prototype.delete = function(callback) {
...
 Subscription.prototype.pull = function(options, callback) {
   var that = this;
-  // TODO(jbd): Should not be racing with other pull.
+
   if (!callback) {
     callback = options;
     options = {};
   }
+
+  var batch = util.is(options.maxCount, 'number');
```
ryanseys: If I pull without specifying a max, maybe I want all the messages?

stephenplusplus: If you don't specify `maxCount`, that's the same as pulling one message (not batch) to keep with the current API.

lib/pubsub/subscription.js
```diff
@@ -296,32 +300,53 @@ Subscription.prototype.delete = function(callback) {
...
+    var messages = response.pullResponses || [response];
+    messages = messages.map(Subscription.formatMessage_);
+
+    var messageResponse = response.pullResponses ? messages : messages[0];
+
     if (that.autoAck) {
-      that.ack(message.id, function(err) {
-        callback(err, message);
+      var ackIds = messages.map(function(message) {
+        return message.id;
```
tmatsuo: Now the message contains messageId, which is different from ackId. The messageId can be used for tracking particular messages.

Can I still access messageId?

regression/pubsub.js
```diff
@@ -198,9 +205,40 @@ describe('pubsub', function() {
         assert.equal(msg.data, 'hello');
         subscription.ack(msg.id, done);
       });
-      topic.publishRaw({
-        data: new Buffer('hello').toString('base64')
-      }, assert.ifError);
+      topic.publish([
```
tmatsuo: publishBatch returns messageIds, which can be used for tracking individual messages by subscribers. Is it possible to test whether the messageIds returned by publishBatch are the same as messageIds in pulled messages?

lib/pubsub/subscription.js
stephenplusplus: Thinking `.pull` should default to `pullBatch`, we just need to decide what to set `maxCount` to (`maxEvents` as the API calls it).

The only way a user would be able to pull a single message would be with `sub.pull({ maxCount: 1 })`

ryanseys: Yes, this is correct. We can document this as a breaking change and help users through the migration if they require. My thought is most people will very welcome the change and like that we are efficiently returning all their messages at the same time.

lib/pubsub/subscription.js
ryanseys: Looks like that functionality will need to be added. message.id should be the `messageId` and `message.ackId` be the `ackId`. Resource definition: https://cloud.google.com/pubsub/v1beta1/subscriptions/pull#response

lib/pubsub/subscription.js
```diff
@@ -280,15 +280,19 @@ Subscription.prototype.delete = function(callback) {
  * Pull messages from the subscribed topic. If messages were found, your
  * callback is executed with the message object.
  *
+ * @todo Should not be racing with other pull.
+ * @todo Fix API to return a list of messages.
```
ryanseys: Is this still a todo?

lib/pubsub/subscription.js
```diff
@@ -280,15 +280,19 @@ Subscription.prototype.delete = function(callback) {
  * Pull messages from the subscribed topic. If messages were found, your
  * callback is executed with the message object.
  *
+ * @todo Should not be racing with other pull.
```
ryanseys: We should open an issue for this with specifics of how to reproduce. I'm not sure the specifics of this TODO.

lib/pubsub/topic.js
```diff
@@ -63,6 +63,27 @@ function Topic(pubsub, options) {
...
 
 /**
+ * Format a message object as the upstream API expects it.
+ *
+ * @private
+ *
+ * @return {object}
+ */
+Topic.formatMessage_ = function(message) {
+  if (!message.data) {
```
ryanseys: Consider rewriting this block as:

``` js
if(!message.data) {
  message = { data: new Buffer(message).toString('base64') };
}
// ...
```

lib/pubsub/topic.js
```diff
@@ -87,25 +108,75 @@ Topic.formatName_ = function(projectId, name) {
...
+ *   // ...as are raw message objects.
+ *   {
+ *     label: 'registration',
+ *     data: new Buffer('new user: stephen').toString('base64')
+ *   }
+ * ];
+ *
+ * topic.publishBatch(messages, function(err) {});
+ */
+Topic.prototype.publishBatch = function(messages, callback) {
```
ryanseys: Wondering if we should make this a `_privateMethod` and just support `Topic#publish` which will accept a message or array of messages.

regression/pubsub.js
```diff
@@ -198,9 +205,40 @@ describe('pubsub', function() {
...
+        { data: new Buffer('hello').toString('base64') },
+        { data: new Buffer('hello').toString('base64') },
+        { data: new Buffer('hello').toString('base64') },
+        { data: new Buffer('hello').toString('base64') },
+        { data: new Buffer('hello').toString('base64') },
+        { data: new Buffer('hello').toString('base64') }
+      ], assert.ifError);
+    });
+
+    it.only('should receive the chosen amount of results', function(done) {
```
ryanseys: remove `.only`

regression/pubsub.js
```diff
@@ -198,9 +205,40 @@ describe('pubsub', function() {
...
+      var subscription = topic.subscription(subscriptions[0].name);
+      var opts = { returnImmediately: true, maxCount: 3 };
+
+      subscription.pull(opts, function(err, messages) {
```
ryanseys: We pull before we publish? Should this be a callback on topic.publish call further below?

ryanseys: If you want, you could set `returnImmediately` to false and then it should wait until messages are available to return.

lib/pubsub/subscription.js
stephenplusplus: I'm not either. Just moved @rakyll's old comments up to where we keep TODOs.

lib/pubsub/topic.js
stephenplusplus: :+1:

lib/pubsub/topic.js
stephenplusplus: :+1:

regression/pubsub.js
stephenplusplus: :(

regression/pubsub.js
stephenplusplus: I'll just move publish before pull.

stephenplusplus:
Please give another look.

lib/pubsub/subscription.js
```diff
@@ -278,50 +287,98 @@ Subscription.prototype.delete = function(callback) {
...
   var that = this;
-  // TODO(jbd): Should not be racing with other pull.
+
   if (!callback) {
     callback = options;
     options = {};
   }
+
+  if (!util.is(options.maxCount, 'number')) {
+    options.maxCount = 99999;
```
ryanseys: I'm confused about this. Is maxEvents a required field? I would assume it is not.

lib/pubsub/subscription.js
```diff
@@ -278,50 +287,98 @@ Subscription.prototype.delete = function(callback) {
...
+  }
+
+  var single = options.maxCount === 1;
+  var apiEndpoint = single ? 'subscriptions/pull' : 'subscriptions/pullBatch';
```
ryanseys: I still don't understand why we don't just use pullBatch for everything, seems like it would simplify this logic a bit. Is it more efficient to use `/pull` or am I missing something?

lib/pubsub/topic.js
```diff
@@ -63,6 +63,27 @@ function Topic(pubsub, options) {
...
+ */
+Topic.formatMessage_ = function(message) {
+  if (!message.data) {
+    message = {
+      data: new Buffer(message).toString('base64')
+    };
+  }
+
+  if (!util.is(message.data, 'string') && !util.is(message.data, 'buffer')) {
+    message.data = JSON.stringify(message.data);
```
ryanseys: Would this work if the data is a number?

regression/pubsub.js
```diff
@@ -174,33 +175,98 @@ describe('pubsub', function() {
...
-        subscription.ack(msg.id, done);
+        subscription.ack(msg.ackId, done);
+      });
+    });
+
+    it('should receive the chosen amount of results', function(done) {
+      var subscription = topic.subscription(subscriptions[0].name);
+      var opts = { returnImmediately: true, maxCount: 3 };
+
+      topic.publish([
```
ryanseys: I'm still convinced this call is race-y. Why isn't pull happening in the callback to publish?

ryanseys:
Overall looks great! Just a couple concerns this won't really fix the race-y tests we've been having. How's your experience running them on your machine?

Also, seems unnecessary to rename `maxEvents` to `maxCount`... maybe if we are going to rename, just rename it to `max` ? Just a thought. I wasn't sure the motivation for that change.

ryanseys:
Yeah I ran the regression tests and got the following error:

```
1) pubsub Subscription should receive the published message:
     Uncaught Error: The subscription has no more messages available (subscription="/subscriptions/project-id/sub1").
      at handleResp (/Users/ryanseys/gcloud-node/lib/common/util.js:176:14)
      at Request._callback (/Users/ryanseys/gcloud-node/lib/common/util.js:406:11)
      at Request.self.callback (/Users/ryanseys/gcloud-node/node_modules/request/request.js:344:22)
      at Request.emit (events.js:110:17)
      at Request.<anonymous> (/Users/ryanseys/gcloud-node/node_modules/request/request.js:1239:14)
      at Request.emit (events.js:129:20)
      at IncomingMessage.<anonymous> (/Users/ryanseys/gcloud-node/node_modules/request/request.js:1187:12)
      at IncomingMessage.emit (events.js:129:20)
      at _stream_readable.js:908:16
      at process._tickCallback (node.js:355:11)
```

ryanseys:
I just put the pull in the callback and it worked great!

``` js
it('should receive the published message', function(done) {

      var subscription = topic.subscription(subscriptions[0].name);

      topic.publish([
        { data: new Buffer('hello').toString('base64') },
        { data: new Buffer('hello').toString('base64') },
        { data: new Buffer('hello').toString('base64') },
        { data: new Buffer('hello').toString('base64') },
        { data: new Buffer('hello').toString('base64') },
        { data: new Buffer('hello').toString('base64') },
        { data: new Buffer('hello').toString('base64') },
        { data: new Buffer('hello').toString('base64') },
        { data: new Buffer('hello').toString('base64') },
        { data: new Buffer('hello').toString('base64') }
      ], function() {
        subscription.pull({
          returnImmediately: true,
          maxCount: 1
        }, function(err, msg) {
          assert.ifError(err);
          assert.equal(msg.data, 'hello');
          subscription.ack(msg.ackId, done);
        });
      });
    });
```

lib/pubsub/subscription.js
stephenplusplus: This is just allowing the dev to hit both API endpoints based on what they're asking for. Ask the Pub/Sub team why there are two.

lib/pubsub/topic.js
stephenplusplus: I can add a check in here for a number.

regression/pubsub.js
stephenplusplus: That's a good idea.

stephenplusplus:
> Also, seems unnecessary to rename maxEvents to maxCount... maybe if we are going to rename, just rename it to max ? Just a thought. I wasn't sure the motivation for that change.

As a user of our Pub/Sub abstraction API, I'm not thinking "I need to pull for events", I'm thinking "I need to pull for messages". Our users will likely never interface directly with the upstream API, so I don't place a lot of value on being consistent with it in places where I think we can make it more logical. RE: `max` vs `maxCount`, I'd rather use maxCount for maximum clarity. Other places in our library use `maxResults` for example.

lib/pubsub/subscription.js
stephenplusplus: I hope not! I assumed that it was. I'll test.

stephenplusplus:
```
Uncaught Error: A required argument is missing in the request: (argument="max_events").
```

stephenplusplus:
Updated the PR to move pulls to the publish callbacks.

lib/pubsub/subscription.js
ryanseys: @tmatsuo What would you suggest we do here?

lib/pubsub/subscription.js
ryanseys: D'oh!

lib/pubsub/topic.js
stephenplusplus: Going to think out loud here in case anyone can help me out with this.

This is to format a message before being published to the API. The API expects a base64 encoded string value for message contents (`message.data`). What I'm struggling with is:

- how to know if the user gave us a value already encoded,
- base64 encoding a number (`new Buffer(JSON.stringify(3))`?)
- base64 encoding an object (`new Buffer(JSON.stringify({ userId: 3 })`?)

Please let me know if I've said something silly.

ryanseys: Would be nice to have a `topic#formatMessage_` and whatever its inverse is. It seems the inverse logic is in `subscription#formatMessage_`, I believe. If those methods could be separated and tested together, that might simplify the pain we have here with understanding the different cases. Overall I think it should work because `JSON.parse(JSON.stringify(2))` is `2` but would be nice to let the tests tell us.

ryanseys: If they pass a buffer, we can just call .toString('base64') on it, otherwise, `JSON.stringify` it, wrap in a buffer and `.toString('base64')` it.

stephenplusplus: This?
```js
Topic.formatMessage_ = function(message) {
  if (!message.data) {
    message = {
      data: message
    };
  }

  if (!util.is(message, 'buffer')) {
    message.data = new Buffer(JSON.stringify(message));
  }

  message.data = message.data.toString('base64');

  return message;
};
```

ryanseys: Looks like that would work.

ryanseys:
Yeah I think consistency would be nice here. `maxResults` makes sense too.

lib/pubsub/topic.js
stephenplusplus: I'm not sure if that supports the case where a user gives a string they've already converted to base64, like:

```js
topic.publishRaw({
  data: new Buffer("hey").toString("base64")
})
```

Maybe we should just not document publishRaw as taking buffers/non-base64 enc strings?

ryanseys: Why are they converting it? Just say whatever value they pass in on a push they will get out on a pull, which isn't the case if they are pre-converting themselves.

ryanseys: I haven't met a developer who can read base64 strings so I don't think it's reasonable to talk to our API client using them.

ryanseys: What's the purpose of publishRaw? Why is it even a public method?

stephenplusplus: It's to publish a message in the API format, which in practicality means you can attach a label to a message. We could easily allow that with `.publish()`, though it would force `publish` to accept an object:

```js
.publish({
  data: "hi",
  label: ["messages"]
})
```

ryanseys: Well could accomplish with `.publish(message, [labels], callback) { } where labels is optional?

ryanseys: Yeah just keep publishRaw for the catch-all use case in the scenario that pubsub adds more optional parameters though I don't see why we need to support the case where they have a base64'd string they want to publish REALLY RAW.

lib/pubsub/subscription.js
tmatsuo: The only reason we still have pull, is not to break old code. We introduced pullBatch and publishBatch without changing the version.

So please just use only pullBatch and publishBatch.

Thanks!

ryanseys: Thanks! Is there some value we can put for maxEvents that will return them
all?

tmatsuo: Unfortunately there is no such value. The default is 100, and the max value is 1,000.

ryanseys: What do you mean by default? Stephen said that the value is required.

tmatsuo: Ah, sorry I was confused with the different thing. No default, and the max is 1,000 as far as I remember correctly.

stephenplusplus: Thanks @tmatsuo!

lib/pubsub/topic.js
stephenplusplus: > We could accomplish with `.publish(message, [labels], callback) { } where labels is optional?

:+1:

> Yeah just keep publishRaw for the catch-all use case in the scenario that pubsub adds more optional parameters though I don't see why we need to support the case where they have a base64'd string they want to publish REALLY RAW.

I would rather ditch it. Instead of just accepting labels, we can accept an object to future proof against more API options.

```js
.publish('message', [  opts: { labels: [] }  ], function(err) {})
```

ryanseys: > `.publish('message', [  { labels: [] }  ], function(err) {})`

That won't work for batch publish will it?

ryanseys: Sorry that `[ ]` threw me off... It's just meant to be optional, not an array.

So we have:
```js
.publish('message', cb); // one message, no labels

.publish('message', ['label1', 'label2'], cb); // one message with labels

.publish([ { data: 'message1' }, { data: 'message2' } ], cb); // multiple messages, no labels.

.publish([{
  data: 'message1'
  labels: ['label1', 'label2']
}, {
 data: 'message2'
 labels: ['label3', 'label4']
}], cb); // multiple messages, with labels
```

stephenplusplus: Aw man, I realized that yesterday and forgot by the time I posted.

I think that forces us to use:

```js
// multiple
topic.publish([
  { data: "anyType", labels: [] },
  { data: "anyType", labels: [] }
], callback)

// single
topic.publish({ data: "anyType", labels: [] }, callback)
```

We could use a message type?

```js
var message = topic.message("Hi!")
message.labels = ["user", "group3"]
// message.addLabel("user")
// message.addLabel("group3")

topic.publish(message, callback)
```

stephenplusplus: In general, I don't mind optional parameters (though I prefer to avoid when possible), but I really don't like changing the signature based on an argument's type.

ryanseys: > ``` js
// multiple
topic.publish([
 { data: "anyType", labels: [] },
 { data: "anyType", labels: [] }
], callback)

> // single
topic.publish({ data: "anyType", labels: [] }, callback)
```

Seems fine to me. And it's understandable. We could change `data:` to `message:` to make it a little friendlier. We do a similar thing with datastore already so I don't see a problem. And very easy to extend in the future! :)

Comment (stephenplusplus) [inline lib/pubsub/topic.js]: Woot!

**Comment by ryanseys** [comment]
It this ready for another look?

**Comment by stephenplusplus** [comment]
Not yet.

**Comment by stephenplusplus** [comment]
Okay, ready for a look :)

File: lib/common/util.js [hunk 14]
```diff
@@ -92,9 +92,14 @@ module.exports.extendGlobalConfig = extendGlobalConfig;
  * // [ 'Hi' ]
  */
 function arrayize(input) {
+  if (input == null) {
```
ryanseys: Could we just rewrite as `input === null || input === undefined` for readability and to limit type coercion?

stephenplusplus: Sure, it's just so boring!

lib/pubsub/subscription.js
```diff
@@ -278,50 +287,86 @@ Subscription.prototype.delete = function(callback) {
...
+ * var opts = {
+ *   maxResults: 1
+ * };
+ *
+ * subscription.pull(opts, function(err, messages) {});
  */
 Subscription.prototype.pull = function(options, callback) {
   var that = this;
-  // TODO(jbd): Should not be racing with other pull.
+  var MAX_EVENTS_PULLED = 1000;
```
ryanseys: MAX_EVENTS_LIMIT

stephenplusplus: Much better

lib/pubsub/topic.js
```diff
@@ -83,59 +100,55 @@ Topic.formatName_ = function(projectId, name) {
  *
  * @throws {Error} If no message is provided.
  *
- * @param {*} message - The message to publish.
+ * @param {object|object[]} message - The message(s) to publish.
+ * @param {*} message.data - The contents of the message in any format.
```
ryanseys: Clarify what "format" means. Strings, numbers, objects.

test/pubsub/subscription.js
```diff
@@ -211,6 +205,119 @@ describe('Subscription', function() {
...
+              }
+            }
+          ]
+        };
+
+        subscription.makeReq_ = function(method, path, query, body, callback) {
+          callback(null, apiResponse);
+        };
+
+        subscription.pull({ maxCount: 3 }, function(err, messages) {
```
ryanseys: maxResults, not maxCount, not that it makes a difference here anyway.

ryanseys:
Looks good to me. I'm good with you self-merging after addressing the couple issues I mentioned.

lib/pubsub/topic.js
stephenplusplus: Actually, I'll just say "The contents of the message." - saying the type is done from the `*`.

stephenplusplus:
Addressed those issues and the tests pass, so all aboard the merge town train. :train:

ryanseys:
Woot! Now let's hope this fixes the build! :)

stephenplusplus:
[Nope :(](https://travis-ci.org/GoogleCloudPlatform/gcloud-node/jobs/51419871#L1754)

PR #390, opened 2015-02-11
Make files public and private

ryanseys:
Included in this PR is my interpretation of how we should handle making Google Cloud Storage objects public and private.

I am not putting it under the `file.acl` object because if the user understand acls then they probably don't need to use these helper methods. These are slap-em-on permissions that require minimal configuration.

Buckets, by default, apply a default ACL of `project-private`, that is it is private only to the maintainers of the project. This is the "default" setting. By running `file.makePrivate()` we are reverting to this default. If you want to REALLY make it private, that is only to the object owner user, then you can pass in `{ strict: true }` (ideas on a better param name are welcome). `makePrivate()` removes all other permissions. It DNGAF.

`makePublic()` just slaps on `allUsers: READER` permission, maintaining all other permissions. It's nicer than `makePrivate()` in this regard.

Edit: Added tests.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -798,6 +798,73 @@ File.prototype.setMetadata = function(metadata, callback) {
 };
 
 /**
+ * Make a file private to the project and remove all other permissions.
+ * Set `options.strict` to true to make the file private to only the owner.
+ *
+ * @param {object=} options - The configuration object.
+ * @param {boolean=} options.strict - If true, set the file to be private to
+ *     only the owner user. Otherwise, it will be private to the project.
+ * @param {function=} callback - The callback function.
+ *
+ * @example
+ *
+ * //-
+ * // Set the file private so only project maintainers can see and modify it.
+ * //-
+ * file.makePrivate(function(err) {});
+ *
+ * //-
+ * // Set the file private so only the owner can see and modify it.
+ * //-
+ * file.makePrivate({ strict: true }, function(err) {});
+ */
+File.prototype.makePrivate = function(options, callback) {
+  var that = this;
+  if (util.is(options, 'function')) {
+      callback = options;
+      options = {};
+  }
+  var path = '/o/' + encodeURIComponent(this.name);
+  var query = { predefinedAcl: options.strict ? 'private' : 'projectPrivate' };
+
+  // You aren't allowed to set both predefinedAcl & acl properties on a file, so
+  // acl must explicitly be nullified, destroying all previous acls on the file.
+  var metadata = { acl: null };
+
+  callback = callback || util.noop;
+
+  this.makeReq_('PATCH', path, query, metadata, function(err, resp) {
+    if (err) {
+      callback(err);
+      return;
+    }
+
+    that.metadata = resp;
+
+    callback(null);
+  });
+};
+
+/**
+ * Set a file to be publicly readable and maintain all previous permissions.
+ *
+ * @param {function=} callback - The callback function.
+ *
+ * @example
+ * file.makePublic(function(err) {});
+ */
+File.prototype.makePublic = function(callback) {
+  callback = callback || util.noop;
+
+  this.acl.add({
+    entity: 'allUsers',
+    role: 'READER'
+  }, function(err) {
+    callback(err);
+  });
+};
+
+/**
  * `startResumableUpload_` uses the Resumable Upload API: http://goo.gl/jb0e9D.
  *
  * The process involves these steps:
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -1025,6 +1025,59 @@ describe('File', function() {
     });
   });
 
+  describe('makePublic', function() {
+    it('should execute callback', function(done) {
+      file.acl.add = function(options, callback) {
+        callback();
+      };
+
+      file.makePublic(done);
+    });
+
+    it('should make the file public', function(done) {
+      file.acl.add = function(options) {
+        assert.deepEqual(options, { entity: 'allUsers', role: 'READER' });
+        done();
+      };
+
+      file.makePublic(util.noop);
+    });
+  });
+
+  describe('makePrivate', function() {
+    it('should execute callback', function(done) {
+      file.makeReq_ = function(method, path, query, body, callback) {
+        callback();
+      };
+
+      file.makePrivate(done);
+    });
+
+    it('should make the file private to project by default', function(done) {
+      file.makeReq_ = function(method, path, query, body) {
+        assert.equal(method, 'PATCH');
+        assert.equal(path, '/o/' + encodeURIComponent(file.name));
+        assert.deepEqual(query, { predefinedAcl: 'projectPrivate' });
+        assert.deepEqual(body, { acl: null });
+        done();
+      };
+
+      file.makePrivate(util.noop);
+    });
+
+    it('should make the file private to user if strict = true', function(done) {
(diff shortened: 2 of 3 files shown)


stephenplusplus:
All sounds good to me.

ryanseys:
Cool beans, I'll add some tests to this bad boy. Might leave makePublic and makePrivate for buckets out for now as a way to break the problem down.

test/storage/file.js
```diff
@@ -1025,6 +1025,22 @@ describe('File', function() {
     });
   });
 
+  describe('makePublic', function() {
+    it('should make the file public', function(done) {
+
+      // If this mocked method (THAT DOES NOTHING!) is removed, nothing works.
+      file.acl.makeAclObject_ = function() {};
```
ryanseys: @stephenplusplus Maybe you can explain this. If I take this line out, this test does not pass. So much is being mocked, it's hard to follow where things are getting overwritten and why this is required or what it's doing!

stephenplusplus: Quick answer: You should be testing that when you call `file.makePublic` it calls `this.acl.add` with the correct arguments.

Philosophical explanation: It's not the business of this test to be concerned with how `this.acl.add` does its job - we have tests for that. That's why you have to mock `file.acl.makeAclObject_`, because `this.acl.add` is using that method to do its job. And again, our test of `makePublic` only cares that `this.acl.add` is getting what it needs, and if we trust our acl test suite, then we don't need to question it from way up here.

ryanseys: Is this a side-effect of using mockery? I had a feeling I should have been doing that instead. I'll update it to do that.

stephenplusplus: This is just how I see effective unit testing working for our library, not specific to this situation. `file.acl.makeAclObject_` is being called because when you call `file.makePublic`, it in turn calls `file.acl.add`, which calls `file.acl.makeAclObject_`. Instead of letting it fall through all those places, we should interrupt it on its way to `file.acl.add`.

ryanseys: That makes sense. Just doesn't make sense that I added an empty function and it still worked.

ryanseys:
This should be good for a second look. I've added tests. I've omitted a test for testing makePrivate with `strict: true` because I only have project-level access, not user access, so the file would be completely inaccessible to me after making it private. I do test makePrivate on a project level though.

lib/storage/file.js
```diff
@@ -798,6 +798,68 @@ File.prototype.setMetadata = function(metadata, callback) {
...
+ * file.makePrivate({ strict: true }, function(err) {});
+ */
+File.prototype.makePrivate = function(options, callback) {
+  var that = this;
+  if(util.is(options, 'function')) {
+      callback = options;
+      options = {};
+  }
+  var path = '/o/' + encodeURIComponent(this.name);
+  var query = { predefinedAcl: (options.strict ? '' : 'project') + 'private' };
```
stephenplusplus: In [the docs](https://cloud.google.com/storage/docs/json_api/v1/objects/patch), it asks for `projectPrivate` - does the casing matter?

lib/storage/file.js
```diff
@@ -798,6 +798,68 @@ File.prototype.setMetadata = function(metadata, callback) {
...
+ * file.makePrivate(function(err) {});
+ *
+ * //-
+ * // Set the file private so only owner can see and modify it.
+ * //-
+ * file.makePrivate({ strict: true }, function(err) {});
+ */
+File.prototype.makePrivate = function(options, callback) {
+  var that = this;
+  if(util.is(options, 'function')) {
```
stephenplusplus: `if (`

lib/storage/file.js
```diff
@@ -798,6 +798,68 @@ File.prototype.setMetadata = function(metadata, callback) {
...
+      callback = options;
+      options = {};
+  }
+  var path = '/o/' + encodeURIComponent(this.name);
+  var query = { predefinedAcl: (options.strict ? '' : 'project') + 'private' };
+  var metadata = { acl: null };
```
stephenplusplus: Can you leave a quick comment in the code explaining why we're doing this?

lib/storage/file.js
```diff
@@ -798,6 +798,68 @@ File.prototype.setMetadata = function(metadata, callback) {
...
+
+/**
+ * Set a file to be publicly readable and maintains all previous permissions.
+ *
+ * @param {function=} callback - The callback function.
+ *
+ * @example
+ * file.makePublic(function(err) {});
+ */
+File.prototype.makePublic = function(callback) {
```
stephenplusplus: We need a `callback = callback || util.noop`

lib/storage/file.js
```diff
@@ -798,6 +798,68 @@ File.prototype.setMetadata = function(metadata, callback) {
 };
 
 /**
+ * Set a file to be private to the project and removes all other permissions.
+ * If you set `options.strict` to true, it will set it to be private to the
+ * owner user only.
```
stephenplusplus: Hope you don't mind these small change suggestions:

```diff
- Set a file to be private to the project and removes all other permissions.
+ Make a file private to the project and remove all other permissions.

- If you set `options.strict` to true, it will set it to be private to the owner user only.
+ Set `options.strict` to true to make the file private to only the owner.
```

lib/storage/file.js
```diff
@@ -798,6 +798,68 @@ File.prototype.setMetadata = function(metadata, callback) {
...
+ *
+ * @example
+ *
+ * //-
+ * // Set the file private so only project maintainers can see and modify it.
+ * //-
+ * file.makePrivate(function(err) {});
+ *
+ * //-
+ * // Set the file private so only owner can see and modify it.
```
stephenplusplus: `only the owner`

lib/storage/file.js
```diff
@@ -798,6 +798,68 @@ File.prototype.setMetadata = function(metadata, callback) {
...
+    }
+
+    that.metadata = resp;
+
+    callback(null);
+  });
+};
+
+/**
+ * Set a file to be publicly readable and maintains all previous permissions.
```
stephenplusplus: `maintain`

lib/storage/file.js
ryanseys: Nope doesn't matter, I'll update though for clarity.

lib/storage/file.js
ryanseys: Done.

ryanseys:
PTAL.

lib/storage/file.js
```diff
@@ -798,6 +798,73 @@ File.prototype.setMetadata = function(metadata, callback) {
...
+ * //-
+ * file.makePrivate(function(err) {});
+ *
+ * //-
+ * // Set the file private so only the owner can see and modify it.
+ * //-
+ * file.makePrivate({ strict: true }, function(err) {});
+ */
+File.prototype.makePrivate = function(options, callback) {
+  var that = this;
```
stephenplusplus: We're coming for you, `that`!!!

stephenplusplus:
LGTM but I'll leave it unmerged for now to confirm the way it works is what @jgeewax had in mind.

stephenplusplus:
Yolo merge?

ryanseys:
YEP

stephenplusplus:
Woooooo!

jgeewax:
can i just say for a minute how great this thread is? @ryanseys @stephenplusplus

ryanseys:
Glad you like! :+1: :dancer:

PR #393, opened 2015-02-16
datastore: add declarative methods - fixes #392

### LINKED ISSUE #392: Add declarative datastore methods
RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/387#issuecomment-73526494

Add `dataset.upsert()`, `dataset.insert()`, and `dataset.update()`.

stephenplusplus:
Adds:
- `dataset.insert()`
- `dataset.update()`
- `dataset.upsert()`

### DIFF
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -151,9 +151,8 @@ DatastoreRequest.prototype.get = function(keys, callback) {
  * Maps to {module:datastore/dataset#save}, forcing the method to be `insert`.
  */
 DatastoreRequest.prototype.insert = function(entities, callback) {
-  var isArray = util.is(entities, 'array');
   entities = util.arrayize(entities).map(util.propAssign('method', 'insert'));
-  this.save(isArray ? entities : entities[0], callback);
+  this.save(entities, callback);
 };
 
 /**
@@ -252,8 +251,7 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
  * ], function(err) {});
  */
 DatastoreRequest.prototype.save = function(entities, callback) {
-  var isMultipleRequest = Array.isArray(entities);
-  entities = isMultipleRequest ? entities : [entities];
+  entities = util.arrayize(entities);
 
   var insertIndexes = [];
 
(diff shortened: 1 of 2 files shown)


lib/datastore/request.js
```diff
@@ -148,6 +148,15 @@ DatastoreRequest.prototype.get = function(keys, callback) {
 };
 
 /**
+ * Maps to {module:datastore/dataset#save}, forcing the method to be `insert`.
+ */
+DatastoreRequest.prototype.insert = function(entities, callback) {
+  var isArray = util.is(entities, 'array');
+  entities = util.arrayize(entities).map(util.propAssign('method', 'insert'));
+  this.save(isArray ? entities : entities[0], callback);
```
ryanseys: We shouldn't need any of this logic, it gets put into an array in `save` anyway. Snippet:

``` js
DatastoreRequest.prototype.save = function(entities, callback) {
  var isMultipleRequest = Array.isArray(entities);
  entities = isMultipleRequest ? entities : [entities];
  // ...
```
Actually, while we are here, mind using `util.arrayize()` in `save()` instead of those two lines?

stephenplusplus: `isMultipleRequest` is used to determine if the user should get one value returned in the callback or an array.

e.g. `function callback(err, result) {}` vs `function callback(err, results) {}`

ryanseys: I see that being used in `get` and `delete` but not in `save`? Could you point me to the logic?

stephenplusplus: Oh, my bad. I'll remove pretty much 99% of this PR :)

stephenplusplus:
Updated to remove the array back and forthing.

lib/datastore/request.js
```diff
@@ -148,6 +148,14 @@ DatastoreRequest.prototype.get = function(keys, callback) {
 };
 
 /**
+ * Maps to {module:datastore/dataset#save}, forcing the method to be `insert`.
+ */
+DatastoreRequest.prototype.insert = function(entities, callback) {
+  entities = util.arrayize(entities).map(util.propAssign('method', 'insert'));
```
ryanseys: :heart_eyes:

PR #394, opened 2015-02-16
add last built timestamp to master docs - fixes #385

### LINKED ISSUE #385: Show the last-updated time in the generated gh-pages docs?
I know we have the version history, but it'd be really cool if we could see the date that the docs were last updated. Any chance we can put that into the travis deployment build?

stephenplusplus:
Live: http://stephenplusplus.github.io/gcloud-node/#/docs/master


docs/site/components/docs/docs.js
```diff
@@ -322,6 +322,13 @@ angular
       return doc.toLowerCase() === $routeParams.module;
     };
 
+    if ($routeParams.version === 'master') {
+      $http.get('https://api.github.com/repos/GoogleCloudPlatform/gcloud-node/commits?sha=gh-pages&per_page=1')
```
stephenplusplus: *hattip* @ryanseys

ryanseys:
Code looks great! Only problem I could see was I looked at the live example first and could not find the date. Whipped up an alternative that might separate itself from the regular content a little better.

screen shot 2015-02-16 at 12 42 57 pm

docs/site/components/docs/docs.js
ryanseys: yay for :octocat:!

stephenplusplus:
Would you want to do a separate PR for that after this lands? Only asking because I've had quite some trouble with the CSS putting text there in the past.

ryanseys:
Absolutely.

PR #398, opened 2015-02-17
Add dataset.save example to README

ryanseys:
Ripped this from the tests! It's a good example using lots of different types :)


stephenplusplus:
:pizza: :+1:

PR #402, opened 2015-02-19
fix regression tests.

stephenplusplus:
Our merge of #388 didn't end up helping our regression tests. Well, it may have, but it seems a new one has popped up.

When I run locally, I don't get the error, but Travis does. It gets an error in one of the many functions being called in a `before` hook. So, I started looking deeper into what the troubled hook does, and questioned what was necessary so that it could be uncomplicated.

This PR includes my findings.

### DIFF
--- a/lib/pubsub/topic.js
+++ b/lib/pubsub/topic.js
@@ -152,7 +152,7 @@ Topic.prototype.publish = function(messages, callback) {
 };
 
 /**
- * Delete the topic.
+ * Delete the topic. This will also delete any subscriptions to this topic.
  *
  * @param {function=} callback - The callback function.
  *


regression/pubsub.js
```diff
@@ -28,8 +29,20 @@ var Subscription = require('../lib/pubsub/subscription.js');
 
 var pubsub = gcloud.pubsub();
 
+function generateTopicName() {
+  return 'test-topic-' + uuid.v4();
+}
+
+function generateSubName() {
+  return 'test-subscription-' + uuid.v4();
+}
```
stephenplusplus: Instead of using names like `topic1` and `topic2`, I made quick helper methods to generate unique names.

regression/pubsub.js
```diff
@@ -91,45 +104,21 @@ describe('pubsub', function() {
...
-      pubsub.getTopics(function(err, topics) {
-        if (err) {
-          callback(err);
-          return;
-        }
-        async.parallel(topics.map(function(topic) {
-          return topic.delete.bind(topic);
-        }), callback);
-      });
-    }
```
stephenplusplus: Turns out this function was already defined earlier in the script.

regression/pubsub.js
```diff
@@ -91,45 +104,21 @@ describe('pubsub', function() {
...
-      pubsub.getSubscriptions(function(err, subs) {
-        if (err) {
-          callback(err);
-          return;
-        }
-        async.parallel(subs.map(function(sub) {
-          return sub.delete.bind(sub);
-        }), callback);
-      });
-    }
```
stephenplusplus: According to the API docs, deleting a topic deletes all subscriptions to it, so this isn't necessary.

regression/pubsub.js
```diff
@@ -91,45 +104,21 @@ describe('pubsub', function() {
...
-
     before(function(done) {
-      async.parallel([deleteAllTopics, deleteAllSubscriptions], function(err) {
+      deleteAllTopics(function(err) {
```
stephenplusplus: This is the troubled `before` hook. It used to delete topics and subscriptions in parallel, which doesn't seem too smart in retrospect. Especially since I know now that deleting a topic will delete any subscriptions to it.

regression/storage.js
```diff
@@ -370,7 +370,7 @@ describe('storage', function() {
 
         file.download(function(err, remoteContents) {
           assert.ifError(err);
-          assert.equal(fileContents, remoteContents);
+          assert.equal(String(fileContents), String(remoteContents));
```
stephenplusplus: This popped up for me. `fileContents` and `remoteContents` were buffers, which are problematic to compare. We do have a utility library we could use, but it's just as easy to convert to Strings to compare.

lib/pubsub/topic.js
```diff
@@ -190,7 +190,13 @@ Topic.prototype.delete = function(callback) {
...
  */
 Topic.prototype.getSubscriptions = function(query, callback) {
+  if (util.is(query, 'function')) {
+    callback = query;
+    query = {};
+  }
+
   query.query = 'pubsub.googleapis.com/topic in (' + this.name + ')';
+
   this.pubsub.getSubscriptions(query, callback);
```
stephenplusplus: Before this change, it was impossible to get only subscriptions registered to a topic if you did not provide a `query` to `topic.getSubscriptions`.

ryanseys:
Yay! Can you write a comment [here](https://github.com/GoogleCloudPlatform/gcloud-node/blob/c5ecbf7de62511a9d00d1c713465c60911b6a723/lib/pubsub/topic.js#L155) so that developers know that deleting a topic will destroy all subscriptions for that topic?

stephenplusplus:
:+1:

lib/pubsub/topic.js
```diff
@@ -152,7 +152,7 @@ Topic.prototype.publish = function(messages, callback) {
 };
 
 /**
- * Delete the topic.
+ * Delete the topic. This will also delete any subscriptions to this topic.
```
stephenplusplus: @ryanseys :dancers:

ryanseys:
LGTM! :steam_locomotive: Here goes!

PR #405, opened 2015-02-20
Use unique pubsub subscriptions and topics in reg tests

ryanseys:
This PR changes the pubsub regression tests so they only use unique topic and subscription names so names do not conflict between parallel-running Travis instances (largely the reason for our existing issues with these tests). Also they only delete the topics and subscriptions they create (and they should delete every single one they create).

PS: More deletions than additions! :dancer:


stephenplusplus:
Sweet!

PR #407, opened 2015-02-23
support local datastore

### LINKED ISSUE #389: Support datastore on local development server
https://cloud.google.com/datastore/docs/tools/devserver

stephenplusplus:
Fixes #389 

While running the gcd server, this now works:

``` js
// gcd is running on localhost:8080

var gcloud = require("gcloud")

var ds = gcloud.datastore.dataset({
  apiEndpoint: "localhost:8080",
  projectId: "nth-circlet-705"
})

var company = ds.key("company")

ds.save({
  key: company,
  data: { name: "google" }
}, function (err) {
  if (err) throw err

  ds.get(company, console.log)
})
```

I'm not sure `host` and `port` are technically correct argument names, since for this PR, "host" needs to be in the format of protocol + host ("http://localhost"). Ideas I can think of:
- Ditch `host` and `port` in favor of one argument name to include protocol + host + port (but not a path):
  - `uri`
  - `resource`
  - `api`
  - `apiPrefix`
- Accept `protocol`, `host`, and `port`
- Not worry about it

Interested in your thoughts!

### DIFF
--- a/lib/datastore/dataset.js
+++ b/lib/datastore/dataset.js
@@ -80,13 +80,21 @@ var SCOPES = [
  *     a `keyFilename`.
  * @param {string=} options.apiEndpoint - Override the default API endpoint used
  *     to reach Datastore. This is useful for connecting to your local Datastore
- *     server.
+ *     server (usually "http://localhost:8080").
  * @param {string} options.namespace - Namespace to isolate transactions to.
  *
  * @example
  * var dataset = datastore.dataset({
  *   projectId: 'my-project',
  *   keyFilename: '/path/to/keyfile.json'
+ * });
+ *
+ * //-
+ * // Connect to your local Datastore server.
+ * //-
+ * var dataset = datastore.dataset({
+ *   projectId: 'my-project',
+ *   apiEndpoint: 'http://localhost:8080'
  * });
  */
 function Dataset(options) {


lib/common/util.js
```diff
@@ -380,6 +380,14 @@ function makeAuthorizedRequest(config) {
 
   var authorize = gsa(config);
 
+  if (config.customApi) {
```
ryanseys: put the `var authorize = gsa(config);` at L381 in the else block of this if. Yes, that means declaring `var authorize;` but also means the call to `gsa` can't mess things up because its called every time unnecessarily

ryanseys:
Re: `protocol`, `host` and `port` why not just accept a `customEndpoint` value that can be set to something like `"http://localhost:8080"` and our default will be `"https://www.googleapis.com"` where the port of `443` is implied. Right now if they just specify a host, the default port is assumed 443 which won't work for them if they are on http.

test/datastore/request.js
```diff
@@ -58,10 +53,12 @@ describe('Request', function() {
   var Request;
   var key;
   var request;
+  var HOST = 'http://localhost';
+  var PORT = 8080;
 
   before(function() {
     mockery.registerMock('./pb.js', pb);
-    mockery.registerMock('https', https);
```
ryanseys: YAY!

ryanseys:
Looks good to me otherwise. Let me know what you think about just one string for configuration e.g. "http://localhost:8080" instead of two or three.

lib/common/util.js
stephenplusplus: :+1:

stephenplusplus:
> Re: protocol, host and port why not just accept a customEndpoint value that can be set to something like "http://localhost:8080" and our default will be "https://www.googleapis.com" where the port of 443 is implied.

SGTM. :+1:

stephenplusplus:
Sent in an update using `apiEndpoint`. Also, instead of throwing if a user doesn't include a protocol, it will default to `http://`. Let me know if you catch anything I overlooked.

test/datastore/dataset.js
```diff
@@ -23,6 +23,33 @@ var Dataset = require('../../lib/datastore/dataset');
...
+
+    it('should remove slashes from the apiEndpoint', function() {
+      var ds1 = new Dataset({ apiEndpoint: 'http://localhost:8080' });
+      var ds2 = new Dataset({ apiEndpoint: 'http://localhost:8080/' });
+      var ds3 = new Dataset({ apiEndpoint: 'http://localhost:8080//' });
+
+      assert.equal(ds1.apiEndpoint, 'http://localhost:8080');
+      assert.equal(ds2.apiEndpoint, 'http://localhost:8080');
+      assert.equal(ds3.apiEndpoint, 'http://localhost:8080');
+    });
```
stephenplusplus: New behavior.

lib/datastore/dataset.js
```diff
@@ -78,6 +78,9 @@ var SCOPES = [
  *     `credentials` object.
  * @param {object=} options.credentials - Credentials object, used in place of
  *     a `keyFilename`.
+ * @param {string=} options.apiEndpoint - Override the default API endpoint used
+ *     to reach Datastore. This is useful for connecting to your local Datastore
+ *     server.
```
ryanseys: Put an example here, preferably with the default port of whatever the local datastore runs on out of the box.

lib/datastore/dataset.js
```diff
@@ -78,6 +78,9 @@ var SCOPES = [
...
  * @param {string} options.namespace - Namespace to isolate transactions to.
  *
  * @example
```
ryanseys: Add an example of using the custom datastore.

```js
var dataset = datastore.dataset({
  projectId: 'my-project',
  apiEndpoint: process.env.DATASTORE_HOST // 'http://localhost:8080'
});
```

test/datastore/dataset.js
ryanseys: :+1:

ryanseys:
Cool beans. Looks good. Just could benefit from a couple example as I commented.

stephenplusplus:
Good idea. Added those: https://github.com/GoogleCloudPlatform/gcloud-node/commit/78358d426cda42813232012b612137b9f8f66adb

ryanseys:
Thanks! :dancer:

PR #408, opened 2015-02-23
storage: support file.compose - fixes #401

### LINKED ISSUE #401: Support Objects: compose as myBucket.combine
Here's a suggested usage:

``` js
var myBucket = storage.bucket('my-server-logs');
var log1 = myBucket.file('server-log-1');
var log2 = myBucket.file('server-log-2');

myBucket.combine({
  files: [log1, log2], // must all be in myBucket
  destination: 'combined-log-1-2' // new File in myBucket
}, function(err, newFile) {
  console.log(err, newFile);
});

### PR DESCRIPTION
**Pull request by stephenplusplus**
Fixes: #401
Refs: #375 

``` js
var 2013logs = bucket.file('2013-logs.txt');
var 2014logs = bucket.file('2014-logs.txt');

var allLogs = bucket.file('all-logs.txt');

bucket.combine([
  2013logs,
  2014logs
], allLogs, function(err, newFile) {
  // newFile === allLogs
});
```

### DIFF
--- a/test/storage/bucket.js
+++ b/test/storage/bucket.js
@@ -135,7 +135,7 @@ describe('Bucket', function() {
         bucket.file('destination.txt')
       ];
 
-      async.each(destinations, function (destination, next) {
+      async.each(destinations, function(destination, next) {
         bucket.storage.makeAuthorizedRequest_ = function(reqOpts) {
           assert(reqOpts.uri.indexOf(bucket.name + '/o/destination.txt') > -1);
           next();


lib/common/util.js
```diff
@@ -209,6 +209,52 @@ function getType(value) {
...
+ */
+function prop(name) {
+  return function(item) {
+    if (name in item) {
+      return item[name];
+    }
+  };
+}
+
+module.exports.prop = prop;
```
stephenplusplus: I've wanted this in our library many times in the past. I figured I'd add it in sooner than later.

lib/common/util.js
```diff
@@ -209,6 +209,52 @@ function getType(value) {
...
+ *
+ * @param {string} name - Name of the property to return.
+ * @return {function}
+ *
+ * @example
+ * var people = [
+ *   {
+ *     name: 'Stephen',
+ *     origin: 'USA',
+ *     beenToNYC: false
```
ryanseys: You definitely gotta go!

lib/common/util.js
```diff
@@ -209,6 +209,52 @@ function getType(value) {
...
+ * var peopleInUSA = people.filter(prop('beenToNYC'));
+ * // [
+ * //   {
+ * //     name: 'Ryan',
+ * //     origin: 'Canada',
+ * //     beenToNYC: true
+ * //   }
+ * // ]
+ */
+function prop(name) {
```
ryanseys: :+1:

lib/storage/bucket.js
```diff
@@ -203,6 +203,100 @@ function Bucket(storage, name) {
...
+Bucket.prototype.combine = function(sources, destination, callback) {
+  if (!util.is(sources, 'array') || sources.length < 2) {
+    throw new Error('You must provide at least two source files.');
+  }
+
+  if (!destination) {
+    throw new Error('A destination file must be specified.');
+  }
+
+  var self = this;
```
ryanseys: omgomg thank you!! I hate `var that = this;`

lib/storage/bucket.js
```diff
@@ -203,6 +203,100 @@ function Bucket(storage, name) {
...
+ * var allLogs = bucket.file('all-logs.txt');
+ *
+ * bucket.combine([
+ *   2013logs,
+ *   2014logs
+ * ], allLogs, function(err, newFile) {
+ *   // newFile === allLogs
+ * });
+ */
+Bucket.prototype.combine = function(sources, destination, callback) {
```
ryanseys: No way to specify optional parameters like `destinationPredefinedAcl` or `ifGenerationMatch` etc?

lib/storage/bucket.js
```diff
@@ -203,6 +203,100 @@ function Bucket(storage, name) {
...
+  destination = convertToFile(destination);
+  callback = callback || util.noop;
+
+  if (!destination.metadata.contentType) {
+    var destinationContentType = mime.contentType(destination.name);
+
+    if (destinationContentType) {
+      destination.metadata.contentType = destinationContentType;
+    } else {
+      throw new Error(
```
ryanseys: Seems a little violent to throw because of this, no? Given that the content type is apparently optional to begin with?

regression/storage.js
```diff
@@ -25,8 +25,11 @@ var fs = require('fs');
 var request = require('request');
 var through = require('through2');
 var tmp = require('tmp');
+var util = require('../lib/common/util');
 var uuid = require('node-uuid');
 
+var prop = util.prop;
```
ryanseys: Not necessary, just use `util.prop`

test/storage/bucket.js
```diff
@@ -93,6 +95,189 @@ describe('Bucket', function() {
...
+      bucket.combine([file1, file2], 'destination.txt');
+    });
+
+    it('should accept string or file input for destination', function(done) {
+      var destinations = [
+        'destination.txt',
+        bucket.file('destination.txt')
+      ];
+
+      async.each(destinations, function (destination, next) {
```
ryanseys: Extra space between `function (`

ryanseys:
This function is very trigger happy when it comes to throwing which is fine I think, in fact, I can think of a couple more scenarios where it may like to throw:

When destination is given as a `File` but it's for a different bucket. And when sources come from a different bucket then the one being called on... but I'm not 100% sure we need to worry about that, we can always just default to the bucket the method is called on. Just becomes awkward when that file doesn't exist in that bucket or they think they can do cross-bucket composition.

lib/storage/bucket.js
stephenplusplus: Woops! This was actually a mistake. I hate `that = this`, too, but would have done it to match the codebase. I'll be so happy to keep this and work towards removing other that's.

lib/storage/bucket.js
stephenplusplus: I was thinking not. Another most common use cases > being an API passthrough case, which matches other method we have in the library (which don't allow complete access to the underlying API call). I think if we want to be 1:1 with the API, that'll require a new issue for discussion and careful examination of every single method we have to make sure it offers it.

lib/storage/bucket.js
stephenplusplus: It is. Without that, the API returns an unhelpful "Required argument missing." Took a long time of playing around to figure out what was missing. Turns out it's content type.

stephenplusplus:
> This function is very trigger happy when it comes to throwing

Only throws in situations where we don't have the minimum required data to let the API call have a chance to succeed.

> When destination is given as a File but it's for a different bucket.
> And when sources come from a different bucket then the one being called on.

Yeah, I thought of this as well, but avoided as I *think* the API error should be helpful in that regard. Can implement though if anyone thinks it's important.

ryanseys:
I agree, delegate to the API here.

lib/storage/bucket.js
ryanseys: ok sounds good

lib/storage/bucket.js
ryanseys: Yeah just leave it and label the switch to `self` as a sort of technical debt.

PR #409, opened 2015-02-24
storage: support specifying a generation

### LINKED ISSUE #406: Support generation and other metadata in Bucket#file
Would be great to be able to specify the `generation` of a file like so:

```js
myBucket.file('my-logs.txt', { generation: 'some older generation id' });
```

Should be pretty straight forward to support this.

### LINKED ISSUE #498: Object versioning support
I don't see anything about object versioning on the docs. Is there support for listing/retrieving the revision history of a file?

stephenplusplus:
Fixes #406
Fixes #498
- [x] Docs
- [x] Unit tests
  - [x] Bucket
  - [x] File
- [x] System tests
- [x] Needs to work

This is a first attempt at allowing a user to lock a File instance to a generation:

``` js
var gcloud = require("gcloud")({ projectId: "my-project" })
var storage = gcloud.storage()

var myFile = storage.file("yay.txt", { generation: 234234324 })
// use myFile as normally, however all operations are locked to that generation
```
## 
##### Resolved :+1:

**https://github.com/GoogleCloudPlatform/gcloud-node/pull/409#issuecomment-75832802**

How should normal operations on a File be changed to match a file locked to a generation?

These are pretty straightforward:
- `copy()` - specify `sourceGeneration` in the query
- `createReadStream()` / `download()` - specify `generation` in the query
- `delete()` - specify `generation` in the query
- `getMetadata()` - specify `generation` in the query
- `setMetadata()` - specify `generation` in the query

However, when you update the file with `createWriteStream()`, it uses `ifGenerationMatch` in the query to mean: only write to this file if the current generation on the remote file is the same as the generation specified for the File instance. So, if you're 3 generations back trying to update the file, it wouldn't work.
- should we use `ifGenerationMatch` in this way or should we just let the user write to the file regardless?
- should we update the generation lock on the File instance to the new generation?
## 
##### Resolved :+1:

**Bucket must be created with `{versioning: { enabled: true }}`.**

This has also been difficult to test with the real API. Here is a script I wrote to attempt to create a file, update it, then access its first state:

``` js
var async = require("async")
var gcloud = require("gcloud")({
  keyFilename: "/Users/stephen/dev/keyfile.json",
  projectId: "nth-circlet-705"
})

var storage = gcloud.storage()
var bucket = storage.bucket("stephens-new-bucket")

var fileExists = false
var generations = []

var fileName = Date.now()
var file = bucket.file(fileName)

async.eachSeries("ab".split(""), upload, log)

function upload(appendData, callback) {
  var writeStream = file.createWriteStream()
  writeStream.on("error", callback)
  writeStream.on("complete", function () {
    file.getMetadata(function (err, metadata) {
      if (err) return callback(err)

      console.log(appendData, "=>", metadata.generation)

      // store the generation of the file after writing to it
      generations.push(metadata.generation)

      // continue to the next append
      setTimeout(callback, 60000 * 20) // :-o - 20 minutes
    })
  })

  if (!fileExists) {
    // create the file
    writeStream.end(appendData)
  } else {
    // 1. get the current contents of the file
    // 2. append data to it
    // 3. write it to the remote file
    file.download(function (err, data) {
      if (err) return callback(err)

      writeStream.end(String(data) + appendData)
    })
  }
}

function log(err) {
  if (err) throw err

  async.eachSeries(generations, function (generation, callback) {
    console.log("downloading generation " + generation + "...")

    bucket.file(fileName, { generation: generation }).download(function (err, contents) {
      if (err) return callback(err)
      console.log(String(contents))
      callback()
    })
  }, console.log)
}
```

But that results in:

``` sh
> node ./
a => 1424723174921000
b => 1424724376222000
downloading generation 1424723174921000...
{ [Error: Not Found]
  errors: [ { domain: 'global', reason: 'notFound', message: 'Not Found' } ],
  code: 404,
  message: 'Not Found',
  response: undefined }
```

I can use `bucket.file(fileName, { generation: /* b's value */ })` fine, but I can't get a's. I'm inclined to believe this is me not understanding how to use generations, and not a bug in the code.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -72,7 +72,7 @@ function File(bucket, name, options) {
   options = options || {};
 
   this.bucket = bucket;
-  this.explicitGeneration = parseInt(options.generation, 10);
+  this.generation = parseInt(options.generation, 10);
   this.makeReq_ = bucket.makeReq_.bind(bucket);
   this.metadata = {};
 
@@ -216,8 +216,8 @@ File.prototype.copy = function(destination, callback) {
 
   var query = {};
 
-  if (this.explicitGeneration) {
-    query.sourceGeneration = this.explicitGeneration;
+  if (this.generation) {
+    query.sourceGeneration = this.generation;
   }
 
   this.makeReq_('POST', path, query, null, function(err, resp) {
@@ -343,9 +343,9 @@ File.prototype.createReadStream = function(options) {
       uri: uri
     };
 
-    if (that.explicitGeneration) {
+    if (that.generation) {
       reqOpts.qs = {
-        generation: that.explicitGeneration
+        generation: that.generation
       };
     }
 
@@ -665,8 +665,8 @@ File.prototype.delete = function(callback) {
 
   var query = {};
 
-  if (this.explicitGeneration) {
-    query.generation = this.explicitGeneration;
+  if (this.generation) {
+    query.generation = this.generation;
   }
 
   this.makeReq_('DELETE', path, query, null, function(err, resp) {
@@ -751,8 +751,8 @@ File.prototype.getMetadata = function(callback) {
 
   var query = {};
 
-  if (this.explicitGeneration) {
-    query.generation = this.explicitGeneration;
+  if (this.generation) {
+    query.generation = this.generation;
   }
 
   this.makeReq_('GET', path, query, null, function(err, resp) {
@@ -856,8 +856,8 @@ File.prototype.setMetadata = function(metadata, callback) {
   var path = '/o/' + encodeURIComponent(this.name);
   var query = {};
 
-  if (this.explicitGeneration) {
-    query.generation = this.explicitGeneration;
+  if (this.generation) {
+    query.generation = this.generation;
   }
 
   this.makeReq_('PATCH', path, query, metadata, function(err, resp) {
@@ -1020,8 +1020,8 @@ File.prototype.startResumableUpload_ = function(stream, metadata) {
       json: metadata
     };
 
-    if (that.explicitGeneration) {
-      reqOpts.qs.ifGenerationMatch = that.explicitGeneration;
+    if (that.generation) {
+      reqOpts.qs.ifGenerationMatch = that.generation;
     }
 
     makeAuthorizedRequest(reqOpts, function(err, res, body) {
@@ -1232,8 +1232,8 @@ File.prototype.startSimpleUpload_ = function(stream, metadata) {
     })
   };
 
-  if (this.explicitGeneration) {
-    reqOpts.qs.ifGenerationMatch = this.explicitGeneration;
+  if (this.generation) {
+    reqOpts.qs.ifGenerationMatch = this.generation;
   }
 
   util.makeWritableStream(stream, {
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -137,7 +137,7 @@ describe('File', function() {
 
     it('should accept specifying a generation', function() {
       var file = new File(bucket, 'name', { generation: 2 });
-      assert.equal(file.explicitGeneration, 2);
+      assert.equal(file.generation, 2);
     });
   });
 


lib/storage/file.js
```diff
@@ -1070,18 +1124,24 @@ File.prototype.startResumableUpload_ = function(stream, metadata) {
...
+      name: that.name
+    },
+    uri: util.format('{base}/{bucket}/o', {
+      base: STORAGE_UPLOAD_BASE_URL,
+      bucket: that.bucket.name
+    })
+  };
+
+  if (this.explicitGeneration) {
+    reqOpts.qs.ifGenerationMatch = this.explicitGeneration;
```
stephenplusplus: This means: Only begin writing to the remote file if the current generation of that file matches the generation of this File instance.

lib/storage/file.js
```diff
@@ -872,11 +920,17 @@ File.prototype.startResumableUpload_ = function(stream, metadata) {
...
-        uploadType: 'resumable'
+        uploadType: 'resumable',
       },
       headers: headers,
       json: metadata
-    }, function(err, res, body) {
+    };
+
+    if (that.explicitGeneration) {
+      reqOpts.qs.ifGenerationMatch = that.explicitGeneration;
```
stephenplusplus: This means: Only update the metadata of the remote file if the current generation of that file matches the generation of this File instance.

ryanseys:
> should we use ifGenerationMatch in this way or should we just let the user write to the file regardless?

Yes, we should. And the API call can fail, and they can handle that in some way.

> should we update the generation lock on the File instance to the new generation?

Nope I think if they specify a generation, keep it at that generation. If they don't specify a generation, it always represents the latest version. If they specify a generation that HAPPENS to be the current generation, still keep it at that generation.

lib/storage/file.js
ryanseys: Try to write regardless, but if a generation is given and doesn't match the current, it should fail because that makes the most sense.

lib/storage/file.js
ryanseys: Makes sense.

ryanseys:
I think generation is an edge-case so having it fail in most scenarios is okay, only when it makes the most sense to use, should it be used. Also instead of trying to figure out or guess how it will be used by developers, I say delegate the logic to the API.

lib/storage/file.js
stephenplusplus: I think that's what's happening. The API request is still made, but if the current generation isn't the one the user gave, the API call should fail.

stephenplusplus:
Revisiting this one to clean up the PR queue. Issues were resolved. Most importantly, versioning has to be enabled at the bucket level. We can perhaps create a convenience method for this (`bucket.enableVersioning()`) or just let it be done via metadata `storage.createBucket('name', { versioning: { enabled: true }})` / `storage.bucket('name').setMetadata({ versioning: { enabled: true }})`.

Either way, writing some tests and will ping when it's ready for a closer look.

lib/storage/bucket.js
```diff
@@ -365,25 +370,38 @@ Bucket.prototype.file = function(name) {
...
       return;
     }
+
     var files = (resp.items || []).map(function(item) {
-      var file = that.file(item.name);
+      var options = {};
+
+      if (query.versions) {
+        options.generation = item.generation;
+      }
```
stephenplusplus: If a user specified that they want to retrieve all the versions of a file, they will get back files scoped to their generation.

stephenplusplus:
This is ready for a closer review.

Calling `bucket.getFiles({ versions: true })` will return version-locked File objects to your callback.

// @javorosas

ryanseys:
Seems the tests are failing?

stephenplusplus:
Not anymore!

lib/storage/file.js
ryanseys: SGTM

lib/storage/file.js
```diff
@@ -65,14 +64,17 @@ var STORAGE_UPLOAD_BASE_URL = 'https://www.googleapis.com/upload/storage/v1/b';
...
-function File(bucket, name, metadata) {
+function File(bucket, name, options) {
   if (!name) {
     throw Error('A file name must be specified.');
   }
 
+  options = options || {};
+
   this.bucket = bucket;
+  this.explicitGeneration = parseInt(options.generation, 10);
```
ryanseys: Shouldn't this be private?

ryanseys: I don't see an issue of exposing `file.generation` or even just `file.metadata.generation` but I wouldn't expose this as a value that someone can rely on (not that we would recommend it anyway)

stephenplusplus: I agree. Must have been a relic of some earlier implementation where I thought it was a bad idea. Good reviewing! `file.metadata.generation` could be confusing since there is such a thing as metadata generations in addition to file generations. Stay tuned for an update.

ryanseys: > there is such a thing as metadata generations in addition to file generations

Weird....

stephenplusplus: Updated in a450f0c4049147f6ccf433893eac11af9bc2e5d6.

ryanseys: Thanks!

lib/storage/file.js
```diff
@@ -54,9 +54,8 @@ var STORAGE_UPLOAD_BASE_URL = 'https://www.googleapis.com/upload/storage/v1/b';
  * @param {module:storage/bucket} bucket - The Bucket instance this file is
  *     attached to.
  * @param {string} name - The name of the remote file.
- * @param {object=} metadata - Metadata to set on the object. This is useful
```
ryanseys: This is no longer supported? Not even with `options.metadata` ? Any particular reason for the removal?

stephenplusplus: It wasn't actually used anywhere or even a thing. Back when we were working out the details of how creating a File object would work, we considered `new File()` having network side-effects. The operation was ultimately shut down, and I believe the `metadata` argument was briefly re-purposed as a way to set metadata from `bucket.getFiles()` when the API responds with the metadata. I decided it was most clear to just use `file.metadata = {/*...*/}`, and not even make it look like we accept a "metadata" object to a "new File()" call, because what would that imply? Would the metadata get set immediately if specified? That's wacky. Being that it's a privately documented property, the artifact of yesteryear was forgotten.

ryanseys: Haha yeah I figured that's what happened because I didn't think we used it. Good to know this won't be a breaking change.

ryanseys:
This looks good to me. Is it merge-able? The main description still seems to suggest otherwise, but I'm guessing that's old.

stephenplusplus:
Fixed!

ryanseys:
BOOM! :dancer:

PR #416, opened 2015-02-28
Accept 0 for start and end of storage range requests

### LINKED ISSUE #415: Found a bug in file.js gcloud storage
I found a bug in file.js at line 318

Range: 'bytes=' + [options.start || '', options.end || ''].join('-')

If I I use a read stream with start and end, it wont work if start = 0
aFile.createReadStream({start: 0, end: 123}).pipe(res);

Here the header Range will be set to "bytes:-123" but google storage will return nothing, the header should be set to "bytes:0-123" to get valid data from gcloud storage.

Make sure you set nothing in the header Range only if options.start is UNDEFINED but not if it is equal to 0

ryanseys:
Fixes #415

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -235,9 +235,10 @@ File.prototype.copy = function(destination, callback) {
  *     with less reliability. You may also choose to skip validation completely,
  *     however this is **not recommended**.
  * @param {number} options.start - A byte offset to begin the file's download
- *     from. NOTE: Byte ranges are inclusive; that is, `options.start = 0` and
- *     `options.end = 999` represent the first 1000 bytes in a file or object.
- *     NOTE: when specifying a byte range, data integrity is not available.
+ *     from. Default is 0. NOTE: Byte ranges are inclusive; that is,
+ *     `options.start = 0` and `options.end = 999` represent the first 1000
+ *     bytes in a file or object. NOTE: when specifying a byte range, data
+ *     integrity is not available.
  * @param {number} options.end - A byte offset to stop reading the file at.
  *     NOTE: Byte ranges are inclusive; that is, `options.start = 0` and
  *     `options.end = 999` represent the first 1000 bytes in a file or object.
@@ -325,8 +326,10 @@ File.prototype.createReadStream = function(options) {
     };
 
     if (rangeRequest) {
+      var start = util.is(options.start, 'number') ? options.start : '0';
+      var end = util.is(options.end, 'number') ? options.end : '';
       reqOpts.headers = {
-        Range: 'bytes=' + [options.start || '', options.end || ''].join('-')
+        Range: 'bytes=' + start + '-' + end
       };
     }
 
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -454,12 +454,12 @@ describe('File', function() {
       file.createReadStream({ start: startOffset });
     });
 
-    it('should accept an end range', function(done) {
+    it('should accept an end range and set start to 0', function(done) {
       var endOffset = 100;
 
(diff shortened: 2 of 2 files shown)


test/storage/file.js
```diff
@@ -485,6 +485,23 @@ describe('File', function() {
...
       file.createReadStream({ start: startOffset, end: endOffset });
     });
+
+    it('should accept range start and end as 0', function(done) {
+      var startOffset = 0;
+      var endOffset = 0;
+
+      request_Override = function(opts) {
+        setImmediate(function () {
+          var expectedRange = 'bytes=0-0';
```
ryanseys: Google Storage ranges [are inclusive](https://cloud.google.com/storage/docs/reference-headers#range), so `range=0-999` is the first 1000 bytes. `range=0-0` would be the first byte.

lib/storage/file.js
```diff
@@ -325,8 +325,10 @@ File.prototype.createReadStream = function(options) {
     };
 
     if (rangeRequest) {
+      var start = typeof options.start === 'undefined' ? '' : options.start;
+      var end = typeof options.end === 'undefined' ? '' : options.end;
```
stephenplusplus: How about "util.is(options.end, 'number')"?

lib/storage/file.js
```diff
@@ -325,8 +325,10 @@ File.prototype.createReadStream = function(options) {
...
       reqOpts.headers = {
-        Range: 'bytes=' + [options.start || '', options.end || ''].join('-')
+        Range: 'bytes=' + [start, end].join('-')
```
stephenplusplus: Might as well just do string concatenation, start + '-' + end

lib/storage/file.js
ryanseys: I considered that. But they could technically pass a string and it would still work as long as it was a number e.g. `"3"` I can switch to using `util.is()` if you prefer that though.

stephenplusplus:
Looks good, but API wise, I would like to see 0 as default. This would block a user from receiving the tail bytes, but an api where you use "end" to specify that doesn't make sense (to me). Any thoughts?

lib/storage/file.js
stephenplusplus: Eh, readability is fine either way. Keep it :+1:

ryanseys:
I don't like 0 as default start because that just destroys potentially useful functionality. It's not difficult to specify `start: 0` in my opinion.

lib/storage/file.js
ryanseys: :+1: Done.

lib/storage/file.js
ryanseys: Keepin it as-is.

stephenplusplus:
The way the json API expects a range is "start-end" or "number", where "-200" having a dash is just a coincidence. It's not really the "end". I too would like to continue offering this funtionality, but not accidentally as we are currently. Maybe we should just accept a single string/number range (-200 or "3-10", for example) and remove the start/end niceties?

ryanseys:
@stephenplusplus Added documentation to describe the situation where a start is not specified. I can change the default start to zero, I don't feel too strongly either way, but I guess I just like to stick close to the specification in that regard. I'd assume if someone wants the first 100 bytes they would just say `'start: 0, end: 99'` especially considering we give that example anyway in the docs.

ryanseys:
Ah, I get what you're saying. I'll fix it to do what you assumed then we can discuss alternatives after.

ryanseys:
Should be ready for another look. Let me know if the documentation should be clarified or added to any further as well.

stephenplusplus:
Perfecto!

PR #417, opened 2015-02-28
Allow 'excludeFromIndexes' on listValues

SerafinTech:
Need to be able to turn off indexes on lists to prevent exploding indexes

### DIFF
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -260,10 +260,17 @@ DatastoreRequest.prototype.save = function(entities, callback) {
           data.value = entity.valueToProperty(data.value);
 
           if (util.is(data.excludeFromIndexes, 'boolean')) {
-            data.value.indexed = !data.excludeFromIndexes;
+            
+            if(util.is(data.value.list_value, 'array')){
+              
+              for(var listValueIndex in data.value.list_value){
+                data.value.list_value[listValueIndex]['indexed'] = !data.excludeFromIndexes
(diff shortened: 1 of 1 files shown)


ryanseys:
Thanks for the PR. Looks like a good submission, just please lint the code with `npm run lint` before submitting :+1: and then your tests should pass here.

lib/datastore/request.js
```diff
@@ -260,10 +260,17 @@ DatastoreRequest.prototype.save = function(entities, callback) {
           data.value = entity.valueToProperty(data.value);
 
           if (util.is(data.excludeFromIndexes, 'boolean')) {
-            data.value.indexed = !data.excludeFromIndexes;
+            
+            if(util.is(data.value.list_value, 'array')){
+              
+              for(var listValueIndex in data.value.list_value){
+                data.value.list_value[listValueIndex]['indexed'] = !data.excludeFromIndexes
+              }
```
stephenplusplus: You can use our helper, `util.propAssign` here:

```js
var exclude = !data.excludeFromIndexes;
var list = data.value.list_value;

data.value.list_value = list.map(util.propAssign('indexed', exclude));
```

stephenplusplus:
We took over the boring stuff: #430 :)

Thanks again!

jlouthan:
Thanks for this! Will these updates be included in the version I get from `npm install gcloud` soon?

PR #418, opened 2015-02-28
Add bucket#makePublic and bucket#makePrivate

ryanseys:
**Do not merge!** No tests yet!

This is the implementation to make a bucket public and private. In my preliminary manually testing, it works great!

Optionally a `recursive` option can be set to `true` to set all the files in the bucket to public as well. QUESTION: Should this be default true? Also, I'm returning either a single error after the first error occurs if a file cannot be made public for some reason, or if they set `force` to `true` then it will add those errors to an array and return that array of errors at the end. 

Additionally, I'm returning a list of the file objects that were successfully publicized/privatized.

I'm looking for feedback on general approach and big issues you can see.

**Issues:** bucket#makePrivate with `strict: true` always throws an error forbidden before it finishes, because it somehow doesn't recognize my account using service account key json to authenticate as an owner. Any idea how I can become the root owner of the bucket?

**TODO**: Add `options.auto` to apply a predefinedObjectACL on the bucket to allow new files to be created with the same permissions.

Depends on #390.

### DIFF
--- a/lib/storage/bucket.js
+++ b/lib/storage/bucket.js
@@ -371,22 +371,58 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
  * });
  */
 Bucket.prototype.makePublic = function(options, callback) {
-  var self = this;
-  var errors = [];
-  var filesProcessed = [];
-  var nextQuery = {};
-
   if (util.is(options, 'function')) {
     callback = options;
     options = {};
   }
 
+  options.public = true;
+  this._makePublicPrivate(options, callback);
+};
+
+Bucket.prototype.makePrivate = function(options, callback) {
+  if (util.is(options, 'function')) {
+    callback = options;
+    options = {};
+  }
+
+  options.private = true;
+  this._makePublicPrivate(options, callback);
+};
+
+Bucket.prototype._makePublicPrivate = function(options, callback) {
+  var self = this;
+  var errors = [];
+  var filesProcessed = [];
+  var nextQuery = {};
+
   // Allow public to read bucket contents
   // while preserving original permissions
-  this.acl.add({
-    entity: 'allUsers',
-    role: 'READER'
-  }, function(err) {
+  if (options.public) {
+    this.acl.add({
+      entity: 'allUsers',
+      role: 'READER'
+    }, processAllFiles);
+  } else if (options.private) {
+    var query = {
+      predefinedAcl: options.strict ? 'private' : 'projectPrivate'
+    };
+
+    // You aren't allowed to set both predefinedAcl & acl properties on a bucket
+    // so acl must explicitly be nullified.
+    var metadata = { acl: null };
+
+    this.makeReq_('PATCH', '', query, metadata, function(err, resp) {
+      if (err) {
+        processAllFiles(err);
+        return;
+      }
+      self.metadata = resp;
+      processAllFiles(null);
+    });
+  }
+
+  function processAllFiles(err) {
     if (err) {
       callback(err);
       return;
@@ -408,7 +444,7 @@ Bucket.prototype.makePublic = function(options, callback) {
     } else {
       callback(null, filesProcessed);
     }
-  });
+  }
 
   function processMoreFiles(query, callback) {
     self.getFiles(query, function(err, files, nextPageQuery) {
@@ -419,7 +455,13 @@ Bucket.prototype.makePublic = function(options, callback) {
 
       // Iterate through each file and make it public.
       async.eachLimit(files, 10, function(file, cb) {
-        file.makePublic(function(err) {
+        if (options.public) {
+          file.makePublic(processedCallback);
+        } else if (options.private) {
+          file.makePrivate({ strict: options.strict }, processedCallback);
+        }
+
+        function processedCallback(err) {
           if (err) {
             if (!options.force) {
               cb(err);
@@ -431,7 +473,7 @@ Bucket.prototype.makePublic = function(options, callback) {
             filesProcessed.push(file);
             cb();
           }
-        });
+        }
       }, function(err) {
         if (err) {
           callback(err);


lib/storage/bucket.js
```diff
@@ -349,6 +351,105 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
...
+  this.acl.add({
+    entity: 'allUsers',
+    role: 'READER'
+  }, function(err) {
+    if (err) {
+      callback(err);
+      return;
+    }
+
+    if (options.recursive) {
```
ryanseys: Not really recursive, I do it iteratively. I'm just following the `-R` option name convention that `gsutil` uses here.

lib/storage/bucket.js
```diff
@@ -349,6 +351,105 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
...
+
+      // Iterate through each file and make it public.
+      async.eachLimit(files, 10, function(file, cb) {
+        file.makePublic(function(err) {
+          if (err) {
+            if (!options.force) {
+              cb(err);
+            } else {
+              errors.push(err);
+              cb();
```
ryanseys: This could probably use some more inline docs, but basically add errors to an array when they explicitly want to `force` to keep processing.

lib/storage/bucket.js
```diff
@@ -349,6 +351,142 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
...
+  }
+
+  function processAllFiles(err) {
+    if (err) {
+      callback(err);
+      return;
+    }
+
+    if (options.recursive) {
+      async.whilst(function() {
```
stephenplusplus: I knew you'd find cool stuff in `async` :)

lib/storage/bucket.js
```diff
@@ -349,6 +351,142 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
...
+
+      // Iterate through each file and make it public.
+      async.eachLimit(files, 10, function(file, cb) {
+        if (options.public) {
+          file.makePublic(processedCallback);
+        } else if (options.private) {
+          file.makePrivate({ strict: options.strict }, processedCallback);
+        }
+
+        function processedCallback(err) {
```
ryanseys: I don't know enough about the JS runtime here but will this function definition be reinitialised for each file? Or maybe I should access the objects in these shared functions through `this` and then bind the function to the appropriate `this` at runtime? Idk what's better

lib/storage/bucket.js
```diff
@@ -349,6 +351,142 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
...
+        processMoreFiles(nextQuery, cb);
+      }, function(err) {
+        if (err || errors.length) {
+          callback(err || errors, filesProcessed);
+        } else {
+          callback(null, filesProcessed);
+        }
+      });
+    } else {
+      callback(null, filesProcessed);
```
stephenplusplus: I think `filesProcessed` here will always be an empty array. Is that confusing for a user who doesn't do recursive?

ryanseys: Yeah that's probably confusing.

lib/storage/bucket.js
stephenplusplus: Just leave it. Micro optimizations are generally worthless (unless the DOM is concerned) and just lead to weird code in JS.

lib/storage/bucket.js
stephenplusplus: I guess the alternative is just doing `bucket.getFiles()` and returning all of the files? But that feels silly. Alternative alternative is documenting that `files` is only populated if you do recursive?

lib/storage/bucket.js
ryanseys: :+1:

lib/storage/bucket.js
ryanseys: No the idea you return the list of files that successfully made public/private.

lib/storage/bucket.js
```diff
@@ -349,6 +351,142 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
...
+
+  function processMoreFiles(query, callback) {
+    self.getFiles(query, function(err, files, nextPageQuery) {
+      if (err) {
+        callback(err);
+        return;
+      }
+
+      // Iterate through each file and make it public.
+      async.eachLimit(files, 10, function(file, cb) {
```
stephenplusplus: Maybe make `10` a constant for readability.

lib/storage/bucket.js
stephenplusplus: Sorry, ACL confusion again. An empty array makes sense then, since no files were affected. Forget I started this thread :man::hammer:

stephenplusplus:
I know the worst thing someone reviewing your PR can do is be like "here's a whole other bunch of code that renders yours useless!" I didn't intend to let it get to that point- I just wanted to drop the PR in my editor, which makes it easier for me to make sense of. I found some things I thought I could improve by demonstrating as opposed to trying to explain. And, before you know it... https://gist.github.com/stephenplusplus/66bfc7ba04ba97a8a15a

I think either approach is perfectly fine, I just wanted to hold less state outside of each function, and make it more modular in general. As an example, I let `bucket#makePrivate()` and `bucket#makePublic()` handle each of their metadata calls, then changed the behavior of `_makePublicPrivate` to a more focused intent: `_makeAllFilesPublicPrivate`.

It's not tested (didn't even run it), but if it's something that doesn't look hideous to you, I can send a PR with it to evolve from this one, or you can send me a meme that explains what a jerk I am, and I'll delete the gist and pretend it never happened!

ryanseys:
Hey this is great! I don't mind one bit because mine was more POC. One thing that is still TODO that needs to be thought of is the `options.auto` which sets the default object ACL for the bucket as well. I've got a midterm that requires my focus today so can't really care for this PR so I'm 100% ok with abandoning it and letting you run with your gist. We can close this PR, and you can reopen a new one if you want to keep running with this. That's the easiest way forward.

ryanseys:
Upon closer inspection your async.eachLimit is recursive and doesn't really make sense. You don't want to run with the nextQuery on each file, only after all files are processed.

stephenplusplus:
Ooh, good catch. I'll have to think my way out of that one!

ryanseys:
Actually you're right, I was messing up the callback count.

stephenplusplus:
Wait:

```js
async.eachLimit(files, MAX_PARALLEL_LIMIT, processFile, function(err) {
```

Won't that process files (max 10 at a time) until all files are processed, then call processFiles again for the nextQuery?

stephenplusplus:
We're a mess

ryanseys:
Yeah my bad... :( I think you'll be good. Or you'll figure it out! :smile:

stephenplusplus:
Going to close this to let #429 carry it on. If I missed anything in the transfer, please update the issue.

Issue #422, opened 2015-03-03
Lack of error handling while uploading a file.


teddybearz:
If the network connection is lost (e.g. cable pulled), the node process will exit because of an error even that has no handler registered.

Need some fix like:
```javascript
diff --git a/node_modules/gcloud/lib/storage/file.js b/node_modules/gcloud/lib/storage/file.js
index 8e1a26e..4aaec72 100644
--- a/node_modules/gcloud/lib/storage/file.js
+++ b/node_modules/gcloud/lib/storage/file.js
@@ -913,11 +915,20 @@
 
     var writeStream = request(reqOpts);
     writeStream.callback = util.noop;
-
+    var errorHandled = false;
+    writeStream.on('error', function(err) {
+      if (!errorHandled) {
+         errorHandled = true;
+         handleError(err);
+      }
+    });
     writeStream.on('complete', function(res) {
       util.handleResp(null, res, res.body, function(err, data) {
         if (err) {
-          handleError(err);
+          if (!errorHandled) {
+             errorHandled = true;
+             handleError(err);
+          }
           return;
         }
```

ryanseys:
I'll look into this.

ryanseys:
Fixed in #426

PR #426, opened 2015-03-04
Emit errors from upstream on file upload

ryanseys:
I (think I) have a fix here for #422  but writing the test for it has been troublesome for me, most likely due to my lack of understanding with streams. Hopefully @stephenplusplus can help!

The test is passing (all asserts pass and done is called) and then immediately **failing** due to stream.end being called on a stream that (I guess) no longer exists?

I'm getting the following error when running the test included in this PR and don't know how to fix:

```
> mocha test/*

  File
    createWriteStream
      ✓ should re-emit errors
      1) should re-emit errors

  1 passing (23ms)
  1 failing

  1) File createWriteStream should re-emit errors:
     Uncaught TypeError: Cannot read property 'ending' of undefined
      at Duplexify.end (/Users/ryanseys/gcloud-node/node_modules/duplexify/index.js:218:27)
      at handleError (/Users/ryanseys/gcloud-node/lib/storage/file.js:1064:12)
      at DestroyableTransform.<anonymous> (/Users/ryanseys/gcloud-node/lib/storage/file.js:992:7)
      at DestroyableTransform.emit (events.js:129:20)
      at Immediate._onImmediate (/Users/ryanseys/gcloud-node/test/storage/file.js:553:18)
      at processImmediate [as _immediateCallback] (timers.js:358:17)
```

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -988,6 +988,10 @@ File.prototype.startResumableUpload_ = function(stream, metadata) {
     var writeStream = request(reqOpts);
     writeStream.callback = util.noop;
 
+    writeStream.on('error', function(err) {
+      handleError(err);
+    });
+
     writeStream.on('complete', function(res) {
       util.handleResp(null, res, res.body, function(err, data) {
         if (err) {
@@ -1056,7 +1060,6 @@ File.prototype.startResumableUpload_ = function(stream, metadata) {
     }
 
     stream.emit('error', err);
-    stream.end();
   }
 };
 
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -528,6 +528,44 @@ describe('File', function() {
       writable.write('data');
     });
 
+    it('should re-emit errors', function(done) {
+      var error = new Error('Error.');
+      var requestCount = 0;
+      file.bucket.storage.makeAuthorizedRequest_ = function(reqOpts, cb) {
+        requestCount++;
+
+        // respond to creation POST.
+        if (requestCount === 1) {
+          cb(null, null, { headers: { location: 'http://resume' }});
+          return;
+        }
+
+        // create an authorized request for the first PUT.
+        if (requestCount === 2) {
+          cb.onAuthorized(null, { headers: {} });
+        }
+      };
+
+      // respond to first upload PUT request.
+      request_Override = function() {
+        var stream = through();
+        setImmediate(function() {
+          stream.emit('error', error);
+        });
+        return stream;
+      };
+
+      var stream = duplexify();
+
+      stream
+        .on('error', function(err) {
+          assert.equal(err, error);
+          done();
+        });
+
+      file.startResumableUpload_(stream);
+    });
+
     it('should start a simple upload if specified', function(done) {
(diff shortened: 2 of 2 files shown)


test/storage/file.js
```diff
@@ -528,6 +528,44 @@ describe('File', function() {
       writable.write('data');
     });
 
+    it('should re-emit errors', function(done) {
+      var error = new Error('Error.');
```
stephenplusplus: Love that my boring world's most boring error is catching on.

We should just have a custom Type, BoringTestError.

lib/storage/file.js
```diff
@@ -988,6 +988,10 @@ File.prototype.startResumableUpload_ = function(stream, metadata) {
     var writeStream = request(reqOpts);
     writeStream.callback = util.noop;
 
+    writeStream.on('error', function(err) {
+      handleError(err);
+    });
+
```
stephenplusplus: `writeStream.on('error', handleError)` should work

stephenplusplus:
I believe handleError doesn't need to end the stream. Errors are emitted back to the stream that was passed into `startResumableUpload_`, which [ends the stream there](https://github.com/GoogleCloudPlatform/gcloud-node/blob/1a7737bb8898398d6b9d865ed0853de7131ceb84/lib/storage/file.js#L544).

... streams of tears.

lib/storage/file.js
```diff
@@ -1056,7 +1060,6 @@ File.prototype.startResumableUpload_ = function(stream, metadata) {
     }
 
     stream.emit('error', err);
-    stream.end();
```
ryanseys: Yep this fixed it.

ryanseys:
PTAL :smile:

stephenplusplus:
:+1:

teddybearz:
Is it possible that handleError() being called twice? i.e. one error event and one complete event with error come back to back?

ryanseys:
It was that the stream was being closed too many times. I believe we have addressed this issue fully. If you want to test against master, that would be wonderful!

teddybearz:
I wonder in real network scenario (not just the test mock environment), it is possible that there will be one error event and one complete event+error being emitted back to back with error code all equals to 404 or 499-600. If it happens, the handleError() will be called twice and startUpload() resumeUpload() will be called twice.

stephenplusplus:
The write stream will either emit an error or complete, it can't do both.

Regardless, somewhat related, we should aim to "once-ify" our user provided callbacks.

stephenplusplus:
@teddybearz I can't help but feel I'm missing your point. If I am, thanks for your patience.

teddybearz:
> The write stream will either emit an error or complete, it can't do both.
In theory yes, but for event driven code, it takes some efforts to enforce and maintain that. Maybe better not assume too much and prepare for the worst.

BTW, is there a mock environment I can use to test my code that uses this library? So I can simulate bad-upload/bad-download/network-connection-reset those kinds of error cases programmatically? I am tired of pulling cable now :smile:

ryanseys:
@teddybearz You might have some luck trying out https://github.com/tylertreat/Comcast :smile:

PR #429, opened 2015-03-05
allow making a bucket public/private

stephenplusplus:
Evolves: #418

This introduces:
##### options.includeFiles

Type: `boolean`
Default: `false`

Run `file.makePublic()` against every in the bucket.
##### options.force

Type: `boolean`
Default: `false`

Queue errors occurred while making files public until all files have been processed.
### callback(error, [files])

Type: `function`

`error` can be `null`, `Error`, or `Error[]`.

If you don't specify `includeFiles: true`, `files` doesn't exist. If you did, it contains all files successfully made public.
## 
##### options.strict

Type: `boolean`
Default: `false`

Make the bucket private to only the owner user. Otherwise, it will be private to the project.
##### options.includeFiles

Type: `boolean`
Default: `false`

Run `file.makePrivate()` against every in the bucket.
##### options.force

Type: `boolean`
Default: `false`

Queue errors occurred while making files public until all files have been processed.
### callback(error, [files])

Type: `function`

`error` can be `null`, `Error`, or `Error[]`.

If you don't specify `includeFiles: true`, `files` doesn't exist. If you did, it contains all files successfully made private.
## 
### Todos
- [ ] Regression tests
- [x] Unit Tests
- [x] Docs
- [x] ~~Add `options.auto` to apply a predefinedObjectACL on the bucket to allow new files to be created with the same permissions.~~ (this behavior will become default)
## 
### Issues
1. bucket#makePrivate with `strict: true` always throws an error forbidden before it finishes, because it somehow doesn't recognize my account using service account key json to authenticate as an owner. Any idea how I can become the root owner of the bucket? (applies to @ryanseys and me as well)
   - **Resolution**: Removed `strict` option: https://github.com/GoogleCloudPlatform/gcloud-node/pull/429#issuecomment-77442932

### DIFF
--- a/lib/storage/bucket.js
+++ b/lib/storage/bucket.js
@@ -485,7 +485,24 @@ Bucket.prototype.makePrivate = function(options, callback) {
   async.parallel([setPredefinedAcl, makeFilesPrivate], callback);
 
   function setPredefinedAcl(done) {
-    self.setPredefinedAcl_(options, done);
+    var query = {
+      predefinedAcl: 'projectPrivate'
+    };
+
+    // You aren't allowed to set both predefinedAcl & acl properties on a bucket
+    // so acl must explicitly be nullified.
+    var metadata = { acl: null };
+
+    self.makeReq_('PATCH', '', query, metadata, function(err, resp) {
+      if (err) {
+        done(err);
+        return;
+      }
+
+      self.metadata = resp;
+
+      done();
+    });
   }
 
   function makeFilesPrivate(done) {
@@ -570,10 +587,21 @@ Bucket.prototype.makePublic = function(options, callback) {
   options = options || {};
   options.public = true;
 
-  async.parallel([addAclPermissions, makeFilesPublic], callback);
+  async.parallel([
+    addAclPermissions,
+    addDefaultAclPermissions,
+    makeFilesPublic
+  ], callback);
 
   function addAclPermissions(done) {
     // Allow reading bucket contents while preserving original permissions.
+    self.acl.add({
+      entity: 'allUsers',
+      role: 'READER'
+    }, done);
+  }
+
+  function addDefaultAclPermissions(done) {
     self.acl.default.add({
       entity: 'allUsers',
       role: 'READER'
@@ -838,43 +866,6 @@ Bucket.prototype.makeAllFilesPublicPrivate_ = function(options, callback) {
       callback();
     }
   }
-};
-
-/**
- * Set the predefinedAcl for this bucket to `publicRead` (public) or
- * `projectPrivate` (private).
- *
- * This will also update the metadata on the object with the APIresponse.
- *
- * @private
- *
- * @param {object} options - Configuration object.
- * @param {boolean} options.private - Set the predefinedAcl to projectPrivate.
- * @param {boolean} options.public - Set the predefinedAcl to publicRead.
- */
-Bucket.prototype.setPredefinedAcl_ = function(options, callback) {
-  var self = this;
-
-  var predefinedValue = options.public ? 'publicRead' : 'projectPrivate';
-
-  var query = {
-    predefinedAcl: predefinedValue
-  };
-
-  // You aren't allowed to set both predefinedAcl & acl properties on a bucket
-  // so acl must explicitly be nullified.
-  var metadata = { acl: null };
-
-  this.makeReq_('PATCH', '', query, metadata, function(err, resp) {
-    if (err) {
-      callback(err);
-      return;
-    }
-
-    self.metadata = resp;
-
-    callback();
-  });
 };
 
 /**
--- a/test/storage/bucket.js
+++ b/test/storage/bucket.js
@@ -446,26 +446,49 @@ describe('Bucket', function() {
       };
     });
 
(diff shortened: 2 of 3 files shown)


ryanseys:
So thoughts on private issues... we could just remove `strict: true`? Private to the project seems like smarter than private to specific user anyway.

stephenplusplus:
Makes sense to me.

ryanseys:
make-it-so-captain

stephenplusplus:
2/3 todos done. Regarding the last one:

> Add options.auto to apply a predefinedObjectACL on the bucket to allow new files to be created with the same permissions.

Should we just do this by default? Seems if I say "make my bucket private", it should shut people out until I re-open it (same logic in reverse as well).

ryanseys:
I'd say yes for auto.

stephenplusplus:
(saying that to myself)

lib/storage/bucket.js
```diff
@@ -410,6 +411,194 @@ Bucket.prototype.getMetadata = function(callback) {
...
+    options = {};
+  }
+
+  options = options || {};
+  options.public = true;
+
+  async.parallel([
+    setPredefinedAcl,
+    addAclPermissions,
+    makeFilesPublic
```
stephenplusplus: Do I need to set "allUsers" to "READER" (what `addAclPermissions` does) if I'm changing the `predefinedAcl`?

stephenplusplus: // @ryanseys

ryanseys: There's three stages to this:
1. Set the bucket to be publicly readable (list contents)
2. Set bucket to automatically make new objects publicly readable
3. Set each file to be publicly readable (read body of objects)

Steps 1 and 3 are done. Step 2 is not done. I believe this can be covered by a default object acl or predefined default object acl. You have to choose one or the other. After a brief skim it looks like setPredefinedAcl should accomplish Step 1 and 2 in a single request as long as both predefinedAcl and predefinedDefaultObjectAcl can be both set at the same time.

stephenplusplus:
```
[Error: Cannot provide both a predefinedDefaultObjectAcl and a defaultObjectAcl.]
```

Halp.

ryanseys: You have to choose one or the other.

What are you trying to do?

stephenplusplus: > There's three stages to this:
> 1. Set the bucket to be publicly readable (list contents)
> 2. Set bucket to automatically make new objects publicly readable
> 3. Set each existing file to be publicly readable (read body of objects)

Would you mind adding next to these points the way each step is accomplished?

ryanseys: 1. Set the bucket to be publicly readable (list contents) - Set predefinedAcl = 'publicRead'
2. Set bucket to automatically make new objects publicly readable - Set defaultObjectAcl to { entity: 'allUsers', role: 'READER' }
3. Set each existing file to be publicly readable (read body of objects) - Loop through each file and add acl { entity: 'allUsers', role: 'READER' }

ryanseys: (I think that should work)

lib/storage/bucket.js
```diff
@@ -576,6 +765,98 @@ Bucket.prototype.upload = function(localPath, options, callback) {
...
+};
+
+Bucket.prototype.setPredefinedAcl_ = function(options, callback) {
+  var self = this;
+
+  var predefinedValue = options.public ? 'publicRead' : 'projectPrivate';
+
+  var query = {
+    predefinedAcl: predefinedValue,
+    predefinedDefaultObjectAcl: predefinedValue
```
ryanseys: You can't do both here.

stephenplusplus:
PTAL!

lib/storage/bucket.js
```diff
@@ -410,6 +411,186 @@ Bucket.prototype.getMetadata = function(callback) {
...
+  }
+
+  options = options || {};
+  options.public = true;
+
+  async.parallel([addAclPermissions, makeFilesPublic], callback);
+
+  function addAclPermissions(done) {
+    // Allow reading bucket contents while preserving original permissions.
+    self.acl.default.add({
```
ryanseys: Does this additionally make the bucket listing public readable? I'd assume we'd have to additionally do 
``` js
self.acl.add({
  entity: 'allUsers',
  role: 'READER'
}, cb);
```

test/storage/bucket.js
```diff
@@ -623,6 +727,236 @@ describe('Bucket', function() {
...
+      }, function(errs, files) {
+        assert.deepEqual(errs, [error, error]);
+        assert.deepEqual(files, successFiles);
+        done();
+      });
+    });
+  });
+
+  describe('setPredefinedAcl_', function() {
+    it('should make correct public PATCH request', function(done) {
```
ryanseys: Is this test necessary anymore?

stephenplusplus: I can remove the `setPredefinedAcl_` function altogether since we're just using it for `private` now.

ryanseys: SGTM

ryanseys:
:ship: it. (after tests pass)

PR #430, opened 2015-03-05
datastore: allow indexing list values

stephenplusplus:
Fixes: #417


stephenplusplus:
This is probably a simple enough change for a self merge, but @ryanseys for sanity check :)

ryanseys:
:+1:

Issue #427, opened 2015-03-05
Total Rows in  job.getQueryResults()


sunnychow:
When we get make a pagination, we want to know the total number of page.

According to google, it could return the totalRows, such that, we could know the total number of page. 
https://cloud.google.com/bigquery/docs/reference/v2/jobs/getQueryResults

Unfortunately, in this project, job.getQueryResults() do not returns the total number of records, it's better to return the total rows.

stephenplusplus:
Good find! If you come across anything else we're not passing back that would be useful, please continue to let us know.

stephenplusplus:
This will be solved by #448.

ryanseys:
#448 recently was merged into master. If you'd like to give it a shot now, feel free and let us know how it goes. Otherwise you may wait until we release the next version.

PR #431, opened 2015-03-06
return Total Rows

sunnychow:
return Total Rows
#427

### DIFF
--- a/lib/bigquery/index.js
+++ b/lib/bigquery/index.js
@@ -391,23 +391,22 @@ BigQuery.prototype.query = function(options, callback) {
         return;
       }
 
-      var nextQuery = null;
-	  var rows = Table.mergeSchemaWithRows_(resp.schema, resp.rows || []);
+	    var nextQuery = null;
+	    var rows = Table.mergeSchemaWithRows_(resp.schema, resp.rows || []);
 
-      if (resp.jobComplete === false) {
-        // Query is still running.
-        nextQuery = extend({
-          job: that.job(options.job.id)
-        }, options);
-      } else if (resp.pageToken) {
-		      // More results exist.
-		      nextQuery = extend({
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Thank you for taking a crack at the issue! Can you explain what this PR is doing? I'm not sure I see how total rows is being returned, but I might be distracted by the formatting changes. Could you please keep with the style of the project (2 space indentation)?

Thanks!

ryanseys:
Seems the tests are failing now, so those will need to be addressed in your PR. You can run the tests locally with `npm test`. Regarding the approach of this PR, we don't return the raw response object for any requests so I can't see us making an exception for this case, the object that returns rows will need to be adjusted to return the totalRows property as well. This would be a backward-incompatible change which sucks, but seems necessary. 

Aside: @stephenplusplus we have no clear way on how to handle these regressions caused by the addition of more attributes in the return body. We should likely discuss a best-practices approach to this because I anticipate it will happen a lot in the future and I'd prefer to not break everyone's code when it does happen and we need to change things.

sunnychow:
I think the best practices approach to this is returns the response object. For example, in the line
https://github.com/sunnychow/gcloud-node/blob/26bca6bd7d7f909e6eeaf77cab938bf1c56a6910/lib/bigquery/index.js#L412
we should change it to
```
onComplete(null, rows, nextQuery, resp);
```
This is because google big query returns useful information in the resp. https://cloud.google.com/bigquery/docs/reference/v2/jobs/getQueryResults
For example, `kind`, `etag`, `totalRows`

stephenplusplus:
This will be solved by #448, where we return the raw API response for all calls. Thanks for getting this started, @sunnychow!

sunnychow:
thank you.

PR #433, opened 2015-03-07
Use pubsub v1beta2

### LINKED ISSUE #414: Update pubsub to v1beta2
The following is a checklist for migrating over to the new v1beta2 pubsub API. It lists the incompatible changes that occurred between the two APIs.

- [x] All requests now using a new url scheme: `https://pubsub.googleapis.com/v1beta2/...`

### Subscriptions:
- [x] acknowledge: Move `subscription` to path parameter from request body.
- [x] acknowledge: `ackId` changes to `ackIds`.
- [x] create: No longer accepts a name in request body (subscription resource).
- [x] list: `pageSize` instead of `maxResults`.
- [x] list: project in path parameter.
- [x] list: no `query` parameter.
- [x] list: returns `subscriptions` instead of `subscription` in response.
- [x] pull: subscription moved to path parameter instead of request body.
- [x] pull: added `maxMessages` to set max number of messages to pull.
- [x] pull: response now contains top level `receivedMessages` and is drastically different response.
- [x] pullBatch: removed pullBatch.

### Topics:
- [x] create: name now a path parameter instead of in request body.
- [x] delete: different url scheme for topic name.
- [x] get: topic name now in path parameter instead of in request body.
- [x] list: require project in path parameter.
- [x] list: no `query` parameter.
- [x] list: `pageSize` instead of `maxResults`.
- [x] list: returns `topics` instead of `topic` in response.
- [x] publish: topic now a path parameter instead of in request body.
- [x] publish: request body now takes `messages` array of messages.
- [x] publish: response will now return list of `messageIds` rather than an empty response.
- [x] publishBatch: removed publishBatch.

### NEW:
- [x] [topics.subscriptions.list](https://cloud.google.com/pubsub/reference/rest/v1beta2/projects/topics/subscriptions/list) which results the name of the subscriptions for a particular topic.
- [x] [PubSubMessage description](https://cloud.google.com/pubsub/reference/rest/google/pubsub/v1beta2/PubsubMessage) which describes the pubsub message response for various requests.

ryanseys:
This PR updates PubSub to `v1beta2`. Some of the parameters for calls have changed to match the API in hopes that it more understandable.

I've also made a couple changes that are irrespective of the change in API:
1. When pulling messages to be emitted on the event listener, pull for the maximum number of messages and emit each one received rather than pulling 1 message at a time. This greatly increases the throughput of messages and reduces the overhead of making the JSON request for each message individually.
2. Added `Subscription#setAckDeadline` to set the deadline for a message. I've added respective unit and regression tests in this PR.

Note: We have still left out PushConfig various usage and endpoints from this API.

Fixes #414

### DIFF
--- a/lib/pubsub/subscription.js
+++ b/lib/pubsub/subscription.js
@@ -380,17 +380,17 @@ Subscription.prototype.pull = function(options, callback) {
  * to make the message available for redelivery if the processing was
  * interrupted.
  *
- * @param {object}   options - The configuration object.
- * @param {number}   options.ackId - The ack id to change.
- * @param {number}   options.seconds - Number of seconds after call is made to
+ * @param {object} options - The configuration object.
+ * @param {number} options.ackId - The ack id to change.
+ * @param {number} options.seconds - Number of seconds after call is made to
  *     set the deadline of the ack.
  * @param {Function=} callback - The callback function.
  *
  * @example
  * var options = {
- * 	ackId: 123,
- * 	seconds: 10 // Expire in 10 seconds from call.
- * }
+ *   ackId: 123,
+ *   seconds: 10 // Expire in 10 seconds from call.
+ * };
  * subscription.setAckDeadline(options, function(err) {});
  */
 Subscription.prototype.setAckDeadline = function(options, callback) {
--- a/lib/pubsub/topic.js
+++ b/lib/pubsub/topic.js
@@ -115,8 +115,8 @@ Topic.formatName_ = function(projectId, name) {
  *     event: 'new user'
  *   },
  *   attributes: {
- *     'key': 'value',
- *     'hello': 'world'
+ *     key: 'value',
+ *     hello: 'world'
  *   }
  * };
  * topic.publish(registrationMessage, function(err, messageIds) {});
@@ -174,7 +174,7 @@ Topic.prototype.delete = function(callback) {
 };
 
 /**
- * Lists the name of the subscriptions for this topic. You may optionally
+ * Get a list of the subscriptions registered to this topic. You may optionally
  * provide a query object as the first argument to customize the response.
  *
  * Your provided callback will either be invoked with an error object, if an API


lib/pubsub/subscription.js
```diff
@@ -311,9 +314,10 @@ Subscription.prototype.delete = function(callback) {
  * subscription.pull(function(err, messages) {
  *   // messages = [
  *   //   {
- *   //     ackId: '', // ID used to acknowledge its receival.
- *   //     id: '',    // Unique message ID.
- *   //     data: ''   // Contents of the message.
+ *   //     ackId: '',     // ID used to acknowledge its receival.
```
jgeewax: `ackIds`* ?

lib/pubsub/topic.js
```diff
@@ -144,11 +147,17 @@ Topic.prototype.publish = function(messages, callback) {
   callback = callback || util.noop;
 
   var body = {
-    topic: this.name,
     messages: messages.map(Topic.formatMessage_)
   };
 
-  this.makeReq_('POST', 'topics/publishBatch', null, body, callback);
+  var path = this.name + ':publish';
```
jgeewax: Is `:publish` right ? For a URL path? If this isn't a typo, doesn't that seem off for an API URL?

ryanseys: Yeah it's a little different than what we are used to seeing but it's right. See [here](https://cloud.google.com/pubsub/reference/rest/v1beta2/projects/topics/publish). I believe pubsub can get away with this different URL path because they are running their API out of https://pubsub.googleapis.com/ and not simply https://googleapis.com/ like most other APIs.

lib/pubsub/subscription.js
ryanseys: Each message has a single `ackId` but when you make the  `acknowledge` API call, you can acknowledge many `ackIds`. Here is the format: https://cloud.google.com/pubsub/reference/rest/v1beta2/projects/subscriptions/pull#google.pubsub.v1beta2.ReceivedMessage

lib/pubsub/topic.js
ryanseys: @tmatsuo Any idea why this `:publish` style for API URLs is used? Personally, I kinda like it, but perhaps you can fill us in on the reasoning for using it?

It feels very much like I'm making a method call. The resource is `/projects/my-project/topics/my-topic` and my action on that resource is `:publish` :smiley:

tmatsuo: I think this is for conforming to the new internal API, which is more rest style. I'll send you the link in an e-mail.

stephenplusplus:
> When pulling messages to be emitted on the event listener, pull for the maximum number of messages and emit each one received rather than pulling 1 message at a time. This greatly increases the throughput of messages and reduces the overhead of making the JSON request for each message individually. 

:+1:

lib/pubsub/subscription.js
```diff
@@ -371,4 +374,35 @@ Subscription.prototype.pull = function(options, callback) {
...
+ * @param {number}   options.ackId - The ack id to change.
+ * @param {number}   options.seconds - Number of seconds after call is made to
+ *     set the deadline of the ack.
+ * @param {Function=} callback - The callback function.
+ *
+ * @example
+ * var options = {
+ * 	ackId: 123,
+ * 	seconds: 10 // Expire in 10 seconds from call.
+ * }
```
stephenplusplus: `};`

lib/pubsub/subscription.js
```diff
@@ -371,4 +374,35 @@ Subscription.prototype.pull = function(options, callback) {
...
+ * @param {object}   options - The configuration object.
+ * @param {number}   options.ackId - The ack id to change.
+ * @param {number}   options.seconds - Number of seconds after call is made to
+ *     set the deadline of the ack.
+ * @param {Function=} callback - The callback function.
+ *
+ * @example
+ * var options = {
+ * 	ackId: 123,
+ * 	seconds: 10 // Expire in 10 seconds from call.
```
stephenplusplus: 2 spaces.

lib/pubsub/subscription.js
```diff
@@ -371,4 +374,35 @@ Subscription.prototype.pull = function(options, callback) {
...
+ * Modify the ack deadline for a specific message. This method is useful to
+ * indicate that more time is needed to process a message by the subscriber, or
+ * to make the message available for redelivery if the processing was
+ * interrupted.
+ *
+ * @param {object}   options - The configuration object.
+ * @param {number}   options.ackId - The ack id to change.
+ * @param {number}   options.seconds - Number of seconds after call is made to
+ *     set the deadline of the ack.
+ * @param {Function=} callback - The callback function.
```
stephenplusplus: Formatting here is inconsistent with other areas of the docs. To match, this would be (note the spacing and casing):

```js
@param {type} label - Description.
```

lib/pubsub/topic.js
```diff
@@ -112,11 +114,12 @@ Topic.formatName_ = function(projectId, name) {
  *     name: 'Stephen',
  *     event: 'new user'
  *   },
- *   labels: [
- *     'registration'
- *   ]
+ *   attributes: {
+ *     'key': 'value',
+ *     'hello': 'world'
```
stephenplusplus: Unquote the keys.

lib/pubsub/subscription.js
ryanseys: Oh geez. :( I blame my editor's auto-creation of these blocks which I mindlessly filled in. I'll definitely fix all these.

lib/pubsub/topic.js
```diff
@@ -161,11 +170,11 @@ Topic.prototype.publish = function(messages, callback) {
  */
 Topic.prototype.delete = function(callback) {
   callback = callback || util.noop;
-  this.makeReq_('DELETE', 'topics/' + this.name, null, true, callback);
+  this.makeReq_('DELETE', this.name, null, null, callback);
 };
 
 /**
- * Get a list of the subscriptions registered to this topic. You may optionally
+ * Lists the name of the subscriptions for this topic. You may optionally
```
stephenplusplus: `List the names of the subscriptions for this topic.` or something like it. Mostly looking at "name of the subscriptions", where instead, each subscription has a name.

stephenplusplus:
It looks great! I left non-issue comments on formatting and phrasing, but after settling those, we're both driving straight to big green button city.

lib/pubsub/topic.js
ryanseys: Oh yes this is actually still returning subscription objects and not just the names of the subscriptions which is what the API returns (which is weird but I digress). I should leave out "name" and all that jazz and just talk about subscriptions.

ryanseys:
PTAL at the latest diff in this PR for my fixes.

Issue #423, opened 2015-03-03
http response from google may not always has 'x-goog-hash' header


teddybearz:
The http response from google may not always has 'x-goog-hash' header. When it happens, the node process will exit.

Need fix like:
```javascript
diff --git a/node_modules/gcloud/lib/storage/file.js b/node_modules/gcloud/lib/storage/file.js
index 8e1a26e..4aaec72 100644
--- a/node_modules/gcloud/lib/storage/file.js
+++ b/node_modules/gcloud/lib/storage/file.js
@@ -361,10 +361,12 @@
             var md5Fail = true;
 
             var hashes = {};
-            res.headers['x-goog-hash'].split(',').forEach(function(hash) {
-              var hashType = hash.split('=')[0];
-              hashes[hashType] = hash.substr(hash.indexOf('=') + 1);
-            });
+            if (res.headers && res.headers['x-goog-hash']) {
+              res.headers['x-goog-hash'].split(',').forEach(function(hash) {
+                var hashType = hash.split('=')[0];
+                hashes[hashType] = hash.substr(hash.indexOf('=') + 1);
+              });
+            }
 
             var remoteMd5 = hashes.md5;
             var remoteCrc = hashes.crc32c && hashes.crc32c.substr(4);
```

stephenplusplus:
Did you run into a situation where it wasn't returned?

> Google Cloud Storage stores MD5 hashes for all non-composite objects. CRC32Cs are available for all objects.
> \- https://cloud.google.com/storage/docs/reference-headers#xgooghash

ryanseys:
Seems like a reasonable request for a sanity check.

stephenplusplus:
Still would like to know if this can come up. Would affect more of our code, if we can't always trust that a CRC32C hash is available.

teddybearz:
It happens reliably if you try to fetch an object right before the upload of the object is done.

ryanseys:
Why would you try to fetch an object before it was uploaded? Can you provide a snippet of code that causes this issue?

teddybearz:
It was not intentionally. The bug was triggered because of in my old code, I registered on the "finish" event instead of "complete" event.  A normal writeablestream will emit 'finish' when the data has been flushed to the underlying system, apparently 'finish' and 'complete' have different semantic in your lib. 'finish' event doesn't mean the object is committed. There is no document about this 'complete" event and I have to dig into the lib code to find out that.

[...]

PR #435, opened 2015-03-10
storage: explain `complete` event.

stephenplusplus:
Related: #423 

Edit (@ryanseys): and also another example: #436


ryanseys:
Merging because LGTM and docs-only fix and too lazy to wait for travis to finish.

PR #438, opened 2015-03-10
Check retry limit on 404 errors

### LINKED ISSUE #436: File write stream does not error when writing to a file in a non-existent bucket
When using file.createWriteStream when the file was created on a non-existent bucket, the stream does not error and essentially acts as if everything is fine. The only indication of error is the fact that no file is uploaded.

In the example below, one would expect the error event to trigger:
```javascript

var storage = gcloud.storage({
  keyFilename: process.env.GOOGLE_APPLICATION_CREDENTIALS,
  projectId: process.env.GCLOUD_PROJECT_ID
})

var bucket = storage.bucket('this-bucket-does-not-exist');

var file = bucket.file('file.ext');
var stream = file.createWriteStream();

stream.on('error', function(){ console.log('Error'); });
stream.on('finish', function(){ console.log('Done'); });

stream.write('data');
stream.end();
```

ryanseys:
This is a patch fix for uploading to a bucket that doesn't exist. There's no differentiation between a 404 file not found and 404 bucket not found, it's all just blanketed under a "404 Not Found" error, so this will attempt to upload to the non-existent bucket 5 times before giving up and returning an error. We could pinpoint if the bucket does really exist and give a better error if we used https://cloud.google.com/storage/docs/json_api/v1/buckets/get ahead of time to determine if it exists. I'd consider this more overhead and not ideal though so I think this is sufficient.

Fixes #436

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -1107,7 +1107,8 @@ File.prototype.startResumableUpload_ = function(stream, metadata) {
   // Handle an error from API calls following the recommended best practices:
   // http://goo.gl/AajKku
   function handleError(err) {
-    if (err.code === 404) {
+    if (err.code === 404 && retries < RETRY_LIMIT) {
+      retries++;
       startUpload();
       return;
     }


stephenplusplus:
Looks good to me.

PR #439, opened 2015-03-11
Only test 0.12.0

ryanseys:
Remove 0.10.0 and iojs testing from travis in hopes that we no longer have any issues with our regression tests so we can release!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
-- 0.10
 - 0.12
-- iojs
 script: ./scripts/build.sh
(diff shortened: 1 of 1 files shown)


stephenplusplus:
This is the start of something amazing.

stephenplusplus:
Or not: https://travis-ci.org/GoogleCloudPlatform/gcloud-node/#L1863
This error looks legit though.

ryanseys:
Hmm... I couldn't reproduce on my local machine. Looks like more race conditions.

ryanseys:
Created #440 in hopes to kill this issue.

PR #440, opened 2015-03-11
Remove possibility of race conditions in pubsub reg tests

ryanseys:
The pubsub API must still suffer from race conditions or limitations because we saw an issue in #439 that resulted in a message being pulled with returnImmediately = true unsuccessfully after being published. :sob: 

This PR forces the regression tests to wait until the number of messages they require are available, at the expense of speed. returnImmediately should probably not be used for regression testing purposes because its functionality is, by definition, unpredictable and based solely on the APIs discretion.

### DIFF
--- a/regression/pubsub.js
+++ b/regression/pubsub.js
@@ -200,7 +200,7 @@ describe('pubsub', function() {
         assert.ifError(err);
 
         subscription.pull({
-          returnImmediately: true,
+          returnImmediately: false,
           maxResults: 1
         }, function(err, msgs) {
           assert.ifError(err);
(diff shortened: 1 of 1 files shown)


stephenplusplus:
What's the timeout for the callbacks to execute and could it be longer than the 20 seconds we allow each test to max out at? Should we bump the test max timeout / cap the pull response time?

ryanseys:
30 seconds?

ryanseys:
It wasn't a timing issue though, I'm not sure how fast it is on Travis but these tests took around 5-6 seconds on my local machine.

stephenplusplus:
Worth a shot, then!

PR #448, opened 2015-03-17
Add apiResponses

### LINKED ISSUE #434: Return raw response object for all API calls
As a way to support the more unique use-cases that developers can come up with that our library doesn't support out of the box (see below for example bugs), and continue to support backward compatibility to our existing users, we'd like to return the raw response object from the API to all API calls. This will additionally provide an less intrusive way for developers to migrate their code from the generic libraries to our more idiomatic library.

Proposal for documented object name in docs: `rawBody`, `rawResponse`, `response`, `resp`, `apiResponse` ?

Addresses #427 
Addresses #342

ryanseys:
Adding `apiResponse` to the callbacks of every API.
- [x] BigQuery
- [x] Datastore
- [x] PubSub
- [x] Storage
- [x] Unit tests
- [x] ~~Regression tests~~ (doesn't make sense)

Fixes #434

### DIFF
--- a/lib/pubsub/topic.js
+++ b/lib/pubsub/topic.js
@@ -143,7 +143,7 @@ Topic.prototype.autoCreateWrapper_ = function(method, path, q, body, callback) {
  * @param {function=} callback - The callback function.
  *
  * @example
- * var registrationMessage = {
+ * var registerMessage = {
  *   data: {
  *     userId: 3,
  *     name: 'Stephen',
@@ -154,7 +154,7 @@ Topic.prototype.autoCreateWrapper_ = function(method, path, q, body, callback) {
  *     hello: 'world'
  *   }
  * };
- * topic.publish(registrationMessage, function(err, messageIds) {});
+ * topic.publish(registerMessage, function(err, messageIds, apiResponse) {});
  *
  * //-
  * // You can publish a batch of messages at once by supplying an array.
@@ -170,7 +170,7 @@ Topic.prototype.autoCreateWrapper_ = function(method, path, q, body, callback) {
  * topic.publish([
  *   registrationMessage,
  *   purchaseMessage
- * ], function(err, messageIds) {});
+ * ], function(err, messageIds, apiResponse) {});
  */
 Topic.prototype.publish = function(messages, callback) {
   messages = util.arrayize(messages);
@@ -192,10 +192,10 @@ Topic.prototype.publish = function(messages, callback) {
   var path = this.name + ':publish';
   this.makeReq_('POST', path, null, body, function(err, result) {
     if (err) {
-      callback(err);
+      callback(err, null, result);
       return;
     }
-    callback(null, result && result.messageIds || []);
+    callback(null, result && result.messageIds || [], result);
   });
 };
 
@@ -205,7 +205,7 @@ Topic.prototype.publish = function(messages, callback) {
  * @param {function=} callback - The callback function.
  *
  * @example
- * topic.delete(function(err) {});
+ * topic.delete(function(err, apiResponse) {});
  */
 Topic.prototype.delete = function(callback) {
   callback = callback || util.noop;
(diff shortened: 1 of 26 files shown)


stephenplusplus:
Looking clean so far :+1:

ryanseys:
Thanks! Do you think this is appropriate for datastore or what approach should we take there?

stephenplusplus:
We do a lot of convenience wrapping of the response (all that `entityTo___` stuff), so I think the decoded response will be the closest thing to a "raw" response.

test/storage/index.js
```diff
@@ -192,8 +192,4 @@ describe('Storage', function() {
       });
     });
   });
-
-  describe('makeReq_', function() {
```
ryanseys: Uhhh?

stephenplusplus: Wat.

ryanseys: I found it like this.

ryanseys: Git blame blames you @stephenplusplus . https://github.com/GoogleCloudPlatform/gcloud-node/blame/master/test/storage/index.js#L196

stephenplusplus: Blames me for what? It looks like this PR removes the tests for makeReq_, and the blame shows where I added them. Feel like I'm definitely missing out on something. Halp.

ryanseys: Sorry, hard to tell given that I commented on the `describe` line but it's an empty `describe` block with no tests in it. I'm just removing it because it's useless. Not sure what happened to those tests if they ever existed in the first place.

ryanseys:
TADA! :fireworks: :rotating_light: 

This PR should be ready for a review. This PR goes down in my books as one of the most tedious and boring PRs of all time. :smile:

ryanseys:
I plan to squash, but I figure it might be easier to review with the nice separate diffs.

ryanseys:
I left out `file#createReadStream` and `file#createWriteStream` because there's many different types of responses or paths that can be taken where the apiResponses aren't really necessary. And we're dealing with responses that probably won't be any use to the end users. In your @stephenplusplus refactor of these methods, it might be nice to consider how the apiResponse objects could be returned with these methods.

lib/pubsub/topic.js
```diff
@@ -169,7 +169,7 @@ Topic.prototype.autoCreateWrapper_ = function(method, path, q, body, callback) {
  * topic.publish([
  *   registrationMessage,
```
stephenplusplus: `registerMessage`

stephenplusplus:
@sindresorhus We've been designing gcloud-node to give users curated, massaged responses from their API calls. The upstream API response can be quite intimidating/non-user friendly, which is why we try to clean it up. However, we've hit a couple places where we forgot to include something and/or didn't want to include it, because it doesn't fit the majority use case. To support cases like these, we want to give users the raw API response in a "it's there if you need it, ignore it if you don't care" kind of way.

This PR is our solution, which tacks on that raw response as the final argument to their callback:

```js
// old
bucket.upload("file.jpg", function(err, file) {})
// new
bucket.upload("file.jpg", function(err, file, apiResponse) {})
```

Quite a bit uglier, but we have methods that return more than just two arguments to the callback:

```js
// old
dataset.get(key, function(err, results, nextQuery) {})
// new
dataset.get(key, function(err, results, nextQuery, apiResults) {})
```

I'm wondering if you've come across APIs that have solved for the same problem, and if our solution seems reasonable. Thanks!

ryanseys:
Seems this has gotten out of sync with master. I'll rebase once we hear back from sindre.

sindresorhus:
> I'm wondering if you've come across APIs that have solved for the same problem, and if our solution seems reasonable. Thanks!

I guess that makes sense. I do that in my [`got`](https://github.com/sindresorhus/got#callbackerr-data-response) module too, and so does [`request`](https://github.com/request/request), albeit different order.

stephenplusplus:
Cool, thanks!

Once the out of sync-ness is resolved, I think we're merge-ready.

ryanseys:
Rebased and squashed. Merge when ready!

stephenplusplus:
Yay!

ryanseys:
Wewt!

sunnychow:
thanks!

PR #449, opened 2015-03-17
/lib/common/ & /lib/pubsub/: cleaned a duplicate line and a useless .bind(this)

Pampattitude:
Mostly about cleaning the actual code a bit.

---

First diff is a duplicate line removal (see `lib/common/util.js` `+331`).

Second one is a useless bind, which might result in slightly degraded perfs on each call.

---

By the way, how far are we allowed to make changes to the code? I'd like to rework a few other parts that may lead to a few milliseconds loss (e.g. error messages constructed from an `array.join()` during a function call). These changes are somewhat trivial, but require clogging the global scope a bit and may result in a slightly less clean code.

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -331,7 +331,6 @@ function makeWritableStream(dup, options, onComplete) {
 
   options = options || {};
   options.metadata = options.metadata || {};
-  onComplete = onComplete || noop;
 
   var boundary = uuid.v4();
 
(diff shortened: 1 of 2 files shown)


Pampattitude:
Done!

ryanseys:
Hey thanks for your contribution! You can make any change you want, though
I'd advise you consult with the maintainers first if you're going to make
some breaking changes or major feature additions. You can simply open an
issue about the changes you want to make and we'll see about the
feasibility of it and get back to you.

In regards to the global scope, just don't put anything in the global
scope. Always use var for initializing your variables. And use the linter
with `npm run lint`. Happy contributing! :)

>

Pampattitude:
Thanks for the tips!

I'll make a new PR with the correct info.

PR #450, opened 2015-03-17
/lib/common/ & /lib/pubsub/: cleaned a duplicate line and a useless .bind(this)

Pampattitude:
Mostly about cleaning the actual code a bit.

---

First diff is a duplicate line removal (see `lib/common/util.js` `+331`).

Second one is a useless bind, which might result in slightly degraded perfs on each call.

---

Note: this is a duplicate of [PR #449](https://github.com/GoogleCloudPlatform/gcloud-node/pull/449) with proper Git configuration (CLA validation would not pass). Sorry!

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -331,7 +331,6 @@ function makeWritableStream(dup, options, onComplete) {
 
   options = options || {};
   options.metadata = options.metadata || {};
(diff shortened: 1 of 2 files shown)


ryanseys:
Thanks @Pampattitude!

PR #451, opened 2015-03-18
Make less buckets and change ACL on new files

ryanseys:
In general, just slow down regression tests and make all ACL tests on different files so we're very likely to not hit the rate limit for modifying a single file a boatload of times.


ryanseys:
Self merging. Revert if necessary.

PR #456, opened 2015-03-22
Do not autoAck when no messages are pulled

### LINKED ISSUE #455: pubsub will throw during Subscription.pull when autoAck is on and no messages avail
I'm porting to v1beta2 of the pubsub service. In recent versions of gcloud-node, my previous code will abort in Subscription.pull with:

    Error: At least one ID must be specified before it can be acknowledged.
        at Subscription.ack     (/Users/jared/work/other/deep/node_modules/gcloud/lib/pubsub/subscription.js:258:11)

I use autoAck and right now there are very few messages on the topic. So I think what's happening is the long-poll completes without receiving a message, and this block in Subscription.pull:

```javascript
  this.makeReq_('POST', path, null, body, function(err, response) {
    if (err) {
      callback(err);
      return;
    }

    var messages = response.receivedMessages || [];
    messages = messages.map(Subscription.formatMessage_);

    if (self.autoAck) {
      var ackIds = messages.map(function(message) {
        return message.ackId;
      });

      self.ack(ackIds, function(err) {
        callback(err, messages);
      });
```

calls self.ack with ackIds empty because receivedMessages is empty. I fixed this in my local copy with:

    if (self.autoAck && messages.length != 0) {

since there's no point in trying to ack if there are no messages. If I had a bit more time right now, I'd offer a PR with a test, but let's get an issue opened at least to see if the problem has been characterized correctly first.

ryanseys:
Fixes #455

### DIFF
--- a/lib/pubsub/subscription.js
+++ b/lib/pubsub/subscription.js
@@ -360,7 +360,7 @@ Subscription.prototype.pull = function(options, callback) {
     var messages = response.receivedMessages || [];
     messages = messages.map(Subscription.formatMessage_);
 
-    if (self.autoAck) {
+    if (self.autoAck && messages.length !== 0) {
       var ackIds = messages.map(function(message) {
         return message.ackId;
       });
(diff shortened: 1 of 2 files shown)


jaredrhine:
LGTM.

PR #459, opened 2015-03-23
Use PATCH when setting BigQuery dataset metadata

### LINKED ISSUE #454: Access Denied: Cannot delete BigQuery dataset in regression test.
It seems that we suddenly have lost permission to delete BigQuery datasets in our regression tests. It doesn't work on local or Travis.

Error is as follows:

```
BigQuery "after all" hook:
     Error: Access Denied: Dataset long-door-651:gcloud_test_dataset_temp187832e0_XXX: 
The user USER@developer.gserviceaccount.com does not have permission to delete 
dataset PROJECT_ID:gcloud_test_dataset_temp187832e0_XXX
```

https://travis-ci.org/GoogleCloudPlatform/gcloud-node/builds/55070457

Google Status shows no issues and all functioning normally. https://status.cloud.google.com/

ryanseys:
Fixes #454

### DIFF
--- a/lib/bigquery/dataset.js
+++ b/lib/bigquery/dataset.js
@@ -244,7 +244,7 @@ Dataset.prototype.query = function(options, callback) {
 Dataset.prototype.setMetadata = function(metadata, callback) {
   var that = this;
 
-  this.makeReq_('PUT', '', null, metadata, function(err, resp) {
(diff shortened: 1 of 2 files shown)


stephenplusplus:
Good detective work, sir!

Issue #445, opened 2015-03-13
Automatically create topic / subscription if not found.


jonparrott:
Presently I have some code to publish to a particular topic, creating it if necessary:

```javascript
function publish(message, cb){
  topic.publish(message, function(err){
    if(err && err.code == 404){
      logging.info("Creating pub/sub topic.");
      pubsub.createTopic(topicName, function(err, _){
        if(err) return cb(err);
        publish(message, cb);
      });
    } else {
      cb(err);
    }
  });
}
```

And similar code to listen for messages to a topic. It goes the other way around - trying to create one at first then re-using if it's already there.

```javascript
function subscribe(cb){
  topic.subscribe(subscriptionName, function(err, subscription){
    if(!err) return wireSubscription(subscription);
    if(err && err.code == 409) return wireSubscription(topic.subscription(subscriptionName));
    throw err;
  });
  ...
}
```

According to [this](https://github.com/GoogleCloudPlatform/gcloud-node/issues/98#issuecomment-51552633) comment, it seems that the intended usage is to create the topic and subscription outside of the application content and never touch them afterwards. It's not clear in the documentation that this is expected.

I think it would be great to either:
 1. Consolidate subscribe/subscription createTopic/topic into combined create/reuse functions (e.g. .topic() .subscription()). There can be an optional argument for erring instead of re-using. This hides all of this somewhat ugly logic from users that *do* want this use-case.
 2. Explicitly document that topics and subscriptions should be created beforehand.

I would personally prefer the former as it removes the need to write a "bootstrapping" script and simplifies the user-facing API.

stephenplusplus:
Do you mean a Topic and Subscription object should create itself if necessary after you attempt to use it? I'm ok with this idea, but what's the trouble with manually creating the topic/sub before using it?

jonparrott:
Yes. The trouble with manually creating it before use leads to having to write unnecessary code to handle creating it if necessary- unless you meant creating it with a bootstrap script beforehand outside of the application context. In that case, it's just inconvenient for no real benefit. I could see myself forgetting to create new subscriptions/topics while iterating new features.

ryanseys:
The approach we have taken for every object in this library is that you can create one or reference an existing one. If we changed this, it would need to be changed everywhere and would require two API requests rather than one, causing increased delays for responses. 

[...]

PR #465, opened 2015-03-25
Add option to automatically create a topic and re-use a subscription.

theacodes:
This adds `autoCreate` to `pubsub.topic`:

``` js
var topic = pubsub.topic('does-not-yet-exist', {autoCreate: true});
// topic will be created on first RPC.
topic.publish({data: {hello: 'world'});
```

And `reuseExisting` to `topic.subscribe`:

``` js
topic.subscribe('existingSub', {reuseExisting: true}, function(err, sub){ ... }));
```

Combined, these two features completely eliminate any bootstrapping code/steps required for the Getting Started on GCP sample.

### DIFF
--- a/test/pubsub/topic.js
+++ b/test/pubsub/topic.js
@@ -166,7 +166,7 @@ describe('Topic', function() {
 
     it('should generate 404 error without autoCreate', function(done) {
       topic.makeReq_ = function(method, path, query, body, callback) {
-        callback({code: 404});
+        callback({ code: 404 });
       };
 
       topic.publish(messageObject, function(err){
@@ -183,7 +183,7 @@ describe('Topic', function() {
 
       acTopic.origMakeReq_ = function(method, path, query, body, callback) {
         if (!created) {
-          callback({code: 404});
+          callback({ code: 404 });
         }
         else {
           callback(null);
@@ -321,7 +321,7 @@ describe('Topic', function() {
         };
 
         topic.makeReq_ = function(method, path, qs, body, callback){
-          callback({code: 409});
+          callback({ code: 409 });
         };
 
         topic.subscribe(SUB_NAME, function(err) {


lib/pubsub/index.js
```diff
@@ -196,18 +196,24 @@ PubSub.prototype.createTopic = function(name, callback) {
...
  * @param {string} name - The name of the topic.
+ * @param {object=} options - Configuration object.
+ * @param {bool=} options.autoCreate - Automatically create topic if
+ *     it doesn't exist. Note that messages published to a topic with
+ *     no subscribers will not be delivered.
  * @return {module:pubsub/topic}
  *
  * @example
  * var topic = pubsub.topic('my-existing-topic');
  * topic.publish('New message!');
```
ryanseys: Could you add an example here showing how it would work?

lib/pubsub/topic.js
```diff
@@ -96,6 +102,37 @@ Topic.formatName_ = function(projectId, name) {
...
+      if (err) {
+        callback(err);
+        return;
+      }
+      self.origMakeReq_(method, path, q, body, callback);
+    });
+  };
+
+  this.origMakeReq_(method, path, q, body, function(err, res) {
+    if (method !== 'DELETE' && err && err.code === 404) {
```
tmatsuo: Can method !== 'DELETE' be the last condition?

jonparrott:
Fixed the lint issues and updated after the review.
Also, forgot to mention in the PR - this fixes #445.

lib/pubsub/topic.js
tmatsuo: ping?
If we changed the if clause to:
```javascript
if (err && err.code === 404 && method !== 'DELETE') {
}
```
We can remove one comparison in normal cases. gcloud-node already has lots of abstraction and I feel it's slower than expected (see https://github.com/GoogleCloudPlatform/gcloud-node/issues/441), this single change might not contribute much in terms of performance, but it is important to always keep performance-wise implementation in mind, otherwise we'll end up having such small inefficiency everywhere and it will get slower and slower.

lib/pubsub/topic.js
```diff
@@ -276,7 +313,10 @@ Topic.prototype.subscribe = function(name, options, callback) {
 
   var path = Subscription.formatName_(this.projectId, name);
   this.makeReq_('PUT', path, null, body, function(err) {
-    if (err) {
+    if (options.reuseExisting && err && err.code === 409) {
```
ryanseys: What is `options.reuseExisting` ?

lib/pubsub/topic.js
tmatsuo: Ah, sorry I was referring to the outdated diff! It's already changed.
LGTM, but please wait for the maintainers' response.

lib/pubsub/topic.js
jonparrott: Updated with docs.

ryanseys: Is it used anywhere? Test or otherwise?

jonparrott: It's not in the tests. Honestly, the fallback path is to call topic.subscription(). We can add a test that'll throw a 409 but it's hard to tell if .subscription was called from the 409 branch or the 200 branch.

ryanseys: Just add a test for both. In our unit tests, both should be covered. Coverage tool will determine that the branch was taken.

ryanseys: I'm assuming 409 Conflict will only occur when the subscription already exists?

jonparrott: Right.
I'll ask the test tomorrow. Thanks.

jonparrott: Done.

ryanseys:
Overall LGTM. One stylistic nitpick: we add spaces between our keys in an object.

Like this: `{ hello: 'world' }` not like this: `{hello: 'world'}`

stephenplusplus:
LGTM. Thanks for doing this!

ryanseys:
Merged in 3f8f60fd404a1bc6e0837356c6af8ac2f09f1b43

matthewmueller:
I wasn't able to get this working from the example you provided. Here's my code:

```js
const PubSub = require('@google-cloud/pubsub')

const pubsub = PubSub({
  projectId: require('./cred.json').project_id,
  credentials: require('./creds.json')
})

const topic = pubsub.topic('hi-there', { autoCreate: true })

topic.publish({
  data: 'New message!'
}, function(err) {
  console.log(err)
});
```

`pubsub.topic` doesn't actually take an 2nd options parameter

stephenplusplus:
A similar example came up on StackOverflow (http://stackoverflow.com/questions/38864043/gcloud-check-if-a-topic-exist-and-ability-to-reuse-the-topic), so I'll quote my response there:

--

I believe the problem you'll run into is that if a message is published to a topic that doesn't exist, it is immediately dropped. So, it won't hang around and wait for a subscription to be created; it'll just disappear.

However, gcloud-node does have methods that will create a topic if necessary:

```js
var topic = pubsub.topic('topic-that-maybe-exists');
topic.get({ autoCreate: true }, function(err, topic) {
  // topic.publish(...
});
```

In fact, almost all gcloud-node objects have the get method that will work the same way as above, i.e. a Pub/Sub subscription or a Storage bucket or a BigQuery dataset, etc.

Here's a link to the topic.get() method in the docs: https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.37.0/pubsub/topic?method=get

gustawdaniel-acaisoft:
I have error

```
TS2345: Argument of type '{ autoCreate: boolean; }' is not assignable to parameter of type 'PublishOptions'.   Object literal may only specify known properties, and 'autoCreate' does not exist in type 'PublishOptions'.
```

gustawdaniel-acaisoft:
these are publish options

```
export interface PublishOptions {
    batching?: BatchPublishOptions;
    flowControlOptions?: FlowControlOptions;
    gaxOpts?: CallOptions;
    messageOrdering?: boolean;
    enableOpenTelemetryTracing?: boolean;
}
```

PR #468, opened 2015-03-28
Exponential backoff and retry certain requests

### LINKED ISSUE #460: Automatic retry request for all rate limit responses
We should have a global setting to automatically retry requests (with exponential back-off) that receive a rate limit response. It should be turned on by default, have a retry limit of 3-5 that is configurable and additionally be configured to be turned off altogether.

Something like:

```js
var gcloud = require('gcloud')({ retryLimit: 5 });
```

and

```js
var gcloud = require('gcloud')({ autoRetry: false });
```

We should be clear that this setting is only for responses related to rate limits being reached and not that it will retry every request regardless of response error.

ryanseys:
This should work. No regression tests because proof it works is our regression tests will no longer suffer from rate limit errors! :see_no_evil: 

gcloud config now accepts: `autoRetry: false` (default: `true`) to turn off and `maxRetries: 5` (default: `3`) to change number of retries.

All retries are exponentially backed-off. That is:

```
1st retry: 1000-2000 ms delay
2nd retry: 2000-3000 ms delay
3rd retry: 4000-5000 ms delay (default: this is the last try by default)
4th retry: 8000-9000 ms delay
...
```

Fixes #460

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -409,8 +409,39 @@ function makeWritableStream(dup, options, onComplete) {
 
 module.exports.makeWritableStream = makeWritableStream;
 
+/**
+ * Returns an exponential distributed time to wait given the number of retries
+ * that have been previously been attempted on the request.
+ *
+ * @param  {number} retryNumber - The number of retries previously attempted.
+ * @return {number} An exponentially distributed time to wait E.g. for use with
+ *     exponential backoff.
+ */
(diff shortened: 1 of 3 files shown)


ryanseys:
PS I'm so excited for this! :lollipop: :candy: :moneybag:

stephenplusplus:
This will go down as one of the best PRs gcloud has ever had.

PR #471, opened 2015-03-31
Do not run regression tests on tagged release commits

ryanseys:
If our release is passing, and we want to release, then we shall release.


scripts/build.sh
```diff
@@ -22,8 +22,12 @@ npm run test
 # if merging to master and not a pull request, execute regression tests, create coverage report and update docs
 if [ "${TRAVIS_BRANCH}" == "master" ] && [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
   openssl aes-256-cbc -K $encrypted_b8aa0887832a_key -iv $encrypted_b8aa0887832a_iv -in key.json.enc -out key.json -d
-  # create new coverage report (executes regression tests)
-  npm run coveralls
+
+  # Only run regression tests on non-release (non-tagged) commits.
+  if [ "${TRAVIS_TAG}" == "" ]; then
```
ryanseys: I'm not 100% sure this works, but I have no better idea.

ryanseys:
Closing for now. We might not need it. But release this Kragle if going gets tough.

PR #475, opened 2015-04-01
docs: fix markup for dox.

stephenplusplus:
Dox was failing due to some markup mistakes in our JSDocs.

Self-merging, as this is quite trivial. Just wanted it to be on record, in case we run into something similar down the road.

### DIFF
--- a/lib/bigquery/index.js
+++ b/lib/bigquery/index.js
@@ -96,7 +96,7 @@ var SCOPES = ['https://www.googleapis.com/auth/bigquery'];
  * //-
  * // In the following examples from this page and the other modules (Dataset,
  * // Table, etc.), we are going to be using a dataset from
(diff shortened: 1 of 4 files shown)


lib/bigquery/index.js
```diff
@@ -96,7 +96,7 @@ var SCOPES = ['https://www.googleapis.com/auth/bigquery'];
  * //-
  * // In the following examples from this page and the other modules (Dataset,
  * // Table, etc.), we are going to be using a dataset from
- * // [data.gov](http://goo.gl/f2SXcb) of higher education institutions.
+ * // data.gov (http://goo.gl/f2SXcb) of higher education institutions.
```
stephenplusplus: (Links don't render in example blocks)

PR #490, opened 2015-04-09
Storage: add https support to getSignedUrl

mziccard:
Added `https` field to getSignedUrl `options` parameter.
If `https` is set the returned url uses HTTPS.
Otherwise we stick to HTTP.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -786,14 +786,8 @@ File.prototype.getSignedUrl = function(options, callback) {
     ].join('\n'));
     var signature = sign.sign(credentials.private_key, 'base64');
 
-    var base = 'http://storage.googleapis.com';
-
-    if (options.https) {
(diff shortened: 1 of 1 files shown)


ryanseys:
You'll have to lint your changes to ensure they meet the requirements. In this case it appears you missed a semi-colon.

Regarding this PR, I'm wondering why we don't just use HTTPS by default. Thoughts @stephenplusplus ?

stephenplusplus:
Always https sounds good to me :+1:

mziccard:
Sounds perfect to me.

ryanseys:
@mziccard Would you mind updating this PR to just use https always? :)

mziccard:
In the end, we just added an 's' :)

stephenplusplus:
Merged via 3ce39f676cf2d7290e0157fd5ef07436d8851989.

Thanks, @mziccard!

PR #492, opened 2015-04-13
support uploading directories

### LINKED ISSUE #412: support globs in `upload`
I have error Error: EISDIR

Is it  relevance with upload folder features?
Can I request that feature? if not, could you explain me why?

Thanks

stephenplusplus:
Fixes #412
#### To Dos
- [x] Unit Tests
- [x] System Tests
- [x] Docs

This is a work in progress, but is an attempt to make a sane API for uploading directories.

```
zoo/
  chipmunks/
    chipmunk-1.jpg
  pandas/
    panda-1.jpg
  zoo-photos.zip
```

``` js
bucket.uploadDirectory("zoo", function (err, files) {
  files = [
    File("chipmunks/chipmunk-1.jpg"),
    File("pandas/panda-1.jpg"),
    File("zoo-photos.jpg")
  ]
})
```
#### Globs

`bucket.uploadDirectory` is a wrapper around `bucket.upload`, which now supports globs:

``` js
bucket.upload("zoo/**/*.jpg", function (err, files) {
  files = [
    File("chipmunk-1.jpg"),
    File("panda-1.jpg")
  ]
})
```

Note that the resulting file names aren't prefixed with a directory path (e.g. "chipmunks/chipmunk-1.jpg" vs "chipmunk-1.jpg"). Since a glob can include several files and directories in various locations on a hard drive, by default, just the filename is used as the uploaded file's name.

By specifying `options.basePath`, a user can tell us the common parent of all of the locations their glob matches, which is then used as a prefix for the file. This is how `bucket.uploadDirectory` works (simplified):

``` js
Bucket.prototype.uploadDirectory = function (directoryPath) {
  bucket.upload(path.join(directoryPath, "**/*"), { basePath: directoryPath })
}
```
#### Extending globs

bucket.upload accepts either a single glob or array of globs. It's passed directly to [sindresorhus/globby](https://github.com/sindresorhus/globby), which uses [isaacs/minimatch](https://github.com/isaacs/minimatch) and [isaacs/node-glob](https://github.com/isaacs/node-glob) to return the matching paths. All [minimatch glob patterns](https://github.com/sindresorhus/globby#globbing-patterns) are accepted, as well as [node-glob options](https://github.com/isaacs/node-glob#options) via `options.globOptions`.

### DIFF
--- a/regression/storage.js
+++ b/regression/storage.js
@@ -22,6 +22,7 @@ var assert = require('assert');
 var async = require('async');
 var crypto = require('crypto');
 var fs = require('fs');
+var path = require('path');
 var request = require('request');
 var through = require('through2');
 var tmp = require('tmp');
@@ -301,7 +302,7 @@ describe('storage', function() {
         var options = {
           destination: uuid.v1() + '.png'
         };
-        bucket.upload(files.logo.path, options, function(err, f) {
+        bucket.uploadFile(files.logo.path, options, function(err, f) {
           assert.ifError(err);
           file = f;
           done();
@@ -490,7 +491,7 @@ describe('storage', function() {
     });
 
     it('should read a byte range from a file', function(done) {
-      bucket.upload(files.big.path, function(err, file) {
+      bucket.uploadFile(files.big.path, function(err, file) {
         assert.ifError(err);
 
         var fileSize = file.metadata.size;
@@ -516,7 +517,7 @@ describe('storage', function() {
     it('should download a file to memory', function(done) {
       var fileContents = fs.readFileSync(files.big.path);
 
-      bucket.upload(files.big.path, function(err, file) {
+      bucket.uploadFile(files.big.path, function(err, file) {
         assert.ifError(err);
 
         file.download(function(err, remoteContents) {
@@ -545,7 +546,7 @@ describe('storage', function() {
           resumable: false
         };
 
-        bucket.upload(files.logo.path, options, function(err, file) {
+        bucket.uploadFile(files.logo.path, options, function(err, file) {
           assert.ifError(err);
 
           file.getMetadata(function(err, metadata) {
@@ -631,7 +632,7 @@ describe('storage', function() {
     });
 
     it('should copy an existing file', function(done) {
-      bucket.upload(files.logo.path, 'CloudLogo', function(err, file) {
+      bucket.uploadFile(files.logo.path, 'CloudLogo', function(err, file) {
         assert.ifError(err);
         file.copy('CloudLogoCopy', function(err, copiedFile) {
           assert.ifError(err);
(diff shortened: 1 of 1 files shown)


stephenplusplus:
@ryanseys - If you can take a quick look at the first post to see if this is a useful addition, I'll start working on the tests for these to get it ready for code review & merge.

ryanseys:
Yes this looks ok. What approach is taken when filenames conflict. E.g. `/img/chipmunks/chippy-1.jpg` and `/img/woodchippers/chippy-1.jpg` will both be uploaded as `chippy-1.jpg` when manually using `bucket.upload('/img/**/*.jpg', function(err, files) {});` Does this just mean that one will be overwritten or can we somehow tell `bucket.upload` to be smart about it so that nothing is overwritten? 

I'm thinking maybe we don't even need `bucket.uploadDirectory` (???), why not just fs.stat to figure out if the name given is a directory, and then take the approach of using basePath in that case. If they give a glob, maybe they can specify a { flatten: true } to explictly say they want to overwrite files, otherwise upload and maintain the structure from their base path?

Sorry if this doesn't make sense.

ryanseys:
Or just always maintain structure when given a directory. If given a glob, always flatten. Seems reasonable.

ryanseys:
I think that's probably what you're doing already except its strewn across upload and uploadDirectory where I'm not convinced we need both.

stephenplusplus:
> What approach is taken when filenames conflict

The last one will win.

> I'm thinking maybe we don't even need bucket.uploadDirectory (???), why not just fs.stat to figure out if the name given is a directory, and then take the approach of using basePath in that case.

I like this!

ryanseys:
If it makes the logic of upload infinitely more difficult, I'd say it's not worth it, so if this is already effectively doing the same thing as:

> just always maintain structure when given a directory. If given a glob, always flatten.

then I don't see any reason to change it now. But if it's easy to change, that'd be cool :)

stephenplusplus:
I played around with it for a bit, but I couldn't find a way to make the change elegantly. For now, I'll continue with the two methods (`upload` for globs, `uploadDirectory` for directories), and maybe revisit later if a solution strikes me.

ryanseys:
No worries. Might be beneficial and more intuitive for the users anyway. At
least that can be our excuse :)

stephenplusplus:
Ready for a closer look :eyeglasses:

lib/storage/bucket.js
```diff
@@ -723,68 +734,255 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
...
+          } else {
+            files.push(file);
+          }
+
+          next(options.force ? null : err);
+        });
+      };
+    });
+
+    async.parallelLimit(uploadFileFns, MAX_PARALLEL_UPLOADS, function() {
```
ryanseys: Can you just use `async.eachLimit(filePaths, MAX_PARALLEL_UPLOADS, function(next) {}, function(){});` (where `function(next) {}` is the function returned by the map call, that way for N files, we don't have to create an array of N functions which we then iterate through, we just create one function and reuse it for each of the filenames?

stephenplusplus: Yes! Much nicer idea. I get overwhelmed by the async API.

ryanseys: Yeah, it's convoluted.

lib/storage/bucket.js
```diff
@@ -723,68 +734,255 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
...
+ * });
+ *
+ * //-
+ * // You can also specify a `basePath` if you need more control.
+ * //-
+ * var options = {
+ *   basePath: '/Users/stephen/Photos';
+ * };
+ *
+ * bucket.uploadDirectory(zooPhotosPath, options, function(err, files) {
```
ryanseys: What would make more sense here is:

``` js
var options = {
  remotePath: 'zoo'
}
bucket.uploadDirectory('/Users/stephen/Photos/zoo', options, callback);
```

That would upload all files in `'/Users/stephen/Photos/zoo'` to the `'zoo'` folder in my bucket. This would accomplish the same thing as the above example but I believe it's more clear.

stephenplusplus: Definitely more clear :+1: I will work on implementing this. Support for `remotePath` in `upload` and `uploadFile` would be nice, as well.

ryanseys: Oh I forgot about `destinationPath` which I think is the same thing? Unless that also specifies the name of the file? `remotePath` would just be the folder portion of the file path.

stephenplusplus: What's destinationPath? I suggested it earlier in our chat just as another name idea for remotePath.

stephenplusplus: The other methods have "destination" which is a string file name or File object. It would definitely be overwhelming to have destination and destinationPath. Maybe "destinationDirectory" is the clearest option. Or just basePath :)

ryanseys: Yeah basePath with destination seems like best option after thinking about it for a bit. That means they can specify their entire path technically in destination but for uploading globs, a basePath is useful.

stephenplusplus: Methods that deal with multiple files don't accept destination, since that [historically] refers to a single file. So, uploadDirectory and upload just accept basePath, while uploadFile accepts destination.

At the end of the day, a user might want to handle all of this logic on their own, directly using uploadFile or even the bare-bones createWriteStream. It's not too late to scrap the PR and just let the user handle all this themselves :)

stephenplusplus: It's too bad this is so complicated, since it's such a simple premise... upload a directory into my bucket. Maybe I should pull out the glob stuff, and just use `uploadFile` and `uploadDirectory` with `upload` doing a stat to see which way to send the incoming file path?

ryanseys: I think we should provide it because it is complicated.

stephenplusplus:
(cont'd from https://github.com/GoogleCloudPlatform/gcloud-node/pull/492#discussion_r28647097)

After much thought and attempts to implement this, I don't think we need to support uploading directories or globs.

It's quite straightforward for a user to implement uploading a directory in their app:

```js
var async = require("async")
var fs = require("fs")
var storage = require("gcloud")

var myBucket = gcloud.storage().bucket("pics")
uploadDirectory(myBucket, "/Users/stephen/Photos", function (err, files) {})

function uploadDirectory(bucket, directory, callback) {
  fs.readdir(directory, function (err, files) {
    async.map(files, bucket.upload, callback)
  })
}
```

If they want other things to happen, like uploading the files into a new directory in their bucket, recursing through child directories and maintaining that hierarchy, etc., their code is the best place to handle that. Our supporting it It leads to either too much guess work or confusing options.

ryanseys:
Globs I agree are a little fancy and require weird options but just
uploading a path where the path is a directory should be easy to support.
May I take a stab at it? :) I'll work off of what you've already created.

stephenplusplus:
Sure, your code has yet to leave me unimpressed!

stephenplusplus:
I still think this is outside the scope of what our lib should support. We allow uploading files via `upload()` and a writable stream. With those two things, a user who wants to mass upload multiple files can make it happen using any utility library / native looping mechanism they choose. I'm going to close as I don't think this is a large priority.

@ryanseys with all of your free time, if you're still interesting in taking a stab at a smaller implementation of this, please feel free :)

ryanseys:
> @ryanseys with all of your free time

Hahaha!

Like you said in #412, I believe this is not too difficult for a user to implement given our library and would allow them more flexibility than we could provide through our library in terms of handling errors and other responses on the fly.

PR #496, opened 2015-04-13
Fixed bug in BigQuery's table.import() callback invocation

### LINKED ISSUE #495: BigQuery's table.import() callback is invoked improperly
If you pass a string filename into BigQuery's `table.import()` method, the callback is invoked with the improper arguments. The documentation says the callback should be called with this format:

```
function(error, job) { ... }
```

But it is actually invoked like this:

```
function(job) { ... }
```

This is caused by improperly handling the `complete` event of `createWriteStream()` on [this line](https://github.com/GoogleCloudPlatform/gcloud-node/blob/f9cab1ec380f1a22256018d2aed1a9b89b326dda/lib/bigquery/table.js#L635). The complete event fires with no `error` argument, but instead with the `job` as the first argument. So just forwarding the arguments onto `callback` results in the undocumented behavior.

A PR with a fix is coming right up!

jwngr:
Fixes #495. Details are provided there.

Google ldap: jwenger

### DIFF
--- a/test/bigquery/table.js
+++ b/test/bigquery/table.js
@@ -698,16 +698,22 @@ describe('BigQuery/Table', function() {
     }, 'file.json');
 
     it('should accept just a File and a callback', function(done) {
+      var mockJob = { id: 'foo' };
+
       table.createWriteStream = function() {
         var ws = new stream.Writable();
         setImmediate(function() {
-          ws.emit('complete');
+          ws.emit('complete', mockJob);
           ws.end();
         });
         return ws;
       };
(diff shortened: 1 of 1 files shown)


jwngr:
Let me know if I actually need to sign the CLA. I would imagine not since I am a Google employee. My ldap is jwenger.

ryanseys:
Probably should anyway just to shut up the Google bot. I know you're already under NDA.

ryanseys:
LGTM.

PR #497, opened 2015-04-14
datastore: support local devserver automatically - fixes #104

### LINKED ISSUE #104: Support App Engine local server

stephenplusplus:
This checks for the presence of a `DATASTORE_HOST` environment variable, used with gcd: https://cloud.google.com/datastore/docs/tools/devserver.


ryanseys:
LGTM! :)

PR #502, opened 2015-04-15
storage: add getSignedPolicy function to generate a signed policy

mziccard:
Added a function to generate a signed policy for POST requests.
So far it supports several conditions (according to the [docs](http://goo.gl/JWJEkG)):
- Equality: `["eq", "$<field>", "<value>"]`
- Starts with: `["starts-with", "$<field>", "<value>"]`
- Content length: `["content-length-range", <min_range>, <max_range>]`
- Bucket: `{"bucket": "<bucket_name>"}`
- Acl: `{"acl": "<acl_name>"}`
- Success redirect: `{"success_action_redirect": "<url>"}`
- Success status: `{"success_action_status": "<status>"}`

Documentation is provided describing function arguments.
Unit tests are added.

### DIFF
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -1164,7 +1164,6 @@ describe('File', function() {
           expiration: expiration,
           startsWith: [['$<field>', '<value>']]
         }, function(err, signedPolicy) {
-          console.log(signedPolicy);
           var conditionString = '[\"starts-with\",\"$<field>\",\"<value>\"]';
           assert.ifError(err);
           assert(signedPolicy.string.indexOf(conditionString) > -1);
@@ -1178,7 +1177,6 @@ describe('File', function() {
           expiration: expiration,
           startsWith: ['$<field>', '<value>']
         }, function(err, signedPolicy) {
-          console.log(signedPolicy);
           var conditionString = '[\"starts-with\",\"$<field>\",\"<value>\"]';
           assert.ifError(err);
           assert(signedPolicy.string.indexOf(conditionString) > -1);


lib/storage/file.js
```diff
@@ -796,6 +796,145 @@ File.prototype.getSignedUrl = function(options, callback) {
 };
 
 /**
+ * Get a signed policy document to allow user to upload data
+ * with a POST.
+ *
+ * *[Reference](http://goo.gl/JWJEkG).*
+ *
+ * @throws {Error} if an expiration timestamp from the past is given or 
+ *     option parameter does not respect
```
ryanseys: does not respect what?

lib/storage/file.js
```diff
@@ -796,6 +796,145 @@ File.prototype.getSignedUrl = function(options, callback) {
...
+      if (!(startCondition instanceof Array) || startCondition.length !== 2) {
+        throw new Error('StartsWith condition must be an array of 2 elements.');
+      } else {
+        conditions.push(['starts-with', startCondition[0], startCondition[1]]);
+      }
+    }
+  }
+
+  // Eq condition for object name
+  conditions.push(['eq', '$key', this.name]);
```
ryanseys: Is this required or is there any way that getSignedPolicy can be made for an arbitrary file name to a given bucket? Sorry for my naive question, I'm not too familiar with this functionality yet.

lib/storage/file.js
```diff
@@ -796,6 +796,145 @@ File.prototype.getSignedUrl = function(options, callback) {
...
+ * *[Reference](http://goo.gl/JWJEkG).*
+ *
+ * @throws {Error} if an expiration timestamp from the past is given or 
+ *     option parameter does not respect
+ *
+ * @param {object} options - Configuration object.
+ * @param {object} options.expiration - Timestamp (seconds since epoch) 
+ *     when this policy will expire.
+ * @param {object[][]=} options.equals - Array of request parameters and 
+ *     their expected value (e.g. [["$<field>", "<value>"]]). Values are
```
ryanseys: If the `$` is required in `$<field>` then can it be omitted and added ourselves? Would this make sense from a user-experience point of view or is it more understandable if the `$` is left in place?

lib/storage/file.js
mziccard: "the expected format". Just fixed.

test/storage/file.js
```diff
@@ -1041,6 +1041,174 @@ describe('File', function() {
...
+        file.getSignedPolicy({
+          expiration: expiration,
+          equals: [['$<field>', '<value>']]
+        }, function(err, signedPolicy) {
+          var conditionString = '[\"eq\",\"$<field>\",\"<value>\"]';
+          assert.ifError(err);
+          assert(signedPolicy.string.indexOf(conditionString) > -1);
+          done();
+        });
+      });
```
ryanseys: For consistency, spaces between these tests would be great to add.

lib/storage/file.js
mziccard: getSignedPolicy can be made for an arbitrary file inside a bucket. However, I would prefer to have it (also?) as a method of a file, in order to be consistent with getSignedUrl or createWriteStream.

lib/storage/file.js
mziccard: The `$` is required. As a user I would love to get rid of it. However, for a user coming from the official [docs](http://goo.gl/JWJEkG) removing it might be counter-intuitive.

ryanseys: My thoughts exactly. I'd consider leaving it in just to remain more consistent with the official docs.

lib/storage/file.js
ryanseys: Perhaps this logic could be moved to a method in Bucket by default, and then added here to File where we reuse that logic in Bucket and add this additional name constraint? Does that make sense or am I totally off the wall here with that idea?

mziccard: We could do it but I feel that is the opposite of what the library does so far.
Let me try to explain myself with an example but please correct me if I am wrong. Uploading a file can be done for an arbitrary file in a bucket. Now Bucket has a convenience method `upload` that takes the path to the local file and the remote file as either an instance of storage File or a string. Such `upload` function is implemented using the logic provided by File's `createWriteStream`.
My point is: we can keep the logic in File and provide a convenience method in Bucket.
I'm cool with any of the two approaches:)

ryanseys: That makes more sense than what I was thinking :) adding that to bucket to
take an arbitrary name would be a great addition.

ryanseys: Wait, I think if we do that, it's no different than just creating a file
and passing it a name. I'm wondering if the name can be omitted altogether?

mziccard: Uh now I got what you were trying to say, sorry. According to the docs on policies you can omit the equality condition on the name, so creating a name-agnostic policy should be possible. However, I really cannot imagine why one would do that: a malicious user intercepting a name-agnostic policy could be able to upload any content to the bucket respecting the policy constraints until expiration. If the name is specified then a malicious user is "only" able to update the file specified. 
However, if we want to allow name-agnostic policies then we should move the logic to Bucket, as you suggested.

stephenplusplus:
All looks good to me. Thanks for adding this @mziccard!

@ryanseys anything left to resolve before merge? I'll handle the small style issues, so we don't bug him with that :)

mziccard:
@stephenplusplus Feel free to bug me:)

BTW, I think it is still open for discussion whether to move this method to Bucket (allowing signed policies not associated to a specific filename) or not. 

Adding a regression test might be useful as well.

lib/storage/file.js
```diff
@@ -796,6 +796,145 @@ File.prototype.getSignedUrl = function(options, callback) {
...
+File.prototype.getSignedPolicy = function(options, callback) {
+  if (options.expiration < Math.floor(Date.now() / 1000)) {
+    throw new Error('An expiration date cannot be in the past.');
+  }
+
+  var expirationString = new Date(options.expiration).toISOString();
+  var conditions = []; 
+
+  // Populate eq conditions
+  if (options.equals && options.equals instanceof Array) {
```
stephenplusplus: For these, would you mind just doing:

```js
if (util.is(options.condition, 'array')) {
  options.condition.forEach(function(condition) {
    if (!util.is(condition, 'array') || condition.length !== 2) {
      // throw ...
    }

    // conditions.push...
  });
}
```

(It's not too late if you want to make me do these things :))

lib/storage/file.js
```diff
@@ -796,6 +796,145 @@ File.prototype.getSignedUrl = function(options, callback) {
...
+
+  // Eq condition for object name
+  conditions.push(['eq', '$key', this.name]);
+
+  // The bucket that will hold the object
+  conditions.push({'bucket': this.bucket.name});
+  
+  // The ACL for the file, if any provided
+  if (options.acl) {  
+    conditions.push({'acl': options.acl});
```
stephenplusplus: No need to quote the key (for this and some other ones that follow). Also, please use a new line:

```js
conditions.push({
  acl: options.acl
});

conditions.push({
  success_action_redirect: options.successRedirect
});

// ...
```

lib/storage/file.js
```diff
@@ -796,6 +796,145 @@ File.prototype.getSignedUrl = function(options, callback) {
...
+        conditions.push(['starts-with', startCondition[0], startCondition[1]]);
+      }
+    }
+  }
+
+  // Eq condition for object name
+  conditions.push(['eq', '$key', this.name]);
+
+  // The bucket that will hold the object
+  conditions.push({'bucket': this.bucket.name});
```
stephenplusplus: Can you place this in the original declaration of `var conditions = [];`?:

```js
var conditions = [
  {
    bucket: this.bucket.name
  }
];

File: lib/storage/file.js [hunk 8]
```diff
@@ -796,6 +796,145 @@ File.prototype.getSignedUrl = function(options, callback) {
...
+  if (options.successStatus) {
+    conditions.push({'success_action_status': options.successStatus});
+  }
+
+  // Range for the content length
+  if (options.contentLengthRange) {
+    var min = options.contentLengthRange.min;
+    var max = options.contentLengthRange.max;
+    if (typeof min === 'undefined' || 
+      typeof max === 'undefined') {
```
stephenplusplus: Maybe check if they are numbers:
```js
if (!util.is(min, 'number') || !util.is(max, 'number')) {
  throw new Error('ContentLengthRange must have numeric min and max fields.');
}
```

lib/storage/file.js
```diff
@@ -796,6 +796,145 @@ File.prototype.getSignedUrl = function(options, callback) {
...
+
+    var signature = sign.sign(credentials.private_key, 'base64');
+
+    callback(null, {
+      string: policyString,
+      base64: policyBase64,
+      signature: signature  
+    });
+  });
+};
```
stephenplusplus: I think you can remove the comments you used in this method. Your architecture of the variable names and logical handling of the input is self-documenting :)

lib/storage/file.js
```diff
@@ -796,6 +796,145 @@ File.prototype.getSignedUrl = function(options, callback) {
...
+ * @param {string=} options.successStatus - The status of the Google Storage 
+ *     response if the upload is successfull (must be string)
+ * @param {object=} options.contentLengthRange - Object providing 
+ *     minimum (options.contentLengthRange.min) and maximum 
+ *     (options.contentLengthRange.max) value for the request's 
+ *     content length
+ *
+ * @example
+ * file.getSignedPolicy({
+ *   equals: [["$Content-Type", "image/jpeg"]],
```
stephenplusplus: Can we also support accepting just the inner array for cases like this, where there is only one condition, so that it's...

```js
file.getSignedPolicy({
  equals: ['$Content-Type', 'image/jpeg']
  // ...
```

?

mziccard:
Should be ok now, even the boring stuff :)
 Both 
```javascript
file.getSignedPolicy({
  equals: ['$Content-Type', 'image/jpeg']
  // ...
```
and
```javascript
file.getSignedPolicy({
  equals: [['$Content-Type', 'image/jpeg']]
  // ...
```
are now supported if only one condition is specified. (also for `startsWith` condition).

lib/storage/file.js
```diff
@@ -796,6 +796,162 @@ File.prototype.getSignedUrl = function(options, callback) {
...
+      options.equals.forEach(function(condition) {
+        if (!util.is(condition, 'array') || condition.length !== 2) {
+          throw new Error(
+            'Equals condition must be an array of 2 elements.'
+          );
+        } 
+        conditions.push(['eq', condition[0], condition[1]]);
+      });
+    }
+  }
```
stephenplusplus: To handle supporting the `[]` and `[][]` thing, this should make it a little less noisy:

```js
if (util.is(options.condition, 'array')) {
  if (!util.is(options.condition[0], 'array')) {
    options.condition = [options.condition];
  }

  options.condition.forEach(function(condition) {
    // ...
  });
}
```

stephenplusplus:
Thank you very much!

lib/storage/file.js
mziccard: This is much better. Just pushed an update.

mziccard:
Everything ok on my side, are there any more thoughts?

test/storage/file.js
```diff
@@ -1041,6 +1041,208 @@ describe('File', function() {
...
+    });
+
+    describe('prefix conditions', function() {
+      it('should add prefix conditions (array of arrays)', function(done) {
+        var expiration = Math.round(Date.now() / 1000) + 5;
+        file.getSignedPolicy({
+          expiration: expiration,
+          startsWith: [['$<field>', '<value>']]
+        }, function(err, signedPolicy) {
+          console.log(signedPolicy);
```
ryanseys: Remove this `console.log`

test/storage/file.js
```diff
@@ -1041,6 +1041,208 @@ describe('File', function() {
...
+        });
+      });
+
+      it('should add prefix condition (array)', function(done) {
+        var expiration = Math.round(Date.now() / 1000) + 5;
+        file.getSignedPolicy({
+          expiration: expiration,
+          startsWith: ['$<field>', '<value>']
+        }, function(err, signedPolicy) {
+          console.log(signedPolicy);
```
ryanseys: Remove this `console.log`

ryanseys:
Just a couple `console.log`s left in the tests, but otherwise looks ready for merge! :)

mziccard:
Oops, thanks for spotting them. Just removed them:)

stephenplusplus:
Merged via 6cbdbe9c0bb7d3b1eb47154c772ae078d4838fad. Thanks again for this!

PR #504, opened 2015-04-20
localize highlight - fixes #503

stephenplusplus:
Fixes #503

Issue #484, opened 2015-04-06
Change GCLOUD_TESTS_KEY to "./key.json" in Travis


ryanseys:
Should change `GCLOUD_TESTS_KEY` environment variable to "./key.json" in Travis.

stephenplusplus:
We really don't need this after #505.

PR #505, opened 2015-04-21
use path.resolve to find keyfile path

stephenplusplus:
For #484 

This will create a full path to the keyfile, re-enabling our tests to run after merge. Going to merge this straight away so we can merge #504 and have the docs built.

This should be future-proof for when the env var changes as well.


ryanseys:
Smart! Didn't even think of this.

PR #507, opened 2015-04-21
add authorization guide + links to Google docs

### LINKED ISSUE #479: Lack of developer documentation
See https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.13.1/datastore
It mentions "some convenience methods"  

Convenience based on What?
Do I need to master the low level Datastore APIs? Imo, the gcloud layer should be self sufficient and complete for documentation, exposing all the Datastore concepts assuming the user knows nothing, and without the need for the user to read other docs....

The doc issue seems generic for all gcloud APIs for all the runtimes (same for Java, I did not look at ruby)

### LINKED ISSUE #483: Overview of Authentication in docs
Should have a Authentication section in docs to make it clear how to authenticate.

stephenplusplus:
fixes #483
fixes #479

Docs: http://stephenplusplus.github.io/gcloud-node/#/docs/master
Auth: http://stephenplusplus.github.io/gcloud-node/#/authorization

### DIFF
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -1,0 +1,52 @@
+With `gcloud-node` it's incredibly easy to get authorized and start using Google's APIs. You can set your credentials on a global basis as well as on a per-API basis. See each individual API section below to see how you can auth on a per-API-basis. This is useful if you want to use different accounts for different Google Cloud services.
+
+## On Google Compute Engine
+
+If you are running this client on Google Compute Engine, we handle authorization for you with no configuration. You just need to make sure that when you [set up the GCE instance][gce-how-to], you add the correct scopes for the APIs you want to access.
+
+``` js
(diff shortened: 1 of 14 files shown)


ryanseys:
Only thing I notice is that example code on the Authorization page doesn't have a background around it, it's just on the white background which is different that the `@example` blocks in the other parts of the docs. Otherwise, it looks great :)

stephenplusplus:
It turns out that implementing syntax highlighting is quite complicated. We're using a library to turn the Markdown into HTML, which spits out a plain `<pre class="javascript"></pre>` block. I did make a change to carry the base styles (background-color, padding, color) to the Markdown code blocks, however, which helps.

ryanseys:
Sounds good.

PR #488, opened 2015-04-09
Update CRC module to SSE4_CRC32 v3.1.0

anandsuresh:
- v3.1.0 fixes issues with hardware CRC detection and provides software fallback
  for platforms lacking hardware CRC support.
- Reduces the number of layers between core code and the CRC functions
- Fixes #425

PR #512, opened 2015-04-27
Update to CRC module and fix self register

ryanseys:
Inherits from #488

### DIFF
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -92,6 +92,7 @@ describe('File', function() {
   var bucket;
 
   before(function() {
+    mockery.registerMock('sse4_crc32', crc);
     mockery.registerMock('configstore', FakeConfigStore);
     mockery.registerMock('duplexify', FakeDuplexify);
     mockery.registerMock('request', fakeRequest);


ryanseys:
Fear not, dear @googlebot.

ryanseys:
@anandsuresh Will wait for Travis to confirm this actually works, then will merge and close other PR.

anandsuresh:
Aaah! I see now that Mockery needed to be told about the new module in order to get that to work. My bad. I didn't realize that was a requirement.

Having said that, a more descriptive error message might have made this easier. :tired_face:

ryanseys:
Yeah something like that. I don't actually know much about the register and deregister system in node regarding requiring modules and caches and all that jazz, but this seemed to fix it on my machine so I'm running with it. I believe it's an issue because it's required both inside the test file and the file.js. That combined with the fact that it's a native module on node 0.12.0 is a recipe for disaster.

stephenplusplus:
It passed! It passed!

ryanseys:
:D

anandsuresh:
Yaay!! Finally done! :)

PR #520, opened 2015-05-01
fix message.id and message.ackId documentation

### LINKED ISSUE #518: Update pubsub.Subscription.ack()'s docstring to say the ID is explicitly an ackId
@pierre-b wrote in #487:

> Could you please update the doc, the message.id should be message.ackId https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.13.2/pubsub/subscription?method=ack

At first I thought we had this right, but it seems like the `messageId` might not be the same as the `ackId` (see [RecievedMessage](https://cloud.google.com/pubsub/reference/rest/v1beta2/projects/subscriptions/pull#google.pubsub.v1beta2.ReceivedMessage) which says "ackId: This ID can be used to acknowledge the received message.")

**Note**: I still think we should call the parameter "ids", but that we should update the docstring.

On https://github.com/GoogleCloudPlatform/gcloud-node/blob/686fe62e7103f80dc053417e08c13abfc501f67f/lib/pubsub/subscription.js#L250

```
s/message IDs/message ackIds/
```

ryanseys:
Fixes #518 

There actually was an issue I found where we were incorrectly documenting message.id as the ackId.
I clarified it everywhere I could, including in the code because it doesn't affect the end user and makes things more clear all around.

### DIFF
--- a/lib/pubsub/subscription.js
+++ b/lib/pubsub/subscription.js
@@ -102,7 +102,8 @@ var util = require('../common/util.js');
  * // Register a listener for `message` events.
  * function onMessage(message) {
  *   // Called every time a message is received.
- *   // message.id = ID used to acknowledge its receival.
+ *   // message.id = ID of the message.
+ *   // message.ackId = ID used to acknowledge the message receival.
  *   // message.data = Contents of the message.
  * }
  * subscription.on('message', onMessage);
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Cool, cool

PR #525, opened 2015-05-04
storage: use standard URL instead of calling getMetadata - fixes #524

### LINKED ISSUE #524: Change download link to not require storage.getMetadata() ?
Looks like we can auto-generate the download link without pulling all the metadata in a separate request:

`https://<bucket>.storage.googleapis.com/<object>`

Can we just make this a standard URL?

@robertdimarco says:

> Remove the getMetadata() round-trip from the createReadStream() code path, in the interests of reducing overall request latency.
> 
> In comparison to the S3 implementation we were using, gcloud-node performs an additional call to the storage backend in series to get the metadata to obtain a download link. This download link, however, is able to be statically generated. Removing the extra round-trip to the server would halve the request latency for serving small files.

stephenplusplus:
#524 gave us the good news we can ditch the extra call just to get something we can figure out on our own.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -321,18 +321,12 @@ File.prototype.createReadStream = function(options) {
   var crc32c = validation === 'crc32c' || validation === 'all';
   var md5 = validation === 'md5' || validation === 'all';
 
-  if (this.metadata.mediaLink) {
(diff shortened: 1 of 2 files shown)


ryanseys:
:+1:

PR #528, opened 2015-05-04
global: remove socket limit - fixes #522

### LINKED ISSUE #522: Remover 5 concurrent connection limit for GCS?
Using the following sample code:

```javascript
// Include required NPM dependencies
var express = require('express');

// Initialize app and GCloud GCS module
var app = express();
var gcs = gcloud.storage({ projectId: '...', keyFilename: './key.json' });

// Create one route to proxy a file from GCS
app.get('/my-file.zip', function(req, res) {
  var file = bucket.file('my-file.zip');
  var readStream = file.createReadStream();
  return readStream.pipe(res);
});

// Listen on a port
app.listen(8080);
```

We're seeing that we're limited to 5 concurrent connections to GCS. This sucks when the server running this code can handle far more than 5 concurrent connections -- as we're shoving N connections through a 5-connection pipe.

From @robertdimarco:

> By default, requests to GCS are pooled and limited to 5 concurrent connections.
> 
> Node.js, prior to v0.12 or io.js, automatically pools outbound requests to five concurrent requests per host. This can be overridden globally per-application or on a per-package basis. Our prior integration with AWS / S3 had disabled this limit at the package level, anticipating our use case. Node.js v0.12 and io.js both remove this pooling altogether.
 
Can we remove the connection pooling limits for our server process altogether ? (also see https://github.com/substack/hyperquest#rant)

stephenplusplus:
#522

`request` has a defaults option where we can mimic the Node 0.12.0 behavior of not limiting open sockets to 5 at once.

@ryanseys - you're ~~kind~~_king_ of benching - do you have any tips how I might test that this is working as intended?

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -23,7 +23,11 @@
 
 var extend = require('extend');
 var GoogleAuth = require('google-auth-library');
-var request = require('request');
+var request = require('request').defaults({
+  pool: {
+    maxSockets: Infinity
+  }
+});
 var nodeutil = require('util');
 var uuid = require('node-uuid');
 
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -21,7 +21,11 @@
 'use strict';
 
 var streamEvents = require('stream-events');
-var request = require('request');
+var request = require('request').defaults({
+  pool: {
+    maxSockets: Infinity
+  }
+});
 var through = require('through2');
 
 /**
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -27,7 +27,11 @@ var crypto = require('crypto');
 var duplexify = require('duplexify');
 var fs = require('fs');
 var once = require('once');
-var request = require('request');
+var request = require('request').defaults({
+  pool: {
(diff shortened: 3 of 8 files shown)


ryanseys:
I can't think of any way to test that's easy, other than writing a unit test to make sure the proper defaults are set on the request object. I feel we should just trust the module to "do the right thing". The tests in this PR aren't all passing either so if you could take a look at that too.

stephenplusplus:
Aye.

stephenplusplus:
Fixed and added tests. No more 666! :tada: 

Also, please see my typo correction from the initial post. That was probably weird.

ryanseys:
Just datastore and storage get these defaults? Or am I reading something wrong here? Should it be added to every API?

stephenplusplus:
Most importantly, it's added to common/util, which is where non-fancy API requests all go through (the makeAuthorizedRequest flow).

Everywhere you see `require('request')` in our library code, we're using the defaults.

ryanseys:
Perfect! Definitely glossed over that. :gem:

PR #529, opened 2015-05-05
Storage: add move method to File object

mziccard:
Just a simple move function for a File object. It uses the already available copy and delete.
I also added some unit tests.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -235,6 +235,87 @@ File.prototype.copy = function(destination, callback) {
 };
 
 /**
+ * Move this file to another location. By default, this will move 
+ * the file to the same bucket, but you can choose to move it 
+ * to another Bucket by providing either a Bucket or File object.
+ *
+ * @throws {Error} If the destination file is not provided.
+ *
+ * @param {string|module:storage/bucket|module:storage/file} destination -
+ *     Destination file.
+ * @param {function=} callback - The callback function.
+ *
+ * @example
+ * //-
+ * // You can pass in a variety of types for the destination.
+ * //
+ * // For all of the below examples, assume we are working with the following
+ * // Bucket and File objects.
+ * //-
+ * var bucket = storage.bucket('my-bucket');
+ * var file = bucket.file('my-image.png');
+ *
+ * //-
+ * // If you pass in a string for the destination, the file is moved to its
+ * // current bucket, under the new name provided.
+ * //-
+ * file.move('my-image-new.png', function(err, destinationFile, apiResponse) {
+ *   // `my-bucket` no longer contains:
+ *   // - "my-image.png"
+ *   // but contains instead:
+ *   // - "my-image-new.png"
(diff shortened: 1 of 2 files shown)


lib/storage/file.js
```diff
@@ -235,6 +235,87 @@ File.prototype.copy = function(destination, callback) {
...
+ *   // Note:
+ *   // The `destinationFile` parameter is equal to `anotherFile`.
+ * });
+ */
+File.prototype.move = function(destination, callback) {
+  callback = callback || util.noop;
+  var deleteFunction = File.prototype.delete.bind(this);
+  this.copy(destination, function(err, destinationFile, apiResponse) {
+    if (err) {
+      callback(err, destinationFile, apiResponse);
```
ryanseys: If this fails, destinationFile here should just be an explicit `null`.

lib/storage/file.js
```diff
@@ -235,6 +235,87 @@ File.prototype.copy = function(destination, callback) {
...
+    } else {
+      deleteFunction(function(err, apiResponse) {
+        callback(err, destinationFile, apiResponse);
```
ryanseys: If this passes, `err` here should just be an explicit `null`.

ryanseys:
I like the idea. And code looks great! Can you squash your commits to get rid of those messy merge commits :smiley: thanks!

stephenplusplus:
I like these helpers, too. Thanks!

test/storage/file.js
```diff
@@ -281,6 +281,126 @@ describe('File', function() {
...
   });
 
+  describe('move', function() {
+    it('should throw if no destination is provided', function() {
+      assert.throws(function() {
+        file.move();
+      }, /should have a name/);
+    });
+
+    describe('destination types', function() {
```
stephenplusplus: It's hard to give a negative review of the tests, because they're great at testing the functionality you added. But, we tend to test a bit differently in the library, making use of stubs wherever possible, not caring about if the methods we use do their job properly, as we have other tests for that.

As an example of how we would test this change, we would stub the "copy" method, and see that it was called with the correct arguments. We would then see that our callback was executed with the response we mocked. We also check for errors wherever they can occur, and make sure the provided callback gets them.

I can elaborate with some examples tomorrow, and as always, you're free to leave the PR as-is and let me worry about it :)

Thanks again for the contribution!

ryanseys: @stephenplusplus Good catch on the test style. 

@mziccard I fumble over the style we use sometimes as well. The stubs can really convolute things, but I do feel it pays off in the end. I find it really helps to look at the existing tests and copy the style used within. Hope that helps :)

mziccard:
I fixed the mess I did with merges and made other corrections. Tests should now be better but please double-check them:)

stephenplusplus:
Awesome, thank you!

Look how happy you made gcloud:


(Those are tears of joy)

Issue #523, opened 2015-05-04
Connections not cleaned up if disconnected mid-download? 


jgeewax:
See #522 for the sample code.

In that example, if a client disconnects mid-way through a download, the connection apparently doesn't get cleaned up, and we're left with N-1 available connections in the pool.

Can we make sure that if a connection is broken, we clean up the socket and toss it ?

/cc @robertdimarco

robertdimarco:
Thanks for filing these issues and resolving them so quickly! :clap: 

This one is the trickiest of the bunch, and (in my view) doesn't have the clearest resolution. In the prior example on #522, without removing the connection limits, the issue is easily reproduced by having a client request a large file and hang-up part-way through that file.

Our short-term resolution for this issue was (a) remove the Node v0.10 socket pooling limits, and (b) listen for the client hang-up / socket close, and try to invoke `stream.abort()` on the through-stream returned from this module. That last bit improves the situation, but appears to be error-prone and I *think* does not pass-through and call `stream.destroy()` on the stream returned from the request module.

Happy to expand further or help out however we can.

robertdimarco:
Thanks everyone for your help here. :clap: :clap: :clap:

PR #530, opened 2015-05-05
Ensure File#createReadStream sockets are properly closed on error

ryanseys:
Trying to address #523 

Don't merge this, it's not complete.

This seems to mitigate sockets staying open when a request is ended early (only scenario I tried to get the request to hang).

Question for @robertdimarco: What do you mean by:

> and hang-up part-way through that file

in your comment [here](https://github.com/GoogleCloudPlatform/gcloud-node/issues/523#issuecomment-98822620). I don't know how to "hang up" part way, and is it the server that is hanging up or the client? (I'm assuming client). Any other scenarios you can think of that we should test?

@stephenplusplus Seems that `done` at the bottom of `File#createReadStream` isn't called when throughStream gets an error event like you can see in this PR. I listen for this event specifically and destroy the socket related to the request at that time. This seems to destroy all the sockets laying around and fires an error events saying there is a "Download mismatch" according to verification.

My test code can be found here: https://gist.github.com/ryanseys/0e2c3244175db6525d7a

Set `maxSockets: 2` when running test code because currently infinite sockets are allowed to remain open so the 4th request will succeed.

I have a feeling this is on the right track but I perhaps don't clean up everything appropriately due to the craziness that is Node.js Streams and `request`. In any case, it still seems to mitigate the issue a bit so far.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -301,9 +301,10 @@ File.prototype.createReadStream = function(options) {
     util.is(options.start, 'number') || util.is(options.end, 'number');
   var throughStream = streamEvents(through());
 
+  var requestStream;
+
   var validations = ['crc32c', 'md5'];
   var validation;
-  var socket;
 
   if (util.is(options.validation, 'string')) {
     options.validation = options.validation.toLowerCase();
@@ -333,6 +334,22 @@ File.prototype.createReadStream = function(options) {
 
   createAuthorizedReq(remoteFilePath);
 
+  // End the stream, first emitting an error or complete event.
+  var endThroughStream = once(function(err, resp) {
+    if (err) {
+      throughStream.emit('error', err, resp);
+    } else {
+      throughStream.emit('complete', resp);
+    }
+
+    throughStream.destroy();
+  });
+
+  var endRequestStream = once(function() {
+    requestStream.abort();
+    requestStream.destroy();
+  });
+
   return throughStream;
 
   // Authenticate the request, then pipe the remote API request to the stream
@@ -359,7 +376,7 @@ File.prototype.createReadStream = function(options) {
     that.bucket.storage.makeAuthorizedRequest_(reqOpts, {
       onAuthorized: function(err, authorizedReqOpts) {
         if (err) {
-          done(err, null);
+          endThroughStream(err, null);
           return;
         }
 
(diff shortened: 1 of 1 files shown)


robertdimarco:
@ryanseys I'm referring to a client disconnect before the write-stream was able to complete. We came across this issue by pointing a browser at the proxy server, let the browser connect and begin downloading, and then close the tab / refresh / etc. to cause the client to disconnect the socket. That is generally sufficient to reproduce the issue.

The handling for this I imagine is a little tricky, because the error might not be propagated from Express's write-stream to the read-stream / through-stream returned from `file.createReadStream()`. That part around piping likely just needs some documentation around it, as other devs will have to handle that client disconnect and propagate to this module.

ryanseys:
The problem with using that scenario (i.e. closing a tab) is I have no idea what that means in regards to what happens to either the read or write stream. A code-only test to reproduce this is most ideal, especially for ensuring this issue doesn't arise again as a regression. That's what I'm trying to accomplish here with my test script. I will try your Express method and perhaps watch the events that occur on the streams to get an idea of where things can go wrong so we thoroughly patch this thing.

robertdimarco:
@ryanseys Understood.

stephenplusplus:
@ryanseys I have a bit of a different approach going on: https://github.com/stephenplusplus/gcloud-node/commit/7b4ab4fa51905b24b90c6c137b0b1c7ec124367c - can you try this out with your tests?

ryanseys:
@stephenplusplus That's much cleaner and appears to work with my tests when `maxSockets: Infinity` (no open sockets at the end). When `maxSockets: 2`, weird things happen though. The third request that get's prematurely ended doesn't throw an error, or the 4th request that should succeed just never does, BUT there are no sockets left open as well. Considering we are going to have `maxSockets: Infinity` this shouldn't be an issue.

ryanseys:
Actually that might be just because `maxSockets` is doing its job, in which case we just set it to `Infinity` and hope nobody has more than `Infinity` simultaneous sockets open on gcloud (but if this happens, that person should probably get a nobel prize of the node.js variety).

ryanseys:
tl;dr: looks good to me. I'll cherry pick into this PR.

ryanseys:
@googlebot confirm.

ryanseys:
Apparently @googlebot doesn't like you @stephenplusplus

stephenplusplus:
Neither does Travis. I'm not big with robots, I guess :/

stephenplusplus:
@ryanseys I suppose I should fix the tests, since I'm the one who wrecked your PR!

stephenplusplus:
Carrying this on in #636.

PR #536, opened 2015-05-07
Fix #535 - Docs committed by travis-ci instead of selfiebot.

### LINKED ISSUE #535: Make our docs published by travis-ci
Following what gcloud-python does: https://github.com/GoogleCloudPlatform/gcloud-python/blob/master/scripts/update_docs.sh#L72

and what we do: https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/scripts/build.sh#L43

Might be good to be a bit consistent so let's make the docs published by `travis-ci <travis@travis-ci.org>`

jgeewax:


scripts/build.sh
```diff
@@ -36,11 +36,10 @@ if [ "${TRAVIS_BRANCH}" == "master" ] && [ "${TRAVIS_PULL_REQUEST}" == "false" ]
   set +e
   git add .
   set -e
-  git status
```
jgeewax: @ryanseys : Is this required? I removed so that I could comment on it in this PR.

scripts/build.sh
```diff
@@ -36,11 +36,10 @@ if [ "${TRAVIS_BRANCH}" == "master" ] && [ "${TRAVIS_PULL_REQUEST}" == "false" ]
...
-  # H/T: https://github.com/dhermes
```
jgeewax: Removed this as it doesn't seem to apply to gcloud-node.

scripts/build.sh
ryanseys: I think it's more for debugging purposes. Removing it won't hurt anything but in the future we might need to add it back if we're (for whatever reason) not having our docs built properly. Leave it removed, we should be ok.

scripts/build.sh
ryanseys: I believe this is just a hat-tip to dhermes because we stole this code (written by him) from gcloud-python.

PR #539, opened 2015-05-07
Fix #538 - Homepage code sample has tabs for different runtime envs

### LINKED ISSUE #538: Make the homepage example have two tabs (one for "anywhere" another for "GCE")
Would be really cool if we could toggle between the "Anywhere" state (where you need to specify authentication stuff) and the "inside GCE" state (where auth happens automagically.

jgeewax:


ryanseys:
Ahhh... I misunderstood the original idea. Just tried this and this is awesome! I have a small concern I'll address inline.

docs/site/home.html
```diff
@@ -80,9 +80,18 @@ <h3 class="block-title">What is it?</h3>
...
       <div class="quote-box--supplementary">
-        <h4>Retrieve Datastore Entities</h4>
-
-        <div hljs language="js">
+        <h4>Example: Retrieve an entity from Cloud Datastore</h4>
+
+        <div class="code-sample-tabs" ng-init="selected = 'compute engine'">
+          <a ng-click="selected = 'anywhere'"
+             ng-class="{selected: selected == 'anywhere'}"
+             >Run anywhere</a>
```
ryanseys: "anywhere" to me suggests anywhere including Compute Engine... I suppose that's correct, this would work anywhere including Compute Engine, but that's not what we are recommending. Maybe "elsewhere" is better (if it's the second tab)? Or keep it this way, that's suffice with me.

docs/site/home.html
```diff
@@ -95,6 +104,14 @@ <h3 class="block-title">What is it?</h3>
   console.log(err, entity);
 });</div>
 
+        <div hljs language="js" ng-show="selected == 'compute engine'">
+// Auth is handled automatically inside GCE.
+var dataset = require('gcloud').datastore.dataset();
```
ryanseys: Our code samples should always reflect our best practices too. I believe creating a separate `gcloud` object is always what we should suggest to the users. So the code should be fixed as such:

``` js
// ... rest as normal
```

docs/site/home.html
jgeewax: Cool - tweaking to:

1. Tabs will be "Run in Compute Engine" and "Run elsewhere"
1. Order of tabs will be "compute engine", then "elsewhere"

Important to note that you can use the "elsewhere" one anywhere (ie, you can be explicit in compute engine about which project and key you want).

docs/site/home.html
jgeewax: Fixed.

ryanseys:
Awesome! :dancer:

PR #540, opened 2015-05-07
Fix #534 - Nav bar is fixed and scrollable.

### LINKED ISSUE #534: Make the sidebar in docs always visible regardless of scrolling.
Looks like when I scroll down I lose the sidebar and have to scroll back to the top :( Would be awesome if it stayed with me as I read the docs.

jgeewax:
Updated left navigation bar:
- On all screens, the navbar is visible when scrolling down in the docs.
- On long vertical screens (if the whole navbar fits), no scrolling.
- On short vertical screens (if the whole navbar doesn't fit), you can scroll down inside the navbar.

Tested on...
- Chrome (Linux) with a full-sized (large) window, and a smaller window)
- Chrome (Mac OS X) with a 13" laptop screen (requires vertical scrolling)

### DIFF
--- a/docs/site/css/main.css
+++ b/docs/site/css/main.css
@@ -1359,13 +1359,14 @@ ul {
     }
 
     .side-nav {
-        position: absolute;
+        position: fixed;
         top: 0;
         left: 0;
         bottom: 0;
         width: 240px;
         padding-top: 7.7em;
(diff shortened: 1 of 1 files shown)


ryanseys:
Cool, works on Firefox, Safari and Opera as well.

PR #543, opened 2015-05-08
docs: sidebar links should stay consistent across doc pages

### LINKED ISSUE #541: Make the sidebar consistent regardless of location
Looks like we change the side bar of the documentation depending on where we are in the hierarchy (ie, click on Authorization and the whole side bar is different), but not in the way that it looks like certain sections are collapsed and just need to be expanded.

Can we keep the sidebar consistent regardless of what page we're currently looking at?

stephenplusplus:
Fixes #541 

Preview: http://stephenplusplus.github.io/gcloud-node/#/authorization


ryanseys:
LGTM!

jgeewax:
Wow... that was easy... Thanks @stephenplusplus

PR #544, opened 2015-05-08
Add a warning to the storage/file#move method.

jgeewax:
We should make it clear that this is a composition of two methods, and if the second (storage/file#delete) were to fail, a lingering file would be left around (which could cost money).

This also removes some trailing whitespace.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -235,9 +235,17 @@ File.prototype.copy = function(destination, callback) {
 };
 
 /**
- * Move this file to another location. By default, this will move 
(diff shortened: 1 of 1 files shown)


stephenplusplus:
:+1: JSHint reports a "line too long" warning, but I'll fix it up and manually merge this in.

stephenplusplus:
Merged in 6c650e17134b8eef5fd93d2cdeb62571c9e27a58 with the fixed up spacing & just a small tweak to the phrasing of the last sentence.

ryanseys:
:+1:

PR #560, opened 2015-05-08
Add Issue Stats badges to README.md.

jgeewax:
See example here: https://github.com/jgeewax/gcloud-node/blob/issue-stats/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,9 @@
 [![Travis Build Status](https://travis-ci.org/GoogleCloudPlatform/gcloud-node.svg)](https://travis-ci.org/GoogleCloudPlatform/gcloud-node/)
 [![Coverage Status](https://img.shields.io/coveralls/GoogleCloudPlatform/gcloud-node.svg)](https://coveralls.io/r/GoogleCloudPlatform/gcloud-node?branch=master)
 
+[![Issue Stats](http://issuestats.com/github/GoogleCloudPlatform/gcloud-node/badge/pr?style=flat)](http://issuestats.com/github/GoogleCloudPlatform/gcloud-node)
+[![Issue Stats](http://issuestats.com/github/GoogleCloudPlatform/gcloud-node/badge/issue?style=flat)](http://issuestats.com/github/GoogleCloudPlatform/gcloud-node)
+
 * [Homepage][gcloud-homepage]
 * [API Documentation][gcloud-docs]
 


stephenplusplus:
I'll sound like a total whiner here, but I'm not a big fan of badges. I think they're just noise, and not something users actually find useful. If it were up to me, I would actually get rid of all of the badges we have now, so I'm of course opposed to this, but would be okay with it under the same pretense of "well, we already have *n* badges, so why not *y* more?" :)

If I wind up at a project's readme, I don't use a badge to show me the stability of a package... personally, I go to the Issues and PRs pages, look at the recent commits, and of course, the star count. If I were interested in stats, I could go to the [pulse](https://github.com/GoogleCloudPlatform/gcloud-node/pulse) page.

jgeewax:
I'm going to pull out one of those "I'd like to keep the project funded" cards -- and these stats are an awesome thing to point to as justification for that... I don't love badges either, but I think that these help show involvement and put some numbers next to all the work we do. Does that sway anything?

stephenplusplus:
I was going to add in "This badge feels very 'project manager-y'" but left that off :) If this is the best way to prove our value to whoever needs to see it, I will commit to helping us set some high scores on those badges :)

jgeewax:
They are **super** project manager-y... Like right up there on the list of "shit that leads to optimizing to make the numbers change"...

That said, when people go "well how effective are you guys at dealing with questions from users?", etc (which people have asked), this makes them go away. I like that.

stephenplusplus:
It's nice to hear how things work on that side. Thanks for the insight, and I'll try to be more considerate of that going forward!

ryanseys:
I'm late to this party but I'm with Stephen here. I'm curious why we have to put these stats in our README, surely if people (project managers) want to see these stats, they are far better tools and sources with more descriptive data than a single badge. Badges are inherently for the users, not really for us and shouldn't be for project managers.

jgeewax:
Speaking as a user also, it gives me an idea of PR and issue latency at a glance. Pulse + Graphs don't really provide that latency indicator.

If you guys are really against this -- as in, you think it detracts from the value of the README -- we should roll back. If it's more of a "seems like it doesn't add that much", then I'd want to chirp in with the "I think it adds something nice". It certainly isn't an absolute necessity though.

ryanseys:
Also, this is much like the "code coverage means nothing" premise. Blanket averages, blanket stats like this are completely useless and don't actually say anything about the VALUE of the responses we are giving to our users. Some ideas are great but take time. If we try to drive these numbers down by closing issues quickly and being inconsiderate to the idea, we risk using more users than what we would gain from ever showing these stats in the first place (which I will argue is ~= zero).

I'm all for a roll-back.

ryanseys:
It's noise. I don't work on a project because its responses are fast, I work on them because my ideas are carefully considered. I do believe it distracts from the value of the README because it puts emphasis on speed over quality. Surely our response time numbers aren't bad, but we've never had them yelling at us in the face. PR and Issue latency is not the problem here. It has never been a problem.

jgeewax:
I'll leave this to you guys -- didn't think it was too much noise but it's all you :)

ryanseys:
Okay I killed it. :)

PR #564, opened 2015-05-08
Fix #555 - Permalinks have larger click target.

### LINKED ISSUE #555: Permalink link in the docs is... pretty small. Could we make it a larger click target?
To permalink, I need to hover, which makes a # appear, and then click that... Might be easier if the heading itself was a link...

jgeewax:


stephenplusplus:
I like it, even though it kind of looks like clicking it will take me to Twitter :)

jgeewax:
Ha - that's fair. Should we use a different symbol? like a pilcrow ? (¶)

stephenplusplus:
Sorry, should have waited to merge for a response! (But hey, the stats!!) I think it's okay with #, really. But if @ryanseys comes along and thinks it's weird, I'm happy with whatever the team decides!

ryanseys:
The # comes from the # in the anchor URL but I think that's probably not really the case here because it goes to a new page. I'm really indifferent but a pilcrow (that's a fun word) looks nice too!

PR #566, opened 2015-05-08
Fix #563 - Add breadcrumbs to single method docs.

### LINKED ISSUE #563: Add breadcrumbs when you're "zoomed in" to a specific method?

- h1: Datastore
- h2: Transaction
- h3: allocateIds

Thoughts?

jgeewax:
Need a review... there's some weirdness with this:
1. ACL methods are weird. They show Storage > Acl > acl.add
2. Default ACL methods are even weirder: They show Storage > acl.default.add

Also I'm not 100% happy with the code, but not sure how I could make this better. Guidance would be really helpful.....

### DIFF
--- a/docs/site/components/docs/docs.html
+++ b/docs/site/components/docs/docs.html
@@ -118,6 +118,7 @@ var gcloud = require('gcloud')({
     <article
       ng-repeat="method in methods"
       ng-hide="singleMethod && method.name !== singleMethod">
+
       <h2 ng-if="method.name[0].toUpperCase() === method.name[0]">
         {{method.name}}
         <span class="view-code-link" docs-code-link></span>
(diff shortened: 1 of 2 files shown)


ryanseys:
Yeah this can be resolved by putting `acl` on `Bucket.prototype` instead of initialized in the constructor of `Bucket`. (I quickly fooled around with it and that seemed to resolve it, but left the code in a non-working state). @stephenplusplus can perhaps describe the reason why this was done this way. Seems lots of `this` scope magic prevents us from moving these methods around more freely I believe.

stephenplusplus:
Another issue is http://localhost:8000/docs/#/docs/master/datastore/dataset?method=allocateIds showing `DatastoreRequest` as its parent. Breadcrumbs sounded nice at first, but I wouldn't mind ditching them in favor of the sidebar being the "hierarchy of truth". It's also a little odd why the breadcrumbs are only used on one page, and not all.

stephenplusplus:
@ryanseys The main issue is to keep Acl generic, it needs information that we don't have until instantiation of our modules.

ryanseys:
Yeah makes sense. I think sidebar of truth sounds okay to me. Breadcrumbs are messy when we don't use a very specific style of object hierarchy.

ryanseys:
> (edit: can't believe I gave a localhost link)

oh man :see_no_evil: 

@jgeewax You ok with closing this due to the complexity of this fix given the circumstances noted above? Neat idea, but hard in practice.

jgeewax:
:( Makes me sad, but if it's not worth the effort, so be it.

It just seems weird that we can detect, for every method, what module and class should light up in the sidebar, but we can pick up that same thing and just put in the class (Bucket, File, ...).

I'll close out this PR but leave the main issue open. Maybe there's a better way.

PR #576, opened 2015-05-09
Make query reference links more explanatory.

jgeewax:


ryanseys:
This is great! It's these small improvements that make our docs *that* much better.

PR #578, opened 2015-05-10
storage: add response-content-* parameters to signed url

mziccard:
This lets the user add the `response-content-type` and `response-content-disposition` parameters to a signed url. 
It really is a minor addition but I think it might be useful (at least to me :)). It took me a while to find those two parameters in the docs so having them at hand might save some time to others.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -1083,12 +1083,6 @@ File.prototype.getSignedUrl = function(options, callback) {
         encodeURIComponent(options.promptSaveAs) + '"';
     }
     if (util.is(options.responseDisposition, 'string')) {
-      if (responseContentDisposition !== '') {
-        console.warn([
-          'getSignedUrl: the promptSaveAs option is ignored',
-          'when responseDisposition is provided.'
-        ].join(' '));
-      }
       responseContentDisposition = 
         '&response-content-disposition=' + 
         encodeURIComponent(options.responseDisposition);
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -1389,14 +1389,9 @@ describe('File', function() {
         });
       });
 
-      it('should warn and ignore promptSaveAs if set', function(done) {
+      it('should ignore promptSaveAs if set', function(done) {
         var disposition = 'attachment; filename="fname.ext"';
         var saveAs = 'fname2.ext';
-        var oldWarn = console.warn;
-        console.warn = function(message) {
-          assert(message.indexOf('promptSaveAs') > -1);
-          console.warn = oldWarn;
-        };
         directoryFile.getSignedUrl({
           action: 'read',
           expires: Math.round(Date.now() / 1000) + 5,
@@ -1405,7 +1400,6 @@ describe('File', function() {
         }, function(err, signedUrl) {
           assert(signedUrl.indexOf(encodeURIComponent(disposition)) > -1);
           assert(signedUrl.indexOf(encodeURIComponent(saveAs)) === -1);
-          assert.equal(console.warn, oldWarn);
           done();
         });
       });


test/storage/file.js
```diff
@@ -1350,6 +1350,30 @@ describe('File', function() {
...
+        done();
+      });
+    });
+
+    it('should add response content type parameter', function(done) {
+      var type = 'application/json';
+      directoryFile.getSignedUrl({
+        action: 'read',
+        expires: Math.round(Date.now() / 1000) + 5,
+        responseType:  type
```
ryanseys: nit: Change double space to single space: `responseType: type`

lib/storage/file.js
```diff
@@ -1019,12 +1019,18 @@ File.prototype.getSignedPolicy = function(options, callback) {
  *     link will expire.
  * @param {string=} options.extensionHeaders - If these headers are used, the
  *     server will check to make sure that the client provides matching values.
+ * @param {string=} options.responseDisposition - If you provide this value,
```
ryanseys: Can you also explain why the user may want to set these two parameters in the first place (i.e. so they don't need to find additional documentation elsewhere).

lib/storage/file.js
```diff
@@ -1019,12 +1019,18 @@ File.prototype.getSignedPolicy = function(options, callback) {
...
+ * @param {string=} options.responseType - If you provide this value, the 
+ *     response-content-type parameter of the signed url is set accordingly.
  * @param {function=} callback - The callback function.
  *
  * @example
  * file.getSignedUrl({
  *   action: 'read',
- *   expires: Math.round(Date.now() / 1000) + (60 * 60 * 24 * 14) // 2 weeks.
+ *   expires: Math.round(Date.now() / 1000) + (60 * 60 * 24 * 14), // 2 weeks.
+ *   responseDisposition: 'attachment; filename="filename.ext"'
```
ryanseys: Yikes. This looks very much like a raw header. Obviously if you think that flexibility is needed then ok, otherwise perhaps maybe the option could be renamed as something more user friendly and descriptive like  `promptSaveAs: 'filename.ext'` where filename.ext is the name of the file they want the browser to prompt them to save as, which is what I understand this option does.

ryanseys:
I like this PR. The code looks ok. I'm curious about exploring the idea of exposing these options, specifically content-disposition header as something a little more user friendly. I've commented my concerns inline with the code.

lib/storage/file.js
mziccard: I left the raw header as it was more flexible but I agree with you. 
I had a look [here](http://www.iana.org/assignments/cont-disp/cont-disp.xhtm) at what values content-disposition can take and they are quite a lot. 
Explicitly handling all of them is a bit an overkill as I have the feeling that nobody uses them/they're not really supported by browsers/they're not really useful in our context. 
So I think we can go in two directions: 1) leaving it as it is (PRO: flexible, CONS: not very node-oriented, ugly) 2) name it `promptSaveAs` as you suggest and support only the `attachment` disposition (PRO: easy to understand and nice, CONS: less flexible).
Or we could allow both options and simply check which one the user specified thus providing both simplicity and flexibility. 
I'm open for suggestions!

jgeewax: Thanks for this PR! I **really** like the idea of making the common case ("make this a download prompt") obvious and simple. Could we possibly provide both?

When we open that door, the question becomes what happens when you specify both fields... Some options are...

1. `responseDisposition` wins (as it's the more specific of the two)
1. `responseDisposition` wins and logs a warning about `promptSaveAs` being ignored
1. We raise an exception saying "Sorry you can't do that"

mziccard: Providing both sounds like the best choice.
I would avoid the exception as it is not a severe error. Can we let `responseDisposition` win and maybe log a warning?

jgeewax: I'm down with that. @ryanseys ?

lib/storage/file.js
jgeewax: The link you included below (http://www.iana.org/assignments/cont-disp/cont-disp.xhtml) seems like it'd be really really useful to someone mucking with this particular header. Maybe we can reference that?

lib/storage/file.js
ryanseys: SGTM

mziccard:
I applied the changed we discussed. 
As usual, please, double check them as it has been a long day :)

lib/storage/file.js
```diff
@@ -1019,12 +1019,22 @@ File.prototype.getSignedPolicy = function(options, callback) {
  *     link will expire.
  * @param {string=} options.extensionHeaders - If these headers are used, the
  *     server will check to make sure that the client provides matching values.
+ * @param {string=} options.promptSaveAs - If you provide this value the SaveAs 
+ *     prompt will be displayed when accessing the signed url and the file name 
+ *     will be set to the provided value. If options.responseDisposition is set 
+ *     this option is ignored.
+ * @param {string=} options.responseDisposition - If you provide this value,
```
ryanseys: Seems wordy. Might be your style of param descriptions but I'd just say what the parameter represents, not really the consequences of setting it... that is implied through its usage. Something like: "The filename to prompt the user to save the file as when the signed url is accessed. This is ignored if options.responseDisposition is set."

lib/storage/file.js
```diff
@@ -1019,12 +1019,22 @@ File.prototype.getSignedPolicy = function(options, callback) {
...
  * @param {string=} options.extensionHeaders - If these headers are used, the
  *     server will check to make sure that the client provides matching values.
+ * @param {string=} options.promptSaveAs - If you provide this value the SaveAs 
+ *     prompt will be displayed when accessing the signed url and the file name 
+ *     will be set to the provided value. If options.responseDisposition is set 
+ *     this option is ignored.
+ * @param {string=} options.responseDisposition - If you provide this value, 
+ *     the response-content-disposition parameter (http://goo.gl/yMWxQV) of the 
+ *     signed url is set accordingly.
+ * @param {string=} options.responseType - If you provide this value, the
```
ryanseys: Similarily here, just say "The response-content-type parameter of the signed url."

lib/storage/file.js
ryanseys: Oops this is for the parameter above this one. But likewise for all of these params, the style of description is strange. Sorry I'm really just nitpicking at this point. It's really ok, you can leave as-is. :)

ryanseys:
This is basically mergeable. Couple small issues above I noted. Nothing even slightly alarming. Looks really good :)

mziccard:
@ryanseys Thanks for pointing that out, I added an assertion to check that console.warn is reset to the non-stub version. 
I also made the docs less wordy :)

ryanseys:
LGTM. Hopefully Travis will finish in the next millennium, and we'll merge this in.

lib/storage/file.js
```diff
@@ -1060,11 +1069,38 @@ File.prototype.getSignedUrl = function(options, callback) {
...
+    if (util.is(options.promptSaveAs, 'string')) {
+      responseContentDisposition =
+        '&response-content-disposition=attachment; filename="' + 
+        encodeURIComponent(options.promptSaveAs) + '"';
+    }
+    if (util.is(options.responseDisposition, 'string')) {
+      if (responseContentDisposition !== '') {
+        console.warn([
+          'getSignedUrl: the promptSaveAs option is ignored',
+          'when responseDisposition is provided.'
```
stephenplusplus: Our library shouldn't log anything to stdout. gcloud could be used as a dependency's dependency, and the end user shouldn't be seeing this. We can either throw if it's serious. This warning can be noted in the docs.

ryanseys: That's a fair note. I'd be for not throwing

mziccard: The fact that `promptSaveAs` is ignored if `responseDisposition` is provided is already noted in the docs. Would you like to make it more explicit?

jgeewax: Is there a way to silently warn...? As in... when you have logging level == debug, you see the warning?

If not, I'm fine to silently smash this, but also make it clear in a comment that we're blasting out the `promptSaveAs` value.

ryanseys: Yeah the only reason someone would set either (or both) of these
parameters is if they read our docs and so we should be ok to assume they
also read the note to go along with it saying that you can only set one or
the other. I say just rip out the console.warn goodies and we're good to go

mziccard:
I removed the logging, the docs should already be clear enough.

stephenplusplus:
Thanks! Merged in via 8da19e3976b7f46e8546cffb605bbc4f7f407618

PR #580, opened 2015-05-11
Fix #579 - Overview docs must work in production.

### LINKED ISSUE #579: Overview pages not showing up in production...
Looks like we have a leading `/` in the include URL which means that we're looking for `http://googlecloudplatform.github.io/site/components/docs/datastore-overview.html` instead of `http://googlecloudplatform.github.io/gcloud-node/site/components/docs/datastore-overview.html` (note the missing `gcloud-node`...

This makes the overview sections not show up and a 404 appear in the console :(

jgeewax:


stephenplusplus:
:+1:

PR #581, opened 2015-05-11
docs: publicly document Datastore.dataset - fixes #547

### LINKED ISSUE #547: datastore#dataset is ignored in documentation?
Looking at https://github.com/GoogleCloudPlatform/gcloud-node/blob/1f57dd607b4c52a48f88781cd185220d40683184/lib/datastore/index.js#L94

Looks like we chose to make this "Developer documentation"?

Seems weird that we'd have pubsub#topic documented but not datastore#dataset? (http://googlecloudplatform.github.io/gcloud-node/#/docs/master/pubsub?method=topic)

Was there a reason we chose to `!` ignore that docstring?

stephenplusplus:
Fixes #547

The developer documentation was actually correctly marked as such. Since our code supports two ways of using gcloud (`var gcloud = require('gcloud');` & `var gcloud = require('gcloud')(config)`), we have both static and instance methods for creating Datasets.

### DIFF
--- a/lib/datastore/index.js
+++ b/lib/datastore/index.js
@@ -99,6 +99,16 @@ Datastore.dataset = Dataset;
  * @param  {object=} options - Configuration object.
  * @return {module:datastore/dataset}
  */
+/**
+ * Create a Dataset object to reference an existing dataset.
+ *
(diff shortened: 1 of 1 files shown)


ryanseys:
Thanks!

PR #584, opened 2015-05-11
docs: fix broken code links - fixes #547

### LINKED ISSUE #547: datastore#dataset is ignored in documentation?

stephenplusplus:
~~Fixes #547~~
Fixes #584

Preview: http://stephenplusplus.github.io/gcloud-node-gh-pages-again/#/docs/master

### DIFF
--- a/docs/site/components/docs/docs-directives.js
+++ b/docs/site/components/docs/docs-directives.js
@@ -11,16 +11,43 @@ angular.module('gcloud.docs')
     };
   })
 
-  .directive('docsCodeLink', function($compile) {
+  .directive('docsCodeLink', function() {
     'use strict';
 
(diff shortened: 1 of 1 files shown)


ryanseys:
Looks good! :) That's an annoying work-around but whatever works!

PR #585, opened 2015-05-12
Fix #563 - Add breadcrumbs to the header.

### LINKED ISSUE #563: Add breadcrumbs when you're "zoomed in" to a specific method?

jgeewax:
OK, new attempt here... Some issues with capitalization in this attempt (Bigquery instead of BigQuery, and likewise with PubSub vs Pubsub).

But... doesn't suffer from the same concerns as before with ACLs and whatnot...  Thoughts?

### DIFF
--- a/docs/site/components/docs/docs.html
+++ b/docs/site/components/docs/docs.html
@@ -1,5 +1,5 @@
 <script id="docs-header.html" type="text/ng-template">
-  <header header title="Node.js">
+  <header header title="{{pageTitle}}">
     <div class="row row--right">
       <div class="col margin-vertical">
         <a href="https://github.com/GoogleCloudPlatform/gcloud-node/issues/new"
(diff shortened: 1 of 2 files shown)


ryanseys:
@stephenplusplus any idea how to get the module title with correct capitalization in this scope? Something like `$scope.moduleProper` would be nice here. I tried to tackle it for a bit but couldn't wrangle any sense out of it. Seems $scope.links contains objects that have the sidebar links which might be massage-able to get the current active module, however I'm sure there's a more efficient way of doing this.

stephenplusplus:
https://github.com/stephenplusplus/gcloud-node/commit/5d9c4387394b4d601b87c1960a092d3bfeeb65b4 !

ryanseys:
:+1: Can you put this in a new PR? This PR will close when the commit here is merged in yours.

jgeewax:
Is that @ me or @stephenplusplus  ?

ryanseys:
@stephenplusplus

PR #586, opened 2015-05-12
Breadcrumbs attempt 2

### LINKED ISSUE #563: Add breadcrumbs when you're "zoomed in" to a specific method?

stephenplusplus:
Closes  #585
Fixes #563


jgeewax:
LGTM

PR #589, opened 2015-05-12
Fix #587 - Move buttons to top blue header.

### LINKED ISSUE #587: Move Report an Issue and Version History buttons into the blue header
See how Python has the Report an issue button in the top right http://googlecloudplatform.github.io/gcloud-python/latest/

jgeewax:
Here's what this looks like for me....

header button move

Note this leaves the "last build date" stuff in the same place.

### DIFF
--- a/docs/site/components/docs/docs.html
+++ b/docs/site/components/docs/docs.html
@@ -1,20 +1,5 @@
 <script id="docs-header.html" type="text/ng-template">
   <header header title="{{pageTitle}}">
-    <div class="row row--right">
-      <div class="col margin-vertical">
-        <a href="https://github.com/GoogleCloudPlatform/gcloud-node/issues/new"
-          class="v-btn">
-          <img src="site/img/icon-link-github.svg" />
-          Report an Issue
-        </a>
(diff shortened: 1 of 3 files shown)


ryanseys:
> Note this leaves the "last build date" stuff in the same place.

Do you think we should move it to somewhere else or it's good there? Looks pretty off, perhaps it could be better aligned with the Datastore >> Transaction text?

stephenplusplus:
Looks good, but busted at smaller viewports:

screen shot 2015-05-12 at 3 15 17 pm

jgeewax:
Ohh good call - I'll fix for smaller dimensions.

ryanseys:
@jgeewax any update on this?

stephenplusplus:
We are now down to just "Report an Issue", since the Version History is gone we're just using a dropdown in the sidebar. Not sure if that will make this easier, but feel free to assign to me if you have other stuff to work on @jgeewax :)

jgeewax:
Punting to you @stephenplusplus as I'm mostly out this week.

Issue #554, opened 2015-05-08
Walk through the "Saving data" in gcloud.datastore's Docs?


jgeewax:
Looks like it's not 100% clear to people how they can save data of different types. Maybe we should write some better documentation in the overview page explaining common tasks you want to do with Datastore? (Reference: #553)

/cc @ryanseys

ryanseys:
Assigning to myself to take a look at making these better.

ryanseys:
@jgeewax Does this suffice or would you like something more?

jgeewax:
LGTM

PR #591, opened 2015-05-12
Add example of using dataset.save with different data types

ryanseys:
Addresses #554 

I added an example to dataset#save here showing how to save with different types. We can copy this example to somewhere else in the docs if necessary? @jgeewax You mentioned something about putting it in the overview docs, but there's no code there right now so I'm not sure how out of place that would feel and I don't think it would have any code highlighting in this case. Ideas? What feels like the best place to put this?

### DIFF
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -196,6 +196,76 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
  *   data: {
  *     rating: '10'
  *   }
+ * }, function(err) {
+ *   console.log(key.path); // [ 'Company', 5669468231434240 ]
+ *   console.log(key.namespace); // undefined
+ * });
+ *
+ * //-
+ * // Save a single entity using a provided name instead of auto-generated ID.
+ * //
+ * // Here we are providing a key with name instead of an ID. After saving, the
+ * // original Key object used to save will be updated to contain the path with
+ * // the name instead of a generated ID.
+ * //-
+ * var key = dataset.key(['Company', 'donutshack']);
+ *
+ * dataset.save({
+ *   key: key,
+ *   data: {
+ *     name: 'DonutShack',
+ *     rating: 8
+ *   }
+ * }, function(err) {
+ *   console.log(key.path); // ['Company', 'donutshack']
+ *   console.log(key.namespace); // undefined
+ * });
+ *
+ * //-
+ * // Save a single entity with a provided namespace. Namespaces allow for
+ * // multitenancy. To read more about this, see
+ * // [the Datastore docs on key concepts](https://goo.gl/M1LUAu).
+ * //
+ * // Here we are providing a key with namespace.
+ * //-
+ * var key = dataset.key({
+ *   namespace: 'my-namespace',
+ *   path: ['Company', 'donutshack']
+ * });
+ *
+ * dataset.save({
+ *   key: key,
+ *   data: {
+ *     name: 'DonutShack',
+ *     rating: 8
+ *   }
+ * }, function(err) {
+ *   console.log(key.path); // ['Company', 'donutshack']
+ *   console.log(key.namespace); // 'my-namespace'
+ * });
+ *
+ * //-
+ * // Save different types of data, including ints, doubles, dates, booleans,
+ * // blobs, and lists.
+ * //
+ * // Notice that we are providing an incomplete key. After saving, the original
+ * // Key object used to save will be updated to contain the path with its
+ * // generated ID.
+ * //-
+ * var key = dataset.key('Company');
+ *
+ * dataset.save({
+ *   key: key,
+ *   data: {
+ *     name: 'DonutShack', // strings
+ *     rating: datastore.int(8), // ints
+ *     worth: datastore.double(123456.78), // doubles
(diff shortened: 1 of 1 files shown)


lib/datastore/request.js
```diff
@@ -199,6 +199,30 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
...
+ *     name: 'DonutShack', // strings
+ *     rating: datastore.int(8), // ints
+ *     worth: datastore.double(123456.78), // doubles
+ *     numDonutsServed: 45, // detect number type (int or double)
+ *     founded: new Date('Tue May 12 2015 15:30:00 GMT-0400 (EDT)'), // dates
+ *     isStartup: true, // booleans
+ *     donutEmoji: new Buffer('\uD83C\uDF69'), // buffers
+ *     keywords: ['donut', 'coffee', 'yum'] // lists of objects
+ *   }
+ * }, function(err) {});
```
jgeewax: Could we add a line here that shows how to mess with the new key? like `console.log(key.<something>)` ?

ryanseys: See the recently pushed version for these examples.

lib/datastore/request.js
```diff
@@ -196,6 +196,75 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
...
+ * });
+ *
+ * //-
+ * // Save a single entity using a provided name instead of auto-generated ID.
+ * //
+ * // Here we are providing a key with name instead of an id. After saving, the
+ * // original Key object used to save will be updated to contain the path with
+ * // the name instead of a generated ID.
+ * //-
+ * var key = dataset.key(['Company', 'DonutShack']);
```
jgeewax: Can we use a lower-case Key name? I think that's been the standard style.

lib/datastore/request.js
```diff
@@ -196,6 +196,75 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
...
+ *     rating: 8
+ *   }
+ * }, function(err) {
+ *   console.log(key.path); // ['Company', 'DonutShack']
+ *   console.log(key.namespace); // undefined
+ * });
+ *
+ * //-
+ * // Save a single entity with a provided namespace. Namespaces allow for
+ * // multitenancy. To read more about this, see [here](https://goo.gl/M1LUAu).
```
jgeewax: Instead of "here" can we put "the Datastore documentation on key concepts" ? (or something?)

lib/datastore/request.js
```diff
@@ -196,6 +196,75 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
...
+ * dataset.save({
+ *   key: key,
+ *   data: {
+ *     name: 'DonutShack',
+ *     rating: 8
+ *   }
+ * }, function(err) {
+ *   console.log(key.path); // ['Company', 'DonutShack']
+ *   console.log(key.namespace); // 'my-namespace'
+ * });s
```
jgeewax: typo?

lib/datastore/request.js
ryanseys: Done

lib/datastore/request.js
```diff
@@ -196,6 +196,75 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
...
+ * dataset.save({
+ *   key: key,
+ *   data: {
+ *     name: 'DonutShack',
+ *     rating: 8
+ *   }
+ * }, function(err) {
+ *   console.log(key.path); // ['Company', 'donutshack']
+ *   console.log(key.namespace); // 'my-namespace'
+ * });s
```
jgeewax: Typo on line 244. :)

lib/datastore/request.js
jgeewax: Damn you're quick...

lib/datastore/request.js
ryanseys: Done

lib/datastore/request.js
ryanseys: Thanks! Good catch! Fixed.

lib/datastore/request.js
ryanseys: :)

lib/datastore/request.js
```diff
@@ -196,6 +196,76 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
...
  *   }
+ * }, function(err) {
+ *   console.log(key.path); // [ 'Company', 5669468231434240 ]
+ *   console.log(key.namespace); // undefined
+ * });
+ *
+ * //-
+ * // Save a single entity using a provided name instead of auto-generated ID.
+ * //
+ * // Here we are providing a key with name instead of an id. After saving, the
```
stephenplusplus: Total nit, but should we use `ID` or `id`? Just like when I write docs, they are both used interchangeably here.

ryanseys: I changed to ID. I think I associated `name` and `id` in that sentence with a property like `key.name` and `key.id` which are naturally lowercase. Otherwise I like uppercase ID. I changed here to ID to remain consistent.

stephenplusplus:
Looks awesome!

PR #596, opened 2015-05-13
Fix https to http redirect

### LINKED ISSUE #559: Should we switch all of our links to use https://googlecloudplatform.github.io/.... ?
Seems that https://googlecloudplatform.github.io/gcloud-node/#/ works just fine, should we use that?

I don't think there's anything we really need to protect, but should we help out with the "https everywhere" thing?

ryanseys:
Fixes #559

Apparently GitHub Pages redirects https://googlecloudplatform.github.io/gcloud-node to http://googlecloudplatform.github.io/gcloud-node/ (note the trailing `/`) so if we change our URLs to include the `/` then it should remain on https.

``` sh
$ curl -I https://googlecloudplatform.github.io/gcloud-node
HTTP/1.1 301 Moved Permanently
Server: GitHub.com
Content-Type: text/html
Location: http://googlecloudplatform.github.io/gcloud-node/
```

I've already changed the URL in our repo too:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ See [CONTRIBUTING](CONTRIBUTING.md) for more information on how to get started.
 
 Apache 2.0 - See [COPYING](COPYING) for more information.
 
-[gcloud-homepage]: https://googlecloudplatform.github.io/gcloud-node
+[gcloud-homepage]: https://googlecloudplatform.github.io/gcloud-node/
 [gcloud-docs]: https://googlecloudplatform.github.io/gcloud-node/#/docs
 [gcloud-bigquery-docs]: https://googlecloudplatform.github.io/gcloud-node/#/docs/bigquery
 [gcloud-datastore-docs]: https://googlecloudplatform.github.io/gcloud-node/#/docs/datastore


jgeewax:
Woah -- good to know. LGTM

ryanseys:
I additionally contacted GitHub about the issue. I'll report back if they say anything but otherwise this should fix things on our end.

ryanseys:
Here's the reply I received:

> Hi Ryan,
>
> Sorry for the trouble. Currently, we aren't officially supporting HTTPS on GitHub Pages.
>
> While HTTPS requests may appear to work, our CDN provider is adding and removing the encryption at their end, and then the request is transmitted over the open internet from our CDN provider to our GitHub Pages infrastructure, creating the appearance of trustability.
>
> This is why we do not yet officially support HTTPS for GitHub Pages. We definitely appreciate the feedback and I'll add a +1 to this item on out internal Feature Request List.

> Thanks,
> Jamie

So there's really no protection except just to make our users feel warm and fuzzy inside.

jgeewax:
Ugh, thumbs down GitHub! But at least they'll look into it. Welp, warm fuzzy green lock thing is there at least...

PR #602, opened 2015-05-17
Remove #294 from troubleshooting advice

### LINKED ISSUE #590: Add a clearer resolution to #294 (or remove from the docs?)
We link to #294 in our docs for a particular error, but I can't see the correct resolution there -- it doesn't seem obvious to me what I have to do to get this to work, or what the underlying problem was.

For reference, on gcloud-python, we had a similar issue, and it was due to the Project ID vs Dataset ID distinction not being clear (the user was setting the project ID correctly, but not setting the Dataset ID correctly -- and they should be the same).

Can we add a comment saying "BTW, the issue here was X and the way to fix this in the future is to do Y" ?

ryanseys:
Fixes #590 

I agree that this issue is kinda strange and the exact resolution isn't clearly defined. How can we be sure that they are using the correct key as well. It might be that projects created under the old console have different incorrect keys our library cannot use.


jgeewax:
Well to be fair, the question is a good one. We had this same problem in gcloud-python and the resolution was that someone was not using the correct dataset ID (== project ID).

But I guess I can't argue with this resolution for now. If it comes up again, we can try to jot down the steps we took to get it fixed... LGTM

PR #603, opened 2015-05-17
Do not document queries as supporting multiple kinds

### LINKED ISSUE #572: Datastore querying by multiple kinds supported in gcloud-node, but not by Datastore.
When running the following code:

```javascript
var dataset = gcloud.datastore.dataset({
  projectId: 'my-project-id'
});

var query = dataset.createQuery(['Kind1', 'Kind2']);
dataset.runQuery(query, function(err, entities) {
  console.log(err);
  console.log(entities); // null
});
```

I get the following error returned: `message: 'multiple kinds not supported'`

According to https://cloud.google.com/datastore/docs/concepts/queries, Datastore allows querying over a single kind, not multiple.

Maybe we should remove the ability to construct a query with multiple kinds?

ryanseys:
Fixes #572 

Updated the documentation to not make people think we support multiple kinds. The code automatically arrayizes the `kind` given as required by the backend API.

### DIFF
--- a/lib/datastore/dataset.js
+++ b/lib/datastore/dataset.js
@@ -185,7 +185,7 @@ Dataset.prototype.key = function(options) {
 };
 
 /**
- * Create a query from the current dataset to query the specified kinds, scoped
+ * Create a query from the current dataset to query the specified kind, scoped
  * to the namespace provided at the initialization of the dataset.
  *
  * *[Reference](http://goo.gl/Cag0r6).*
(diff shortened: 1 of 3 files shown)


jgeewax:
LGTM if this is all the places where `['Kind']` appears in user-facing land.

ryanseys:
I'll double check.

ryanseys:
I added a couple changes so that the code more closely matches the documentation. I also removed a test that will continue to pass, but is against what our library should support.

ryanseys:
PTAL

jgeewax:
If those are all the places we talk about multiple kinds, then LGTM.

ryanseys:
Yep!

PR #607, opened 2015-05-19
Update setMetadata docs to clarify merging. And whitespace issues

### LINKED ISSUE #597: setMetadata() method merges existing with new data?
Looks like if I call `setMetadata` on an object, it merges the data with what's in production...

```javascript
var bucket = ...; // Get a bucket somehow.
var file = bucket.file('my-file');

file.setMetadata({
  metadata: {
    first: 'first'
  }
});

file.setMetadata({
  metadata: {
    second: 'second'
  }
}, function(err, metadata) {
  // metadata here is {first: 'first', second: 'second'}
  // Expected it to be just {second: 'second'} ....
});
```

I suspect that we can't .. fix this all that much... but maybe we can at least document it ?

---

The options we have so far:

- Make metadata immutable: remove `setMetadata`
- Operate on a single attribute at a time: `getMetadata(key)` and `setMetadata(key, value)`
- Rename `setMetadata` to `patchMetadata` or `mergeMetadata`
- Add some documentation about the behavior of `setMetadata`

/cc @capstan @aozarov

ryanseys:
Fixes #597 

@jgeewax Thoughts on this? Does this clarify things?

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -1025,7 +1025,7 @@ File.prototype.getSignedPolicy = function(options, callback) {
  * @param {string=} options.responseDisposition - The
  *     response-content-disposition parameter (http://goo.gl/yMWxQV) of the
  *     signed url.
- * @param {string=} options.responseType - The response-content-type parameter 
+ * @param {string=} options.responseType - The response-content-type parameter
  *     of the signed url.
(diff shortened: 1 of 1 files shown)


stephenplusplus:
LCTM.

PR #608, opened 2015-05-19
Remove auth examples from the README + clean up examples.

stephenplusplus:
Just some clean up of the readme. I think we document how authorization works in enough places (including this readme), that it doesn't consume such a large portion of each example snippet.

Otherwise, the changes are various and small. PTAL!


ryanseys:
LGTM

Issue #606, opened 2015-05-19
Propagate 'response' event through file.createReadStream()


robertdimarco:
***Background***:
We run a proxy service that serves a large number of files of varying size to downstream clients. In cases where we encounter an error serving the requested file (such as 404s) we intercept the error to the downstream client and pass through a custom error and template. See #522 for an oversimplified version.

***Request***:
The [previous request](https://github.com/GoogleCloudPlatform/gcloud-node/issues/524) to remove the required call to `file.getMetadata()` before invoking `file.createReadStream()` has been a big help in reducing mean latency since it removed a roundtrip for each download request. Our next challenge has been intercepting errors (namely those 404s) in a way that continues to let us simply pipe the result to the downstream client (and not buffer the download file until completion), and send the appropriate status code + headers.

Currently, in the case of a 404, the stream returned from `file.createReadStream()` will first emit a `data` event (an XML response containing an error) and then emits an `error` event (the formatted error object). Ideally, we could inspect the response headers before deciding whether or not to stream the result from GCS down to the client, and currently we have to buffer some number of `data` event payloads until we're sure we haven't encountered a GCS error.

***Details***:
The Node `http.ClientRequest` class specifies a `response` event that fires once and exposes (amongst other things) the HTTP response headers. If we could subscribe to that event, it would allow us to inspect the response from GCS, send the appropriate status code and headers to our downstream client, prior to piping any data.

```diff
diff --git a/lib/storage/file.js b/lib/storage/file.js
index d64aa09..137baba 100644
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -464,6 +464,8 @@ File.prototype.createReadStream = function(options) {
         request(authorizedReqOpts)
           .on('error', done)
 
+          .on('response', throughStream.emit.bind(throughStream, 'response'))
+
           .on('data', function(chunk) {
             if (crc32c) {
               localCrcHash = crc.calculate(chunk, localCrcHash);
```

Don't hesitate to let me know if you have any thoughts or questions, and thanks for all of your hard work on this library!

ryanseys:
Interesting use case. I don't see why we couldn't support this. I'm all for exposing this functionality if it helps our users :) @stephenplusplus what do you say?

ryanseys:
@robertdimarco I'm going to make an executive decision here and say "why not!". Could you create a PR for this change with a unit test or two and assign me to review when you're ready? Thanks!

[...]

PR #610, opened 2015-05-19
Emit response event from file#createReadStream()

robertdimarco:
See #606

### DIFF
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -432,7 +432,7 @@ describe('File', function() {
         .on('response', function(res) {
           assert.deepEqual(response, res);
           done();
-        })
+        });
     });
 
     it('should get readable stream from request', function(done) {


test/storage/file.js
```diff
@@ -370,6 +370,30 @@ describe('File', function() {
...
+
+        stream.Readable.call(this);
+        this._read = function() {
+          this.push(data);
+          this.push(null);
+        };
+
+        setImmediate(function() {
+          that.emit('response', fakeResponse);
+          that.emit('complete', fakeResponse);
```
ryanseys: We should probably remove response from the `complete` at sometime then you think?

ryanseys:
Can you please document this functionality in the docs for `file#createReadStream` :)

robertdimarco:
@ryanseys I added a line to the `file#createReadStream` header docs but let me know if you'd like to see more there or there's something I need to do to autogenerate-then-commit those docs.

test/storage/file.js
robertdimarco: Maybe... but I suspect it isn't worth it. People seem to love the `request` module, so perhaps sticking to its conventions isn't a bad thing. =)

ryanseys: Oh is that a convention in the `request` module? I thought we just added it! :)

robertdimarco: Yeah, you're in good company!

ryanseys:
Thanks @robertdimarco ! :dancer:

PR #612, opened 2015-05-19
add version switcher dropdown

### LINKED ISSUE #588: Make the version history button a drop-down with all of the different versions to switch between
It'd be really cool if switching versions was really really easy (See how Python has this as a drop-down at the top left: https://docs.python.org/2/ -- careful though , it's not the dropdown for all versions...)

stephenplusplus:
Fixes #588

Preview: http://stephenplusplus.github.io/gcloud-node-gh-pages-again/#/docs/

screen shot 2015-05-19 at 3 33 10 pm

Got rid of the History view in favor of a dropdown. The dropdown and last built date are now in the sidebar. Let me know what ya think.

### DIFF
--- a/docs/site/components/docs/docs.html
+++ b/docs/site/components/docs/docs.html
@@ -8,16 +8,6 @@
           Report an Issue
         </a>
       </div>
-      <div class="col margin-vertical">
-        <a href="#/docs/history" ng-show="showReference"
-          class="v-btn">
-          <img src="site/img/icon-arrow-bullet.svg" />
-          Version History (<i>{{version}}</i>)
-        </a>
-      </div>
-      <div>
-        <div ng-if="lastBuiltDate" class="build-date">Docs last built {{lastBuiltDate}}.</div>
-      </div>
     </div>
   </header>
 </script>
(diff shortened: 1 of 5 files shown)


ryanseys:
I only see `master` when I try locally. Any reason why that could be happening?

stephenplusplus:
It's always been that way, because we don't carry the built json for all versions locally. To test, you can:

1. clone gcloud-node under a different name (gcloud-node-gh-pages)
2. switch to the gh-pages branch and `git checkout -- versions.txt json`
3. switch back to my branch in your normal gcloud-node directory
4. npm run docs
5. cp -R docs/* ../gcloud-node-gh-pages

ryanseys:
Oooh, I found `git checkout -- versions.txt json` helped to do last after the `cp -R` but looks great! :)

stephenplusplus:
Oops :D Thanks!

ryanseys:
I noticed that the versions sit low on the page when in mobile mode due to being in the sidebar. This is certainly not ideal for mobile I think, but perhaps changing versions is not as important for mobile? Should we simply hide the version dropdown on mobile?

stephenplusplus:
It might be nice to move it up top in the smaller viewports. I would rather leave it where it is or move it up than remove it completely though.

ryanseys:
Can you move it up without moving the whole sidebar to the top?

stephenplusplus:
I’ll try :) I’m thinking a directive in two places, whose visibility toggles via CSS at the appropriate dimensions.

PR #617, opened 2015-05-20
docs: switch homepage example to GCS

### LINKED ISSUE #615: Is the GCE + Datastore example code functional ?
The example code we have is

```javascript
// Auth is handled automatically inside GCE.
var productKey = dataset.key(['Product', 123]);

dataset.get(productKey, function(err, entity) {
  console.log(err, entity);
});
```

Is that accurate? Don't we need to define a project ID and/or dataset ID ?

I ask because we had the debate about how we don't auto-discover the project ID...

stephenplusplus:
Fixes #615 

I think a GCS example might more exciting to the Node developer.

Also tweaked the style a bit to make it pop out a little more.

screen shot 2015-05-20 at 1 58 47 pm

Preview: http://stephenplusplus.github.io/gcloud-node-gh-pages-again/

### DIFF
--- a/docs/site/home.html
+++ b/docs/site/home.html
@@ -84,7 +84,7 @@
       </div>
 
       <div class="quote-box--supplementary">
-        <h4>Example: Retrieve an entity from Cloud Datastore</h4>
+        <h4>Example: Upload a file to Cloud Storage</h4>
 
         <div class="code-sample-tabs" ng-init="selected = 'compute engine'">
           <a ng-click="selected = 'compute engine'"
@@ -96,26 +96,35 @@
         </div>
 
         <div hljs language="js" ng-show="selected == 'compute engine'">
-// Auth is handled automatically inside GCE.
-var gcloud = require('gcloud');
-var dataset = gcloud.datastore.dataset();
-var productKey = dataset.key(['Product', 123]);
+// Authentication is automatic
+// inside Google Compute Engine.
+var gcloud = require('gcloud')({
+  projectId: 'grape-spaceship-123'
+});
 
-dataset.get(productKey, function(err, entity) {
-  console.log(err, entity);
+var gcs = gcloud.storage();
+
+var backups = gcs.bucket('backups');
+backups.upload('db.zip', function(err, file) {
+  // file.createReadStream();
+  // file.getMetadata();
+  // ...
 });</div>
 
         <div hljs language="js" ng-show="selected == 'elsewhere'">
 var gcloud = require('gcloud')({
-  projectId: 'my-project',
+  projectId: 'grape-spaceship-123',
+  // Specify a path to a keyfile.
   keyFilename: '/path/to/keyfile.json'
 });
 
-var dataset = gcloud.datastore.dataset();
-var productKey = dataset.key(['Product', 123]);
+var gcs = gcloud.storage();
 
-dataset.get(productKey, function(err, entity) {
-  console.log(err, entity);
+var backups = gcs.bucket('backups');
+backups.upload('db.zip', function(err, file) {
+  // file.createReadStream();
+  // file.getMetadata();
+  // ...
 });</div>
 
       </div>
--- a/docs/site/css/main.css
+++ b/docs/site/css/main.css
@@ -131,11 +131,6 @@ body {
    ========================================================================== */
 
 pre {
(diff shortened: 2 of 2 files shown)


ryanseys:
`grape-spaceship-123` lives on!

docs/site/home.html
```diff
@@ -96,26 +96,35 @@ <h3 class="block-title">What is it?</h3>
...
+var gcloud = require('gcloud')({
+  projectId: 'grape-spaceship-123'
+});
+
+var gcs = gcloud.storage();
 
-dataset.get(productKey, function(err, entity) {
-  console.log(err, entity);
+var backups = gcs.bucket('backups');
+backups.upload('db.zip', function(err, file) {
```
jgeewax: Just to be clear, this uploads the local file, right?

Wouldn't the logical next action be something logging that the upload worked or checking if there were errors?

```javascript
if (err) {
  pager.create(/* Wake up your dev team! */);
} else {
  log('Backup of ' + file + ' successful.');
}
```

docs/site/home.html
```diff
@@ -96,26 +96,35 @@ <h3 class="block-title">What is it?</h3>
...
 var gcloud = require('gcloud')({
-  projectId: 'my-project',
+  projectId: 'grape-spaceship-123',
+  // Specify a path to a keyfile.
   keyFilename: '/path/to/keyfile.json'
 });
 
-var dataset = gcloud.datastore.dataset();
-var productKey = dataset.key(['Product', 123]);
+var gcs = gcloud.storage();
```
jgeewax: I think our docs use 'storage' as the variable name, right? (`var storage = gcloud.storage()`)

docs/site/home.html
stephenplusplus: I thought it was better to quit the funny business and get to the stuff a dev wants to see; examples of what a file object can do.

docs/site/home.html
stephenplusplus: Yes, and I forgot to draw attention to this, thank you. This started as an effort in preventing wrapping lines, but then I began to think it made more sense than "storage". It does have a bit of mystery at first for the non-gcs user, but once you are one, gcs might be a bit more helpful for a frame of context than the more general "storage" term. Happy to go either way.

jgeewax: How about we err on consistency's side now, and discuss a big sweeping "change all var storage -> var gcs" change as a separate issue?

docs/site/home.html
jgeewax: I was confused because I was like... wait... I uploaded the file, why do I want to do more with it... ?

docs/site/home.html
stephenplusplus: Good idea: #619

docs/site/home.html
stephenplusplus: I don't think it's important to live up to a fictional scenario, so much that it becomes believable/copy-and-paste-able. I think a good example has to be somewhat real to make the user feel comfortable with the concept, but also needs to show what can be done with our API to spark some ideas for how the user can use it in their own project.

jgeewax: Sure, but I'm confused what I'd want to do with `file`... I'd be fine if the example said `if (!err) { keep_calm_and_carry_on() }`... saying `file.createReadStream()` makes me thing... wait... is `upload` just giving me access to a file... and I still need to pipe it's contents somewhere?

stephenplusplus: Let's pull @ryanseys over. Did this confuse you? Any ideas for a better example?

ryanseys: It didn't confuse me but I didn't consider if it made much sense. Thinking now, the file that is returned already has its metadata so getting metadata again actually doesn't make sense. Also creating a read stream doesn't make sense either because it would be smarter to create a read stream from the local file, not the remote one. 

The best idea I think is that once I upload a file to storage, I want to store its media url for download or embedding (e.g. in an image tag) somewhere, and we can get that from the metadata (technically its a well-known, but we should verify that it uploaded properly first).

So something like this I think would be useful:

``` js
var assets = gcs.bucket('my-app-assets');
assets.upload('kitten.jpg', function(err, file) {
  if (!err) {
    // file.metadata.mediaLink
  }
});
```

PR #618, opened 2015-05-20
Move subscription methods up a level

### LINKED ISSUE #605: Add pubsub#subscription, pubsub#subscribe and pubsub#getSubscriptions
Add `pubsub#subscription` and `pubsub#subscribe` and `pubsub#getSubscriptions`.

All the logic I feel should move to these requests and then `topic#subscription` and `topic#subscribe` and `topic#getSubscriptions` should either be removed or rely on the pubsub version with an optional `topic` field automatically passed in.

How's this sound @stephenplusplus?

Reference: #487 

/cc @tmatsuo

stephenplusplus:
Fixes #605
- [x] Logic
- [x] Unit tests
- [x] System tests
  - [x] Old functionality
  - [ ] :warning: New functionality - https://github.com/GoogleCloudPlatform/gcloud-node/pull/618#issuecomment-108022564
- [x] Docs

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -124,7 +124,7 @@ util.arrayize = arrayize;
  */
 function format(template, args) {
   return template.replace(/{([^}]*)}/g, function(match, key) {
-    return args[key] || match;
+    return key in args ? args[key] : match;
   });
 }
 
--- a/lib/pubsub/index.js
+++ b/lib/pubsub/index.js
@@ -163,6 +163,138 @@ PubSub.prototype.createTopic = function(name, callback) {
 };
 
 /**
+ * Create a subscription to a topic. You may optionally provide an object to
+ * customize the subscription.
+ *
+ * Your provided callback will be invoked with an error object if an API error
+ * occurred or a {@linkcode module:pubsub/subscription} object.
+ *
+ * @throws {Error} If a Topic instance or topic name is not provided.
+ * @throws {Error} If a subName is not provided.
+ *
+ * @param {module:pubsub/topic|string} - topic - The Topic to create a
+ *     subscription to.
+ * @param {string} subName - The name of the subscription.
+ * @param {object=} options - Configuration object.
+ * @param {number} options.ackDeadlineSeconds - The maximum time after receiving
+ *     a message that you must ack a message before it is redelivered.
+ * @param {boolean} options.autoAck - Automatically acknowledge the message once
+ *     it's pulled. (default: false)
+ * @param {number} options.interval - Interval in milliseconds to check for new
+ *     messages. (default: 10)
+ * @param {boolean} options.reuseExisting - If the subscription already exists,
+ *     reuse it. The options of the existing subscription are not changed. If
+ *     false, attempting to create a subscription that already exists will fail.
+ *     (default: false)
+ * @param {function} callback - The callback function.
+ *
+ * @example
+ * //-
+ * // Subscribe to a topic. (Also see {module:pubsub/topic#subscribe}).
+ * //-
+ * var topic = 'messageCenter';
+ * var name = 'newMessages';
+ *
+ * pubsub.subscribe(topic, name, function(err, subscription, apiResponse) {});
+ *
+ * //-
+ * // Customize the subscription.
+ * //-
+ * pubsub.subscribe(topic, name, {
+ *   ackDeadlineSeconds: 90,
+ *   autoAck: true,
+ *   interval: 30
+ * }, function(err, subscription, apiResponse) {});
+ *
+ * //-
+ * // Create a subscription to a topic from another project.
+ * //-
+ * var anotherProject = gcloud.pubsub({
+ *   projectId: 'another-project'
+ * });
+ *
+ * var topic = anotherProject.topic('messageCenter');
+ *
+ * pubsub.subscribe(topic, name, function(err, subscription, apiResponse) {});
+ */
+PubSub.prototype.subscribe = function(topic, subName, options, callback) {
+  if (!util.is(topic, 'string') && !(topic instanceof Topic)) {
+    throw new Error('A Topic is required for a new subscription.');
+  }
+
+  if (!util.is(subName, 'string')) {
+    throw new Error('A subscription name is required for a new subscription.');
+  }
+
+  if (!callback) {
+    callback = options;
+    options = {};
+  }
+
+  options = options || {};
+
+  if (util.is(topic, 'string')) {
+    topic = this.topic(topic);
+  }
+
+  var body = {
+    topic: topic.name
+  };
+
+  if (options.ackDeadlineSeconds) {
+    body.ackDeadlineSeconds = options.ackDeadlineSeconds;
+  }
+
+  var subscription = this.subscription(subName, options);
+
+  this.makeReq_('PUT', subscription.name, null, body, function(err, result) {
+    if (err && !(err.code === 409 && options.reuseExisting)) {
+      callback(err, null, result);
+      return;
+    }
+
+    callback(null, subscription, result);
+  });
+};
+
+/**
+ * Create a Subscription object in reference to an existing subscription. This
+ * command by itself will not run any API requests. You will receive a
+ * {@linkcode module:pubsub/subscription} object, which will allow you to
+ * interact with your subscription.
+ *
+ * @throws {Error} If a name is not provided.
+ *
+ * @param {string} name - The name of the subscription.
+ * @param {object=} options - Configuration object.
+ * @param {boolean} options.autoAck - Automatically acknowledge the message once
+ *     it's pulled.
+ * @param {number} options.interval - Interval in milliseconds to check for new
+ *     messages.
+ * @return {module:pubsub/subscription}
+ *
+ * @example
+ * var subscription = pubsub.subscription('my-existing-subscription');
+ *
+ * // Register a listener for `message` events.
+ * subscription.on('message', function(message) {
+ *   // Called every time a message is received.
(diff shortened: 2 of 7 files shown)


lib/pubsub/index.js
```diff
@@ -163,6 +163,124 @@ PubSub.prototype.createTopic = function(name, callback) {
...
 /**
+ * Create a subscription to this topic. You may optionally provide an object to
+ * customize the subscription.
+ *
+ * Your provided callback will either be invoked with an error object, if an API
+ * error occurred, or a {@linkcode module:pubsub/subscription} object.
+ *
+ * @throws {Error} If a name is not provided.
+ *
+ * @param {string} name - The name of the subscription.
```
ryanseys: Seems these params are still old.

ryanseys:
Different than what I had in mind.

I was thinking more like:

``` js
// topic is in options
pubsub.subscribe('my-sub', { topic: pubsub.topic('some-topic') }, callback);
```

Thinking now, `pubsub.subscribe('sub-name', topicObj, options, callback)` makes more sense because its not an option in this case BUT a topic name is specified by its project and topic name like so: `'projects/project-id/topics/topic-name'` so with your string version here, that wouldn't be the best unless that assumes that the topic is in the same project. It would be better to pass in a whole topic object.

Here's a partial attempt I made a while ago: https://github.com/ryanseys/gcloud-node/commit/78c4cfe6b06a7dcc1d44e226b00dbbc51a31d096 don't let it confuse you because it's not complete but something to gander at.

lib/pubsub/index.js
stephenplusplus: Sorry, should have mentioned these were half-worked on.

stephenplusplus:
> BUT a topic name is specified by its project and topic name like so: 'projects/project-id/topics/topic-name

I'm just getting the topic name (`topic-name`), then if the user gives a project-id, that project-id is used to make the full topic name, otherwise, use what's on the PubSub instance.

Quick API comparison :)

```js
// ---
// This PR:
// ---
var pubsub = gcloud.pubsub({ projectId: 'grape-spaceship-123' });

// Each of these next two lines do the same thing:
pubsub.topic('some-topic').subscribe('my-sub', callback);
pubsub.subscribe('some-topic', 'new-sub-name', callback);

// Subscribe to a topic from another project.
pubsub.subscribe('another-topic', 'another-new-sub-name', { projectId: 'squishy-butter-334' }, callback);

// ---
// Your Idea:
// ---
var pubsub = gcloud.pubsub({ projectId: 'grape-spaceship-123' });

// Each of these next two lines do the same thing:
pubsub.topic('some-topic').subscribe('my-sub', callback);
pubsub.subscribe('new-sub-name', pubsub.topic('some-topic'), callback);

// Subscribe to a topic from another project.
var anotherPubsub = gcloud.pubsub({ projectId: 'squishy-butter-334' });

pubsub.subscribe('another-new-sub-name', anotherPubsub.topic('another-topic'), callback);
```

The main difference is requiring a Topic object vs. using explicit string params. I like both. Halp.

jgeewax:
Couldn't we take bits of both?

```javascript
var pubsub1 = gcloud.pubsub({ projectId: 'project1' });

// These all do the same thing
pubsub1.topic('some-topic').subscribe('my-sub', callback); // Most common pattern
pubsub1.subscribe('new-sub-name', 'some-topic', callback);  // Uncommon, but just a name implies same project
pubsub1.subscribe('new-sub-name', pubsub1.topic('some-topic'), callback); // Uncommon, most explicit.

// So then, if we want to subscribe to a topic in another project,
// the only way to do that is to create a topic object that knows it's project.
var pubsub2 = gcloud.pubsub({ projectId: 'project2' });
pubsub2.subscribe('new-sub-name', pubsub1.topic('some-topic'), callback);
```

stephenplusplus:
I'm cool with that. Now just one smaller detail; shouldn't topic name come before subscription?

```js
pubsub.subscribe('new-sub-name', 'some-topic', callback);
// vs.
pubsub.subscribe('some-topic', 'new-sub-name', callback);
```

I like the latter, just to follow the standard hierarchy "project -> topic -> subscription".

--

Just for completeness, here's the full sig:

```js
pubsub.topic('topic-name', 'sub-name', { autoAck: true }, function(err, subscription, apiResponse) {});
```

jgeewax:
SGTM. Since topic is absolutely required, it should definitely be first.

ryanseys:
> Since topic is absolutely required, it should definitely be first.

Aren't both topic and sub-name required? Either order makes sense to me, so whatever you like :)

jgeewax:
Yea, both required -- topic being higher up in the hierarchy seems like it should come first though. Sorry, didn't articulate that piece too. :P

ryanseys:
Cool SGTM!

stephenplusplus:
PTAL.

lib/pubsub/index.js
```diff
@@ -163,6 +163,123 @@ PubSub.prototype.createTopic = function(name, callback) {
 };
 
 /**
+ * Create a subscription to this topic. You may optionally provide an object to
```
ryanseys: > to this topic.

to a topic.

lib/pubsub/index.js
```diff
@@ -163,6 +163,123 @@ PubSub.prototype.createTopic = function(name, callback) {
...
+
+  var body = {
+    topic: topic.name
+  };
+
+  if (options.ackDeadlineSeconds) {
+    body.ackDeadlineSeconds = options.ackDeadlineSeconds;
+  }
+
+  var subscription = topic.subscription(subName, options);
```
ryanseys: > `topic.subscription(...`

`this.subscription(...` will accomplish the same thing and looks less wrong. But I might be nitpicking this.

stephenplusplus: `topic` comes in as a topic instance, or if a string, see line 222. So, it may not be equal to `this`.

stephenplusplus: And with such a big change, I'm relieved to get nitpicks :)

ryanseys: Our context here is PubSub though, so we're really doing `pubsub.subscription(...` which does the same thing as `topic.subscription(...` but with less layers and like I said, just looks less wrong.

stephenplusplus: But their topic could be bound to a different PubSub instance. It's possible I'm in an infinite confusion loop.

ryanseys: Yeah but our subscription is being made on OUR pubsub object :) just to a topic in another project potentially.

ryanseys: So actually `this.subscription(...` is right and `topic.subscription(...` is wrong.

stephenplusplus: What was that noise? Oh, I know. It was Stephen slamming his head into the desk.

ryanseys: Our subscription object always has to be made on our own pubsub object. But it can subscribe to a topic from another project. Does this make sense?

ryanseys: HAHAHA

stephenplusplus: Yeah, thanks :) Will fix!

stephenplusplus: Can you check if https://github.com/GoogleCloudPlatform/gcloud-node/commit/8068181f52e9fd6bf65cf3e0afe372f6101ac889 is right?

stephenplusplus:
While Travis takes a year to run this build...

I'm not sure how I can write a system test for this functionality, since we only have access to one project. Unit testing covers all of the changes pretty well, since it was mostly just relocating code. To support subscribing from another project, all we're really doing is passing a different topic string to the API. The unit tests cover that as well. Is that enough?

ryanseys:
I think this should be good. I can go ahead with a code review if you're ready?

stephenplusplus:
Yes, please!

lib/pubsub/index.js
```diff
@@ -163,6 +163,135 @@ PubSub.prototype.createTopic = function(name, callback) {
...
+  }
+
+  if (!callback) {
+    callback = options;
+    options = {};
+  }
+
+  options = options || {};
+
+  if (util.is(topic, 'string')) {
```
ryanseys: This way of specifying a topic is undocumented in the examples.

stephenplusplus: Added! https://github.com/GoogleCloudPlatform/gcloud-node/commit/06692899f7fa5a0b42faa8249b830d6a13268a73#diff-61808abb3ded8836d16d00ca68259da8R192

lib/common/util.js
```diff
@@ -124,7 +124,7 @@ util.arrayize = arrayize;
  */
 function format(template, args) {
   return template.replace(/{([^}]*)}/g, function(match, key) {
-    return args[key] || match;
+    return key in args ? args[key] : match;
```
odeke-em: Am curious, how come this change? Doesn't the single lookup and indexing translate directly and in less steps to the intention of the method?

stephenplusplus: It previously used the value of `args[key]` to decide if it should be used in the resulting string. Instead, we should confirm the presence of the key in the provided object and use its value, regardless of if it passes a truthy test.

This wasn't an issue until in this PR where we needed something similar to:

```js
format('/api/path{pathExtension}', {
  pathExtension: condition ? 'extension' : ''
});
```

Previous result if `condition` was false: `/api/path{pathExtension}`
New result: `/api/path`

odeke-em: That was a good catch, now I see what you mean. Thank you for the clarification!

stephenplusplus:
Merge time!

Issue #619, opened 2015-05-21
Rename Storage examples from "var storage" to "var gcs"


stephenplusplus:
Continues https://github.com/GoogleCloudPlatform/gcloud-node/pull/617#discussion-diff-30792128...

It seems a lot of people refer to "Cloud Storage" as "GCS" as opposed to "Storage". In userland, I've seen [`var gcs = gcloud.storage(/*...*/);`](https://github.com/GoogleCloudPlatform/gcloud-node/issues/522) more than once, and tend to prefer that name over the more generic "var storage;" naming.

Any thoughts on changing our Storage examples to use `var gcs = gcloud.storage();` in place of `var storage = gcloud.storage();`? The term "storage" feels vague to me, instead of implying that it offers functionality that sits above a bucket.

jgeewax:
I'm a fan of using `gcs` in our examples.

ryanseys:
I'm ok with this idea.

PR #620, opened 2015-05-21
docs: var gcs -> var storage

stephenplusplus:
While we figure out if we want #619 or not, we should be consistent with our naming.


jgeewax:
LGTM

PR #622, opened 2015-05-22
docs: use a better example of how to rollback

### LINKED ISSUE #621: rollback example in current docs throws exception
The current example under transaction.rollback e.g.

```javascript
transaction.begin(function(err) {
  transaction.rollback(function(err) {
    if (err) {
      // Transaction could not be rolled back.
    }
  });
});
```

will produce the following exception

```
Error: Missing at least one required field for Message .pb.RollbackRequest: .pb.RollbackRequest.transaction
    at Error (<anonymous>)
    at ProtoBuf.Reflect.MessagePrototype.encode (/app/node_modules/gcloud/node_modules/protobufjs/dist/ProtoBuf.js:2345:31)
    at MessagePrototype.encode (/app/node_modules/gcloud/node_modules/protobufjs/dist/ProtoBuf.js:1981:31)
    at MessagePrototype.encodeNB (/app/node_modules/gcloud/node_modules/protobufjs/dist/ProtoBuf.js:2063:41)
    at Transaction.DatastoreRequest.makeReq_ (/app/node_modules/gcloud/lib/datastore/request.js:599:51)
    at Transaction.rollback (/app/node_modules/gcloud/lib/datastore/transaction.js:135:8)
    at Object.queryCb (/app/route-callbacks.js:558:19)
    at /app/node_modules/gcloud/lib/datastore/request.js:459:9
    at /app/node_modules/gcloud/lib/datastore/request.js:640:13
    at Object.handleResp (/app/node_modules/gcloud/lib/common/util.js:193:3)
```

Is that still the proper way to rollback? or should we simply call done()... when should rollback be actually used? after a commit and if the commit returns an error?

Thanks for any advice.

stephenplusplus:
Fixes #621 
Related #623 

This is a band-aid on the docs. This is technically correct in how to use our Transaction API, but isn't very useful for the user (see #623)

### DIFF
--- a/lib/datastore/transaction.js
+++ b/lib/datastore/transaction.js
@@ -119,11 +119,10 @@ Transaction.prototype.begin = function(callback) {
  * @param {function} callback - The callback function.
  *
  * @example
- * transaction.begin(function(err) {
- *   transaction.rollback(function(err) {
- *     if (err) {
- *       // Transaction could not be rolled back.
- *     }
+ * transaction.commit(function(err) {
+ *   if (err) {
+ *     // Rollback the transaction.
+ *     transaction.rollback(function(err, apiResponse) {});
  *   });
  * });
  */


ryanseys:
if `done` is meant to be used instead of `transaction.commit` then should this example not show that instead?

stephenplusplus:
Might as well just let this get fixed by #627.

ryanseys:
This won't be fixed (or at least detected as faulty) by #627 due to the async nature of those calls and the vm's inability to wait for an async call to complete.

stephenplusplus:
True, but I don't think it's important to rush this PR, since that one will stomp any changes here for the better.

ryanseys:
You're right. I'm going to continue building on #627 today.

PR #624, opened 2015-05-23
Use alternate Storage.Object.get API URL pattern to avoid SSL certifi…

stephenzsy:
…cated error if there are '.' in the bucket name

npm test all passed


ryanseys:
This is a weird one. Overall the fix looks good to me. @stephenzsy could you try and add a unit test for this too? If not, let me know and I'll add a test when I merge this in.

stephenplusplus:
Resolved by #638. Thanks @stephenzsy!

PR #627, opened 2015-05-25
Verify @example docs on npm test

### LINKED ISSUE #598: Is there any way to make our example code in docstrings part of the test suite?
I'm worried about code rot and typos -- is there any way we can make it so that the snippets we have are actually tested ? Could we pull them out of a unit test somewhere? or a test suite that is specifically samples and we load based on an ID of sorts?

### LINKED ISSUE #623: Transaction Documentation
Related: #621 

Should we privatize `transaction#begin` and `transaction#commit`, removing them from the public docs and their usage in other Transaction examples?

A user goes through their Dataset instance to get a Transaction:

```js
dataset.runInTransaction(transaction, done) {});
```

This calls `transaction.begin` immediately, and the `done` function returned to their callback is actually `transaction.commit`.

Having these methods appear in the docs is probably quite confusing, as the user shouldn't need (or be able to) begin a transaction that already began, or commit the transaction before calling `done` (commit) again.

ryanseys:
Fixes #598
Fixes #622 
Fixes #623 

I really wanted to build this because it seemed like an interesting problem and this builds on our already-epic docs to make them more epic.

This adds an additional test which will run with `npm test` like normal, though I would suggest we consider changing `npm test` to also generate docs, then run the tests to ensure the test remains up to date with the docs generated.

The idea is that we read in `docs/**/*.json` files and iterate through every appropriate `@example` tag, extract the code string, and run it inside a [node.js vm sandbox](https://nodejs.org/api/vm.html), ensuring that it doesn't throw an error along the way. This means that it will check for general syntax errors, as well as calling methods with the wrong number of parameters or missing parameters or just plain incorrect parameters.

A lot of this PR is cleaning up the docs so the test passes! Yeah, it caught a wide range of issues! :+1: 

One thing to note is that it has no way of running anything asynchronously or waiting in anyway for any async function to be called in the VM because we'd have to then inject our own code into the existing examples and that would be quite difficult to guess.

For example `Dataset#runInTransaction` (which you'll see a lot of in this PR) actually is an async function and so it makes it impossible to test anything that is included in the callback of that function, and unfortunately the only way we recommend they create a transaction object is by running `dataset.runInTransaction()` so I'm not sure what we should do to mitigate this issue and test our transaction object. I'll comment on the PR where this issue takes place.

### DIFF
--- a/lib/bigquery/job.js
+++ b/lib/bigquery/job.js
@@ -29,7 +29,12 @@ var util = require('../common/util');
 /*! Developer Documentation
  *
  * @param {module:bigquery} bigQuery - BigQuery instance.
- * @param {string} id - The ID of the table.
+ * @param {string} id - The ID of the job.
+ *
+ * @example
+ * var bigquery = gcloud.bigquery();
+ * var Job = require('gcloud/lib/bigquery/job');
+ * var job = new Job(bigquery, 'job-id');
  */
 /**
  * Job objects are returned from various places in the BigQuery API:
--- a/lib/datastore/dataset.js
+++ b/lib/datastore/dataset.js
@@ -78,6 +78,7 @@ var SCOPES = [
  *
  * @example
  * var datastore = gcloud.datastore;
+ *
  * var dataset = datastore.dataset({
  *   projectId: 'my-project',
  *   keyFilename: '/path/to/keyfile.json'
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -64,6 +64,15 @@ var MODE_TRANSACTIONAL = 'TRANSACTIONAL';
  *
  * Creates requests to the Dataset endpoint. Designed to be inherited by
  * datastore.Dataset and datastore.Transaction objects.
+ *
+ * @example
+ * // This is how to create a transaction object directly using this Transaction
+ * // class. The following transaction object is created for use in the examples
+ * // in this file below.
+ * var dataset = gcloud.datastore.dataset();
+ * var Transaction = require('gcloud/lib/datastore/transaction');
+ * var transaction = new Transaction(dataset, 'my-project-id');
+ * transaction.id = '1234'; // Give the transaction an ID.
  */
 /*
  * Handle logic for Datastore API operations.
@@ -90,10 +99,10 @@ function DatastoreRequest() {}
  *
  * // Get a single entity.
  * var key = dataset.key(['Company', 123]);
- * dataset.get(key, function(err, entity, apiResponse) {});
+ * transaction.get(key, function(err, entity, apiResponse) {});
  *
  * // Get multiple entities at once.
- * dataset.get([
+ * transaction.get([
  *   dataset.key(['Company', 123]),
  *   dataset.key(['Product', 'Computer'])
  * ], function(err, entities, apiResponse) {});
@@ -415,11 +424,16 @@ DatastoreRequest.prototype.save = function(entities, callback) {
  * @param {function} callback - The callback function.
  *
  * @example
+ * //-
+ * // Where you see `transaction`, assume this is the context that's relevant to
+ * // your use case, whether that be a Dataset or a Transaction object.
+ * //-
+ *
  * // Delete a single entity.
- * dataset.delete(dataset.key(['Company', 123]), function(err) {});
+ * transaction.delete(dataset.key(['Company', 123]), function(err, apiResp) {});
  *
  * // Delete multiple entities at once.
- * dataset.delete([
+ * transaction.delete([
  *   dataset.key(['Company', 123]),
  *   dataset.key(['Product', 'Computer'])
  * ], function(err, apiResponse) {});
@@ -460,13 +474,18 @@ DatastoreRequest.prototype.delete = function(keys, callback) {
  *     stream instance is returned.
  *
  * @example
+ * //-
+ * // Where you see `transaction`, assume this is the context that's relevant to
+ * // your use, whether that be a Dataset or a Transaction object.
+ * //-
+ * var query = dataset.createQuery('Lion');
  *
  * // Retrieve 5 companies.
- * dataset.runQuery(query, function(err, entities, endCursor, apiResponse) {
+ * transaction.runQuery(query, function(err, entities, endCursor, apiResponse) {
  *   // Use `endCursor` as the starting cursor for your next query.
  *   var nextQuery = query.start(endCursor);
  *   var callback = function(err, entities, endCursor, apiResponse) {};
- *   dataset.runQuery(nextQuery, callback);
+ *   transaction.runQuery(nextQuery, callback);
  * });
  *
  * //-
@@ -474,7 +493,7 @@ DatastoreRequest.prototype.delete = function(keys, callback) {
  * // queries until no results remain. Entity objects will be pushed as they are
  * // found.
  * //-
- * dataset.runQuery(query)
+ * transaction.runQuery(query)
  *   .on('data', function (entity) {});
  */
 DatastoreRequest.prototype.runQuery = function(q, callback) {
@@ -556,7 +575,7 @@ DatastoreRequest.prototype.runQuery = function(q, callback) {
  *
  * @example
  * //-
- * // Where you see `dataset`, assume this is the context that's relevant to
+ * // Where you see `transaction`, assume this is the context that's relevant to
  * // your use, whether that be a Dataset or a Transaction object.
  * //-
  *
@@ -564,7 +583,7 @@ DatastoreRequest.prototype.runQuery = function(q, callback) {
  *
  * // The following call will create 100 new IDs from the Company kind, which
  * // exists under the default namespace.
- * dataset.allocateIds(incompleteKey, 100, function(err, keys) {});
+ * transaction.allocateIds(incompleteKey, 100, function(err, keys) {});
  *
  * // You may prefer to create IDs from a non-default namespace by providing an
  * // incomplete key with a namespace. Similar to the previous example, the call
@@ -575,7 +594,7 @@ DatastoreRequest.prototype.runQuery = function(q, callback) {
  *   path: ['Company']
  * });
  * var callback = function(err, keys, apiResponse) {};
- * dataset.allocateIds(incompleteKey, 100, callback);
+ * transaction.allocateIds(incompleteKey, 100, callback);
  */
 DatastoreRequest.prototype.allocateIds = function(incompleteKey, n, callback) {
   if (entity.isKeyComplete(incompleteKey)) {
@@ -635,6 +654,8 @@ DatastoreRequest.prototype.upsert = function(entities, callback) {
  *     delete: [] // datastore key objects.
  *   }
  * };
+ *
+ * var dataset = gcloud.datastore.dataset();
  * var callback = function(err, result, apiResponse) {};
  * var Transaction = require('gcloud/lib/datastore/transaction');
  * var transaction = new Transaction(dataset, 'my-project-id');
--- a/lib/datastore/transaction.js
+++ b/lib/datastore/transaction.js
@@ -47,6 +47,7 @@ var extend = require('extend');
  * // This is how to create a transaction object directly using this Transaction
  * // class. The following transaction object is created for use in the examples
  * // in this file below.
+ * var dataset = gcloud.datastore.dataset();
  * var Transaction = require('gcloud/lib/datastore/transaction');
  * var transaction = new Transaction(dataset, 'my-project-id');
  * transaction.id = '1234'; // Give the transaction an ID.
@@ -263,13 +264,13 @@ Transaction.prototype.commit_ = function(callback) {
  *
  * @example
  * // Delete a single entity.
- * transaction.delete(dataset.key(['Company', 123]), function(err) {});
+ * transaction.delete(dataset.key(['Company', 123]));
  *
  * // Delete multiple entities at once.
  * transaction.delete([
  *   dataset.key(['Company', 123]),
  *   dataset.key(['Product', 'Computer'])
- * ], function(err) {});
+ * ]);
  */
 Transaction.prototype.delete = function(entities) {
   var that = this;
--- a/lib/storage/bucket.js
+++ b/lib/storage/bucket.js
@@ -126,6 +126,7 @@ function Bucket(storage, name) {
    * //-
    * // Make a bucket's contents publicly readable.
    * //-
+   * var myBucket = storage.bucket('my-bucket');
    * myBucket.acl.add({
    *   scope: 'allUsers',
    *   role: storage.acl.READER_ROLE
@@ -218,12 +219,14 @@ function Bucket(storage, name) {
  * @param {function=} callback - The callback function.
  *
  * @example
- * var logs2013 = bucket.file('2013-logs.txt');
- * var logs2014 = bucket.file('2014-logs.txt');
+ * var logBucket = storage.bucket('log-bucket');
  *
- * var allLogs = bucket.file('all-logs.txt');
+ * var logs2013 = logBucket.file('2013-logs.txt');
+ * var logs2014 = logBucket.file('2014-logs.txt');
  *
- * bucket.combine([
+ * var allLogs = logBucket.file('all-logs.txt');
+ *
+ * logBucket.combine([
  *   logs2013,
  *   logs2014
  * ], allLogs, function(err, newFile, apiResponse) {
@@ -303,6 +306,7 @@ Bucket.prototype.combine = function(sources, destination, callback) {
  * @param {function=} callback - The callback function.
  *
  * @example
+ * var bucket = storage.bucket('delete-me');
  * bucket.delete(function(err, apiResponse) {});
  */
 Bucket.prototype.delete = function(callback) {
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -17,7 +17,7 @@
 set -ev
 
 npm run lint
-npm run docs
+# generates docs and runs tests
 npm run test
 
 # if merging to master and not a pull request, execute system tests, create coverage report and update docs
@@ -26,8 +26,7 @@ if [ "${TRAVIS_BRANCH}" == "master" ] && [ "${TRAVIS_PULL_REQUEST}" == "false" ]
   # create new coverage report (executes system tests)
   npm run coveralls
 
-  # generate new set of json files in docs/json/master
-  npm run docs
+  # add new docs to the gh-pages branch
   git submodule add -b gh-pages https://${GH_OAUTH_TOKEN}@github.com/${GH_OWNER}/${GH_PROJECT_NAME} ghpages
   # copy all the docs file that might have changed, excluding versions.txt (to avoid overriding it)
   cd docs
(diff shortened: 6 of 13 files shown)


lib/datastore/request.js
```diff
@@ -258,13 +258,13 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
  *   key: key,
  *   data: {
  *     name: 'DonutShack', // strings
- *     rating: datastore.int(8), // ints
- *     worth: datastore.double(123456.78), // doubles
+ *     rating: gcloud.datastore.int(8), // ints
+ *     worth: gcloud.datastore.double(123456.78), // doubles
  *     numDonutsServed: 45, // detect number type (int or double)
- *     founded: new Date('Tue May 12 2015 15:30:00 GMT-0400 (EDT)'), // dates
+ *     // founded: new Date('Tue May 12 2015 15:30:00 GMT-0400 (EDT)'), // dates
```
ryanseys: Oops these were commented out because I have no idea why it throws when given a date object. I'll need to investigate this.

lib/datastore/transaction.js
```diff
@@ -85,16 +85,18 @@ nodeutil.inherits(Transaction, DatastoreRequest);
  *     a transaction.
  *
  * @example
- * transaction.begin(function(err) {
- *   // Perform Datastore operations as usual.
- *   transaction.get(dataset.key(['Company', 123]), function(err, entity) {
- *     // Commit the transaction.
- *     transaction.finalize(function(err, apiResponse) {});
+ * dataset.runInTransaction(function(transaction, done) {
```
ryanseys: All of the examples in this file have been wrapped with `dataset.runInTransaction()` however the code inside the callback is not tested because the function is async so for example in this one, `transaction.begin`, `transaction.get` and `transaction.finalize` are never tested (run), except for syntax when the code block is compiled (so it's not a total loss).

test/common/docs.js
```diff
@@ -0,0 +1,93 @@
...
+
+/*global describe, it */
+
+'use strict';
+
+var vm = require('vm');
+var assert = require('assert');
+var gcloud = require('../../');
+var glob = require('glob');
+var mitm = require('mitm');
```
ryanseys: Added `glob` and `mitm` dev deps.

test/common/docs.js
```diff
@@ -0,0 +1,93 @@
...
+    throw new Error('\n' + code + '\n\n' + err.message);
+  }
+}
+
+describe('docs', function() {
+  var MITM;
+  var FILES;
+
+  before(function(done) {
+    MITM = mitm();
```
ryanseys: Turns off the network so the VM doesn't actually make real requests to google when running the examples. Perhaps I should add a comment here for that?

test/common/docs.js
```diff
@@ -0,0 +1,93 @@
...
+    glob('docs/json/**/*.json', {}, function(err, files) {
+      assert.ifError(err);
+      FILES = files;
+      done();
+    });
+  });
+
+  after(function() {
+    MITM.disable();
```
ryanseys: Reenable the network.

lib/bigquery/job.js
```diff
@@ -117,12 +118,13 @@ Job.prototype.getMetadata = function(callback) {
...
  * //-
  * var through2 = require('through2');
+ * var fs = require('fs');
  *
  * job.getQueryResults(options)
  *   .pipe(through2.obj(function (row, enc, next) {
  *     this.push(JSON.stringify(row) + '\n');
  *   }))
- *   .pipe(fs.createWriteStream('./backup.json'));
+ *   .pipe(fs.createWriteStream('./test/testdata/testfile.json'));
```
ryanseys: These actually write to a file (mind you they write nothing)... so having the file exist in our repo helps mitigate it creating additional files. Is this ok?

scripts/build.sh
```diff
@@ -17,6 +17,7 @@
 set -ev
 
 npm run lint
+npm run docs
```
ryanseys: Added this so travis will build docs on normal PRs so it will catch docs issues before they make their way into the repo.

jgeewax:
This is awesome. I totally get it that we can't inspect inside callbacks... maybe there's some magic we can figure out later, but for now this is a huge step forward. :+1:

stephenplusplus:
This is amazing. This is the first time I've seen anything like this in OSS/ever. @jgeewax is there some kind of Google Dev Blog we can write about this on after it's merged?

lib/datastore/transaction.js
```diff
@@ -85,16 +85,18 @@ nodeutil.inherits(Transaction, DatastoreRequest);
...
+ *   transaction.begin(function(err) {
```
stephenplusplus: Remove `transaction.begin` from these examples. Begin is called automatically when `dataset.runInTransaction` is called.

stephenplusplus: I'm getting mixed up if this PR is trying to solve #623. It looks like you wanted to because of wrapping things in `dataset.runInTransaction` and some doc wording changes. If you don't want to solve it in this PR, that's fine, you can leave as is and we can resolve after this merges.

ryanseys:
Adding don't merge so I can continue working on this today to investigate mitigating some of the issues I've made apparent here with the current version.

ryanseys:
Whoa! This docs test found a real live bug! Our `transaction#rollback` didn't attach the transaction id like it should have. I added a unit test to catch it :)

lib/datastore/transaction.js
```diff
@@ -84,16 +89,11 @@ nodeutil.inherits(Transaction, DatastoreRequest);
  * @param {function} callback - The function to execute within the context of
  *     a transaction.
  *
+ * @private
```
ryanseys: Privatize `transaction#begin` without changing the name yet. Should I go ahead and change it to `begin_` to follow convention?

lib/datastore/transaction.js
```diff
@@ -143,10 +141,10 @@ Transaction.prototype.rollback = function(callback) {
  * Commit the remote transaction and finalize the current transaction instance.
  *
  * @param {function} callback - The callback function.
+ * @private
```
ryanseys: Privatize `transaction#commit` without changing the name yet. Should I go ahead and change it to `commit_` to follow convention?

test/common/docs.js
```diff
@@ -0,0 +1,94 @@
...
+  it('should run public docs examples without errors', function() {
+    for (var f in FILES) {
+      var fileDocBlocks = require('../../' + FILES[f]);
+      for (var i in fileDocBlocks) {
+        var doc = fileDocBlocks[i];
+
+        // We don't want to test developer docs or private documentation that
+        // are hidden because they use the modules in a less-recommended way
+        // e.g. use class object directly or call obj.makeReq() directly.
+        if (!doc.isPrivate) {
```
ryanseys: I am working to remove this restriction so all the docs are tested! :) The more, the better! Death to bitrot!

lib/datastore/transaction.js
stephenplusplus: Yeah :+1:

lib/datastore/transaction.js
stephenplusplus: Yeah. Maybe our docs with `runInTransaction` should show `done` named as `commit`. Or just explain calling done will perform the commit.

```js
dataset.runInTransaction(function(transaction, commit) {
  // ...
```

test/common/docs.js
stephenplusplus: Yeah!

lib/datastore/transaction.js
```diff
@@ -42,6 +42,11 @@ var extend = require('extend');
  *     connection to Google Cloud Datastore.
  * @param {string} projectId - Dataset ID. This is your project ID from the
  *     Google Developers Console.
+ *
+ * @example
+ * var Transaction = require('gcloud/lib/datastore/transaction');
```
ryanseys: This is added to the developer documentation, which is hidden from generated docs (as far as I know), however because the VM runs this with the other examples in the file, we now have a proper `transaction` variable that can be used for all the examples that follow.

stephenplusplus:
> Whoa! This docs test found a real live bug! Our transaction#rollback didn't attach the transaction id like it should have. I added a unit test to catch it :)

Nice!

lib/datastore/transaction.js
ryanseys: :+1:

lib/datastore/transaction.js
stephenplusplus: Yeah, dev docs are hidden because of the `/*!` header. Can you add a comment / line in the dev doc block about why that's there?

ryanseys: Will do

lib/datastore/transaction.js
ryanseys: Why do we provide both? Why not just tell them to run `transaction.commit()` ?

stephenplusplus: I think we did it to be Node convention-friendly. Async functions usually gets a function to execute when you're done.

ryanseys: Ah, ok. This makes sense.

ryanseys:
PTAL. Should test all our examples, including ignored and private ones. 
Also privatized `Transaction#begin_` and `Transaction#commit_`.

test/common/docs.js
```diff
@@ -0,0 +1,102 @@
...
+    });
+  });
+
+  after(function() {
+    // Re-enable the network.
+    MITM.disable();
+  });
+
+  it('should run docs examples without errors', function() {
+    for (var f in FILES) {
```
stephenplusplus: `FILES` is an array isn't it? You can do FILES.forEach.

test/common/docs.js
```diff
@@ -0,0 +1,102 @@
...
+      var fileContents = fs.readFileSync(FILES[f], {
+        encoding: 'utf8'
+      });
+
+      try {
+        fileDocBlocks = JSON.parse(fileContents);
+      } catch(e) {
+        throw new Error('Failed to parse one of the doc files (' + e.message +
+            ')\n\nFilename: ' + FILES[f] + '\n\nFile contents: \'' +
+            fileContents + '\'');
```
stephenplusplus: Can you make this message an array, maybe joined by '\n\n', so it's a little easier to look at?

stephenplusplus: Maybe `"` instead of `'` to avoid the escaping will help as well.

test/common/docs.js
```diff
@@ -0,0 +1,102 @@
...
+      }
+
+      for (var i in fileDocBlocks) {
```
stephenplusplus: `fileDocBlocks.forEach`

test/common/docs.js
```diff
@@ -0,0 +1,102 @@
...
+        var tags = fileDocBlocks[i].tags;
+        for (var j in tags) {
```
stephenplusplus: Not sure if `tags` is an array, but `tags.forEach` if it is.

test/common/docs.js
ryanseys: How am I just learning about this now?

giphy

test/common/docs.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+describe('documentation', function() {
+  var MITM;
+  var FILES;
+
+  before(function(done) {
+    // Turn off the network so that API calls aren't actually made.
+    MITM = mitm();
+    glob('docs/json/**/*.json', {}, function(err, files) {
```
stephenplusplus: The empty `{}` isn't necessary.

lib/datastore/request.js
```diff
@@ -465,17 +460,13 @@ DatastoreRequest.prototype.delete = function(keys, callback) {
  *     stream instance is returned.
  *
  * @example
- * //-
- * // Where you see `transaction`, assume this is the context that's relevant to
- * // your use, whether that be a Dataset or a Transaction object.
- * //-
```
stephenplusplus: Can you run me through why some of the text of these are changed and others are removed?

lib/datastore/request.js
```diff
@@ -465,17 +460,13 @@ DatastoreRequest.prototype.delete = function(keys, callback) {
...
  *
  * // Retrieve 5 companies.
- * transaction.runQuery(query, function(err, entities, endCursor, apiResponse) {
+ * dataset.runQuery(query, function(err, entities, endCursor, apiResponse) {
```
stephenplusplus: I don't see where `query` is defined here. Should we make a fake one? Also, shouldn't the new test suite have thrown on this?

jgeewax: +1 - seems weird.. Are we properly scoping variables between examples? Maybe it's picking up a previously defined `query` variable?

stephenplusplus:
This isn't a problem in this PR, but I just discovered something I must have messed up before:

https://github.com/ryanseys/gcloud-node/blob/verify-docs/lib/datastore/transaction.js#L266

Those examples shouldn't show `delete` taking a callback. Feel free to fix while you're in there, or I can always get to it after.

lib/datastore/request.js
ryanseys: You're right JJ, this is likely because the sandbox is shared between files meaning that any values defined in one are accessible in another. I will update the test to use a new sandbox for each file as long as it's relatively easy to migrate our existing code over.

lib/datastore/request.js
ryanseys: I think this is a result of originally not having a transaction object. I will fix this.

ryanseys: Why are we using `transaction` in these examples anyway instead of `dataset`?

test/common/docs.js
```diff
@@ -0,0 +1,100 @@
...
+        fileDocBlocks = JSON.parse(fileContents);
+      } catch(e) {
+        throw new Error([
+          'Failed to parse one of the doc files (' + e.message + ')',
+          'Filename: ' + filename,
+          'File contents: "' + fileContents + '"'
+        ].join('\n'));
+      }
+
+      var sandbox = {
```
ryanseys: Using a fresh sandbox for each file now.

lib/datastore/transaction.js
```diff
@@ -260,13 +264,13 @@ Transaction.prototype.commit = function(callback) {
  *
  * @example
  * // Delete a single entity.
- * transaction.delete(dataset.key(['Company', 123]), function(err) {});
+ * transaction.delete(dataset.key(['Company', 123]));
```
ryanseys: @stephenplusplus Removed callbacks.

ryanseys:
PTAL.

test/common/docs.js
stephenplusplus: :+1:

lib/datastore/dataset.js
```diff
@@ -208,7 +210,10 @@ Dataset.prototype.createQuery = function(namespace, kind) {
 
 /**
  * Run a function in the context of a new transaction. Transactions allow you to
- * perform multiple operations, committing your changes atomically.
+ * perform multiple operations, committing your changes atomically. When you are
+ * finished making your changes within the transaction, run the done() function
+ * provided in the callback function to commit your changes. See an example
+ * below for more information.
```
stephenplusplus: :+1:

lib/datastore/entity.js
```diff
@@ -411,7 +411,7 @@ function valueToProperty(v) {
     p.blob_value = v;
     return p;
   }
-  if (v instanceof Array) {
+  if (Array.isArray(v)) {
```
stephenplusplus: :+1: :+1:

lib/datastore/request.js
```diff
@@ -664,6 +680,10 @@ DatastoreRequest.prototype.makeReq_ = function(method, body, callback) {
     }
   }
 
+  if (method === 'rollback') {
+    body.transaction = this.id;
```
stephenplusplus: :+1: :+1: :+1:

lib/datastore/transaction.js
```diff
@@ -119,12 +123,10 @@ Transaction.prototype.begin = function(callback) {
  * @param {function} callback - The callback function.
  *
  * @example
- * transaction.begin(function(err) {
- *   transaction.rollback(function(err) {
- *     if (err) {
- *       // Transaction could not be rolled back.
- *     }
- *   });
+ * transaction.rollback(function(err) {
```
stephenplusplus: This is actually a bit confusing, because a rollback can only be done on a transaction that has been committed (I think?), but that's a problem:

```js
dataset.runInTransaction(function(transaction, _commit) {
  
}, function(err) {
  // no `transaction` here to call `rollback` on.
})
```

Also, is there a use case for a user wanting to rollback a transaction that they didn't just create? 

Sorry all this transaction stuff is getting lumped into this PR.

lib/datastore/transaction.js
stephenplusplus: Thanks!

lib/storage/bucket.js
```diff
@@ -65,7 +65,7 @@ var RESUMABLE_THRESHOLD = 5000000;
  * @constructor
  * @alias module:storage/bucket
  *
- * @throws if a bucket name isn't provided.
+ * @throws {Error} if a bucket name isn't provided.
```
stephenplusplus: Debugging crown goes to @ryanseys :crown:

lib/storage/bucket.js
```diff
@@ -218,14 +219,16 @@ function Bucket(storage, name) {
  * @param {function=} callback - The callback function.
  *
  * @example
- * var 2013logs = bucket.file('2013-logs.txt');
- * var 2014logs = bucket.file('2014-logs.txt');
```
stephenplusplus: I'll learn JavaScript one of these days.

package.json
```diff
@@ -76,7 +78,7 @@
   "scripts": {
     "docs": "./scripts/docs.sh",
     "lint": "jshint lib/ system-test/ test/",
-    "test": "mocha test/*",
+    "test": "npm run docs && mocha test/*",
```
stephenplusplus: Now that's how tests should be done!

lib/datastore/transaction.js
ryanseys: Can you open a separate issue for this?

lib/storage/bucket.js
ryanseys: If you call *deleting code until it passes then seeing what you deleted to find the issue* debugging.

lib/datastore/transaction.js
ryanseys: Just cuz I don't know this issue on hand and like you said, doesn't really belong in this PR.

lib/storage/bucket.js
stephenplusplus: That's *exactly* what I call debugging.

lib/datastore/transaction.js
stephenplusplus: Of course :) #633

lib/storage/bucket.js
ryanseys: While we're requesting features for the JS tokenizer, I'd like to request supporting hyphens.

stephenplusplus:
This is awesome. You made history.

ryanseys:
Yay! This thing is awesome! :) I'm excited to see it break our builds with no remorse.

PR #634, opened 2015-05-29
fix ordering of documentation tests

stephenplusplus:
This assures the doc tests run last. If it runs between other test `describe`s, I start getting random backend errors.


ryanseys:
Thanks :+1:

PR #638, opened 2015-06-04
storage: Use standard URL to use path instead of subdomain for bucket

phw:
Fixes the SSL certificate being invalid for buckets containing dots.

E.g. if you want to access the bucket `mybucket.example.com` without this patch this results in the URL `https://mybucket.example.com.storage.googleapis.com/` for which the SSL certificate is not valid (the certificate captures direct subdomains only). This patch uses the equally generic URL `https://storage.googleapis.com/mybucket.example.com/`.

### DIFF
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -395,7 +395,7 @@ describe('File', function() {
     }
 
     it('should create an authorized request', function(done) {
-      var expectedPath = util.format('https://{b}.storage.googleapis.com/{o}', {
+      var expectedPath = util.format('https://storage.googleapis.com/{b}/{o}', {
         b: file.bucket.name,
         o: encodeURIComponent(file.name)
       });


phw:
For reference: This bugs was introduced in 93d7a12812e35bf3f998fffdc535200a3ec2f656

stephenplusplus:
Cool, thanks. This should resolve #624 then as well, right?

phw:
I signed it!

stephenplusplus:
Do you want to [fix the tests](https://travis-ci.org/GoogleCloudPlatform/gcloud-node/builds/65404916#L1184) or have me worry about it?

phw:
I think that should fix #624 and it looks IMHO better than the fix provided there (seems a bit hacky). Will fix the unit test in a moment.

phw:
Done. Or do you prefer the commits squashed together?

stephenplusplus:
This is perfect, thanks very much!

phw:
Thank you very much, never got anything merged upstream that fast. Saves me some headaches for the deployment today :)

stephenplusplus:
Great! Never fear submitting patches and ideas here, we'll always get back pretty quickly :)

PR #639, opened 2015-06-04
require project ID

### LINKED ISSUE #570: Raise a helpful error if projectID is missing when making a request?
Right now, if I run the following code on an unconfigured machine, I get a scary back-end error...

```javascript
// (We're not running on App Engine or Compute Engine).
var dataset = gcloud.datastore.dataset();  // <-- Here is the issue (projectId === undefined)

dataset.get(datastore.key(['Person', 1], function(result) {
  console.log(result);
});
```

Seems like if we can't auto-discover the project ID or credentials, we fail with a very long error.

Could we update things so that we raise a specific exception saying "You tried to make a request with a null project ID, meaning we couldn't figure out the project ID and you also didn't tell us the project ID..."

stephenplusplus:
Fixes #570 

``` js
// Doesn't throw.
var gcloud = require('gcloud')({ keyFilename: 'key.json' });

// No project-id, will throw:
gcloud.bigquery();
gcloud.datastore.dataset();
gcloud.pubsub();
gcloud.storage();
```

Other pattern:

``` js
// Doesn't throw.

// No project-id, will throw:
gcloud.bigquery({ keyFilename: 'key.json' });
gcloud.datastore.dataset({ keyFilename: 'key.json' });
gcloud.pubsub({ keyFilename: 'key.json' });
gcloud.storage({ keyFilename: 'key.json' });
```

This involved fixing a lot of the examples, but the change is pretty simple.

### DIFF
--- a/test/common/util.js
+++ b/test/common/util.js
@@ -61,6 +61,10 @@ describe('common/util', function() {
     // Override all util methods, allowing them to be mocked. Overrides are
     // removed before each test.
     Object.keys(util).forEach(function(utilMethod) {
+      if (typeof util[utilMethod] !== 'function') {
+        return;
+      }
+
       util[utilMethod] = function() {
         return (utilOverrides[utilMethod] || util_Cached[utilMethod])
           .apply(this, arguments);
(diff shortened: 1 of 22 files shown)


jgeewax:
What about 

```javascript
var gcloud = require('gcloud')({projectId: 'myproject'});
```

Does anything throw now?

test/common/util.js
```diff
@@ -83,6 +87,16 @@ describe('common/util', function() {
     assert.deepEqual(REQUEST_DEFAULT_CONF, { pool: { maxSockets: Infinity } });
   });
 
+  it('should export an error for module instantiation errors', function() {
+    var missingProjectIdError = new Error([
```
jgeewax: Is this really useful? Won't it mean that we have to change the text twice if we want to modify it?

jgeewax:
LGTM

test/common/util.js
stephenplusplus: It's useful if we really love our error message, and don't want anyone to change it one day without at least having the tests catch it.

stephenplusplus:
> What about var gcloud = require('gcloud')({projectId: 'myproject'}); Does anything throw now?

No, we only check at the instantiation of the sub-modules, which lets the user wait until the last minute to worry about giving the projectId. So as long as there's a projectId somewhere up the chain or at the time of calling, no errors will be thrown

PR #645, opened 2015-06-08
Fixed optional spec in the docs.

jgeewax:
Turns out that:
1. optional is already provided by dox, so we don't need to figure it out too.
2. The way we were figuring out optional was broken (it was finding anything with a URL -- and therefore a `=` character and marking it as optional...)

This PR:
1. Takes out the part where we doubly figure out if things are optional.
2. Changes where optional appears (puts it in the middle, right below the parameter type)


jgeewax:
Just found a small positioning problem, hang tight.

stephenplusplus:
Cool, good find!

PR #652, opened 2015-06-10
bigquery#query: stop passing Job object to API

### LINKED ISSUE #647: BigQuery job body included in many requests that don't need it
Filed internally at [b/21557935](http://b/21557935), moving to public here:

> Ryan, looking at some of our API logs for an unrelated issue, I see gcloud-node is including a giant query string of an entire job configuration for BigQuery API methods that do not require it, such as getQueryResults(). Is there any reason for this? Looking at the gcloud-node code, it seems like simply removing it from all the GET call request options should greatly reduce request size.

I'm guessing the issue starts [here](https://github.com/GoogleCloudPlatform/gcloud-node/blob/01ec51f5b37becccdc20130c6389e8201cc1f389/lib/bigquery/job.js#L141).

stephenplusplus:
Fixes #647 

For developer convenience, when BigQuery#query is used first, it creates a job. From the API response, it creates a Job object, and includes it in a `nextQuery` that is given back to the user (or called itself if in stream mode). We were mistakenly passing that serialized Job object as a query string parameter in API requests to `getQueryResults`.

### DIFF
--- a/lib/bigquery/index.js
+++ b/lib/bigquery/index.js
@@ -361,6 +361,9 @@ BigQuery.prototype.query = function(options, callback) {
 
   options = options || {};
 
+  var requestQuery = extend({}, options);
+  delete requestQuery.job;
+
   if (!util.is(callback, 'function')) {
     stream = streamEvents(through.obj());
     stream.once('reading', runQuery);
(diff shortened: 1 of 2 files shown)


ryanseys:
LGTM is this the only place in the code where that happens?

stephenplusplus:
It's the only place I found, but I asked @weslem over in https://github.com/GoogleCloudPlatform/gcloud-node/issues/647#issuecomment-110768352

stephenplusplus:
Confirmed!

PR #655, opened 2015-06-11
Updated padding in the left nav version switcher area.

jgeewax:
Updated the left-nav padding to be a bit smaller (so the version switcher div is flush with the blue header bar). Screenshot:

leftnavpadding


ryanseys:
LGTM

PR #656, opened 2015-06-11
bigquery: handle case where API doesn't respond with jobReference

### LINKED ISSUE #432: No jobReference when timeout
I got the following error when I run a heavy load job which cause a TIMEOUT. Such that the jobComplete is false. And the jobReference is undefined in this case. 
https://cloud.google.com/bigquery/querying-data

```
TypeError: Cannot read property 'jobId' of undefined at responseHandler (.........../node_modules/gcloud/lib/bigquery/index.js:394:43) 
```

stephenplusplus:
Fixes #432 

The issue covers the problem well, but briefly: on GCE, the upstream API responds with a differently formatted response than non-GCE environments (so it seems).

This PR just shifts around the code to get around it by not relying on a [`getQueryResults`](https://cloud.google.com/bigquery/docs/reference/v2/jobs/getQueryResults) response to contain information about the job.

### DIFF
--- a/lib/bigquery/index.js
+++ b/lib/bigquery/index.js
@@ -361,6 +361,8 @@ BigQuery.prototype.query = function(options, callback) {
 
   options = options || {};
 
+  var job = options.job;
+
   var requestQuery = extend({}, options);
   delete requestQuery.job;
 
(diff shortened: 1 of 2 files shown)


ryanseys:
Weird bug... should I file this internally as well? Also, is it difficult to add a test for this case?

stephenplusplus:
That would be cool, especially to confirm if I'm crazy or not.

For tests, really, this is a better way to handle this than how we were doing it. We don't depend on the API call to `getQueryResults` returning anything now.

stephenplusplus:
This should be a harmless change, so going to merge in. If you get any answers about the API's inconsistencies, please share :)

PR #659, opened 2015-06-16
Added some extra padding on the parent-param div

jgeewax:
Before:
screen shot 2015-06-16 at 7 41 11 am

After:
screen shot 2015-06-16 at 7 39 49 am


stephenplusplus:
:clap:

PR #661, opened 2015-06-17
cache auth client

ofrobots:
The authClient from the google-auth-library ought to be cached. Otherwise we end up creating a new AuthClient on each request which ends up fetching fresh credentials on every API request. This can potentially double the latency of each API request.

I can add JSDoc about config.authClient if this (quick) patch seems reasonable.

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -468,6 +468,8 @@ util.shouldRetryRequest = shouldRetryRequest;
  * for authenticating API requests.
  *
  * @param {object} config - Configuration object.
+ * @param {object=} config.authClient - AuthClient object. If not provided,
+ *     it will be created and cached here.
  * @param {object=} config.credentials - Credentials object.
  * @param {string=} config.email - Account email address, required for PEM/P12
  *     usage.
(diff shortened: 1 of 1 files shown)


lib/common/util.js
```diff
@@ -476,6 +476,10 @@ util.shouldRetryRequest = shouldRetryRequest;
  * @param {function} callback - The callback function.
  */
 function getAuthClient(config, callback) {
+  if (config.authClient) { // cached authClient
+    callback(null, config.authClient);
```
stephenplusplus: Please do:
```js
setImmediate(function() {
  callback(null, config.authClient);
});
```

since this function is meant to be async.

lib/common/util.js
```diff
@@ -476,6 +476,10 @@ util.shouldRetryRequest = shouldRetryRequest;
  * @param {function} callback - The callback function.
  */
 function getAuthClient(config, callback) {
+  if (config.authClient) { // cached authClient
```
stephenplusplus: Comment isn't necessary :)

stephenplusplus:
Thanks for catching this! A mention in the JSdocs would be great.

ofrobots:
Thanks. I have addressed the comments.

stephenplusplus:
Great, thanks!

PR #663, opened 2015-06-17
Update ruby SVG

blowmage:
Update the ruby SVG image. The current SVG is for rails, not ruby.


stephenplusplus:
Thanks! Which one #662 or this one?

blowmage:
Both, I suppose. :)

stephenplusplus:
Oh sorry, I realize now they are both the same file on different branches! gh-pages is built from our `docs` directory, so after merging that one, it should get copied over into gh-pages automatically. Thanks again!

blowmage:
Yeah, I wasn't sure if it was replaced or not. I figured it would be easier to give you both and let you sort it out. Thanks!

PR #685, opened 2015-06-24
pubsub: default autoCreate:true

### LINKED ISSUE #669: createTopic vs topic can be confusing?
Right now, to "create a topic", people jump immediately to `createTopic` -- which gives them back a topic, but then the second time they try to create a topic, it'll cause problems, right? What people typically want in their code is `topic()` with `{autoCreate: true}` ...

So... a couple options

1. Remove / privatize `createTopic` (why create a topic to not use it...?)
1. Document in `createTopic` that you *probably* want `topic()`
  1. *and* update `topic()`'s options to auto-create by default

Thoughts?

stephenplusplus:
Fixes #669 

~~This privatizes `createTopic` by removing it from the docs and renaming it to `createTopic_`. It is still used internally with `autoCreate: true`, which is now **`true` by default.**~~

The other changes in this PR include doc/test changes to suit the new behavior, and refactoring how the `autoCreate` behavior is implemented (+ tests).
#### To Dos
- [x] publicize `createTopic`
- [x] add line to `createTopic` docs stating `topic()` is much easier, and likely what the user intends
- [x] fix doc issues mentioned by @ryanseys

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,13 +214,11 @@ var pubsub = gcloud.pubsub({
   keyFilename: '/path/to/keyfile.json'
 });
 
-// Create a new topic.
-pubsub.createTopic('my-new-topic', function(err, topic) {});
-
-// Reference an existing topic.
-var topic = pubsub.topic('my-existing-topic');
+// Reference a topic.
+var topic = pubsub.topic('my-topic');
 
 // Publish a message to the topic.
+// The topic will be created if it doesn't exist.
 topic.publish({
   data: 'New message!'
 }, function(err) {});
--- a/lib/pubsub/index.js
+++ b/lib/pubsub/index.js
@@ -160,8 +160,16 @@ PubSub.prototype.getTopics = function(query, callback) {
  *
  * @example
  * pubsub.createTopic('my-new-topic', function(err, topic, apiResponse) {
- *   topic.publish('New message!', function(err) {});
+ *   topic.publish({
+ *     data: 'New message!'
+ *   }, function(err) {});
  * });
+ *
+ * //-
+ * // <strong>Note:</strong> For cases like the one above, it is simpler to use
+ * // {module:pubsub#topic}, which will create the topic for you at the time you
+ * // publish a message.
+ * //-
  */
 PubSub.prototype.createTopic = function(name, callback) {
   callback = callback || util.noop;
@@ -315,14 +323,40 @@ PubSub.prototype.subscription = function(name, options) {
  *
  * @param {string} name - The name of the topic.
  * @param {object=} options - Configuration object.
- * @param {boolean=} options.autoCreate - Automatically create topic if it
+ * @param {boolean} options.autoCreate - Automatically create topic if it
  *     doesn't exist. Note that messages published to a topic with no
- *     subscribers will not be delivered.
+ *     subscribers will not be delivered. Default: true.
  * @return {module:pubsub/topic}
  *
  * @example
- * var topic = pubsub.topic('my-existing-topic');
- * var topic = pubsub.topic('topic-that-maybe-exists', { autoCreate: true });
+ * //-
+ * // By default, it isn't required to specify a topic that already exists. The
+ * // first time you publish a message, the topic will be created for you.
+ * //
+ * // This will only cost one additional API request at the time of publishing.
+ * // If the topic doesn't need to be created, there is no performance penalty.
+ * //-
+ * var topic = pubsub.topic('my-topic');
+ *
+ * topic.publish({
+ *   data: 'New message!'
+ * }, function(err) {});
+ *
+ * //-
+ * // If you prefer an error when trying to publish to a topic that doesn't
+ * // exist, set `autoCreate` to `false`.
+ * //-
+ * var nonExistentTopic = pubsub.topic('my-non-existent-topic', {
+ *   autoCreate: false
+ * });
+ *
+ * nonExistentTopic.publish({
+ *   data: 'New message!'
+ * }, function(err) {
+ *   if (err) {
+ *     // API error from trying to publish a message to a non-existent topic.
+ *   }
+ * });
  */
 PubSub.prototype.topic = function(name, options) {
   if (!name) {
--- a/lib/pubsub/subscription.js
+++ b/lib/pubsub/subscription.js
@@ -76,7 +76,7 @@ var util = require('../common/util.js');
  * //-
  * // From {@linkcode module:pubsub/topic#getSubscriptions}:
  * //-
- * var topic = pubsub.topic('my-existing-topic');
+ * var topic = pubsub.topic('my-topic');
  * topic.getSubscriptions(function(err, subscriptions) {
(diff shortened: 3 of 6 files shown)


stephenplusplus:
@ryanseys sorry to disturb!, but curious of your thoughts on this one, mostly in terms of if privatizing createTopic and defaulting autoCreate to true sounds like a good plan? Thanks!

jgeewax:
Yep. I'll look at it today. Hang tight :) sorry for the delay.

stephenplusplus:
No rush at all! If this doesn't make it in by tomorrow, we can just pack it up in 0.17.0.

ryanseys:
Doesn't sound like a bad idea unless someone wants to create a lot of topics all the time... e.g. what if I'm making a public chatroom app and want to create a topic for each room being created? In this case, every request to use the topic will result in an extraneous request to attempt to use a non-existent topic first and could just not be the most efficient.

Have you considered leaving createTopic public, making autoCreate true by default and updating the createTopic docs to describe that the first time you try to use a topic it will auto create one by default anyway so you don't need to use the method except in certain cases? This would also allow the method to be more smoothly deprecated.

README.md
```diff
@@ -213,11 +213,7 @@ var pubsub = gcloud.pubsub({
...
 });
 
-// Create a new topic.
-pubsub.createTopic('my-new-topic', function(err, topic) {});
-
-// Reference an existing topic.
-var topic = pubsub.topic('my-existing-topic');
+var topic = pubsub.topic('my-topic');
 
 // Publish a message to the topic.
```
ryanseys: Maybe comment here saying Publish a message to the topic, automatically creating the topic if it doesn't exist yet.

README.md
```diff
@@ -213,11 +213,7 @@ var pubsub = gcloud.pubsub({
   keyFilename: '/path/to/keyfile.json'
 });
 
-// Create a new topic.
-pubsub.createTopic('my-new-topic', function(err, topic) {});
-
-// Reference an existing topic.
-var topic = pubsub.topic('my-existing-topic');
+var topic = pubsub.topic('my-topic');
```
ryanseys: Comment here about what this is doing... e.g. referencing a potentially non-existent topic? Auto-create is true by default... etc?

README.md
stephenplusplus: :+1: :+1:

ryanseys:
Code looks good to me. Not against the idea, just definitely should consider all alternatives such as the one I mentioned above.

stephenplusplus:
Thanks, sir! I'm in favor of keeping the method public as well, but only for the reason that users would actually use it. I don't mind harshly deprecating it since we're still pre-stable.

ryanseys:
It's definitely not like we're getting rid of, or not supporting, the ability to create topics (that would be silly) so I don't see the need to hide it behind some private flag.

stephenplusplus:
`createTopic('should we keep createTopic public?')`

If @jgeewax agrees we should keep it public, I'll make the changes to the PR, fix the docs, and we can be on our mergy way.

jgeewax:
Yea I can see the need to have `createTopic` as a method -- however I think the use case for it is *much* narrower (that the majority of people in a typical use case want `.topic()`), so we shouldn't really advertise it (or say in the docstring "You probably want `.topic()`").

Cool?

stephenplusplus:
Yeah. First post updated with todos.

stephenplusplus:
To Dos have been To Done.

README.md
```diff
@@ -214,13 +214,11 @@ var pubsub = gcloud.pubsub({
...
-// Create a new topic.
-pubsub.createTopic('my-new-topic', function(err, topic) {});
-
-// Reference an existing topic.
-var topic = pubsub.topic('my-existing-topic');
+// Reference a topic.
+var topic = pubsub.topic('my-topic');
 
 // Publish a message to the topic.
+// The topic will be created if it doesn't exist.
```
stephenplusplus: Readme changes.

lib/pubsub/index.js
```diff
@@ -160,8 +160,16 @@ PubSub.prototype.getTopics = function(query, callback) {
...
+ *   topic.publish({
+ *     data: 'New message!'
+ *   }, function(err) {});
  * });
+ *
+ * //-
+ * // <strong>Note:</strong> For cases like the one above, it is simpler to use
+ * // {module:pubsub#topic}, which will create the topic for you at the time you
+ * // publish a message.
+ * //-
```
stephenplusplus: Notes in the `createTopic` doc that `topic` is easier.

lib/pubsub/index.js
```diff
@@ -315,14 +323,40 @@ PubSub.prototype.subscription = function(name, options) {
...
+ *   autoCreate: false
+ * });
+ *
+ * nonExistentTopic.publish({
+ *   data: 'New message!'
+ * }, function(err) {
+ *   if (err) {
+ *     // API error from trying to publish a message to a non-existent topic.
+ *   }
+ * });
```
stephenplusplus: Explanation of `autoCreate`.

lib/pubsub/index.js
ryanseys: :+1:

lib/pubsub/index.js
ryanseys: Question: If I don't create a topic because I don't have any messages to publish yet, and instead I subscribe to that topic assuming it already exists (as this would suggest I do), then will it throw an error in that case saying the topic doesn't exist?

stephenplusplus: ?
```js
nonExistentTopic.subscribe('newMessages', function(err, subscription) {
  // err = {
  //   code: 404,
  //   message: 'Resource not found (resource=blahef).'
  // }
});
```

stephenplusplus: Are you saying I should change "If you prefer an error when trying to interact with a topic that doesn't exist" to "...when trying to publish a message to a topic that doesn't exist..."?

stephenplusplus: Messaging changed:

> If you prefer an error when trying to publish to a topic that doesn't exist, set `autoCreate` to `false`.

lib/pubsub/index.js
```diff
@@ -160,8 +160,16 @@ PubSub.prototype.getTopics = function(query, callback) {
  *
  * @example
  * pubsub.createTopic('my-new-topic', function(err, topic, apiResponse) {
- *   topic.publish('New message!', function(err) {});
+ *   topic.publish({
```
jgeewax: How come we're not using the short-hand here?

stephenplusplus: I don't think we have that yet: https://github.com/GoogleCloudPlatform/gcloud-node/issues/629

jgeewax: OK Cool

jgeewax:
Didn't spend too much time looking through but this looks pretty good to me :+1:

stephenplusplus:
If anyone has doc improvements, we can iterate on them as we always do :)

PR #687, opened 2015-06-24
readme: condense the fluff

### LINKED ISSUE #672: Move examples and docs out of the README, point to docs instead
We're seeing people using the README contents on NPM as a source of real documentation, and missing out on the bigger picture because they're not seeing all of the other pieces.

Any objections to removing the README example content (specific snippets for each service) and pointing to the gh-pages docs instead?

stephenplusplus:
Fixes #672 

Preview: https://github.com/stephenplusplus/gcloud-node/tree/spp--docs-readme-condensation

This removes the introductory text we had for the services and sentence-form links. To help the reader understand there is more to learn, there are now two bullet points with links to our docs and the official docs for the service. Then, above the service's example, we try to drive home that this code from the readme should not be confused with official docs by giving them a header that reads **Preview**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,11 +77,13 @@ var gcloud = require('gcloud')({
 
 You can also set auth on a per-API-instance basis. The examples below show you how.
 
+
 ## Google BigQuery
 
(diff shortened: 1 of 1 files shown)


dhermes:
Love it!

stephenplusplus:
@jgeewax wdyt?

jgeewax:
LGTM

PR #688, opened 2015-06-24
docs: use hambuger for navigation

### LINKED ISSUE #676: Left navigation moves to the bottom when the page gets narrower?
Seems like this is a bit counter-intuitive. Could we maybe have it be something that appears when you click on the 3-horizontal-lined-link at the top right?

People in half-screen mode got super confused when the navigation disappeared, and were surprised to find it at the absolute bottom of the page... :(

stephenplusplus:
Fixes #676 

Demo: http://stephenplusplus.github.io/gcloud-node-gh-pages-again/#/docs/v0.15.0

The navigation will never be placed at the bottom of the page. Anytime there is a hamburger icon, it will expand the full set of links.

### DIFF
--- a/docs/site/components/language-switcher/language-switcher-directive.js
+++ b/docs/site/components/language-switcher/language-switcher-directive.js
@@ -1,13 +1,14 @@
 angular
   .module('gcloud.language-switcher', [])
-  .directive('languageSwitcher', function() {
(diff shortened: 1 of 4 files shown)


jgeewax:
LGTM

PR #692, opened 2015-06-25
streamrouter: implement across library

### LINKED ISSUE #640: Make pagination simpler?
(Going to use Datastore as the example)

In the docs from [`Datastore.runQuery`](http://googlecloudplatform.github.io/gcloud-node/#/docs/v0.15.0/datastore/dataset?method=runQuery)

```javascript
var query = dataset.createQuery('Lion');

// Retrieve 5 companies.
dataset.runQuery(query, function(err, entities, endCursor, apiResponse) {
  // Use `endCursor` as the starting cursor for your next query.
  var nextQuery = query.start(endCursor);
  var callback = function(err, entities, endCursor, apiResponse) {};
  dataset.runQuery(nextQuery, callback);
});
```

Seems like in datastore there's no way to set the page size, however even so, it seems that paging is *super* manual here, which bums me out.

What if we could make it so that...

```javascript
var query = dataset.createQuery('Lion');
dataset.runQuery(query, function(err, entities, nextPage) {
  // entities is the current page.
  nextPage(); // Call this same callback function, with the next page of entities?
});
```

Or even more explicitly, let the user tell you "just get them all please":

```javascript
var query = dataset.createQuery('Lion').setPagination(false);
dataset.runQuery(query, function(err, entities) {
  // entities is the current page.
});
```

?

Let me know if I'm a crazy person here... @ryanseys @stephenplusplus We're having a similar discussion about pagination being explicit versus implicit on https://github.com/GoogleCloudPlatform/gcloud-python/issues/895

### LINKED ISSUE #673: bucket.getFiles should document (at the top) the `on()` way of iterating.
Right now, in http://googlecloudplatform.github.io/gcloud-node/#/docs/v0.15.0/storage/bucket?method=getFiles

```javascript
bucket.getFiles(function(err, files, nextQuery, apiResponse) {
  if (nextQuery) {
    // nextQuery will be non-null if there are more results.
    bucket.getFiles(nextQuery, function(err, files, nextQ, apiResponse) {});
  }
});
```

which is ... absurdly confusing -- I just want to iterate through all the files. 

Can we change the examples to document the way to iterate through all files, and abort the traversal early if necessary?

```javascript
bucket.getFiles().on('data', function(file) {
  console.log(file);
  if (some condition) {
    this.end();
  }
});

stephenplusplus:
Fixes #640
Fixes #673 

This allows all methods with a `nextQuery` to be used:
1. with a callback,
   
   ``` js
   bucket.getFiles({ autoPaginate: false }, function(err, files, nextQuery, apiResponse) {});
   ```
2. as a readable object stream that can be `this.end()`ed early, or
   
   ``` js
   bucket.getFiles().on('data', function(file) { if (something) this.end() });
   ```
3. with a callback, having all of the nextQuerys run automatically, then getting all of the results at once. (DEFAULT)
   
   ``` js
   bucket.getFiles(function(err, files) {});
   ```

Docs: http://stephenplusplus.github.io/gcloud-node/#/docs/master
## Changes
1. **No more `nextQuery`!**
   
   By default, any method that has a nextQuery will be run until there is no more nextQuery. The user's callback will be executed _once_ after all of the results are fetched. To turn it off, the user can specify `autoPaginate: false` in the query/options/config object (e.g. `bucket.getFiles({ autoPaginate: false })` or for running a Dataset query: `var query = dataset.createQuery('Person').autoPaginate(false)`.
   
   When `autoPaginate` is true, the callback doesn't receive a `nextQuery` or `apiResp`. No `nextQuery`, because one won't exist (we run through them all), and no `apiResp`, because there is the potential for there to be many api requests.
2. A stream is now stopped after a user-defined `query.maxResults` (most of our API supports this), `query.limitVal` (datastore), or `query.pageSize` (pubsub) amount of results is reached.
   
   Any time one of these properties is on the query, and a callback is provided, `autoPaginate` is shut off. When a user says "I only want this many", we assume they might want the ability to cursor around after they get their results.
3. `dataset.runQuery` now uses a `nextQuery` object instead of an `endCursor`.
## Methods that are affected
- BigQuery
  - bigquery.getDatasets ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/bigquery?method=getDatasets))
  - bigquery.getJobs ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/bigquery?method=getJobs))
  - bigquery.query ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/bigquery?method=query))
  - table.getRows ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/bigquery/table?method=getRows))
- Datastore
  - dataset.runQuery ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/datastore/dataset?method=runQuery))
- PubSub
  - pubsub.getSubscriptions ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/pubsub?method=getSubscriptions))
  - pubsub.getTopics ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/pubsub?method=getTopics))
- Search
  - search.getIndexes ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/search?method=getIndexes))
  - index.getDocuments ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/search/index?method=getDocuments))
  - index.search ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/search/index?method=search))
- Storage
  - storage.getBuckets ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/storage?method=getBuckets))
  - bucket.getFiles ([docs](http://stephenplusplus.github.io/gcloud-node/#/docs/master/storage/bucket?method=getFiles))

### DIFF
--- a/lib/bigquery/index.js
+++ b/lib/bigquery/index.js
@@ -163,13 +163,23 @@ BigQuery.prototype.dataset = function(id) {
  * @param {object=} query - Configuration object.
  * @param {boolean} query.all - List all datasets, including hidden ones.
  * @param {boolean} query.autoPaginate - Have pagination handled automatically.
- *     Default: false.
+ *     Default: true.
  * @param {number} query.maxResults - Maximum number of results to return.
  * @param {string} query.pageToken - Token returned from a previous call, to
  *     request the next page of results.
  * @param {function} callback - The callback function.
  *
  * @example
+ * bigquery.getDatasets(function(err, datasets) {
+ *   if (!err) {
+ *     // datasets is an array of Dataset objects.
+ *   }
+ * });
+ *
+ * //-
+ * // To control how many API requests are made and page through the results
+ * // manually, set `autoPaginate` to `false`.
+ * //-
  * var callback = function(err, datasets, nextQuery, apiResponse) {
  *   if (nextQuery) {
  *     // More results exist.
@@ -177,17 +187,9 @@ BigQuery.prototype.dataset = function(id) {
  *   }
  * };
  *
- * bigquery.getDatasets(callback);
- *
- * //-
- * // To have pagination handled for you, set `autoPaginate`. Note the changed
- * // callback parameters.
- * //-
  * bigquery.getDatasets({
- *   autoPaginate: true
- * }, function(err, datasets) {
- *   // Called after all datasets have been retrieved.
- * });
+ *   autoPaginate: false
+ * }, callback);
  *
  * //-
  * // Get the datasets from your project as a readable object stream.
@@ -251,7 +253,7 @@ BigQuery.prototype.getDatasets = function(query, callback) {
  * @param {boolean=} options.allUsers - Display jobs owned by all users in the
  *     project.
  * @param {boolean} options.autoPaginate - Have pagination handled
- *     automatically. Default: false.
+ *     automatically. Default: true.
  * @param {number=} options.maxResults - Maximum number of results to return.
  * @param {string=} options.pageToken - Token returned from a previous call, to
  *     request the next page of results.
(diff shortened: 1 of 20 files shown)


stephenplusplus:
Don't merge lifted. Sorry again for such a large PR. The individual commits should make it easier, and I'm happy to answer any questions.

stephenplusplus:
I added one more feature to stream-router: `autoPaginate: true`. Updated the first post with examples.

stephenplusplus:
@jgeewax mind if I merge this one? I feel confident this will be a relief to our users (both the streaming mode with early exiting and the { autoPaginate: true }.

If you're on board with the idea of { autoPaginate: true } (explained in the opening post to the PR), I think we can at least put this out there. If we hear any issues, we can always take it back to the drawing board.

jgeewax:
Is autoPaginate defaulted to true?

I suspect people will want that more frequently........

stephenplusplus:
I can default to true. I think people would want that more as well, but I wasn't sure if we were worried about the unknown additional API calls / associated costs that would be blanketed. I was under the impression that's what made us do this whole nextQuery thing in the first place.

jgeewax:
I think we need to provide control, but should default to convenient.

As people get big and notice lots of API calls, they might think about optimizing, and we should make that possible. But let's make the common case fast (and simple)

stephenplusplus:
I'm so happy.

stephenplusplus:
autoPaginate: true defaulted and `don't merge` lifted.

stephenplusplus:
PTAL at the initial post and let me know if everything sounds sane.

jgeewax:
Looked at your examples here and docs for `bucket.getFiles` which LGTM. :+1:

stephenplusplus:
Thanks!

PR #698, opened 2015-06-30
storage: add deleteFiles()

### LINKED ISSUE #680: Cloud Storage Delete all files with prefix
It would be nice if there was a bulk delete for files that satisfy a query in a bucket. Something like:

    myBucket.deleteFiles({ prefix: 'myDirectory/' }, callback);

stephenplusplus:
Fixes #680 

Docs: http://stephenplusplus.github.io/gcloud-node-gh-pages-again/#/docs/master/storage/bucket?method=deleteFiles
#### Todos
- [x] Explain in the docs that this could be costly and is not atomic.
- [x] Document callback arguments.

This adds a new method on Bucket: `deleteFiles`.

Old way of deleting all of the files in the bucket. No error handling, no throttling, no paging.

``` js
bucket.getFiles(function(err, files) {
  files.forEach(function(file) {
    file.delete();
  });
});
```

New way with error handling and throttling built in.

``` js
bucket.deleteFiles(function(err) {});
```

Behind the scenes, `deleteFiles` just calls `getFiles`. So, you're able to pass your `getFiles` query object to `deleteFiles`.

``` js
bucket.deleteFiles({ prefix: 'images/' }, function(err) {});
```

By default, if an error occurs, no more files will be deleted. This can be shut off.

``` js
bucket.deleteFiles({ force: true }, function(errors) {
  // errors = null | Error[]
});
```

### DIFF
--- a/docs/site/components/docs/docs.html
+++ b/docs/site/components/docs/docs.html
@@ -83,7 +83,10 @@ var gcloud = require('gcloud')({
         <tbody>
           <tr
             ng-repeat="param in method.params"
-            ng-class="{ 'param-optional': param.optional }">
+            ng-class="{
+              'param-optional': param.optional,
+              'param-nullable': param.nullable
+            }">
             <th scope="row" class="param">
               <span ng-if="param.subparam" class="param-parent">
                 <div>{{param.name.split('.').slice(0, -1).join('.')}}</div>
--- a/docs/site/components/docs/docs.js
+++ b/docs/site/components/docs/docs.js
@@ -110,9 +110,19 @@ angular
           'bucket',
           'file',
           'job',
-          'table'
+          'table',
+          'index',
+          'document',
+          'field',
+          'topic',
+          'subscription'
         ];
-        type = type.replace('=', '');
+
+        var arrayRegex = /Array\.<([^>]+)>/g;
+        type = type.replace('=', '')
+          .replace('?', '')
+          .replace(arrayRegex, '$1[]');
+
         if (CUSTOM_TYPES.indexOf(type.toLowerCase()) > -1) {
           if (types[index - 1]) {
             type = types[index - 1] + '/' + type;
--- a/docs/site/css/main.css
+++ b/docs/site/css/main.css
@@ -856,15 +856,25 @@ ul {
   padding-top: 0.7em;
 }
 
-.param-optional .param-types {
+.param-optional .param-types,
+.param-nullable .param-types {
   font-style: italic;
+}
+
+.param-optional .param-types:after,
+.param-nullable .param-types:after {
+  display: block;
+  color: #aaa;
+  font-style: italic;
+  font-size: 85%;
 }
 
 .param-optional .param-types:after {
   content: " (optional)";
-  color: #aaa;
-  font-style: italic;
-  font-size: 85%;
+}
+
+.param-nullable .param-types:after {
+  content: " (may be null)";
 }
 
 .method-heading {


jgeewax:
Sounds good to me. Hopefully later we can have this be atomic. For now, we should make it super clear in the documentation that it isn't atomic, and you might incur cost for a huge number of operations...... (ie, a big warning sign...)

lib/storage/bucket.js
```diff
@@ -311,6 +311,117 @@ Bucket.prototype.delete = function(callback) {
...
+ * `{ force: true }` to suppress the errors.
+ *
+ * The `query` object passed as the first argument will also be passed to
+ * {module:storage/bucket#getFiles}.
+ *
+ * @param {object} query - Query object. See {module:storage/bucket#getFiles}
+ *     for all of the supported properties.
+ * @param {boolean} query.force - Supress errors until all files have been
+ *     processed.
+ * @param {function} callback - The callback function.
```
jgeewax: What parameters come w the callback? Can you add that here?

stephenplusplus: Good call, I need to start doing that.

stephenplusplus:
Ready for another check. Added the docs and also [caught some probs](https://github.com/GoogleCloudPlatform/gcloud-node/commit/4e687079c135108a8d2c1f2f8762c9267bbaf77e) with how array params were being displayed. Docs here: http://stephenplusplus.github.io/gcloud-node-gh-pages-again/#/docs/master/storage/bucket?method=deleteFiles

jgeewax:
OK LGTM but maybe @ryanseys should take a look ?

lib/storage/bucket.js
```diff
@@ -311,6 +311,124 @@ Bucket.prototype.delete = function(callback) {
...
+  });
+
+  function deleteFiles(query, callback) {
+    self.getFiles(query, function(err, files, nextQuery) {
+      if (err) {
+        callback(err);
+        return;
+      }
+
+      // Iterate through each file and make it public or private.
```
ryanseys: Update this comment.

test/storage/bucket.js
```diff
@@ -340,6 +340,124 @@ describe('Bucket', function() {
...
+
+    it('should execute callback with error from getting files', function(done) {
+      var error = new Error('Error.');
+
+      bucket.getFiles = function(query, callback) {
+        callback(error);
+      };
+
+      bucket.deleteFiles({}, function(err) {
+        assert.equal(err, error);
```
ryanseys: It probably doesn't matter that much but should probably use strictEqual here

ryanseys:
LGTM other than the small comment to be updated. equal vs. strictEqual might be a repo-wide change to be made so if you don't change it in this PR that's nbd. Also you're comparing objects so `==` should suffice because type inference doesn't do much on an object, but I digress. LGTM.

stephenplusplus:
Thanks for reviewing! I think we have a mixture, but I've been
doing deepEqual lately. Not sure if any of them are really better than the
other, but I'll update it to one of the others before this ships.

ryanseys:
`equal` compares with `==` so type inference and all that jazz.

`deepEqual` compares each property of one object with the properties of another object without the objects necessarily being the same object -- good for assert.deepEqual(config, { actual: 'config', goes: 'here })

`strictEqual` compares with `===` so either the objects reference the same memory or they are equal primitives.

ryanseys:
I'd say `strictEqual` should be used in place of `equal` everywhere and leave `deepEqual` everywhere else.

stephenplusplus:
Changed to `strictEqual` (will use in the future) and updated the stale comment.

PR #705, opened 2015-07-06
core: add retry logic to read streams

### LINKED ISSUE #701: storage- when return status 503 (backend error) auto retry.
add option in storage (by parameter) to auto retry when google return status 503.

stephenplusplus:
Fixes #701 
#### To Dos
- [x] [Implement exponential backoff](https://github.com/stephenplusplus/retry-request/issues/1) in [retry-request](https://github.com/stephenplusplus/retry-request)
- [x] Tests

file#createReadStream was skipping over the universal response handler logic we have for all API requests our library makes. This is most beneficial for retrying failed requests using the recommended exponential backoff technique. This PR brings retry logic to createReadStream, at the cost of more code complexity, because holy moly, is this tricky with streams.

I wrote up a module that helps with this, [retry-request](https://github.com/stephenplusplus/retry-request), which also implements exponential backoff the way we had it here.

Here's a simplified example of how this works:
retry-request uses a module called `stream-cache` that stores what it's given, then replays it when a new stream is connected to it. Additionally, retry-request uses an intermediary stream while we figure out if the request is going to flat out fail. It kind of looks like this (simplified, pseudo):

``` js
file.createReadStream = function() {
  var intermediaryStream = new Stream();
  tryRequest();
  return intermediaryStream;

  function tryRequest() {
    var cacheStream = new CacheStream();
    var requestStream = request('http://storage.googleapis.com/bucket/my-file.zip');

    requestStream.pipe(cacheStream);

    requestStream.on('response', function (resp) {
      if (resp.statusCode < 200 || resp.statusCode > 299 && shouldRetry) {
        // Try the request again. All new streams are created, except the
        // intermediary stream (the one the user is holding onto), which still
        // hasn't had any data sent to it.
        tryRequest();
      } else {
        // Okay, no errors here, replay the data we've been caching to the
        // user's stream.
        cacheStream.pipe(intermediaryStream);
      }
    });
  }
}
```

Related: [AWS streams](http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-making-requests.html) - See last section "Limitations of Streaming"

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -29,6 +29,9 @@ var request = require('request').defaults({
     maxSockets: Infinity
   }
 });
+var retryRequest = require('retry-request');
+var streamForward = require('stream-forward');
+var through = require('through2');
 var uuid = require('node-uuid');
 
 /** @const {object} gcloud-node's package.json file. */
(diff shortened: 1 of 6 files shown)


ryanseys:
Streams in node:

rffwukr

stephenplusplus:
Both excellent choices :)

stephenplusplus:
`don't merge` label lifted. This should be safe for a review.

stephenplusplus:
@callmehiphop can you take a look over this and let me know if anything looks :+1: or :+1:?

lib/common/util.js
```diff
@@ -600,30 +623,47 @@ function makeAuthorizedRequestFactory(config) {
...
-        callback.onAuthorized(null, reqOpts);
-      } else {
-        util.makeRequest(reqOpts, config, callback);
-      }
+    var streamMode = !callback;
+    var stream;
+    var reqConfig = extend({}, config);
 
-      return;
+    callback = callback || util.noop;
```
callmehiphop: This is probably trivial, but why do set `callback` to a noop if we use streams in its absence?

stephenplusplus: No, that is a great catch! Somewhere along the way, I tried to make the variable naming easier to understand, and I think I failed. I updated the PR. PTAL!

callmehiphop:
LGTM! :+1:

stephenplusplus:
Thanks!

PR #706, opened 2015-07-06
Add Google Cloud DNS

### LINKED ISSUE #486: Support Google DNS API
API: https://cloud.google.com/dns/api/v1beta1/
RFC rrdata format: https://tools.ietf.org/html/rfc1034#section-3.6.1

stephenplusplus:
Fixes #486
#### To Dos
- [x] Add support for SPF records in https://github.com/elgs/dns-zonefile/pull/11
- [x] Tests
  - [x] Unit
    - [x] util.exec
    - [x] dns/index
    - [x] dns/zone
    - [x] dns/record
    - [x] dns/change
  - [x] System
- [x] Add `GCLOUD_TESTS_DNS_DOMAIN` env var to travis (value: `gcloud-node.com.`)
- [x] Docs

Docs: https://stephenplusplus.github.io/gcloud-node-gh-pages-again/#/docs/master/dns

### DIFF
--- a/lib/dns/index.js
+++ b/lib/dns/index.js
@@ -1,0 +1,250 @@
+/*!
+ * Copyright 2014 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*!
+ * @module dns
+ */
+
+'use strict';
+
+var extend = require('extend');
+
+/**
+ * @type {module:common/streamrouter}
+ * @private
+ */
+var streamRouter = require('../common/stream-router.js');
+
+/**
+ * @type {module:common/util}
+ * @private
+ */
+var util = require('../common/util.js');
+
+/**
+ * @type {module:dns/zone}
+ * @private
+ */
+var Zone = require('./zone.js');
+
+/**
+ * @const {string} Base URL for DNS API.
+ * @private
+ */
+var DNS_BASE_URL = 'https://www.googleapis.com/dns/v1/projects/';
+
+/**
+ * @const {array} Required scopes for the DNS API.
+ * @private
+ */
+var SCOPES = [
+  'https://www.googleapis.com/auth/ndev.clouddns.readwrite',
+  'https://www.googleapis.com/auth/cloud-platform'
+];
+
+/**
+ * [Google Cloud DNS](https://cloud.google.com/dns) is a reliable, resilient,
+ * low-latency DNS serving from Google’s worldwide network of Anycast DNS
+ * servers.
+ *
+ * @constructor
+ * @alias module:dns
+ *
+ * @param {object} options - [Configuration object](#/docs/?method=gcloud).
+ *
+ * @example
+ * var gcloud = require('gcloud')({
+ *   keyFilename: '/path/to/keyfile.json',
+ *   projectId: 'grape-spaceship-123'
+ * });
+ *
+ * var dns = gcloud.dns();
+ */
+function DNS(options) {
+  if (!(this instanceof DNS)) {
+    return new DNS(options);
+  }
+
+  options = options || {};
+
+  if (!options.projectId) {
+    throw util.missingProjectIdError;
+  }
+
+  this.makeAuthorizedRequest_ = util.makeAuthorizedRequestFactory({
+    credentials: options.credentials,
+    keyFile: options.keyFilename,
+    scopes: SCOPES,
+    email: options.email
+  });
+
+  this.projectId_ = options.projectId;
+}
+
+/**
+ * Create a managed zone.
+ *
+ * @throws {error} If a zone name is not provided.
+ * @throws {error} If a zone dnsName is not provided.
+ *
+ * @param {string} name - Unique name for the zone. E.g. "my-zone"
+ * @param {object} config - Configuration object.
+ * @param {string} config.dnsName - DNS name for the zone. E.g. "example.com."
+ * @param {string=} config.description - Description text for the zone.
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An API error.
+ * @param {?module:dns/zone} callback.zone - A new {module:dns/zone} object.
+ * @param {object} callback.apiResponse - Raw API response.
+ *
+ * @example
+ * dns.createZone('my-awesome-zone', {
+ *   dnsName: 'example.com.', // note the period at the end of the domain.
+ *   description: 'This zone is awesome!'
+ * }, function(err, zone, apiResponse) {
+ *   if (!err) {
+ *     // The zone was created successfully.
+ *   }
+ * });
+ */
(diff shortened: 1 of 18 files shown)


system-test/dns.js
```diff
@@ -0,0 +1,203 @@
...
+  describe('Records', function() {
+    it('should return 0 or more records', function(done) {
+      ZONE.getRecords(function(err, records) {
+        assert.ifError(err);
+        assert(records.length >= 0);
+        done();
+      });
+    });
+
+    it('should replace records', function(done) {
```
stephenplusplus: See the code in this test for where I went the most off-the-rails :train: from the original specs:

- `zone.record` taking a record type
- `zone.getRecords` taking a type
- `zone.replaceRecords` to delete records of a type and replace them with the ones provided

I added these to try to more easily solve a simple problem: "I got a new server and want to point my site at it" - Let me know if this is wacky.

lib/dns/zone.js
```diff
@@ -0,0 +1,458 @@
...
+    var change = self.change(resp.id);
+    change.metadata = resp;
+
+    callback(null, change, resp);
+  });
+};
+
+/**
+ * Delete the zone. Note that only empty zones can be deleted. You can set
+ * options.force to empty before deleting.
```
jgeewax: Can we include a note that this is not atomic?

lib/dns/zone.js
```diff
@@ -0,0 +1,458 @@
...
+ */
+Zone.prototype.import = function(filename, options, callback) {
+  callback(filename, options);
+};
+
+Zone.prototype.record = function(recordType, metadata) {
+  return new Record(this, recordType, metadata);
+};
+
+Zone.prototype.replaceRecords = function(recordType, newRecords, callback) {
```
jgeewax: Docstring on this one is going to be important... 

I'm all for this existing (ie, "replace whatever I have of type A with this one"), but it's not atomic -- in that if someone adds a new A record between our getRecords() call and the createChange() call... we'll have some wonky results...

stephenplusplus:
Assigned to @callmehiphop for a review, please. Probably easiest to start with [the docs](https://stephenplusplus.github.io/gcloud-node-gh-pages-again/#/docs/master/dns), and make sure the API looks sane.

lib/dns/index.js
```diff
@@ -0,0 +1,250 @@
...
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An API error.
+ * @param {?module:dns/zone[]} callback.zones - An array of {module:dns/zone}
+ *     objects.
+ * @param {object} callback.apiResponse - Raw API response.
+ *
+ * @example
+ * dns.getZones(function(err, zones, apiResponse) {});
+ */
+DNS.prototype.getZones = function(query, callback) {
```
callmehiphop: How come we call this `getZones` vs. something like `listZones`? The only reason I ask is because I wanted to see what options I had available in query and my initial reaction was to look at ManagedZones > get.

stephenplusplus:
@jgeewax Since we're trapped on what we can do with system testing (re: the error `The 'gcloud-node.com.' managed zone is not available to be created.`), here are the options we have that I can think of:

  1. Look for a `GCLOUD_TESTS_DNS_DOMAIN` env var and use it for system testing. This is how it is in this PR currently, and how I've been testing with my `stephenplusplus.com.` domain (which has yet to run into the issues `gcloud-node.com.` has)
  1. If the env var is set, try to create a zone. If it fails with the error above, try to get a list of zones that already exist, and use one of them to complete the tests that don't involve creating/deleting zones with.
  1.  If we can't get a zone, or any of the above failed, skip the DNS tests.

It's not exactly pretty to make our tests run like that, but it's possible.

lib/dns/index.js
stephenplusplus: It's just to offer a nicer API than what the upstream API gives us. We have some patterns that you'll find like that, that are really just about simplifying things / giving users more expected naming conventions.

lib/dns/zone.js
```diff
@@ -0,0 +1,817 @@
...
+
+  if (options.force) {
+    this.empty(this.delete.bind(this, callback));
+    return;
+  }
+
+  var path = '/managedZones/' + this.name;
+  this.makeReq_('DELETE', path, null, null, function(err, resp) {
+    callback(err, resp);
+  });
```
callmehiphop: any reason we don't want to simplify this into something like..

```javascript
this.makeReq_('DELETE', path, null, null, callback);
```

jgeewax:
I can't find anywhere in any of my projects that have a zone for gcloud-node.com ... I opened a ticket asking WTF is going on... because that's super weird. If you know a project that has the zone created... lmk... I'd really like to use gcloud-node as the demo domain...

lib/dns/zone.js
```diff
@@ -0,0 +1,817 @@
...
+Zone.prototype.getChanges = function(query, callback) {
+  var self = this;
+
+  if (util.is(query, 'function')) {
+    callback = query;
+    query = {};
+  }
+
+  if (query.sort) {
+    query.sortBy = 'changeSequence';
```
callmehiphop: Do we need this? It looks like this is the default (and only) `sortBy` value upstream,

lib/dns/zone.js
```diff
@@ -0,0 +1,817 @@
...
+ * zone.replaceRecords('ns', newNsRecords, function(err, change, apiResponse) {
+ *   if (!err) {
+ *     // The change was created successfully.
+ *   }
+ * });
+ */
+Zone.prototype.replaceRecords = function(recordType, newRecords, callback) {
+  var self = this;
+
+  this.getRecords(recordType, function(err, recordsToDelete) {
```
callmehiphop: Not sure how likely this is to occur, but is it possible that not all of the records would be returned via `getRecords` and there would be additional records within `nextQuery`?

lib/dns/zone.js
stephenplusplus: Good catch! https://github.com/GoogleCloudPlatform/gcloud-node/commit/b2f283991e51814f5c245d614850575088c50a29

lib/dns/zone.js
stephenplusplus: When we use `getRecords`, it will actually run it until no more results exist. That's new as of a couple merges ago when we introduced streamRouter across the library. The bottom of this file uses "streamRouter.extend()" to list the methods that should be wrapped.

stephenplusplus: (Because we default `autoPaginate` to true)

callmehiphop: nice!

callmehiphop:
LGTM

stephenplusplus:
> If you know a project that has the zone created... lmk... I'd really like to use gcloud-node as the demo domain...

Don't know of one, and me too re: gcloud-node.com. I assigned to you since this is our last step before merging. If you prefer we merge with the system tests shut off for now and track the domain stuff in a new issue, that's cool too.

lib/dns/zone.js
stephenplusplus: Sorry I missed this. The reason is just to limit the arguments that get back to the user's callback. `this.makeReq_` is invoked with `err, resp, parsedResponseBody`, but it's actually not being used anywhere in the library. This question is spawning a new PR (coming soon), so thank you :)

stephenplusplus:
@jgeewax As this PR sits now, our system tests work like this:

- Is an env var called "GCLOUD_TESTS_DNS_DOMAIN" set?
  - Yes: run the DNS system tests using that domain
  - No: don't run the DNS system tests

Testing locally with GCLOUD_TESTS_DNS_DOMAIN set to `gcloud-node.com.` works for me now. I set the env var in travis, too. Merging this should run the tests and "just work". Want to press the green button? :)

jgeewax:
SGTM - but I'll let you do the merge :P

stephenplusplus:
Yolo!

PR #711, opened 2015-07-09
pubsub: migrate to v1

### LINKED ISSUE #699: Migrate to pubsub v1
> The v1beta2 version of the Cloud Pub/Sub API was part of our Beta launch, and will be deprecated as of July 1, 2015. The v1beta2 API will continue to work until January 11, 2016, and will shut down after that date. Please migrate to the v1 API.

v1: https://cloud.google.com/pubsub/reference/rest/index#pubsub.googleapis.com_1

stephenplusplus:
Fixes #699 

It looks like the only change was enabling multiple ack IDs to be passed to subscriptions.modifyAckDeadline: https://cloud.google.com/pubsub/reference/rest/v1/projects.subscriptions/modifyAckDeadline

### DIFF
--- a/lib/pubsub/subscription.js
+++ b/lib/pubsub/subscription.js
@@ -387,7 +387,7 @@ Subscription.prototype.pull = function(options, callback) {
  * interrupted.
  *
  * @param {object} options - The configuration object.
- * @param {number} options.ackId - The ack id to change.
+ * @param {number|number[]} options.ackIds - The ack id(s) to change.
  * @param {number} options.seconds - Number of seconds after call is made to
  *     set the deadline of the ack.
  * @param {Function=} callback - The callback function.
@@ -401,7 +401,7 @@ Subscription.prototype.pull = function(options, callback) {
  */
 Subscription.prototype.setAckDeadline = function(options, callback) {
   var body = {
-    ackId: options.ackId,
+    ackIds: util.arrayize(options.ackIds),
     ackDeadlineSeconds: options.seconds
   };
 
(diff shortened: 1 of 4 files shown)


stephenplusplus:
@tmatsuo did I miss anything in the migration from v1beta2 to v1?

// @dhermes

dhermes:
Thanks for the cc!

tmatsuo:
Yes, that's the only change. Good eye :)

stephenplusplus:
Woo! Thanks for the review!

lib/pubsub/subscription.js
```diff
@@ -401,7 +401,7 @@ Subscription.prototype.pull = function(options, callback) {
  */
 Subscription.prototype.setAckDeadline = function(options, callback) {
   var body = {
-    ackId: options.ackId,
+    ackIds: util.arrayize(options.ackIds),
```
tmatsuo: Nit:
I think you break the compatibility here, but I don't think it's a big deal.

stephenplusplus: Yeah, we just have to leave a note on the next release.

sindresorhus: @stephenplusplus You could have just made it `ackIds: util.arrayize(options.ackIds || options.ackId),` to not make it a breaking change.

stephenplusplus: Yeah, but pre-1.0, no need to worry about legacy support / documenting two cases. Thanks for reviewing, though :)

PR #713, opened 2015-07-09
Fixed incorrect GitHub Markdown syntax.

jacobsa:
This caused the code samples to look totally broken [in the GitHub viewer](https://github.com/GoogleCloudPlatform/gcloud-node/blob/2ad51ae/docs/authorization.md).


jacobsa:
Oops, I guess `hljs-class` is not intended to be part of the code. But it screws up GitHub's viewer. What's the intent of including it?

stephenplusplus:
Wow, amazing that we missed this. The `hljs-class` is picked up by our docs site: https://googlecloudplatform.github.io/gcloud-node/#/authorization. I found a solution that will fix the issues. More details in a PR coming soon!

PR #714, opened 2015-07-10
docs: correctly use classes in markdown & fix site bug

### LINKED ISSUE #712: Example for GCE in authorization documentation seems wrong
In the section called "On Google Compute Engine", [this](https://googlecloudplatform.github.io/gcloud-node/#/authorization) page says that authorization is handled for me with no configuration. But the example directly below shows using an explicit JSON key file:

```javascript
var config = {
};

var gcloud = require('gcloud')(config);
```

In fact, this is identical to the example used in the "Elsewhere" section.

stephenplusplus:
Fixes #712 
Fixes #713

@jacobsa discovered two bugs:
1. Our `docs.md` code formatting was broken: https://github.com/GoogleCloudPlatform/gcloud-node/blob/2ad51ae/docs/authorization.md
2. Our [authorization](https://googlecloudplatform.github.io/gcloud-node/#/authorization) page had been showing the same code block for both examples.

Thanks, @jacobsa!

### DIFF
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -4,7 +4,7 @@ With `gcloud-node` it's incredibly easy to get authorized and start using Google
 
 If you are running this client on Google Compute Engine, we handle authorization for you with no configuration. You just need to make sure that when you [set up the GCE instance][gce-how-to], you add the correct scopes for the APIs you want to access.
 
-```js hljs-class
+```hljs-class
 var config = {
   projectId: 'grape-spaceship-123'
 };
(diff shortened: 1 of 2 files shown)


docs/authorization.md
```diff
@@ -4,7 +4,7 @@ With `gcloud-node` it's incredibly easy to get authorized and start using Google
 
 If you are running this client on Google Compute Engine, we handle authorization for you with no configuration. You just need to make sure that when you [set up the GCE instance][gce-how-to], you add the correct scopes for the APIs you want to access.
 
-```js hljs-class
+```hljs-class
```
stephenplusplus: Bug 1: We don't need the `js`. `hljs-class` is enough to be picked up by our docs site and appropriately syntax highlighted. The only loss is no syntax highlighting on the GH viewer: https://github.com/stephenplusplus/gcloud-node/blob/b21744fc3ef07eb8a260af7ecd4570c611420d9d/docs/authorization.md but I think we want our docs site to look the best.

docs/site/components/subpage/subpage-directive.js
```diff
@@ -72,6 +72,7 @@ angular
       priority: 1500,
       restrict: 'C',
       transclude: true,
+      scope: {},
```
stephenplusplus: Bug 2: Two `hljsClass` directive instances were sharing the same scope, so assigning the contents of the last would would overwrite all previous instances. This creates an isolate scope where they can each happily retain their contents.

jgeewax:
Nice catch - subtle fixes... LGTM

PR #720, opened 2015-07-13
Updated gcloud sdk installation steps

callmehiphop:
The indexing step threw an error for me and I had to add `index.yaml` to the file path in order for it to run properly.


stephenplusplus:
Good catch. Thanks!

PR #721, opened 2015-07-13
Add Google Compute Engine support

### LINKED ISSUE #341: Investigate Google Compute Engine

### LINKED ISSUE #718: Adding Google Compute Engine support
This is just for discussion purpose. I am not reporting a bug nor an issue. 
I was wondering whether we could add support to Google Compute Engine. I think we can all benefit of a more node-oriented library for managing instances, disks, networks and so on. Personally I needed Compute Engine features in a project and I suspect there might be other people out there. However, I don't know if this fits somehow into your timeline :)  

I started adding some of Compute Engine features to gcloud-node in a branch of my fork (https://github.com/mziccard/gcloud-node/tree/compute-support). So far what I did is:
-  list/create/delete/start/stop/reset instances
-  list/create/delete disks
-  list/create/update firewall rules
-  attach/detach disks to/from instances

I haven't made any pull request yet as the code lacks tests and I wanted your opinion on adding these features first. Any thoughts?

mziccard:
Fixes #341 
Fixes #718 

This pull request adds a module to the library to support Google Compute Engine, as described in #718. 
The module is so far capable of:
-  list/create/delete/start/stop/reset instances
-  list/create/delete disks
-  list/create/update firewall rules
-  attach/detach disks to/from instances

I tried to follow the coding style and added some documentation. The module lacks of tests (both unit and system). I plan to add some unit tests as soon as I have some time to.  

Suggestions are more than welcome. Feel free to ask for changes/fixes/updates or apply them yourselves, I am happy either way:)

### DIFF
--- a/lib/compute/disk.js
+++ b/lib/compute/disk.js
@@ -69,11 +69,23 @@ function Disk(zone, name, metadata) {
  * @param {function} callback - The callback function.
  *
  * @example
- * disk.delete(function(err) {});
+ * disk.delete(function(err, operation) {
+ *   // `operation` is an Operation object and can be used to check the status
+ *   // of disk deletion.
+ * });
  */
 Disk.prototype.delete = function(callback) {
   callback = callback || util.noop;
-  this.makeReq_('DELETE', '', null, true, callback);
+  var zone = this.zone;
+  this.makeReq_('DELETE', '', null, true, function(err, resp) {
+    if (err) {
+      callback(err);
+      return;
+    }
+    var operation = zone.operation(resp.name);
+    operation.metadata = resp;
+    callback(null, operation);
+  });
 };
 
 /**
@@ -106,8 +118,10 @@ Disk.prototype.getMetadata = function(callback) {
  * @param {function} callback - The callback function.
  *
  * @example
- * disk.createSnapshot('my-snapshot', function(err, snapshot, apiResponse) {
+ * disk.createSnapshot('my-snapshot', function(err, snapshot, operation) {
  *   // `snapshot` is a Snapshot object.
+ *   // `operation` is an Operation object and can be used to check the status
+ *   // of snapshot creation.
  * });
  */
 Disk.prototype.createSnapshot = function(options, callback) {
@@ -122,14 +136,16 @@ Disk.prototype.createSnapshot = function(options, callback) {
     throw new Error('Name must match [a-z]([-a-z0-9]{0,61}[a-z0-9])?');
   }
 
-  var compute = this.zone.compute;
+  var self = this;
   this.makeReq_('POST', '/createSnapshot', null, options, function(err, resp) {
     if (err) {
       callback(err);
       return;
     }
-    var snapshot = compute.snapshot(options.name);
-    callback(null, snapshot, resp);
+    var snapshot = self.zone.compute.snapshot(options.name);
+    var operation = self.zone.operation(resp.name);
+    operation.metadata = resp;
+    callback(null, snapshot, operation);
   });
 };
 
--- a/lib/compute/firewall.js
+++ b/lib/compute/firewall.js
@@ -77,11 +77,23 @@ function Firewall(compute, name, options) {
  * @param {function} callback - The callback function.
  *
  * @example
- * firewall.delete(function(err) {});
+ * firewall.delete(function(err, operation) {
+ *   // `operation` is an Operation object and can be used to check the status
+ *   // of firewall rule deletion.
+ * });
  */
 Firewall.prototype.delete = function(callback) {
   callback = callback || util.noop;
-  this.makeReq_('DELETE', '', null, true, callback);
+  var compute = this.compute;
+  this.makeReq_('DELETE', '', null, true, function(err, resp) {
+    if (err) {
+      callback(err);
+      return;
+    }
+    var operation = compute.operation(resp.name);
+    operation.metadata = resp;
+    callback(null, operation);
+  });
 };
 
 /**
@@ -128,8 +140,9 @@ Firewall.prototype.getMetadata = function(callback) {
  * @param {function} callback - The callback function.
  *
  * @example
- * var callback = function(err, apiResponse) {
- *   // ...
+ * var callback = function(err, operation) {
+ *   // `operation` is an Operation object and can be used to check the status
+ *   // of firewall rule update.
  * };
  *
  * firewall.setMetadata(
@@ -153,7 +166,9 @@ Firewall.prototype.setMetadata = function(metadata, callback) {
       return;
     }
     self.metadata = metadata;
-    callback(null, resp);
+    var operation = self.compute.operation(resp.name);
+    operation.metadata = resp;
+    callback(null, operation);
   });
 };
 
--- a/lib/compute/index.js
+++ b/lib/compute/index.js
@@ -53,6 +53,12 @@ var Snapshot = require('./snapshot.js');
 var Network = require('./network.js');
 
 /**
+ * @type {module:compute/operation}
+ * @private
+ */
+var Operation = require('./operation.js');
+
+/**
  * @type {module:common/util}
  * @private
  */
@@ -219,6 +225,25 @@ Compute.prototype.network = function(name) {
 };
 
 /**
+ * Get a reference to a Google Compute Engine operation.
+ *
+ * @param {string} name - Name of the existing operation.
+ * @return {module:compute/operation}
+ *
+ * @example
+ * var gcloud = require('gcloud')({
+ *   keyFilename: '/path/to/keyfile.json'
+ * });
+ *
+ * var compute = gcloud.compute();
+ *
+ * var operation = compute.operation('operation-name');
+ */
+Compute.prototype.operation = function(name) {
+  return new Operation(this, name);
+};
+
+/**
  * Create a firewall rule. For a detailed description of method's options see
  * [API reference](https://goo.gl/kTMHep).
  *
@@ -242,8 +267,10 @@ Compute.prototype.network = function(name) {
  * @param {function} callback - The callback function.
  *
  * @example
- * var callback = function(err, firewall, apiResponse) {
+ * var callback = function(err, firewall, operation) {
  *   // `firewall` is a Firewall object.
+ *   // `operation` is an Operation object and can be used to check the status
+ *   // of firewall rule creation.
  * };
  *
  * var compute = gcloud.compute({
@@ -298,7 +325,9 @@ Compute.prototype.createFirewall = function(name, options, callback) {
       }
       var firewall = self.firewall(name);
       firewall.metadata = options;
-      callback(null, firewall, resp);
+      var operation = self.operation(resp.name);
+      operation.metadata = resp;
+      callback(null, firewall, operation);
     });
 };
 
@@ -316,8 +345,10 @@ Compute.prototype.createFirewall = function(name, options, callback) {
  * @param {function} callback - The callback function.
  *
  * @example
- * var callback = function(err, network, apiResponse) {
+ * var callback = function(err, network, operation) {
  *   // `network` is a Network object.
+ *   // `operation` is an Operation object and can be used to check the status
+ *   // of network creation.
  * };
  *
  * var compute = gcloud.compute({
@@ -349,7 +380,9 @@ Compute.prototype.createNetwork = function(name, options, callback) {
     }
     var network = self.network(name);
     network.metadata = options;
-    callback(null, network, resp);
+    var operation = self.operation(resp.name);
+    operation.metadata = resp;
+    callback(null, network, operation);
   });
 };
 
@@ -944,6 +977,104 @@ Compute.prototype.getNetworks = function(options, callback) {
 };
 
 /**
+ * Get a list of global operations. For a detailed description of method's
+ * options see [API reference](https://goo.gl/gX4C1u).
+ *
+ * @param {object} options - Operation search options.
+ * @param {boolean=} options.autoPaginate - Have pagination handled
+ *     automatically. Default: true.
+ * @param {number=} options.maxResults - Maximum number of operations to return.
+ * @param {string=} options.filter - Search filter. The filter must contain:
+ *     `FIELD_NAME COMPARISON_STRING LITERAL_STRING`. `FIELD_NAME` is the name
+ *     of the field to compare. `COMPARISON` is the comparison operator that can
+ *     be either `eq` or `ne`. `LITERAL_STRING` is the string to filter to.
+ *     For string fields `LITERAL_STRING` can be a regular expression.
+ * @param {string=} pageToken - Page identifier used in paginated search.
+ * @param {function} callback - The callback function.
+ *
+ * @example
+ * var callback = function(err, operations) {
+ *   // `operations` is an array of `Operation` objects.
+ * };
+ *
+ * compute.getOperations(
+ *   {
+ *     filter: 'name eq operation-[0-9]'
+ *   }, callback);
+ *
+ * //-
+ * // To control how many API requests are made and page through the results
+ * // manually, set `autoPaginate` to `false`.
+ * //-
+ * var callback = function(err, operations, nextQuery, apiResponse) {
+ *   if (nextQuery) {
+ *     // More results exist.
+ *     compute.getOperations(nextQuery, callback);
+ *   }
+ * };
+ *
+ * compute.getOperations({
+ *   autoPaginate: false
+ * }, callback);
+ *
+ * //-
+ * // Get the operations from your project as a readable object stream.
+ * //-
+ * compute.getOperations()
+ *   .on('error', console.error)
+ *   .on('data', function(operation) {
+ *     // `operation` is a `Operation` object.
+ *   })
+ *   .on('end', function() {
+ *     // All operations retrieved.
+ *   });
+ *
+ * //-
+ * // If you anticipate many results, you can end a stream early to prevent
+ * // unnecessary processing and API requests.
+ * //-
+ * compute.getOperations()
+ *   .on('data', function(operation) {
+ *     this.end();
+ *   });
+ */
+Compute.prototype.getOperations = function(options, callback) {
+  if (util.is(options, 'function')) {
+    callback = options;
+    options = {};
+  }
+
+  options = options || {};
+
+  var self = this;
+  this.makeReq_(
+    'GET',
+    '/global/operations',
+    options,
+    null, function(err, resp) {
+      if (err) {
+        callback(err);
+        return;
+      }
+
+      var nextQuery = null;
+
+      if (resp.nextPageToken) {
+        nextQuery = extend({}, options, {
+          pageToken: resp.nextPageToken
+        });
+      }
+
+      var operations = (resp.items || []).map(function(operationObject) {
+        var operation = self.operation(operationObject.name);
+        operation.metadata = operationObject;
+        return operation;
+      });
+      callback(null, operations, nextQuery, resp);
+    });
+};
+
+/**
  * Make a new request object from the provided arguments and wrap the callback
  * to intercept non-successful responses.
  *
@@ -980,7 +1111,8 @@ streamRouter.extend(Compute, [
   'getAddresses',
   'getSnapshots',
   'getFirewalls',
-  'getNetworks'
+  'getNetworks',
+  'getOperations'
 ]);
 
 module.exports = Compute;
--- a/lib/compute/operation.js
+++ b/lib/compute/operation.js
@@ -1,0 +1,143 @@
+/*!
+ * Copyright 2014 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*!
+ * @module compute/operation
+ */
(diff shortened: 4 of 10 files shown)


jgeewax:
This is fantastic. Thank you so much :)

stephenplusplus:
A while back (#341), @ryanseys worked on an API for Compute Engine support: https://gist.github.com/ryanseys/28c9a17fea00899f3ac7

Could you take a look over the hierarchy he proposed and let us know if you think keeping it flat like in this PR is better, or if we should switch over to that style? Any other thoughts or comparisons would be great to have!

mziccard:
Personally, I have never thought of regions and zones as real "entities" in a cloud infrastructure. I always considered them as a mere geographical location where a resource (i.e. a real entity) resides. However, this is a personal thought that stems from my own work experience.

From a programming perspective I like @ryanseys hierarchical structure better as it gives a precise role to regions and zones and does not relegate them to string parameters. We should also take into account how we expect users to interact with cloud engine entities. On the one hand, it might be reasonable to assume that a user creates an *instance* via a *zone* entity as instances reside in zones. On the other hand, I also imagine that users tend to manipulate instances/disks in a *project* as a whole (possibly spread across zones/regions): a user might want to apply a tag to all instances in his project, regardless of the zone.   

It might be interesting to implement the hierarchical structure while also providing capabilities to manipulate entities at the global (*project*) level when the APIs allow us to (e.g. listing instances and disks at project level might be super-useful).

jgeewax:
AWS does quite a few things similarly from an API perspective. How do they do this in their Node SDK ?

stephenplusplus:
> It might be interesting to implement the hierarchical structure while also providing capabilities to manipulate entities at the global (project) level when the APIs allow us to (e.g. listing instances and disks at project level might be super-useful).

:+1: That matches the approach we try to take throughout the library.

mziccard:
@jgeewax It has been a while since I used them last but if I am not mistaken in the AWS SDK the `region` is a global configuration parameter (as the API key). All requests issued are then directed to the specified region. Then the structure is quite flat, you have an `EC2` entity that offers methods to manage resources, zones are passed as string parameters to these methods.  

@stephenplusplus It should not be too complicated to make this PR follow the hierarchical structure.

mziccard:
With the last commit I added `zone` and `region` entities. 
-  `listInstances` and `listDisks` methods of a `Zone` object return instances and disks in that zone
-  `createInstance` and `createDisk` are now `Zone` methods
-  `listInstances` and `listDisks` on a `Compute` object perform an aggregated search

I avoided adding default values `compute.allZones` and `compute.defaultZone` as I think that aggregated functions should apply directly to a `Compute` object and that there's no need for a default zone when building a `Compute` object.

stephenplusplus:
Thanks again, @mziccard! I'll be looking at this over the next few days/week. Sorry in advance for the lag!

lib/compute/disk.js
```diff
@@ -0,0 +1,146 @@
...
+ * @type {module:common/util}
+ * @private
+ */
+var util = require('../common/util.js');
+
+/**
+ * @const {string}
+ * @private
+ */
+var COMPUTE_BASE_URL = 'https://www.googleapis.com/compute/v1/projects/';
```
callmehiphop: I think if we could avoid duplicating the `COMPUTE_BASE_URL` across the different classes if we made use of all the `makeReq_` methods and did not use `Compute#makeAuthorizedRequest_` directly

lib/compute/index.js
```diff
@@ -0,0 +1,600 @@
...
+ *     filter: {
+ *       fieldName : 'name',
+ *       operatorString : 'eq',
+ *       literalString : 'instance-[0-9]'
+ *     }],
+ *   }, callback);
+ */
+Compute.prototype.listInstances = function(options, callback) {
+  if (!callback) {
+    callback = options;
```
callmehiphop: I think rather than assuming options is meant to be the callback, we could do something like

```javascript
if (util.is(options, 'function')) { ... }
```

And at the same time we can set a default value for options and skip the check on line 318 altogether.

lib/compute/firewall.js
```diff
@@ -0,0 +1,259 @@
...
+ * var gcloud = require('gcloud');
+ *
+ * var compute = gcloud.compute({
+ *   projectId: 'grape-spaceship-123'
+ * });
+ *
+ * var firewall = compute.firewall('tcp-3000');
+ */
+function Firewall(compute, name, options) {
+  this.name = name;
```
callmehiphop: We could probably simplify a lot of this logic if we did something like..

```javascript
options = options || {};
this.description = options.description || '';
```

lib/compute/firewall.js
```diff
@@ -0,0 +1,259 @@
...
+ * };
+ *
+ * firewall.update(
+ *   {
+ *     patch: true,
+ *     description: 'Brand new description',
+ *   }, callback);
+ */
+Firewall.prototype.update = function(options, callback) {
+  if (!options) {
```
callmehiphop: Probably want to add type checking in case the provide a callback

lib/compute/firewall.js
```diff
@@ -0,0 +1,259 @@
...
+    if(!util.is(options.allowed, 'array')) {
+      throw new Error('Allowed protocols and ports must be provided.');
+    }
+    if (!util.is(options.sourceRanges, 'array') && 
+      !util.is(options.sourceTags, 'array')) {
+      throw new Error('Source ranges or source tags must be provided.');
+    }
+  }
+
+  var body = {};
```
callmehiphop: I think this could be simplified into something like

```javascript
var body = extend({
    name: this.name,
    network: this.network
  }, options);

  var newFirewall = new Firewall(this.compute, body.name, extend({}, this, body));
```

lib/compute/index.js
```diff
@@ -0,0 +1,600 @@
...
+                zoneInstances[i]);
+              instances.push(instance);
+            }
+          }
+        }
+        callback(null, instances);    
+      } else {
+        callback(null, []);
+      }
+    }.bind(this));
```
callmehiphop: While I do love `.bind()` I think our style guide discourages the use of it.

https://github.com/GoogleCloudPlatform/gcloud-node/issues/727#issuecomment-122353264

If you rebase your branch we added a style linter to the `npm run lint` command

lib/compute/firewall.js
mziccard: Actually I spotted an error in the code and fixed it: old values are preserved in the new object only if we are using the PATCH semantics.
While being more complex the code as it is checks the type of data provided by the user. Since such data will be used to build a remote request I think it is better to check input data to avoid adding values that might cause the request to fail. This is the reason why I do not add to the body values in options that do not have the expected type.

lib/compute/index.js
mziccard: Thanks for pointing that out. I linted the code with the new command and removed all the `bind()` in the Compute module.

lib/compute/index.js
```diff
@@ -0,0 +1,602 @@
...
+ *   {
+ *     zone: 'europe-west1-b',
+ *     filter: {
+ *       fieldName : 'name',
+ *       operatorString : 'eq',
+ *       literalString : 'instance-[0-9]'
+ *     }],
+ *   }, callback);
+ */
+Compute.prototype.listInstances = function(options, callback) {
```
callmehiphop: Despite the upstream API specifying something as a list, the preferred verbiage would be `getInstances`. (source: https://github.com/GoogleCloudPlatform/gcloud-node/pull/706#discussion_r35118052)

lib/compute/disk.js
```diff
@@ -0,0 +1,129 @@
...
+ * @param {string=} description - Description of the snapshot.
+ * @param {function} callback - The callback function.
+ *
+ * @example
+ * disk.createSnapshot('my-snapshot', function(err, apiResponse) {
+ *   // Handle error
+ * });
+ */
+Disk.prototype.createSnapshot = function(name, description, callback) {
+  if (!name) {
```
callmehiphop: Instead of making description an optional param, I think it might be better to overload name to be an object/string

```javascript
Disk.prototype.createSnapshot = function (options, callback) {
  if (util.is('string', options)) {
    options = {
      name: options
    };
  }
};
```

lib/compute/disk.js
```diff
@@ -0,0 +1,129 @@
...
+    callback = description;
+  }
+
+  var body = {};
+  body.name = name;
+  if (description) {
+    body.description = description;
+  }
+
+  this.makeReq_('POST', '/createSnapshot', null, body, callback);
```
callmehiphop: I was looking at the Snapshot API and I also saw get/list/delete methods, should we create a `Snapshot` class and return an instance of that after we create one?

```javascript
disk.createSnapshot('my-snapshot', function (err, snapshot, apiResponse) {
  snapshot.delete(function (err, apiResponse) {
    // ...
  });
});
```

lib/compute/firewall.js
callmehiphop: What is the use case for `PATCH`? I see where you're going with this since they essentially do the same thing, I'm just wondering if we actually need it or if it would better as its own method.

mziccard: In the API the PATCH method only updates the provided Firewall fields, the others are left unchanged. UPDATE, instead, updates provided Firewall fields and sets non-specified fields to empty or default values (as done for POST). Right now, we distinguish between PATCH and UPDATE via a boolean options (`options.patch`) but this is totally open for discussion, we could go for adding a patch method as well. Comments?

lib/compute/disk.js
mziccard: We totally should. Some methods in this module are placeholder waiting for time to implement more functionalities. This is one of such methods. I will add `snapshot` obje

lib/compute/firewall.js
stephenplusplus: We usually call this kind of operation `setMetadata`, and use PATCH. While having the option for both offers a complete solution, we are primarily concerned with targeting the one of least surprises, which is usually PATCH. So, I would suggest for consistency with our other APIs, we rename this to setMetadata and use PATCH.

mziccard: I am ok with always using PATCH. Not totally sure that `setMetadata` is the most intuitive name for this method as it can completely reshape a firewall rule (name, description, targets, sources...). If we go for this `setMetadata` for consistency would you also want all `Firewall` fields to be removed (apart from `name`) in favor of  a generic `metadata` field?

stephenplusplus: Yes, please. It's just an easy way to keep track of an API-only representation of the object; a place to find it, and an understanding of how to change it. A usual class looks something like:

```js
function Something(parentClass, name, metadata) {
  this.parentClass_ = parentClass;
  this.name = name;
  this.metadata = metadata || {};
}

Something.prototype.getMetadata = function(callback) {
  var self = this;
  this.makeReq_('GET', '/something/' + this.name, null, null, function(err, resp) {
    // ...

    self.metdata = resp; // Update the instance metadata object

    callback(null, self.metadata, resp);
  });
};

Something.prototype.setMetadata = function(options, callback) {
  var self = this;
  this.makeReq_('PATCH', '/something/' + this.name, null, options, function(err, resp) {
    // ...

    self.metadata = resp;
    callback(null, resp);
  });
};
```

Thanks for working with us on these changes!

mziccard:
Guys, I added some more functionalities (Snapshots and Addresses). I think I'll stop adding stuff and wait for a review of yours so that I can align the code (and myself) to your coding style and apply your suggestions/fixes before going on.  

I think next step would be adding an Operation class. Most APIs are asynchronous and return as a response an operation object. Users can then check the operation status and possible errors. We could also provide something (roughly speaking) like:

```javascript
operation.onComplete(
    period, // The polling interval to update operation metadata
    function(error, operation) {
        // operation has status DONE and error is set if an error occurred
    });
```

Or even something more elaborated to stream also other events (such as the transition from PENDING to RUNNING or the operation's progress).

jgeewax:
Re operations -- I think this is something we should totally do, might be a good idea to look across other APIs for where the concept of long-running operations exist and make sure that what we do fits for those also -- it's definitely going to be a common pattern.

lib/compute/index.js
```diff
@@ -0,0 +1,793 @@
...
+          if (regionAddresses) {
+            for (var i = 0; i < regionAddresses.length; i++) {
+              var metadata = regionAddresses[i];
+              var address = region.address(metadata.name);
+              address.metadata = metadata;
+              addresses.push(address);
+            }
+          }
+        }
+        callback(null, addresses);
```
callmehiphop: Usually for aggregated lists we support auto-pagination by default, this is done in two simple steps.

1. Return an object after the response that contains all of the options needed to make subsequent requests -- [bigquery#L329](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/bigquery/index.js#L329)

2. Extend this method using `streamRouter`, this will automatically make any subsequent requests and also provide a stream if a callback was not supplied. -- [bigquery#603](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/bigquery/index.js#L603)

mziccard: Nice! I'll do this:) I'll also do some refactoring to get rid of some `if` and `for`.
Just to clarify: I think we should use this auto-paginated approach also with `zone.getInstances`, `zone.getDisks` and `region.getAddresses`, can you confirm that? 

We should move elsewhere the *filter validation and creation* code as it is a lot of repeated code across all `getXXXs` methods, any suggestion? Should we add a `Filter` class with a constructor that does validation and something like a `URIencode` method that gives back a correctly formatted string filter?

callmehiphop: It would seem that all of those methods could leverage `streamRouter`!

@stephenplusplus any thoughts on the filter validation stuff?

stephenplusplus: I think I would actually just accept a string for the filter and not do any custom validation or handling of the user's input. I always prefer to let the API catch mistakes on validation errors like these, as those can change and our library would go stale. It also seems easier to me from a developer's perspective:

```js
zone.getInstances({
  filter: 'name eq example-instance'
}, function(err, instances, apiResponse) {});
```

vs having to deconstruct the string:

```js
zone.getInstances({
  filter: {
    fieldName: 'name',
    operatorString: 'eq',
    literalString: 'example-instance'
  }
}, function(err, instances, apiResponse) {});
```

Something we need to do a better job of is linking from this library's docs to the upstream API docs *everywhere possible* so the user can see exactly what the API expects. (I'll create an issue for this)

mziccard: Filters are now simple strings, I added auto pagination support and removed filter validation so that we let the API catch those errors. 
I added no reference to the API docs yet. Were you thinking about adding a reference on a per-parameter basis or something like: *"For a detailed description of method options see [API REFERENCE]()."* on top of the docs that describe a method?

stephenplusplus: Just a link at the method level like you suggested would be great.

mziccard:
Hi guys. As you can see I added some more functionalities (Network) and addressed your comments. I started thinking about how to implement operation and would like to do some brainstorming before starting to code anything.  
I went through the docs and identified 3 types of operations `GlobalOperation`, `RegionOperation` and `ZoneOperation`. The three JSON objects differ only in the fact that `RegionOperation` has a `region` field and `ZoneOperation` has a `zone` field. Given this small difference I am not so convinced that we need 3 classes one for each operation and I would rather go for a single `Operation` class. What do you think?  
Operations can take time to complete (go to `DONE` status) so we need a method for the user to register a callback for the complete event. I had in mind something like:
```javascript
// Interval is the optional polling period, default is 10
operation.onComplete(interval, function(error, warnings) {
  // operation metadata are updated and status==='DONE'
  // If error is set the operation failed.
  // If warnings is set the operation succeeded with warnings.
}); 
```
If no callback is provided the `onComplete` method returns a stream the user can register listeners on.
```javascript
// Interval is the polling period, is optional
operation.onComplete(interval)
  .on('complete', function(error, warnings) {
    // operation metadata are updated and status==='DONE'
    // If error is set the operation failed.
    // If warnings is set the operation succeeded with warnings.
  })
  .on('progress', function(progress) {
    // progress is a monotonically increasing progress indicator in [0, 100]
  });
```
Comments that improve/change/trash my proposal are more than welcome:)

callmehiphop:
> Given this small difference I am not so convinced that we need 3 classes one for each operation and I would rather go for a single Operation class.

That sounds like a good plan to me.

> I am not so happy of the redundancy of operation.onComplete(interval).on('complete', ....

We have a pretty elegant [polling solution in the PubSub module](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/pubsub/subscription.js#L211), I think a similar approach would work well here.

```javascript
var operation = compute.operation('myOperation', {
  type: 'global',
  interval: 99999
});

operation.on('done', function() {});
operation.on('error', function() {});
```

We can use the `EventEmitter` to watch for listeners and start/stop polling according to that.
Thoughts?

mziccard:
@callmehiphop I saw what you did in the PubSub module and might be a good solution here. The difference is that an operation does not necessarily have an `interval` attribute when it is created. Let me clarify, in a method like `getOperations` (I assume we will have it for `Compute`, `Zone` and `Region`), you get and build some `Operation` objects but there's no `interval` provided there. What happens if the user then directly uses `on` are we going to use the default value? This is not a big issue but I feel like an operation has a different semantics from a subscription which is more event-oriented in its nature.

callmehiphop:
We could either pass the `interval` into `getOperations` and apply it to all returned operations (not sure how good of an idea that is) or alternatively they could be set after the fact.

```javascript
compute.getOperations(function(err, operations) {
  operations.forEach(function(operation) {
    operation.interval = 9999;
  });
});
```

/cc @stephenplusplus

stephenplusplus:
I prefer the second option @callmehiphop suggested.

What happens when the user gets an Operation that is already completed or had errors?

```js
getOperations(function(err, operations) {
  var aCompletedOperation = operations[0];

  aCompletedOperation.on('complete', function () {
    // immediately invoked with aCompletedOperation.metadata?
  });

  var aCompletedOperationWithErrors = operations[1];

  aCompletedOperationWithErrors.on('error', function() {
    // invoked once per aCompletedOperationWithErrors.errors[]?
  });
  aCompletedOperationWithErrors.on('complete', function() {
    // immediately invoked with aCompletedOperationWithErrors.metadata?
  });
});
```

Just for reference, a similar concept exists in BigQuery where certain actions return a Job instance which must be pinged for its completion status. We don't give the user a way to listen for an event when it is complete. Instead, the user just calls `job.getMetadata` or `job.getQueryResults` to check the status / get the results.

I think an EventEmitter for just-returned, still-running operations would be handy, but for complete operations would be pretty confusing. I'm leaning more towards putting the task of pinging for a complete status on the user and not developing a solution around this. But it would be great if there was a solution that wasn't awkward when applied to an already-completed operation.

mziccard:
I would say that in case of a just completed operation we immediatelly fire all the events. But I agree we better go with just providing the `getMetadata` metadata and let the user handle the polling until we come  up with a better solution. We might add an example in the docs that uses `getMetadata` in combination with `setTimeout` to check the status until the operation's status is `DONE`.

stephenplusplus:
Sounds good to me :+1:

>

mziccard:
I think I found a problem while adding operations support that might need your intervention:) 
When you do a GET on an operation (our `operation.getMetadata`) it returns a JSON object that describes it. If the operation failed its JSON object has an `error` field set and this causes the `util.parseApiResp` function to assume the API call we are performing failed and `util.handleResp` to run `callback` with error parameter set.  
I wanted to go around this problem by checking in `operation.getMetadata` that the response was actually a well formed `Operation` and not an error. However I realized that in case of a (supposed) error `util.handleResp` calls the callback with only the error parameter set (no `response`). A possible fix could be returning the `response` even when an error is detected and let me handle mis-detected errors in `operation.getMetadata`. This change should be done by modifying line 193 of util.js (double check this as I only had a quick look at util.js):
```javascript
callback(parsedApiResponse.err);
```
to
```javascript
callback(parsedApiResponse.err, parsedApiResponse.body, parsedApiResponse.resp);
```
This change should not affect the library nor the users. Thoughts?

stephenplusplus:
It was actually a bug that handleResp was only executing the callback with the parsed error, so we fixed it recently to return all parts (just like you suggested). See if re-basing fixes the problem.

mziccard:
My bad, I did not notice that you had already fixed the problem. I rebased and everything worked, thanks. We now have operation support. For those APIs that are asynchronous (i.e. those that respond with an operation JSON) now the `callback` takes an `operation` parameter (instead of an `apiResponse` parameter) that can be used to check the state of the request. In the `operation.getMetadata` docs I put an example showing how to use the method to check the operation's status until it's `DONE`.

stephenplusplus:
Perfect, thank you! If it's okay with you, @callmehiphop and I can take this over now to write the tests and get it merged in asap!

mziccard:
@stephenplusplus Yeah that's more than okay with me. I feel a bit bad for making you and @callmehiphop test all that code:) But I guess this is fastest way to do it.

stephenplusplus:
It's no problem, we're ecstatic you contributed all of this code for us!

stephenplusplus:
Thanks again @mziccard! Going to close this PR in favor of #792.

lib/compute/operation.js
```diff
@@ -0,0 +1,143 @@
...
+ *   });
+ * };
+ *
+ * setTimeout(checkStatus, 10);
+ */
+Operation.prototype.getMetadata = function(callback) {
+  callback = callback || util.noop;
+  var self = this;
+  this.makeReq_('GET', '', null, null, function(err, resp) {
+    if (err && (!resp || resp.name !== self.name)) {
```
stephenplusplus: @mziccard what was the reason for this error checking?

mziccard: The JSON that comes back when you do get an operation entity contains an `error` field that causes the core of `gcloud-node` to think that the request failed with an error. With that check I am checking that we actually have an error and not an operation entity: we have an error if the response is empty or the name field in the response is not the one of the operation we are trying to get.  
This is workaround but I think it makes more sense than changing how the core of `gcloud-node` requests works which is already very good IMHO.

stephenplusplus: Oh, that makes perfect sense. Thanks!

PR #732, opened 2015-07-20
Refactor Datastore/Query Tests

### LINKED ISSUE #719: Datastore/Query tests need improving
We could do a better job of deconstructing the Query class and testing that the methods do what we want.

### LINKED ISSUE #731: Is the Datastore Query object supposed to be immutable?
At first glance I assumed that the `Query` class was supposed to be immutable -- each method creates a new query instance, copies all of the properties from the previous instance and then returns it. 

However, I noticed that for methods that can be called multiple times (`filter()` and `order()`) it holds a reference to the previous instance.

e.g.

```javascript
var query1 = new Query(['kind1']);
var query2 = query1.filter('count >', 5);

query1.filters.length; // => 1
```

Is this the desired functionality?

callmehiphop:
Closes #719 
Fixes #731

### DIFF
--- a/test/datastore/query.js
+++ b/test/datastore/query.js
@@ -25,116 +25,324 @@ var Query = require('../../lib/datastore/query.js');
 var queryProto = require('../testdata/proto_query.json');
 
 describe('Query', function() {
-  it('should use null for all falsy namespace values', function() {
-    [
-      new Query('', 'Kind'),
-      new Query(null, 'Kind'),
-      new Query(undefined, 'Kind'),
-      new Query(0, 'Kind'),
-      new Query('Kind')
-    ].forEach(function(query) {
-      assert.strictEqual(query.namespace, null);
-      assert.equal(query.kinds, 'Kind');
+
+  describe('instantiation', function() {
+
+    it('should use null for all falsy namespace values', function() {
+      [
+        new Query('', 'Kind'),
+        new Query(null, 'Kind'),
+        new Query(undefined, 'Kind'),
+        new Query(0, 'Kind'),
+        new Query('Kind')
+      ].forEach(function(query) {
+        assert.strictEqual(query.namespace, null);
+        assert.equal(query.kinds, 'Kind');
+      });
     });
+
+    it('should support custom namespaces', function() {
+      var query = new Query('ns', ['kind1']);
+      assert.equal(query.namespace, 'ns');
+    });
+
+    it('should enable auto pagination by default', function() {
+      var query = new Query(['kind1']);
+      assert.strictEqual(query.autoPaginateVal, true);
+    });
+
   });
 
-  it('should support custom namespaces', function() {
-    var query = new Query('ns', ['kind1']);
-    assert.equal(query.namespace, 'ns');
+  describe('autoPaginate', function() {
+
+    it('should enable auto pagination', function() {
+      var query = new Query(['kind1']).autoPaginate();
+
+      assert.strictEqual(query.autoPaginateVal, true);
+    });
+
+    it('should disable auto pagination when false is passed in', function() {
+      var query = new Query(['kind1']).autoPaginate(false);
+
+      assert.strictEqual(query.autoPaginateVal, false);
+    });
+
+    it('should not disable auto pagination with falsy values', function() {
+      var query = new Query(['kind1']).autoPaginate(null);
+
+      assert.strictEqual(query.autoPaginateVal, true);
+    });
+
+    it('should return a new query', function() {
+      var query = new Query(['kind1']);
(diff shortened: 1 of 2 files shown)


test/datastore/query.js
```diff
@@ -25,116 +25,333 @@ var Query = require('../../lib/datastore/query.js');
...
+      assert.equal(query.filters[1].val, 'Title');
+
+      assert.equal(query.filters[2].name, 'count');
+      assert.equal(query.filters[2].op, '>');
+      assert.strictEqual(query.filters[2].val, 20);
+    });
+
+    it('should remove any whitespace surrounding the filter name', function() {
+      var query = new Query(['kind1'])
+        .filter('   count    >', 123);
```
stephenplusplus: Spacing of `   count    >` here and the next test is a little weird. Was this intentional?

test/datastore/query.js
```diff
@@ -25,116 +25,333 @@ var Query = require('../../lib/datastore/query.js');
...
+      var query = new Query(['kind1'])
+        .order('name');
+
+      assert.equal(query.orders[0].name, 'name');
+      assert.equal(query.orders[0].sign, '+');
+    });
+
+    it('should support ascending order', function() {
+      var query = new Query(['kind1'])
+          .order('+name');
```
stephenplusplus: Let's just go with same line until there are multiple chained filters.

E.g.:

```js
// both good:
var query = new Query(['kind1']).order('name');
var query = new Query(['kind1'])
  .order('name')
  .filter('count <=', 5);
```

test/datastore/query.js
```diff
@@ -25,116 +25,333 @@ var Query = require('../../lib/datastore/query.js');
...
+      query = query.order('size');
+
+      assert.equal(query.orders.length, 1);
+    });
+
+    it('should return a new query', function() {
+      var query = new Query(['kind1']);
+      var newQuery = query.order('count');
+
+      assert.notDeepEqual(query.orders, newQuery.orders);
```
stephenplusplus: Didn't even know `notDeepEqual` was a thing! Can we test notStrictEqual?

```js
assert.notStrictEqual(query, newQuery)
```

callmehiphop: We could, my primary concern is that somewhere down the line a `return` is forgotten and `newQuery` becomes `null`. 

```javascript
assert.notStrictEqual(query, null);
```

stephenplusplus:
```js
assert(nextQuery instanceof Query);
assert.notStrictEqual(query, null);
```

?

test/datastore/query.js
callmehiphop: :+1:

test/datastore/query.js
callmehiphop: Yep, this test checks to make sure all white space around the filter name is removed (before and after)

stephenplusplus: I've only recently started reading. Please forgive me.

test/datastore/query.js
callmehiphop: I don't think comparing the two queries directly is the way to go -- we'll become prone to false positives.

e.g. with the issue described in #731 the two `Query` instances were different but `filters` and `orders` are references. A test where we would do a strict comparison between the two objects would pass, even though a test against the filters/orders would not.

stephenplusplus: Sounds good.

test/datastore/query.js
stephenplusplus: Hmm, I think that regex should be:

```js
/[>|<|=|>=|<=]/g
```

stephenplusplus: Also, please name it in the code, something like `operatorRegex`.

callmehiphop: Sure! Do you think we should add tests for each operator in the regex?

stephenplusplus: It wouldn't hurt, you could wrap them all in one `it` to avoid it being too loud, but either way would be great.

test/datastore/query.js
```diff
@@ -25,116 +25,324 @@ var Query = require('../../lib/datastore/query.js');
...
+      var query = new Query(['kind1']).order('-count');
+
+      assert.equal(query.orders[0].name, 'count');
+      assert.equal(query.orders[0].sign, '-');
+    });
+
+    it('should support both ascending and descending', function() {
+      var query = new Query(['kind1'])
+          .order('+name')
+          .order('-count');
```
stephenplusplus: Two spaces here please.

stephenplusplus:
LGTM, just a small formatting request.

stephenplusplus:
Thanks!

PR #734, opened 2015-07-21
Migrating to gcloud-common docs

### LINKED ISSUE #710: Clarify API Enablement instructions in the Authorization section of readme.md
In the Authorization -> Elsewhere section of the readme, readers are instructed to:

Navigate to APIs & auth > APIs section and turn on the following APIs (you may need to enable billing in order to use these services):
1. Google Cloud Datastore API
2. Google Cloud Storage
3. Google Cloud Storage JSON API

Given this description is API agnostic, it is slightly confusing why these three APIs are chose. I would recommend either:

1. Adding clear instruction to 'Enable the APIs you intend to use, for example, for Cloud Storage you will need to enable Cloud Storage API and Cloud Storage JSON API"

or

2. Including all APIs supported by gcloud-node, with a note that if they do not intend to use all services, you only need to enable those you do need.

As a small nit, lets use the language 'enable' and 'disable' for consistency with the experience when they get to the GUI.

### LINKED ISSUE #717: Pull docs from gcloud-common
Fixes: #710 

We need to update our docs to pull from gcloud-common (dependent on [this PR](https://github.com/GoogleCloudPlatform/gcloud-common/pull/3)).

callmehiphop:
Fixes #710
Fixes #717
- [x] Point to gcloud-common markdown files
  - [x] Authorization
  - [x] FAQs
  - [x] Contributing (dependent on https://github.com/GoogleCloudPlatform/gcloud-common/pull/10)
- [x] Update image markdown within gcloud-common - https://github.com/GoogleCloudPlatform/gcloud-common/pull/8
- [x] Create angular directive to fix relative image paths
- [x] Remove duplicate markdown content from local faq/auth -  https://github.com/GoogleCloudPlatform/gcloud-common/pull/9
- [x] Create Showdown extension for relative anchor tags
- [x] Address markdown rendering errors - https://github.com/GoogleCloudPlatform/gcloud-common/pull/11 & https://github.com/showdownjs/showdown/issues/183
- [x] Implement deeplinking

For your viewing pleasure: http://callmehiphop.github.io/gcloud-node-ghpages/#/docs/master


callmehiphop:
@stephenplusplus Will we be deleting the authorization/faq markdown files within gcloud-node? It contains some information tailored to our library specifically, so I wasn't sure if wanted to have an additional auth link called something like "gcloud-node Authorization"

jgeewax:
Maybe there's a way to append or inject more markdown into these at
doc-generation time?

callmehiphop:
It's just an angular directive, so we could include both on the same page and just trim the node specific one to remove any duplicated information.

stephenplusplus:
For the images, we need to change `gcloud-comon` to use the `` style. For some reason, I used ``. After that, we have a couple options, ranked by how good of a solution I think they are (please correct me if I'm wrong!)

  1. Use absolute image paths in the gcloud-common markdown
  1. Use a [showdown extension](https://github.com/showdownjs/showdown/wiki/extensions) to replace img paths with absolute urls at runtime
  1. Use a directive to replace img paths with absolute urls at runtime

The extension would look like:

```js
angular
  .module('gcloud.authorization', [/*...*/])
  .config(function($routeProvider, markdownConverterProvider) {
    'use strict';

    markdownConverterProvider.config({
      extensions: [function() {
        return [
          {
            type: 'output',
            regex: 'img src="',
            replace: 'img src="https://raw.githubusercontent.com/GoogleCloudPlatform/gcloud-common/master/authorization/'
          }
        ];
      }]
    });

  //...
```

The directive could look like:

```js
angular
  .module('gcloud.detect-images', [])
  .directive('detectImages', function() {
    return function($scope, elem, attrs) {
      [].slice.call(elem.find('img')).forEach(function(img) {
        var src = img.getAttribute('src');
        img.setAttribute('src', attrs.detectImages + src);
      });

      return elem;
    };
  });
```
```html
<btf-markdown
  highlight-markdown
  detect-images="https://raw.githubusercontent.com/GoogleCloudPlatform/gcloud-common/master/authorization/"
  ng-include="'https://raw.githubusercontent.com/GoogleCloudPlatform/gcloud-common/master/authorization/readme.md'">
```

stephenplusplus:
For the Node-specific docs, I think we will have to add them after the common docs.

For Authorization, the common docs end with downloading a JSON key file. We can add on right after that to show how you use it with our library, and the transition should be easy. Same for FAQ, we can just tack any custom questions after the common ones.

It's not awesome to have a dependency on the wording of the docs, so I suppose we will just have to keep our docs in mind anytime the common docs change.

jgeewax:
Yea - maybe a "gcloud-node specific auth" section?

callmehiphop:
I ended up going with the directive option

* The hardcoded absolute URL feels gross to me and it might end up only being useful to us and not the other gcloud-* projects
* Configuring the provider doesn't promote re-use in case we need to do something similar for a different markdown file (e.g. we add images to faqs)

callmehiphop:
So I put the node specific auth stuff back in and it looks pretty good, however the faqs are kinda ugly now. Ideally I think we would split the *Where can I get more help?* section into its own markdown file because it would be better placed at the end of the page. There were also some small formatting differences (h2s became h4s) so it's a little harder to distinguish between questions and answers since the text looks the same.

Any opposition to making some minor changes to the common faq?

// cc @stephenplusplus

stephenplusplus:
Change it up! We're the only ones implementing it so far (afaik), so let's
start with what works for us, and make changes as necessary as others join.
Also, we all share the same template, so it should be difficult to do
something for us that won't also benefit the others.

docs/site/components/authorization/authorization.html
```diff
@@ -2,7 +2,7 @@
   <header header title="Authorization">
     <div class="row row--right">
       <div class="col margin-vertical">
-        <a href="https://github.com/GoogleCloudPlatform/gcloud-node/edit/master/docs/authorization.md"
+        <a href="https://github.com/GoogleCloudPlatform/gcloud-common/blob/master/authorization/readme.md"
```
stephenplusplus: I think we still want "edit" instead of "blob".

stephenplusplus:
> The hardcoded absolute URL feels gross to me and it might end up only being useful to us and not the other gcloud-* projects

I actually think it would help the other projects, since they should incorporate the live/master representation of `gcloud-common` in their docs instead of cloning locally. So, the image should always be remote, requiring an absolute path. Am I not considering something?

callmehiphop:
If another project chose to use submodules over absolute URLs, it would kinda defeat the point of using submodules since the markdown files would be referencing an absolute image path instead of the relative one.

stephenplusplus:
But I think if other projects used submodules, that would kind of defeat the point of the shared docs, since they would constantly have to be on the lookout for gcloud-common updates and manually do a submodule update and re-build of the site.

I know we have it easy, since we just point an Angular directive to a remote MD file, and boom, magic. But, for a site that doesn't have such an easy set-up, they will have to run some kind of MD -> HTML conversion as well, and the remote image will always work for everyone. Our docs are GitHub hosted and so are the -common images, so I think we should just be ¯\_(ツ)_/¯ about it and use the absolute to dramatically simplify our site and wait to know it's a problem with others before making it harder on ourselves than it might need to be.

stephenplusplus:
It looks so awesome. How should we handle the links inside of the common MD? Specifically, these ones are problematic:

**Contributing**

- If you encounter any difficulty during the process, do not hesitate to *get in touch*. (links to http://callmehiphop.github.io/troubleshooting/readme.md)
- If you've tried everything in our *Troubleshooting* guide... (links to http://callmehiphop.github.io/troubleshooting/readme.md)

stephenplusplus:
Is it possible that the common links change from:

`/troubleshooting/readme.md`

to 

`/troubleshooting` (if browsing on GH, the readme will still be displayed)

Then we can have a base href of `http://googlecloudplatform.github.io/gcloud-node/#/`?

callmehiphop:
I think that would be a little bit tricky, `<base>` would mess with all of our dependency paths. Maybe for this one a Showdown plugin would be better?

stephenplusplus:
Oh yeah, that makes more sense.

docs/site/components/relative-link/relative-link.js
```diff
@@ -0,0 +1,17 @@
+(function(Showdown) {
+
+  'use strict';
+
+  Showdown.extensions.relativelink = function () {
```
stephenplusplus: Can you move this to the home.js config block? Example: https://github.com/GoogleCloudPlatform/gcloud-node/pull/734#issuecomment-123827020

docs/site/home.js
```diff
@@ -20,6 +22,20 @@ angular
       templateUrl: 'site/home.html',
       controller: 'HomeCtrl'
     });
+
+    markdownConverterProvider.config({
+      extensions: [function () {
+        return [{
+          type: 'lang',
+          regex: '\\[([^\\]]+)\\]\\((\\/[^)]+)\\)',
```
stephenplusplus: Crazy. Would writing this out as a real RegExp() be possible, then sending the stringified value here?

callmehiphop: I don't think so, calling `toString()` on `RegExp` will include the trailing/leading slashes. Then Showdown will pass that string to the `RegExp` constructor and that will automatically escape things that shouldn't be escaped.

stephenplusplus: I guess we could trim them off safely? It's gross, but not as gross as that string, I'm pretty sure!

stephenplusplus:
@callmehiphop I played with this locally using the new Showdown updates (using the rawgit cdn), and all seems well. Just a little quirk... I had to add this after including the file in index.html:

```html
<script>
  Showdown = showdown
  Showdown.converter = showdown.Converter
 </script>
```

(https://github.com/btford/angular-markdown-directive/pull/39)

callmehiphop:
@stephenplusplus yeah, that's because we're on a really old version of showdown, they changed the API a little in 1.x.x

stephenplusplus:
That's because of angular-markdown-directive. We can use [this file](https://github.com/tivie/angular-markdown-directive/blob/134e34a650e259095736784b6ffc49d153fcb534/markdown.js) from https://github.com/btford/angular-markdown-directive/pull/39.

docs/site/components/markdown/markdown-service.js
```diff
@@ -0,0 +1,28 @@
+angular
+  .module('gcloud.markdown')
+  .factory('mdConverter', function() {
+    return new showdown.Converter({
+      extensions: [function () {
+        return [{
+          type: 'lang',
+          // gah \[([^\]]+)\]\(([\/|\#][^)]+)\)
```
stephenplusplus: gah!

stephenplusplus:
Any ideas how we can make the subheadings non-tiny?

screen shot 2015-08-04 at 4 55 04 pm

stephenplusplus:
Looks like this list got busted. Is there also a way to indent and style those bullet points? (Google Cloud Datastore API, etc)

screen shot 2015-08-04 at 4 55 46 pm

docs/site/components/markdown/markdown-service.js
```diff
@@ -0,0 +1,28 @@
...
+          type: 'lang',
+          // unescaped version \[([^\]]+)\]\(([\/|\#][^)]+)\)
+          regex: '\\[([^\\]]+)\\]\\(([\\/|\\#][^)]+)\\)',
+          replace: function (match, text, href) {
+            var isDeeplink = /^\#/.test(href);
+            var attribute;
+
+            if (isDeeplink) {
+              href = href.split(/[\#|\-]/).join('');
+              attribute = 'href="" ng-click="setSection(\'' + href + '\')"';
```
stephenplusplus: Is there a way to keep a meaningful href value to give the user an idea what will happen when they click?

stephenplusplus:
Since you have figured out deeplinking (yay!!), can we add a clickable "#" such as here: http://callmehiphop.github.io/gcloud-node-ghpages/#/docs/master/bigquery/dataset next to `createTable`?

stephenplusplus:
I noticed when hitting http://callmehiphop.github.io/gcloud-node-ghpages/#/authorization?section=createanewserviceaccount in a new tab directly, it will scroll, but it won't be as accurate as if I clicked the link from within the doc.

stephenplusplus:
Related: 

CONTRIBUTING >> "Finding something to work on"

screen shot 2015-08-04 at 5 03 31 pm

docs/site/components/markdown/markdown-service.js
stephenplusplus: Also, since they're all "nothing", the links appear as if they've been visited already.

callmehiphop:
> Any ideas how we can make the subheadings non-tiny?

Currently they are `<h6>` tags, could we just bump them up to `<h4>`s

> Looks like this list got busted. Is there also a way to indent and style those bullet points? (Google Cloud Datastore API, etc)

Looks like a showdown bug, we can increase the indent and it looks like that will work. (https://jsfiddle.net/xjeeLpmt/)

> I noticed when hitting http://callmehiphop.github.io/gcloud-node-ghpages/#/authorization?section=createanewserviceaccount in a new tab directly, it will scroll, but it won't be as accurate as if I clicked the link from within the doc.

That looks like a race condition from the markdown not being fully rendered yet..

docs/site/components/markdown/markdown-service.js
callmehiphop: Yeah, I think that's doable, before this was a provider so I didn't have access to `$location` but since changing it, we could probably just set the actual link to the href now.

callmehiphop:
> Since you have figured out deeplinking (yay!!), can we add a clickable "#" such as here: http://callmehiphop.github.io/gcloud-node-ghpages/#/docs/master/bigquery/dataset next to createTable?

@stephenplusplus We could, but I find it a little confusing since the doc permalinks hide all other content on the page.

stephenplusplus:
> We could, but I find it a little confusing since the doc permalinks hide all other content on the page.

Makes sense. I think it would actually be better if the docs didn't do that... think that's possible? (Can tackle that in another PR, if so)

callmehiphop:
Yeah I think so - let's do that in another PR, I'd like to put this one to bed soon :)

stephenplusplus:
> Currently they are `<h6>` tags, could we just bump them up to `<h4>`s

That should be fine. I think I just made them 6s so they didn't overpower the numeric heading. Here they are as H4s:

screen shot 2015-08-04 at 9 35 06 pm

callmehiphop:
I ended up making them 3s, even 4s kinda looked a bit small. https://github.com/GoogleCloudPlatform/gcloud-common/pull/14

I refactored the deeplinking a bit and in the process I believe I addressed all of your comments. Let me know what you think!

stephenplusplus:
It's gold. Props for taking on this massive PR and putting up with all of my comments along the way :)

callmehiphop:
woo, my pleasure :dancer:

PR #737, opened 2015-07-23
bigquery: fix error callback arguments

callmehiphop:
`Dataset#setMetadata` was not passing the correct arguments in for request failures


stephenplusplus:
Good catch! Could you also show that the apiResponse comes back in the \@example?

callmehiphop:
@stephenplusplus done!

stephenplusplus:
Thanks!

PR #738, opened 2015-07-23
docs: add documentation for callback arguments

### LINKED ISSUE #628: list what arguments go into the callback
I think this is actually a more widely spread problem, where the examples have been saving us, but...

Is there any way we could more broadly document the arguments that go into the callback function?

callmehiphop:
Fixes #628 
- [x] Document all the things!
  - [x] BigQuery
  - [x] Datastore
  - [x] PubSub
  - [x] Search
  - [x] Storage

### DIFF
--- a/lib/bigquery/dataset.js
+++ b/lib/bigquery/dataset.js
@@ -64,6 +64,9 @@ function Dataset(bigQuery, id) {
  *     JSON array of fields, which allows for nested and repeated fields. See
  *     a [Table resource](http://goo.gl/sl8Dmg) for more detailed information.
  * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request
+ * @param {module:bigquery/table} callback.table - The newly created table.
+ * @param {object} callback.apiResponse - The full API response.
  *
  * @example
  * var tableConfig = {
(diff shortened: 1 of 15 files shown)


lib/bigquery/dataset.js
```diff
@@ -64,6 +64,9 @@ function Dataset(bigQuery, id) {
  *     JSON array of fields, which allows for nested and repeated fields. See
  *     a [Table resource](http://goo.gl/sl8Dmg) for more detailed information.
  * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request
+ * @param {module:bigquery/table} callback.table - The newly created table.
```
stephenplusplus: I'm not sure on the right approach here, as far as if this is `nullable` or not. Technically, the callback will get the `table` argument only in non-error cases, so should `{module:bigquery/table}` be `{?module:bigquery/table}`? The only constant, never-null argument is `callback.apiResponse`.

I don't know if that's just going to make things confusing by being technically correct to that detail. Any thoughts?

callmehiphop: Maybe neither should `nullable`, I think anyone familiar with node will understand the convention

stephenplusplus: True, though I think at least `err` should be. I'm cool with leaving the second argument not marked as nullable. I think that would just create a lot of noise on the site.

callmehiphop: I saw nothing :sunglasses:

stephenplusplus:
This is a huge step towards even more docs-awesomeness. Thanks, sir!

callmehiphop:
:dancer:

jgeewax:
Yes-- huge +1 @callmehiphop  :-D

PR #741, opened 2015-07-24
Pass apiResponse along with errors

stephenplusplus:
As introduced in https://github.com/GoogleCloudPlatform/gcloud-node/pull/706#discussion_r35119766, this fixes a couple of things:
1. **"Decorate" request options in all cases.**
   
   Previously, we were decorating requests only when we were making them automatically for the user. That means our UA header wasn't on the `authorizedReqOpts` that were passed to "onAuthorized(function (authorizedReqOpts) {})". So, this just decorates requests sooner so that anytime we say we "authorized a request", we also put on the UA string and cleaned up gcloud-node-only properties that may have leaked.
2. **Pass back apiResponse when there is an error.**
   
   All of the callbacks our API accepts are advertised as receiving an "apiResponse" even in the event of an error. The problem was our universal response handler ("util.handleResp") was not giving back the response body if an error was caught, which is what becomes the "apiResponse". So now, it works as advertised.

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -182,19 +182,9 @@ nodeutil.inherits(ApiError, Error);
 function handleResp(err, resp, body, callback) {
   callback = callback || noop;
 
-  if (err) {
-    callback(err);
-    return;
-  }
+  var parsedResp = util.parseApiResp(err, resp, body);
 
-  var parsedApiResponse = util.parseApiResp(resp, body);
-
-  if (parsedApiResponse.err) {
-    callback(parsedApiResponse.err);
-    return;
-  }
-
-  callback(null, parsedApiResponse.body, parsedApiResponse.resp);
+  callback(parsedResp.err, parsedResp.body, parsedResp.resp);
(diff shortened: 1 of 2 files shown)


callmehiphop:
This might sound crazy, but should we go through the system-tests and add assertions to test that the proper arguments are being passed in?

stephenplusplus:
Not crazy! The argument passing is thoroughly tested in the unit tests. We ran into this problem because our unit tests are so decoupled that we stub what `handleResp` is supposed to return. So, we have validation that "if there is a response body provided, it is sent to the user's callback". The issue was just without our stubs, `handleResp` was only actually returning one argument if there was an error (the error, not the response). The user's callback would receive `undefined` for `apiResp`.

The leak was in `handleResp`, where we weren't testing properly what was happening when there was an error. Now that it always returns all of the arguments (?error, body, response), and we have it tested, we can safely call this fixed.

callmehiphop:
:+1: thanks!

PR #742, opened 2015-07-25
remove topic.options.autoCreate, add topic.getMetadata

### LINKED ISSUE #696: auto-created topic and subscribe do not work together
``topic.subscribe()`` on a new topic fails. See this code:

```
var topic = pubsub.topic('my-topic');  // my-topic hasn't been created yet.
topic.subscribe('my-sub', function(err, _){ console.log(err); }
```
Will fail with a ``404: Resource not found (resource=my-topic)``,

However, this code works:

```
var topic = pubsub.topic('my-topic'); 
topic.publish({'dummy': 'message'});
topic.subscribe('my-sub', function(err, _){ console.log(err); }
```

Seems ``subscribe`` doesn't call ``createTopic`` if it doesn't exist.

stephenplusplus:
Fixes #696 
Related https://github.com/GoogleCloudPlatform/gcloud-python/issues/949

This:
- removes  `topic.options.autoCreate`
- adds `topic.getMetadata()`
- includes small refactors now possible without the overhead of supporting `autoCreate`

### DIFF
--- a/lib/pubsub/index.js
+++ b/lib/pubsub/index.js
@@ -174,9 +174,9 @@ PubSub.prototype.getTopics = function(query, callback) {
       return;
     }
     var topics = (result.topics || []).map(function(item) {
-      return new Topic(self, {
-        name: item.name
-      });
+      var topicInstance = self.topic(item.name);
+      topicInstance.metadata = item;
+      return topicInstance;
     });
     var nextQuery = null;
     if (result.nextPageToken) {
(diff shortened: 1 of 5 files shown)


stephenplusplus:
Assigned for a review :)

lib/pubsub/index.js
```diff
@@ -199,16 +199,10 @@ PubSub.prototype.getTopics = function(query, callback) {
  *
  * @example
  * pubsub.createTopic('my-new-topic', function(err, topic, apiResponse) {
- *   topic.publish({
- *     data: 'New message!'
- *   }, function(err) {});
```
callmehiphop: Out of curiosity, why are we removing this example? I understand the verbiage below, but this snippet still seems relevant.

stephenplusplus: Publishing a message to a topic you created will likely have no subscriptions registered yet, which results in [the message being lost](https://github.com/GoogleCloudPlatform/gcloud-node/issues/696#issuecomment-116835719). I figured it's better to not even suggest that.

callmehiphop: Gotcha!

callmehiphop:
Looks great, thanks!

PR #744, opened 2015-07-25
only show inner methods when class is active

### LINKED ISSUE #675: Make left-side navigation collapsable?
Some people are having trouble finding the method listings at the top level when the overviews exist, so my first thought was to push the overviews lower on the page so the method listing was there, however that doesn't really fix the problem.

Not to mention, as we add support for more services, we're going to need that left-nav to be able to collapse and only show what's interesting...

Any ideas on how we could make the left-side navigation:

1. Show all the services
1. Expand when you click into one of the services
1. Expand to show the different methods when you click on one of the types

?

Some examples of navigation:

- Angular: https://angular.io/docs/js/latest/api/change_detection/AstTransformer-class.html
- Google Cloud: https://cloud.google.com/storage/docs/json_api/v1/buckets/get
- ArrangoDB: https://docs.arangodb.com/HttpEdge/AddressAndEtag.html
- MongoDB: https://docs.mongodb.org/manual/tutorial/backup-with-filesystem-snapshots/

stephenplusplus:
Fixes #675

This will hide the sub-classes of a parent class (BigQuery, Datastore, etc) until it is active.

Preview: http://stephenplusplus.github.io/gcloud-node/#/docs/

### DIFF
--- a/docs/site/components/subpage/subpage.html
+++ b/docs/site/components/subpage/subpage.html
@@ -84,7 +84,7 @@
           ng-href="{{page.url}}">
           {{page.title}}
         </a>
-        <ul class="sub-sections" ng-if="page.pages">
+        <ul class="sub-sections" ng-if="page.pages && isActiveClass(page)">
           <li ng-repeat="innerPage in page.pages">
             <a
               ng-class="{ current: isActiveUrl(page.url + innerPage.url) }"
--- a/docs/site/components/subpage/subpage-directive.js
+++ b/docs/site/components/subpage/subpage-directive.js
@@ -20,10 +20,13 @@ angular
         };
 
         if ($scope.versionSelected === 'master') {
-          $http.get('https://api.github.com/repos/GoogleCloudPlatform/gcloud-node/commits?sha=gh-pages&per_page=1')
-            .then(function(resp) {
-              $scope.lastBuiltDate = moment(resp.data[0].commit.committer.date);
-            });
+          $http({
+            method: 'get',
+            url: 'https://api.github.com/repos/GoogleCloudPlatform/gcloud-node/commits?sha=gh-pages&per_page=1',
+            cache: true
+          }).then(function(resp) {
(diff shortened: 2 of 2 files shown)


docs/site/components/subpage/subpage.html
```diff
@@ -84,7 +84,7 @@ <h4 class="list-item--heading">API</h4>
           ng-href="{{page.url}}">
           {{page.title}}
         </a>
-        <ul class="sub-sections" ng-if="page.pages">
+        <ul class="sub-sections" ng-if="page.pages" ng-show="isActiveClass(page)">
```
callmehiphop: would it make more sense to just combine them both within the `ng-if`?

stephenplusplus: Hmm, I think it would!

callmehiphop:
Do we want to add an arrow or something to indicate that the items are expandable?

stephenplusplus:
Only using the `ng-if` now, and I also started caching the last built date we were pulling from GH constantly. This stops the page from skipping on every new page load, and should help us from burning out the API rate limit as well.

stephenplusplus:
> Do we want to add an arrow or something to indicate that the items are expandable?

I think this would send the wrong message, UX wise. With an arrow, the impression is a menu will be expanded, but another page won't load. To go the expand/collapse route, it would force these :( changes:

  1. Add a new menu item under each class, likely with the same title:
    - gcloud
    - ▿ BigQuery
      - BigQuery
      - Dataset
      - Job
      - Table
    - ▹ Datastore
    - ▹ PubSub
    - ▹ Search
    - ▹ Storage
  1. It would take 2 clicks to get to the place that currently only takes one.

stephenplusplus:
@jgeewax let me know if this is what you had in mind: http://stephenplusplus.github.io/gcloud-node/#/docs/v0.16.0

jgeewax:
Yep - looked at it the other day. This is super awesome.

/cc @blowmage @dhermes

callmehiphop:
:+1:

PR #749, opened 2015-07-27
docs: added local version of angular-highlightjs

callmehiphop:
It looks like angular-hightlightjs recently [added UMD support](https://github.com/pc035860/angular-highlightjs/commit/787bacd3fa0a3f17da78580dc046da36f52727de) which broke our doc examples since we loaded the directive before the underlying library.

This PR changed the loading order to fix the broken examples and I've also added the directive locally to prevent any future conflicts.


stephenplusplus:
Nice. Good catch. Thanks!

stephenplusplus:
Do you think we should cherry pick this into our gh-pages branch manually so that our docs aren't broken until our Wednesday release?

callmehiphop:
Definitely!

PR #754, opened 2015-07-28
pubsub: fixed subscription pulling bug

callmehiphop:
Every time some one listens for a message event on a `Subscription`, the subscription instance itself begins pulling from the server regardless of whether or not a pulling loop is already in progress.

e.g.

``` javascript
subscription.on('message', function() { ... }); // pulling loop starts
subscription.on('message', function() { ... }); // second pulling loop starts
```

This change aims to only start a pulling loop if one doesn't previously exist.

### DIFF
--- a/test/pubsub/subscription.js
+++ b/test/pubsub/subscription.js
@@ -75,8 +75,8 @@ describe('Subscription', function() {
       assert.strictEqual(sub.maxInProgress, 3);
     });
 
-    it('should not be closed', function() {
-      assert.strictEqual(subscription.closed, false);
+    it('should be closed', function() {
+      assert.strictEqual(subscription.closed, true);
     });
 
     it('should default autoAck to false if not specified', function() {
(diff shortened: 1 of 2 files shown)


test/pubsub/subscription.js
```diff
@@ -145,6 +145,17 @@ describe('Subscription', function() {
       assert.strictEqual(subscription.messageListeners, 0);
     });
 
+    it('should only run a single pulling loop', function() {
+      var stub = subscription.startPulling_ = function() {
+        stub.callCount = (stub.callCount + 1) || 1;
+      };
```
stephenplusplus: Could this be made more readable?

```js
var startPullingCalledTimes = 0;

subscription.startPulling_ = function() {
  startPullingCalledTimes++;
};

// ...

assert.strictEqual(startPullingCalledTimes, 1);
```

stephenplusplus:
Awesome catch, and simple solution. Perfect! Just a comment on the test style.

test/pubsub/subscription.js
stephenplusplus: Maybe better... `numTimesStartedPulling`, `numPulls`, ...

callmehiphop:
Sure, does `startPullingCallCount` sound good?

stephenplusplus:
Yep, looks good. Thanks!

PR #757, opened 2015-07-29
pubsub: support auto-named subscriptions.

### LINKED ISSUE #748: Don't require a name for a subscription
The [Pub/Sub documentation](https://cloud.google.com/pubsub/reference/rest/v1/projects.subscriptions/create) states:

> If the name is not provided in the request, the server will assign a random name for this subscription on the same project as the topic.

\- https://github.com/GoogleCloudPlatform/gcloud-ruby/issues/193 / @blowmage

--

So, let's not require it!

stephenplusplus:
Fixes #748 
Related https://github.com/GoogleCloudPlatform/gcloud-ruby/issues/193

Currently, the JSON API doesn't create subscriptions unless a name is given. However, [the docs](https://cloud.google.com/pubsub/reference/rest/v1/projects.subscriptions/create) suggest they do. This is a PR prepared for when the Pub/Sub auto-named subscription feature is supported.

### DIFF
--- a/lib/pubsub/index.js
+++ b/lib/pubsub/index.js
@@ -20,6 +20,8 @@
 
 'use strict';
 
+var extend = require('extend');
+
 /**
  * @type {module:pubsub/subscription}
  * @private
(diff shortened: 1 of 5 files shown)


stephenplusplus:
We can revisit this if we get some clarity around https://github.com/GoogleCloudPlatform/gcloud-ruby/issues/193.

PR #759, opened 2015-07-30
tests: pubsub: fix flakiness

stephenplusplus:
We've had issues with Pub/Sub tests flaking out, and today we ran into them again: https://travis-ci.org/GoogleCloudPlatform/gcloud-node/builds/73278785#L3237

What appears to be happening is when we pull for a message with returnImmediately:false, no messages are ever returned, and the test times out.

History:
#332
#377

The way it was working until today seemed cleaner, which was basically:

``` js
publishMessage(function() {
  pullMessage(function() {
    // run tests about the message here
  })
})
```

We used to run them like this:

``` js
pullMessage(function() {
  // long-running until a message returns
})

publishMessage() x exactly 6
```

This PR runs them like this:

``` js
beforeTestsRun(function() {
  publishMessage() x 10
})

test(function() {
  pullMessage(function() {
    // run tests about the message here
  })
})

test(function() {
  pullMessage(function() {
    // run tests about the message here
  })
})

// etc...
```

I played with the ordering of publishes and pulls, with returnImmediately to true and false, with different amounts of messages (anywhere from 1-10), with publishing multiple messages in one API call vs many, and this was the only combination of all of the above that I could find that worked.

@tmatsuo -- your wisdom is very welcome :)

### DIFF
--- a/system-test/pubsub.js
+++ b/system-test/pubsub.js
@@ -156,15 +156,24 @@ describe('pubsub', function() {
     var topic;
 
     before(function(done) {
-      // Create a new test topic.
       pubsub.createTopic(TOPIC_NAME, function(err, newTopic) {
         assert.ifError(err);
         topic = newTopic;
 
-        // Create subscriptions.
-        async.parallel(SUBSCRIPTIONS.map(function(sub) {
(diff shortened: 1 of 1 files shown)


callmehiphop:
Looks good to me!

PR #760, opened 2015-07-30
tests: increase common/util coverage

### LINKED ISSUE #747: common/util test coverage is lacking
Could use a sweep to make sure all functions are covered.

Should wait for #746 before tackling :football:.

callmehiphop:
Fixes #747 

I would appreciate any feedback -- especially around `makeWritableStream`.

### DIFF
--- a/test/common/util.js
+++ b/test/common/util.js
@@ -132,6 +132,33 @@ describe('common/util', function() {
     });
   });
 
+  describe('format', function() {
+    it('should replace denoated keys with corresponding values', function() {
+      var formatted = util.format('{greeting} {thing}!', {
+        greeting: 'Hello',
+        thing: 'world'
+      });
+
+      assert.strictEqual(formatted, 'Hello world!');
+    });
+
+    it('should leave any unsatisfied keys unchanged', function() {
+      var formatted = util.format('{greeting} {thing}!', {
+        greeting: 'swamp'
+      });
+
+      assert.strictEqual(formatted, 'swamp {thing}!');
+    });
+
+    it('should inject falsy values', function() {
+      var formatted = util.format('{num}', {
+        num: 0
+      });
+
+      assert.strictEqual(formatted, '0');
+    });
+  });
+
   describe('ApiError', function() {
     it('should build correct ApiError', function() {
       var error = {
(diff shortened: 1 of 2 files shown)


test/common/util.js
```diff
@@ -132,6 +132,25 @@ describe('common/util', function() {
...
+    });
+
+    it('should leave any unsatisfied keys unchanged', function() {
+      var formatted = util.format('{greeting} {thing}!', {
+        greeting: 'swamp'
+      });
+
+      assert.strictEqual(formatted, 'swamp {thing}!');
+    });
+  });
```
stephenplusplus: Can you add a test that a falsy value is still injected? E.g. ```util.format('{num}', { num: 0 }) === '0'```

test/common/util.js
```diff
@@ -420,9 +439,82 @@ describe('common/util', function() {
...
+
+      util.makeWritableStream(dup, options);
+
+      dup.on('error', function(err) {
+        assert.strictEqual(err.code, fakeResponse.statusCode);
+        done();
+      });
+
+      setImmediate(function() {
+        fakeStream.emit('complete', fakeResponse);
```
stephenplusplus: This will trigger this:

```js
stream.on('complete', function(res) {
  handleResp(null, res, res.body, function(err, data) {
    if (err) {
      dup.emit('error', err);
      dup.end();
      return;
    }
    onComplete(data);
  });
});
```

Your tests cover both of these cases well, but I would change the approach a little bit. Instead of letting `handleResp` actually run, make sure it's being called in the first place:

```diff
stream.on('complete', function(res) {
-  handleResp(null, res, res.body, function(err, data) {
+  util.handleResp(null, res, res.body, function(err, data) {
```

Then use `utilOverrides` to execute the callback with the data that will cause the error handling / success handling.

stephenplusplus:
Looks awesome, thanks!

test/common/util.js
callmehiphop: Awesome suggestion! I felt like the setup was a little verbose.

test/common/util.js
callmehiphop: done!

test/common/util.js
callmehiphop: Made a few tweaks to these ones, let me know what your thoughts are!

test/common/util.js
```diff
@@ -420,9 +447,85 @@ describe('common/util', function() {
...
+
+    it('should pass back the response data to the callback', function(done) {
+      var dup = duplexify();
+      var fakeStream = new stream.Writable();
+      var fakeResponse = {};
+
+      fakeStream.write = function() {};
+
+      utilOverrides.handleResp = function(err, res, body, callback) {
+        callback(null, fakeResponse)
```
stephenplusplus: `;` :open_mouth:

callmehiphop: I'm not worthy!

stephenplusplus:
You *are* worthy!

PR #762, opened 2015-07-30
core: use is-stream-ended

stephenplusplus:
Just cleaning up some of our in-house logic. This uses https://github.com/stephenplusplus/is-stream-ended.


callmehiphop:
Looks good, thanks!

PR #771, opened 2015-07-31
core: allow more time for system tests to run

stephenplusplus:
Related to the flaking system tests, this allows more time for them to run (30s -> 45s). We do have more complex processes we run before/after/during tests in attempt to create and then clean up isolated objects. This naturally takes longer.

Running locally on a pretty good connection, I very rarely hit timeout errors. I can't speak for Travis, though, as he seems to be on dial-up. :phone:


callmehiphop:
:+1:

PR #776, opened 2015-07-31
storage: file.createReadStream: only send data when non-error

### LINKED ISSUE #772: Reading a nonexisting file succeeds
I tried to open an nonexisting file. It succeeded and gave me the following contents:

```
<?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message></Error>
```

I'd prefer the function threw an exception of called 'error'.

My code
```
  var content = '';
  var readStream = bucket.file(data.name).createReadStream();
  readStream.on('data', function(buf) { content += buf.toString(); });
  readStream.on('end', function() {
    console.log("got file data: " + content);
  });
  readStream.on('error', function(err) {
    console.log("Error opening a gcs file " + data.name + ": " + err)
  });
```

output:
got file data: <?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message></Error>

EDIT: I actually also got an error message, see below.

stephenplusplus:
Fixes #772 

When we make an HTTP request, we immediately get back the response headers. This PR uses those to detect an error via the status code. Previously, we detected the error _after_ the data had been consumed and pushed to the user's stream.

This affected file.createReadStream(), when we would `GET https://storage.googleapis.com/_bucket_/_object_` a file that didn't exist.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -536,7 +536,15 @@ File.prototype.createReadStream = function(options) {
     requestStream
       .on('error', endThroughStream)
 
-      .on('response', throughStream.emit.bind(throughStream, 'response'))
+      .on('response', function(incomingMessage) {
+        throughStream.emit('response', incomingMessage);
+
+        util.handleResp(null, incomingMessage, null, function(err) {
+          if (err) {
+            requestStream.unpipe(throughStream);
+          }
+        });
+      })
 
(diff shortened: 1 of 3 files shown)


callmehiphop:
Gracias!

PR #777, opened 2015-08-01
core: use split-array-stream

stephenplusplus:
[split-array-stream](http://gitnpm.com/split-array-stream) makes it a little easier to handle pushing results from an array to a stream.

This also simplifies the logic in `DatastoreRequest.get` to always use stream mode behind the scenes, even if the user provides a callback.


callmehiphop:
:+1:

PR #784, opened 2015-08-06
system tests: search: delete documents individually

stephenplusplus:
RE: https://github.com/GoogleCloudPlatform/gcloud-node/issues/775#issuecomment-128358680

The Search API doesn't allow indexes to be deleted. Instead, you can delete all of the documents within them. That's what we were doing previously, but as the amount of created Documents grows, it's taking an excessive amount of time to fetch, then iteratively delete each one that exists.

This PR changes the tests to create/delete a document, instead of letting the before/after hooks handle it the old way.

### DIFF
--- a/system-test/search.js
+++ b/system-test/search.js
@@ -34,24 +34,13 @@ function deleteDocument(document, callback) {
 }
 
 function deleteIndexContents(index, callback) {
-  index.getDocuments(function(err, documents) {
+  index.getDocuments({ view: 'ID_ONLY' }, function(err, documents) {
     if (err) {
       callback(err);
       return;
     }
 
     async.eachLimit(documents, MAX_PARALLEL, deleteDocument, callback);
-  });
-}
-
-function deleteAllDocuments(callback) {
-  search.getIndexes(function(err, indexes) {
-    if (err) {
-      callback(err);
-      return;
-    }
-
-    async.eachLimit(indexes, MAX_PARALLEL, deleteIndexContents, callback);
   });
 }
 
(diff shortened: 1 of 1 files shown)


jgeewax:
That's backwards right ? You can't currently delete indexes, and you
accomplish the same result by deleting all of the documents in the index.
Right ?

stephenplusplus:
Yep, thanks. Will fix that in the first post.

system-test/search.js
```diff
@@ -67,12 +56,8 @@ describe('Search', function() {
   var INDEX_NAME = generateIndexName();
   var index = search.index(INDEX_NAME);
 
-  before(function(done) {
-    deleteAllDocuments(done);
```
callmehiphop: Please forgive my lack of familiarity with the search api - if something like a test timeout occurs on a previous Travis run, would there still be lingering documents?

stephenplusplus: Yes, if a test that created a document timed out and the rest of our test suite finished before that test had a chance to delete it, it would still exist. Those documents would be boxed inside indexes that are created once per `npm run system-test`.

callmehiphop:
Travis will be so happy! (I hope!)

quartzmo:
@stephenplusplus We are trying to follow your work on Search, as we're about to start on it for the Ruby lib. Do you think you could create a `search` label and add it to issues like this one?

stephenplusplus:
@quartzmo sure, I found a few: https://github.com/GoogleCloudPlatform/gcloud-node/issues?q=label%3Asearch+is%3Aclosed I don't think we've had that many. Let me know if any questions come up.

PR #786, opened 2015-08-06
system tests: remove duplicate tests

stephenplusplus:
We had a lot of duplicated tests from when we introduced autoPaginate. Since we ended up defaulting `autoPaginate` to true, these aren't really necessary. Instead, we could test what happens with "autoPaginate: false", but we have that covered well in the unit tests, and use it in various spots in the system tests where we need to.


callmehiphop:
:+1:

PR #790, opened 2015-08-10
docs: upstream links

### LINKED ISSUE #740: Link to upstream JSON API from our docs
Introduced in https://github.com/GoogleCloudPlatform/gcloud-node/pull/721#discussion_r35438977

Our APIs will generally try to make the upstream API a little more user-friendly by using more logical names or trimming some of the verbosity. And then we document the things we handle (things like `maxResults`, `pageSize`, `pageToken`, etc). But, most of our APIs are also a pass-through to the upstream API, which will have many more options that we don't document.

I don't think we should document everything, as there's already a place for that. And I don't think we have to limit the user's input to include only the things "we approve". We should just link to the upstream API's docs to show the full list of options and their official descriptions.

Example:

Link https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.16.0/bigquery/dataset?method=createTable to https://cloud.google.com/bigquery/docs/reference/v2/tables/insert

callmehiphop:
Closes #740

Added links for all upstream counterparts.

[demo](http://callmehiphop.github.io/gcloud-node/#/docs/master/bigquery/dataset)

### DIFF
--- a/lib/bigquery/dataset.js
+++ b/lib/bigquery/dataset.js
@@ -56,6 +56,8 @@ function Dataset(bigQuery, id) {
 /**
  * Create a table given a tableId or configuration object.
  *
+ * @resource [Tables: insert API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/v2/tables/insert}
+ *
  * @param {object} options - Table id or configuration object.
  * @param {string} options.id - The id of the table.
  * @param {string|object} options.schema - A comma-separated list of name:type
@@ -117,6 +119,8 @@ Dataset.prototype.createTable = function(options, callback) {
 /**
  * Delete the dataset.
  *
+ * @resource [Datasets: delete API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/v2/datasets/delete}
+ *
  * @param {object=} options - The configuration object.
  * @param {boolean} options.force - Force delete dataset and all tables.
  *     Default: false.
@@ -151,6 +155,8 @@ Dataset.prototype.delete = function(options, callback) {
 /**
  * Get the metadata for the Dataset.
  *
+ * @resource [Datasets: get API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/v2/datasets/get}
+ *
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request
  * @param {object} callback.metadata - The Dataset metadata
@@ -175,6 +181,8 @@ Dataset.prototype.getMetadata = function(callback) {
 
 /**
  * Get a list of tables.
+ *
+ * @resource [Tables: list API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/v2/tables/list}
  *
  * @param {object=} query - Configuration object.
  * @param {number} query.maxResults - Maximum number of results to return.
@@ -270,6 +278,8 @@ Dataset.prototype.query = function(options, callback) {
 
 /**
  * Sets the metadata of the Dataset object.
+ *
+ * @resource [Datasets: patch API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/v2/datasets/patch}
  *
  * @param {object} metadata - Metadata to save on the Dataset.
  * @param {function} callback - The callback function.
--- a/.jscsrc
+++ b/.jscsrc
@@ -14,5 +14,9 @@
     "try",
     "typeof"
   ],
-  "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties"
+  "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties",
+  "maximumLineLength": {
+    "value": 80,
+    "allowUrlComments": true
+  }
 }
--- a/.jshintrc
+++ b/.jshintrc
@@ -7,7 +7,6 @@
   "immed": true,
   "indent": 2,
   "latedef": "nofunc",
-  "maxlen": 80,
   "newcap": true,
   "node": true,
   "noarg": true,
--- a/docs/site/components/docs/docs-directives.js
+++ b/docs/site/components/docs/docs-directives.js
@@ -26,7 +26,7 @@ angular.module('gcloud.docs')
     }
 
     return {
-      template: '<a href="{{link}}">(code on GitHub)</a>',
+      template: '<a href="{{link}}">Source Code</a>',
 
       link: function($scope) {
         var method = $scope.method;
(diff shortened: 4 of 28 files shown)


lib/bigquery/dataset.js
```diff
@@ -56,6 +56,9 @@ function Dataset(bigQuery, id) {
 /**
  * Create a table given a tableId or configuration object.
  *
+ * See [Tables: insert](https://cloud.google.com/bigquery/docs/reference/v2/tables/insert)
```
stephenplusplus: I use [goo.gl](https://goo.gl) to get shorter links.

callmehiphop: So initially I started to go that route, but I think if we were to upgrade the api version it would be pretty annoying to regenerate all the short urls. I think by including the full path, it'll be easier to update in the future - e.g. `v1` => `v2'

stephenplusplus: That's a good point, but we still have to stay within the 80 char limit, which is pretty impossible with these URLs.

callmehiphop: Yeah, still trying to figure that part out!

callmehiphop: Would it be ok to break that rule specifically for these?

stephenplusplus: How would that look? Our test scripts would break, wouldn't they?

callmehiphop: Yeah, I'm looking for an option to ignore a comment blocks line length or something similar. If there isn't anything like that I'll just generate short links for all of them

callmehiphop: http://jscs.info/rule/maximumLineLength.html - they have a `urlComments` flag to ignore long urls within comments!

stephenplusplus: Wow, that's a microsetting if I've ever seen one. What about JSHint? Do we turn off the rule on JSHint and let JSCS catch breakages?

callmehiphop: That's what I'm thinking, it looks like JSCS was already doing line length checks, so it seems silly to do it twice.

callmehiphop:
@stephenplusplus so I kinda think the overall appearance of the links could be better, what are your feelings on doing something like this..

```js
/**
 * <div class="notice">See [Tables: insert](https://cloud.google.com/bigquery/docs/reference/v2/tables/insert)
 * for more detailed information.</div>
 */
```

stephenplusplus:
Could you post a screenshot of what that renders like?

callmehiphop:
http://i.imgur.com/4QADlNc.png

Initially I tried use blockquote markdown syntax (which works) but there isn't any styling associated with it. Alternatively I could write some CSS for that and we could go that route.

stephenplusplus:
That looks cool to me. Semantics: would just a manual a tag be better?

callmehiphop:
> That looks cool to me. Semantics: would just a manual a tag be better?

Instead of the markdown link syntax? `[text](http://link.com)`?

stephenplusplus:
Yeah, I'm thinking since we have to add a class, we don't need the extra
div wrapper if we just write out the a.

callmehiphop:
Well, if I understand correctly, that's going to turn the entire sentence into a link? I'm probably splitting hairs, but I think that would be kind of ugly. We could just write some styles to make the blockquote look the same as the notice class and go full out markdown.

stephenplusplus:
Nope you're right. I'm just getting myself confused. Sorry!

For the text, wdyt about just "[Official API documentation]()"

And if it's just the one line, we can use the a tag with a class, unless you think that's a bad idea.

\<a href="..." class="notice">
Official API documentation.
\</a>

stephenplusplus:
It looks awesome! Moving the link to a jsdoc property makes it much easier to make changes now without having to go through all of the docs.

I would prefer the links move to the top, as we want to get across "see the API docs for more information about our parameters". Here's an example:

screen shot 2015-08-10 at 8 40 15 pm

Can you think of another way to make it clear that the official docs have more information on the properties we accept?

Also note, I changed "Code on GitHub" to "Source Code". Small change, but "Code on GitHub" just reads like a bad sentence to me. "Source Code" is direct and concise!

callmehiphop:
> I would prefer the links move to the top, as we want to get across "see the API docs for more information about our parameters".

Done!

> Can you think of another way to make it clear that the official docs have more information on the properties we accept?

I think maybe just make a note within the object description? e.g. - *Configuration object. For a complete list of properties please refer to the Official API Documentation*. 

It's also possible to give each resource link a custom title from JSDoc, so we could alter the link title to convey this as well. Perhaps instead of *Official API Documentation* you could say *Complete List of Configuration Options* or something..

> Also note, I changed "Code on GitHub" to "Source Code". Small change, but "Code on GitHub" just reads like a bad sentence to me. "Source Code" is direct and concise!

I can dig it.

stephenplusplus:
LGTM.

@jgeewax wdyt? Did we make things better or regress?

jgeewax:
So it's definitely a step forward, however... I have to say I liked the "See: Tables.insert [ -> ] for more information" since it ties it to the specific method (aka, I know I'm not being dumped off to some marketing landing page somewhere that's ridiculous...)

Is it too much work to do that? (Fully understand if so.)

Ideally, text might be something like...

"See: Tables.insert [ -> ] API method for more information"

callmehiphop:
> Is it too much work to do that?

It definitely isn't!

@stephenplusplus thought it would be nice to move the source code link into the same block and while doing that I also found links for concepts/references/etc. scattered across the api so I decided to consolidate them all. (e.g. http://callmehiphop.github.io/gcloud-node/#/docs/master/storage/bucket?method=acl)

Is that something you like and would want to keep? If so, I think maybe the verbiage "See: Tables.insert for more information" might sound kinda funny in a list of links opposed to something like "Tables.insert REST API Documentation"

Thoughts?

jgeewax:
Gotcha

Phrased as a list of links, definitely "Tables.insert API documentation (official)" or similar makes sense

Is it crazy to move the resources down as well? Logically they make sense above, but when I'm looking at docs, examples are likely going to be ... far more important than the link to the API docs... 

Maybe call it "More information" and put it at the bottom?

callmehiphop:
Personally I think it makes sense - a user is probably only going to want to see the full documentation if the examples don't provide them with the information they need. So having the links below the examples is probably more convenient.

However, I think the whole reason @stephenplusplus opened an issue about this is because we don't document all the available options that a user can pass in for some methods. Having the links right above the options as well as a comment in the `options` description telling the user to visit the official API documentation kinda solved that.

stephenplusplus:
How about this or something similar?

screen shot 2015-08-12 at 9 28 11 pm

callmehiphop:
Updated the link titles & placement and pushed, let me know what you guys think!

stephenplusplus:
Awesome. Just. Awesome.

callmehiphop:
:dancer:

PR #792, opened 2015-08-12
Support Google Compute Engine

stephenplusplus:
Continuing from #721 

This PR puts a comb over the great work @mziccard gave us when he implemented support for Compute Engine. That includes various small documentation tweaks, code style things, unit tests, and system tests.
#### To Dos
- [x] Documentation
  - [x] Callback parameters
  - [x] Upstream API links (depends on #790)
    - [x] Check for misspellings of "@"resource or improper {"@"link} tags
    - [x] Confirm using `v1` and not `latest`
- [x] Tests
  - [x] Unit
    - [x] Address
    - [x] Disk
    - [x] Firewall
    - [x] Index
    - [x] Network
    - [x] Operation
    - [x] Region
    - [x] Snapshot
    - [x] VM
    - [x] Zone
  - [x] System

Changed since #721:
- Added an [`operation.onComplete` method](https://stephenplusplus.github.io/gcloud-node/#/docs/master/compute/operation?method=onComplete). Credit to Marco for the idea -- we couldn't think of a way to implement it un-confusingly at the time, so I hope this is a good solution.
- Removed all throwing where we were validating what the API expects. We typically let the API request go through and have _its_ error clarify what was incorrect about the input. We only throw when _our_ code cannot proceed.
- Added sane defaults in `zone.createVM`. Explained [here](https://stephenplusplus.github.io/gcloud-node/#/docs/master/compute/zone?method=createInstance). That necessitated two new modules: [gce-images](https://github.com/stephenplusplus/gce-images) and to handle auth the same as gcloud does, [google-auto-auth](https://github.com/stephenplusplus/google-auto-auth) (this one can eventually be used in gcloud directly).
- Renamed `Instance` to `VM`: `zone.createVM`, `zone.getVMs`, etc.
- Added `compute.getZones()` and `compute.getRegions()`

Docs preview: https://stephenplusplus.github.io/gcloud-node/#/docs/master/compute


README.md
```diff
@@ -113,6 +114,43 @@ job.getQueryResults().on('data', function(row) {});
 ```
 
 
+## Google Compute Engine
+
+- [API Documentation][gcloud-compute-docs]
+- [Official Documentation][cloud-compute-docs]
```
callmehiphop: Might want to watch out for this https://github.com/GoogleCloudPlatform/gcloud-node/pull/790

callmehiphop: Derp, this is a markdown file :dancer:

lib/compute/address.js
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @private
+ *
+ * @param {string} method - Action.
+ * @param {string} path - Request path.
+ * @param {*} query - Request query object.
+ * @param {*} body - Request body contents.
+ * @param {function} callback - The callback function.
+ */
+Address.prototype.makeReq_ = function(method, path, query, body, callback) {
```
callmehiphop: Any reason we don't take a similar approach to https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/dns/zone.js#L71

Seems like we're pretty inconsistent about creating this method.

lib/compute/index.js
```diff
@@ -0,0 +1,1053 @@
...
+      return acc;
+    }, []);
+
+    callback(null, addresses, nextQuery, resp);
+  });
+};
+
+/**
+ * Get a list of disks. For a detailed description of method's options see
+ * [API reference](https://goo.gl/M9Qjb3).
```
callmehiphop: For these ones you really will have to watch out for https://github.com/GoogleCloudPlatform/gcloud-node/pull/790

lib/compute/index.js
```diff
@@ -0,0 +1,1053 @@
...
+ * function callback(err, firewall, operation, apiResponse) {
+ *   // `firewall` is a Firewall object.
+ *
+ *   // `operation` is an Operation object that can be used to check the status
+ *   // of the request.
+ * }
+ *
+ * compute.createFirewall('new-firewall-name', config, callback);
+ */
+Compute.prototype.createFirewall = function(name, config, callback) {
```
callmehiphop: Again, it seems like there's some inconsistency across the API around this.. I think more often than not we overload our methods, but occasionally (like here) we do not. Any reason we don't do something like..

```javascript
Compute.prototype.createFirewall = function(config, callback) {
  if (is.string(config)) {
    config = { name: config };
  }
};
```

stephenplusplus: Well, the issue here is we can't create a firewall from just a name, so the config argument must always be provided. I used this approach to try to get around throwing, but I think it would be better in this case to throw if we don't get the arguments we need (a string name and an object config). WDYT?

lib/compute/address.js
stephenplusplus: We have to use a different path prefix here.

lib/compute/index.js
stephenplusplus: Oh and this is `(name, config, callback)` instead of `(config, callback)` just to feel more consistent with our API, eg `storage.createBucket(name, options, callback)`

callmehiphop: Yeah, that makes sense, giving config an empty object for a default is what threw me off.

lib/compute/address.js
callmehiphop: I get that, I guess I'm just pointing out that some places we make a new `makeReq_` method and others we build the path within the method

e.g. https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/dns/zone.js#L160

I'm probably splitting hairs here, so feel free to ignore me :)

stephenplusplus: Oh, I see. I should have used a separate method in DNS. It's easier to follow along with if you have a method assigned in the same file. You don't usually expect to find a method assigned in a constructor. Just a style thing, though. Would be nice to be consistent, but in this case, DNS would have to change.

callmehiphop: Gotcha, I was just digging to find out what our "preferred" approach was! (And now I know!)

stephenplusplus:
Anyone interested, check out the system tests for an actual working sample of code!

stephenplusplus:
@callmehiphop if you want to, take a look at the system tests (and the code) if you catch anything weird. It's about time to lock it down with unit tests.

Documentation will be a work in progress all the way up until the end.

callmehiphop:
Sweet, I'll run through those this afternoon :D

stephenplusplus:
Just for crosslinkability: https://groups.google.com/forum/#!topic/gce-discussion/mYwoVHa-mdQ

stephenplusplus:
@mziccard we're getting close! I plan to include two commits when the PR merges, your original one and then all of our tests and such smashed into one more on top of it. Our CI bot is asking that you give us some type of :ok_hand:

mziccard:
You have my permission to merge the commits :)

stephenplusplus:
Phew!

stephenplusplus:
Thanks everyone! This is now one of my favorites :)

PR #795, opened 2015-08-12
Test index.js file

stephenplusplus:
When a user gets a localized gcloud instance:

``` js
var gcloud = require('gcloud')(/*…*/);
```

They have been missing out on subclass static properties:

``` js
gcloud.storage.acl // undefined
```

This fixes that and adds tests.

We should do a patch release after this lands.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -20,6 +20,8 @@
 
 'use strict';
 
+var extend = require('extend');
+
 /**
  * @type {module:bigquery}
  * @private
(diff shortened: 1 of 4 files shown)


callmehiphop:
So what happens when a user creates a new API instance like..

```javascript
var search = gcloud.search({ ... });
```

Since search and pubsub are not taking advantage of the auto instantiation none of the static members are available through that method.

lib/index.js
```diff
@@ -122,29 +124,39 @@ var util = require('./common/util.js');
...
     }
   };
+
+  // Extend each API to pick up static members, e.g. Datastore.int & Storage.acl
+  extend(api.bigquery, BigQuery);
+  extend(api.datastore, Datastore);
+  extend(api.dns, DNS);
+  extend(api.pubsub, PubSub);
+  extend(api.search, Search);
+  extend(api.storage, Storage);
```
callmehiphop: This is maybe a little verbose? What if we did something like..

```javascript
function gcloud (config) {
  return {
    bigquery: createFactory(BigQuery, config),
    datastore: extend(new Datastore(config), Datastore),
    dns: createFactory(DNS, config),
    pubsub: createFactory(PubSub, config),
    search: createFactory(Search, config),
    storage: createFactory(Storage, config)
  };
}

function createFactory (ctor, config) {
  function factory (options) {
    return new ctor(util.extendGlobalConfig(config, options));
  }
  extend(factory, ctor);
  return factory;
}
```

stephenplusplus: Check it out now and let me know if I went too crazy.

callmehiphop: Check what out D:

stephenplusplus: Lol, sorry. Check out #798.

PR #798, opened 2015-08-13
Test index.js file

stephenplusplus:
Continuing from #795. (New PR because I didn't realize I was working off of the origin remote... used the just launched GitHub Desktop app... might lay off of that until I figure out how to use it.)

When a user gets a localized gcloud instance:

``` js
var gcloud = require('gcloud')(/*…*/);
```

They have been missing out on subclass static properties:

``` js
gcloud.storage.acl // undefined
```

This fixes that and adds tests.

We should do a patch release after this lands.

### DIFF
--- a/docs/site/components/docs/docs.html
+++ b/docs/site/components/docs/docs.html
@@ -68,11 +68,11 @@ var gcloud = require('gcloud')({
         <span class="view-code-link" docs-code-link></span>
       </h3>
       <p ng-if="method.description" ng-bind-html="method.description"></p>
-      <div ng-show="method.name[0].toUpperCase() === method.name[0]"
+      <div ng-show="method.constructor"
         class="notice">
         Available methods:
         <span ng-repeat="method in methods">
-          <span ng-show="method.name[0].toUpperCase() !== method.name[0]">
+          <span ng-show="$index > 0">
             <a href="{{activeUrl + '?method=' + method.name}}"
               >{{method.name}}</a>{{$last ? '' : ', '}}
           </span>
--- a/docs/site/components/docs/docs.js
+++ b/docs/site/components/docs/docs.js
@@ -266,7 +266,15 @@ angular
     }
 
     function compareMethods(a, b) {
-      return a.constructor ? -1: a.name > b.name ? 1 : -1;
+      if (a.constructor) {
+        return -1;
+      }
+
+      if (b.constructor) {
+        return 1;
+      }
+
+      return a.name > b.name ? 1 : -1;
     }
 
     $routeProvider
--- a/lib/index.js
+++ b/lib/index.js
@@ -23,46 +23,151 @@
 var extend = require('extend');
 
 /**
- * @type {module:bigquery}
- * @private
- */
-var BigQuery = require('./bigquery');
-
-/**
- * @type {module:datastore}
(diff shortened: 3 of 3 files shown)


stephenplusplus:
@callmehiphop RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/795#discussion_r36916231 let me know how this looks.

docs/site/components/docs/docs.html
```diff
@@ -68,11 +68,11 @@ <h2 ng-if="method.name[0].toUpperCase() === method.name[0]">
         <span class="view-code-link" docs-code-link></span>
       </h3>
       <p ng-if="method.description" ng-bind-html="method.description"></p>
-      <div ng-show="method.name[0].toUpperCase() === method.name[0]"
+      <div ng-show="method.constructor"
```
stephenplusplus: Made these changes so the `module:gcloud` constructor shows the available methods.

docs/site/components/docs/docs.js
```diff
@@ -266,7 +266,15 @@ angular
     }
 
     function compareMethods(a, b) {
-      return a.constructor ? -1: a.name > b.name ? 1 : -1;
+      if (a.constructor) {
+        return -1;
+      }
+
+      if (b.constructor) {
+        return 1;
```
stephenplusplus: This was previously missing. If B is a constructor, it's more important than A, unrelated to its name. This previously worked because all constructors were capitalized, thus ranked higher than anything lowercase. The `gcloud` "constructor" isn't meant to be instantiated, but it is the "head" documentation point for the module.

lib/index.js
```diff
@@ -122,149 +229,23 @@ var util = require('./common/util.js');
...
-      return new Storage(util.extendGlobalConfig(config, options));
+  return Object.keys(apis).reduce(function(gcloudExposedApi, apiName) {
+    var Class = apis[apiName];
+
+    if (scopedApis[apiName]) {
+      gcloudExposedApi[apiName] = new Class(config);
+    } else {
+      gcloudExposedApi[apiName] = function(options) {
+        return new Class(util.extendGlobalConfig(config, options));
+      };
```
callmehiphop: I get why you went this way, but it feels like overkill when considering how many APIs we're actually working with, lol.

stephenplusplus: What in particular do you mean?

callmehiphop: We're using reduce to iterate over 6 APIs and creating a map to construct one of those APIs in a different fashion. Considering how few APIs we actual offer at the moment, this feels sort of over the top. I'm not opposed to it, eventually if we offer more APIs (10+) it'll make more sense, but given how many we offer now it just feels like overkill to me!

stephenplusplus: I can revert to the old way if you prefer. With this way of doing things, I'm only bothered by the odd-ball Datastore instantiation we have, but I actually think this helps explain it and adds documentation around it. If you don't merge by tomorrow, I'll take the hint and go back to the old way :)

stephenplusplus: Oh, and Datastore is [about to get interesting](https://github.com/GoogleCloudPlatform/gcloud-node/issues/593#issuecomment-117288846) which may help it conform to the other style.

PR #799, opened 2015-08-14
docs: added no-escape directive to code blocks

### LINKED ISSUE #797: Filter docs have a &lt; where I think it should be a < ?
Looking at https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.18.0/datastore/query?method=filter

".filter('size &lt;', 400);"

callmehiphop:
Closes #797 

Looks like you have to explicitly state that you do **not** want to escape special characters when using our current code highlighting directive.

Source: https://github.com/pc035860/angular-highlightjs#user-content-directive-usage


stephenplusplus:
Well how about that!

callmehiphop:
Indeed!

PR #800, opened 2015-08-14
Document resumable uploads

### LINKED ISSUE #736: support resumable upload - or documenting it
is the resumable upload function implemented natively?  if not, it'd be great to provide such support.  if it is, we should call it out in the doc.

stephenplusplus:
Fixes #736

### DIFF
--- a/lib/storage/bucket.js
+++ b/lib/storage/bucket.js
@@ -864,9 +864,14 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
 };
 
 /**
- * Upload a file to the bucket. This is a convenience method that wraps the
- * functionality provided by a File object, {module:storage/file}.
+ * Upload a file to the bucket. This is a convenience method that wraps
+ * {module:storage/file#createWriteStream}.
  *
+ * You can specify whether or not an upload is resumable by setting
+ * `options.resumable`. *Resumable uploads are enabled by default if your input
+ * file is larger than 5 MB.*
+ *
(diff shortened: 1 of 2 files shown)


lib/storage/bucket.js
```diff
@@ -864,9 +864,14 @@ Bucket.prototype.setMetadata = function(metadata, callback) {
 };
 
 /**
- * Upload a file to the bucket. This is a convenience method that wraps the
- * functionality provided by a File object, {module:storage/file}.
+ * Upload a file to the bucket. This is a convenience method that wraps
+ * {module:storage/file#createWriteStream}.
  *
+ * Resumable uploads are automatically enabled if your input file is larger than
+ * 5 MB. You can override this logic by specifying `options.resumable` as `true`
```
callmehiphop: If resumable uploads are enabled by default, do we need to say that they can override the options by specifying `true`? Would it be better to just say something like, "You can disable this logic by specifying `options.resumable` as `false`."?

stephenplusplus: Well, if you provide a file that is < 5 MB, false is the default, so this just answers both sides.

callmehiphop: Got it.

I think I found this confusing because we lead with how resumable uploads are auto-enabled, so the sentence after feels directly related to that?

What if we did something like..

You can specify whether or not an upload is resumable by setting `options.resumable` to `true` or `false`. *Resumable uploads are enabled by default if your input file is larger than 5 MB.*

stephenplusplus: :+1: updated

callmehiphop:
Awesome, thanks! :+1:

stephenplusplus:
:clap:

PR #801, opened 2015-08-14
docs: storage: show better getSignedUrl examples

### LINKED ISSUE #695: Document cloud storage getSignedUrl
https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/storage/file.js#L1065

This is an awesome feature enabling client-side uploads and downloads, and I nearly switched to AWS s3 because there was no easy way to figure out how to do it with google cloud storage and Node.

A couple of examples would probably go a long way.

If I get a chance I'll do a PR to the readme, but didn't want to forget to indicate how important this is, in the meantime.

Thanks.

stephenplusplus:
Fixes #695

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -1156,11 +1156,53 @@ File.prototype.getSignedPolicy = function(options, callback) {
  * @param {string} callback.url - The signed URL.
  *
  * @example
+ * var TWO_WEEKS_MS = Math.round(Date.now() / 1000) + (60 * 60 * 24 * 14);
+ *
+ * //-
+ * // Generate a URL that allows temporary access to download your file.
+ * //-
+ * var request = require('request');
+ *
  * file.getSignedUrl({
  *   action: 'read',
- *   expires: Math.round(Date.now() / 1000) + (60 * 60 * 24 * 14), // 2 weeks.
- *   promptSaveAs: 'filename.ext'
- * }, function(err, url) {});
+ *   expires: TWO_WEEKS_MS
+ * }, function(err, url) {
+ *   if (err) {
+ *     console.error(err);
+ *     return;
+ *   }
+ *
+ *   // The file is now available to read from this URL.
+ *   request(url, function(err, resp) {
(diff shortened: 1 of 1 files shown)


lib/storage/file.js
```diff
@@ -1153,11 +1153,52 @@ File.prototype.getSignedPolicy = function(options, callback) {
  * @param {string} callback.url - The signed URL.
  *
  * @example
+ * var request = require('request');
+ * var TWO_WEEKS_MS = Math.round(Date.now() / 1000) + (60 * 60 * 24 * 14);
+ *
+ * //-
+ * // Generate a URL that allows temporary access to download your file.
+ * //-
  * file.getSignedUrl({
```
callmehiphop: Just a nitpick that you can feel free to ignore :)

Would it be better to require your modules within this code block? I feel like it might look a little funny to start off our example with a code block that just creates 2 variables.

stephenplusplus: For sure! I love these nitpicks :)

stephenplusplus:
screen shot 2015-08-17 at 4 47 48 pm

callmehiphop:
Sorry, I'm not trying to be annoying, but why are we leaving `TWO_WEEKS_MS` in its own block?

stephenplusplus:
I guess just because its used in both other examples. Should it just be in the top one?

callmehiphop:
Ah, nope, I think that's fine. Sorries!

eljefedelrodeodeljefe:
I was following the examples on the respective GCE pages, but couldn't find the docs you added here. Source code was fine for me, but maybe this could be promoted more and also put onto GCE, b/c this node feature is far easier then what you find there for gsutil or python

stephenplusplus:
Which GCE pages? The docs for this library are here: https://googlecloudplatform.github.io/gcloud-node/#/docs. The example from this PR ended up here: https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.32.0/storage/file?method=getSignedUrl

PR #807, opened 2015-08-17
core: update through2

stephenplusplus:
through2@2 is streams3-ready with backwards compat for stream2. How's that for confusing?


callmehiphop:
nice!

Issue #806, opened 2015-08-17
Datastrore.transaction.select & Datastore.transaction.filter not working properly in GCloud v0.14+ 


jiminikiz:
I have tried the following code snippets in `v-0.[14,15,16,17,18]`

```javascript
transaction.filter("email = ", email);
```
`.filter` works for `v[14,15,16]` but not `v[17,18]` with no changes to my code base. In `[17,18]` the full amount of rows from the table is returned.

```javascript
transaction.select(["date","account"]);
```
`.select` just hasn't worked for any version `[14+]`; the full amount of columns is always returned. I have not tested `.select` with any version before `[14]`.

stephenplusplus:
Thanks for reporting the bug. We'll look into it asap.

stephenplusplus:
I'm not able to reproduce either failure. We have a suite of tests that run against a real Datastore backend. Here's one that tests [filter](https://github.com/GoogleCloudPlatform/gcloud-node/blob/14523d3494611ea67c6c33794859b986252a2f3e/system-test/datastore.js#L383) and another that tests [select](https://github.com/GoogleCloudPlatform/gcloud-node/blob/14523d3494611ea67c6c33794859b986252a2f3e/system-test/datastore.js#L425). Are we not catching a use case similar to yours?

Pinging @pcostell in case there is a deeper reason a filter or select query might not produce the expected results.

pcostell:
There are requirements around filters / what can be selected, however the Datastore backend will fail if these requirements are not met.

pcostell:
@jiminikiz could you provide a test case that puts some data, then shows this failure when queryingn that data? It would be useful since it seems @stephenplusplus is not able to reproduce.

jiminikiz:
We wrote a function that we use to run queries, it gets exported in node as a module. Here is a snippet: 

```javascript
var DataSet = require('gcloud').datastore.dataset(<ourConfig>);

module.exports = {
    ...
    query: function ( query, callback ) {
        var parameter, filter = query.filter,
            transaction = DataSet.createQuery( query.namespace, query.path );

        if ( filter !== undefined ) {
            for ( parameter in filter ) {
                transaction.filter( parameter, filter[ parameter ] );
            }
        }
        if ( query.sort !== undefined  ) {
            transaction.order( query.sort );
        }
        if ( query.select !== undefined ) {
            transaction.select( query.select );
        }
        if ( query.limit !== undefined ) {
[...]

PR #808, opened 2015-08-17
datastore: make query objects mutable

callmehiphop:
Related to #806 

This makes `Query` objects mutable.

### DIFF
--- a/lib/datastore/query.js
+++ b/lib/datastore/query.js
@@ -21,7 +21,6 @@
 'use strict';
 
 var arrify = require('arrify');
-var extend = require('extend');
 
 /**
  * Build a Query object.
@@ -79,9 +78,8 @@ function Query(namespace, kinds) {
  * @return {module:datastore/query}
  */
 Query.prototype.autoPaginate = function(autoPaginateVal) {
-  var query = extend(true, new Query(), this);
-  query.autoPaginateVal = autoPaginateVal !== false;
-  return query;
+  this.autoPaginateVal = autoPaginateVal !== false;
+  return this;
 };
 
 /**
@@ -109,18 +107,16 @@ Query.prototype.autoPaginate = function(autoPaginateVal) {
  */
 Query.prototype.filter = function(filter, value) {
   // TODO: Add filter validation.
-  var query = extend(true, new Query(), this);
   var operatorRegex = /[><=]/g;
   filter = filter.trim();
   var fieldName = filter.replace(operatorRegex, '').trim();
   var op = filter.substr(fieldName.length, filter.length).trim();
-  query.filters = query.filters || [];
-  query.filters.push({
+  this.filters.push({
     name: fieldName,
     op: op,
     val: value
   });
-  return query;
+  return this;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


lib/datastore/query.js
```diff
@@ -278,9 +269,22 @@ Query.prototype.limit = function(n) {
...
+ * Creates a clone of a query. This is useful if you want to work with a Query
+ * in an immutable fashion.
+ *
+ * @return {module:datastore/query}
+ *
+ * @example
+ * var query = dataset.createQuery('Lion');
+ * var limitedQuery = query.clone().limit(10);
+ */
+Query.prototype.clone = function() {
```
stephenplusplus: Let's just leave this to the user to implement if needed.

callmehiphop: Would it be ok to leave as a private method then? We'll need something similar when running transactions.

https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/datastore/request.js#L623

stephenplusplus: Sure, but wouldn't it be easier to just extend() from right within the method like we normally do?

callmehiphop: Yeah, probably. I guess I just figured that since we need to create a clone and the users might want to do the same thing, a clone method would be convenient for everyone.

stephenplusplus: I prefer getting rid of things that we've found to not really be necessary instead of dragging support for them along. Before we lock onto a major version, it's a great time.

stephenplusplus:
Do any of the tests test for `this` being returned?

lib/datastore/query.js
```diff
@@ -109,18 +107,17 @@ Query.prototype.autoPaginate = function(autoPaginateVal) {
...
 Query.prototype.filter = function(filter, value) {
   // TODO: Add filter validation.
-  var query = extend(true, new Query(), this);
   var operatorRegex = /[><=]/g;
   filter = filter.trim();
   var fieldName = filter.replace(operatorRegex, '').trim();
   var op = filter.substr(fieldName.length, filter.length).trim();
-  query.filters = query.filters || [];
-  query.filters.push({
+  this.filters = this.filters || [];
```
stephenplusplus: These things are in the constructor it looks like. You can just push to them without `|| []`

lib/datastore/query.js
```diff
@@ -157,15 +153,14 @@ Query.prototype.hasAncestor = function(key) {
...
-  var query = extend(true, new Query(), this);
   var sign = '+';
   if (property[0] === '-' || property[0] === '+') {
     sign = property[0];
     property = property.substr(1);
   }
-  query.orders = query.orders || [];
-  query.orders.push({ name: property, sign: sign });
-  return query;
+  this.orders = this.orders || [];
```
stephenplusplus: Same as filter re: `|| []` not needed.

stephenplusplus:
:clap:

callmehiphop:
> Do any of the tests test for this being returned?

They don't explicitly test that what is returned is the same as the original object e.g.

```javascript
var query = new Query(['kind1']);
var limited = query.limit(10);

assert.strictEqual(query, limited);
```

However, there are tests that make use of the chaining - https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/test/datastore/query.js#L345

stephenplusplus:
It's better to test for it explicitly, in case something random goes wrong with that catch all test. (maybe the last one in the chain doesn't return `this`, maybe not all of the methods were used, etc)

lib/datastore/query.js

callmehiphop:
Sure, tests have been created!

PR #814, opened 2015-08-19
update docs to reflect actual events that are fired from a stream cre…

neilcronin:
…ated from a bucket item.

No code change.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -395,7 +395,7 @@ File.prototype.move = function(destination, callback) {
  *   .on('response', function(response) {
  *     // Server connected and responded with the specified status and headers.
  *    })
- *   .on('finish', function() {
+ *   .on('complete', function() {
  *     // The file is fully downloaded.
  *   })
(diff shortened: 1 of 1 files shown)


lib/storage/file.js
```diff
@@ -387,17 +387,18 @@ File.prototype.move = function(destination, callback) {
...
+ * var localFilename = '/Users/stephen/Photos/image.png';
  *
  * image.createReadStream()
- *   .pipe(fs.createWriteStream('/Users/stephen/Photos/image.png'))
  *   .on('error', function(err) {})
  *   .on('response', function(response) {
  *     // Server connected and responded with the specified status and headers.
  *    })
- *   .on('complete', function() {
+ *   .on('finish', function() {
```
stephenplusplus: Now that this event is attached to the gcloud stream, it actually is "complete". We could use a "finish" event on the fs stream below to tell readers that's when their file has been successfully written.

stephenplusplus:
Thank you for this!

stephenplusplus:
Thanks for everything!

PR #823, opened 2015-08-26
core: use google-auto-auth

### LINKED ISSUE #822: Caching of auth client regressed by #705
I believe that PR #705 regressed the earlier fix to cache the auth client I contributed in #661. This means that a new cache client is created on each API request we make, adding an unnecessary performance overhead. Here's some Cloud Trace output I get:

image

Note the extra 141ms added to this request, and every other request.

The cause is that the function [`makeAuthorizedRequest`](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/common/util.js#L465) now passes a brand new object to `util.makeRequest` each time. We cache the authClient on that object, which is temporary and not available the next time through.

Would it be possible add a test for this to make sure this doesn't get regressed?

stephenplusplus:
Fixes #822 

This incorporates https://github.com/stephenplusplus/google-auto-auth into our library, replacing `google-auth-library` (kind of; it's wrapped by `google-auto-auth`).

// cc @ofrobots

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -22,7 +22,7 @@
  */
 
 var extend = require('extend');
-var GoogleAuth = require('google-auth-library');
+var googleAuth = require('google-auto-auth');
 var is = require('is');
 var nodeutil = require('util');
 var request = require('request').defaults({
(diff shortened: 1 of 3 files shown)


jgeewax:
/cc @anthmgoogle (working on the auth libraries)

@stephenplusplus : Any reason why we couldn't push the commits upstream into google-auth-library?

stephenplusplus:
google-auth-library is much more complex and lacks documentation. google-auto-auth was just built to simplify it with a more conventional JS approach. I don't think this approach (high-level userland packages that wrap more low-level ones) is anything to worry about.

jgeewax:
Cool - if it's something specific like this, maybe we can get it shoved under the google or GCP github organization...?

stephenplusplus:
That's fine with me, though seeing two repos that handle authorization in Node under the same org might be a bit confusing to a user. I'm happy to transfer the ownership of the repo though if you don't think that's a concern.

jgeewax:
Eh let's leave it alone for now, and keep it as an idea for discussion
later on...

stephenplusplus:
SGTM :+1:

Issue #362, opened 2015-01-23
Storage streams use 'complete' event for end of stream but built-in streams use 'finish' 


ryanseys:
As was raised in #340, I looked into why there was a discrepancy between what the developer thought and what was the real case. There was a suggestion made to update our docs, but our docs aren't the issue here. A snippet from the tests shows the issue:

Using `'finish'` event:
``` js
file.createReadStream()
.pipe(fs.createWriteStream(tmpFilePath))
.on('error', done)
.on('finish', function() {
  file.delete(function(err) {
    assert.ifError(err);

    fs.readFile(tmpFilePath, function(err, data) {
      assert.equal(data, fileContent);
      done();
    });
  });
});
```

Using `'complete'` event:
``` js
var file = bucket.file(filenames[0]);
fs.createReadStream(files.logo.path)
  .pipe(file.createWriteStream())
  .on('error', done)
  .on('complete', function() {
    file.copy(filenames[1], function(err, copiedFile) {
      assert.ifError(err);
      copiedFile.copy(filenames[2], done);
    });
  });
```

Seems the only difference is the type of file that is getting piped to. In the first case, it's a regular stream from `fs` and in the second it's our implementation of the storage file write stream.

So my question is, should we use a consistent `finish` event everywhere or is this by-design or otherwise okay?

stephenplusplus:
Our complete event comes from `request`: https://github.com/request/request/blob/e33a883dd412dc7a1fdd1e1f282e840faa540609/request.js#L1224

Being consistent with `fs` would be nice, but off hand, I'm not sure we can do that.

ryanseys:
Are you sure? I thought we intercept pretty much every event and emit our own events.

I'm looking at code [here](https://github.com/GoogleCloudPlatform/gcloud-node/blob/fcc6123ddb0678f29d9ccd606e81318f988b4fa8/lib/storage/file.js#L406) and [here](https://github.com/GoogleCloudPlatform/gcloud-node/blob/fcc6123ddb0678f29d9ccd606e81318f988b4fa8/lib/storage/file.js#L593) for example.

stephenplusplus:
Ooh, I forgot about our new proxy stream. Yeah, we should be able to emit whatever we choose in these cases.

stephenplusplus:
This is quite complex.

[...]

PR #824, opened 2015-08-26
Refactor Storage read and write streams

stephenplusplus:
Related: #340, #362, #423, #435, #436 https://github.com/GoogleCloudPlatform/gcloud-node/pull/809#issuecomment-132096737

This PR will **force a breaking change**. But, a breaking change for the better.

We will no longer have to tell users to hook on to the "complete" events. That comes from the request module, and while that is quite a popular module, a user who creates a write or read stream from us doesn't really expect it, as they don't know we're using request underneath. It's like we're forcing them to learn our dependency, which is no good.

In the case of read streams, the user always received "end", but now we've removed the "complete" event. The use can safely listen for end and error, as that will give them all they need. "complete" used to emit with it the response from the API, but we update the file metadata with this object, so they can access the same from "file.metadata" inside their end callback*.

For write streams, we hold off letting the stream "finish" now until we know the data was uploaded successfully. This means we don't have to let it finish, then do post-processing, then do complete. That was messy.

Any questions, please ask! Streams, man.

\* Mention in release notes.
### Read Streams

**Before**

``` js
file.createReadStream()
  .on('data', function() {})
  .on('complete', function(metadata) {});
```

**After**

``` js
file.createReadStream()
  .on('data', function() {})
  .on('end', function() {
    // file.metadata === old "complete.metadata" argument
  });
```
### Write Streams

**Before**

``` js
var writeStream = file.createWriteStream();
writeStream.on('complete', function(metadata) {});
writeStream.end('data');
```

**After**

``` js
var writeStream = file.createWriteStream();
writeStream.on('finish', function() {
  // file.metadata === old "complete.metadata" argument
});
writeStream.end('data');
```

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -442,6 +442,9 @@ File.prototype.createReadStream = function(options) {
   }
 
   if (rangeRequest) {
+    if (is.string(options.validation) || options.validation === true) {
+      throw new Error('Cannot use validation with file ranges (start/end).');
+    }
     // Range requests can't receive data integrity checks.
     crc32c = false;
     md5 = false;
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -86,7 +86,6 @@ describe('File', function() {
     mockery.registerMock('sse4_crc32', crc32c);
 
     mockery.registerMock('configstore', FakeConfigStore);
-    // mockery.registerMock('duplexify', FakeDuplexify);
     mockery.registerMock('request', fakeRequest);
     mockery.registerMock('../common/util.js', fakeUtil);
     mockery.enable({
@@ -450,6 +449,37 @@ describe('File', function() {
 
       return FakeFailedRequest;
     }
+
+    it('should throw if both a range and validation is given', function() {
+      assert.throws(function() {
+        file.createReadStream({
+          validation: true,
+          start: 3,
+          end: 8
+        });
+      }, /Cannot use validation with file ranges/);
+
+      assert.throws(function() {
+        file.createReadStream({
+          validation: true,
+          start: 3
+        });
+      }, /Cannot use validation with file ranges/);
+
+      assert.throws(function() {
+        file.createReadStream({
+          validation: true,
+          end: 8
+        });
+      }, /Cannot use validation with file ranges/);
+
+      assert.doesNotThrow(function() {
+        file.createReadStream({
+          start: 3,
+          end: 8
+        });
+      });
+    });
 
     it('should send query.generation if File has one', function(done) {
       var versionedFile = new File(bucket, 'file.txt', { generation: 1 });


jgeewax:
Do you have an example of before / after code snippets? Just so we can clearly communicate to our users what needs to change?

stephenplusplus:
Yep, updated.

jgeewax:
Hmm some of this seems weird to me.

1. Couldn't we pass the metadata along with the event to keep the callbacks backwards compatible?
1. Read streams: couldn't we support listening on "end" as an alias for "complete"
1. Write streams: couldn't we support listening on "complete" as an alias for "finish" ?

From my (outsider's) perspective, it seems like a bunch of renames that we could have potentially hidden from users. I'm also scared that things don't look super consistent:

- Read: end -> complete
- Write: complete -> finish

These all sound like synonyms to me... Is it crazy to suggest maybe we call everything the same? (Say, all `complete` ?)

stephenplusplus:
Well, the thing I actually did was "make our streams not do anything special". "end" and "finish" are actual, Node streams core events that all streams have. We were dancing around them, doing post-processing, forcing us to hide native events and rename them as "complete".

> Couldn't we pass the metadata along with the event to keep the callbacks backwards compatible?

Because they are *not* custom events now (end for readable, and finish for writable), we can't do anything special, such as passing metadata.

> Read streams: couldn't we support listening on "end" as an alias for "complete"
> Write streams: couldn't we support listening on "complete" as an alias for "finish" ?

We shouldn't do anything custom because:

A) a user will have to learn what we created, instead of using what they expect from all other streams they use in Node
B) compatibility in a pipeline will be broken. If we are stream A and they pipe to stream B, B expects A to follow stream conventions.

I understand backwards compatibility would be nice, but I would urge us not to do this for the same reason we don't fix bugs and carry around "shim bugs". Our code gets weird and our users don't know what the heck's going on. We just need to make a note of it in the docs and play by semver. A breaking API is expected in 0.x bumps.

stephenplusplus:
> These all sound like synonyms to me... Is it crazy to suggest maybe we call everything the same? (Say, all complete ?)

Nope, and I wish they were. Streams are super weird, and they're working to improve on them. I really hope it gets more intuitive. (See https://github.com/stephenplusplus/stream-faqs for tracking how difficult they can be :confused:)

callmehiphop:
huge :+1:

> a user will have to learn what we created, instead of using what they expect from all other streams they use in Node

This sums up why I love this pull request.

stephenplusplus:
Updated the first post with many references of the confusion our event juggling has caused.

lib/storage/file.js
```diff
@@ -439,145 +441,129 @@ File.prototype.createReadStream = function(options) {
     md5 = options.validation === 'md5';
   }
 
-  if (is.undef(options.validation)) {
-    crc32c = true;
-    md5 = true;
+  if (rangeRequest) {
+    // Range requests can't receive data integrity checks.
```
callmehiphop: Do you think it would be useful to throw an exception here if the developer had set some form of validation? While that would obviously be a user error, it might be nice to let them know that what they're trying to do isn't possible.

test/storage/file.js
```diff
@@ -111,7 +86,7 @@ describe('File', function() {
     mockery.registerMock('sse4_crc32', crc32c);
 
     mockery.registerMock('configstore', FakeConfigStore);
-    mockery.registerMock('duplexify', FakeDuplexify);
+    // mockery.registerMock('duplexify', FakeDuplexify);
```
callmehiphop: Do we still want this?

lib/storage/file.js
stephenplusplus: Yes!

test/storage/file.js
stephenplusplus: No!

stephenplusplus:
PTAL

callmehiphop:
Looks good to me! Rebase time?

stephenplusplus:
Does it need to be?

callmehiphop:
It doesn't, this can bask in PR limbo for the rest of time!

stephenplusplus:
Oh, like squashing. Sorry, thought you meant it wouldn't merge without a rebase.

I can rebase, but the commits here are fine separated. It's easier to tear apart later / attribute regressions to if we ever needed to.

callmehiphop:
Fine by me!

jseeley78:
Is anyone else having problems with this change besides me? All my writestreams (uploads) now come back with "The uploaded data did not match the data from the server. As a precaution, the file has been deleted. To be sure the content is the same, you should try uploading the file again."

stephenplusplus:
We haven't heard of anything, and our tests seem to work properly. Are there any commonalities between the types of transfers you're doing that fail?

jseeley78:
all my cases are piping from imagemagick; 

var wr = gfs.createWriteStream(dst, {bucket: gfs.Buckets.archiveBucket});
gm(src)
	.setFormat('png')
	.stream(function(err, stdout, stderr) {
	if (err) return cb(err);
	// pipe output to writestream
	stdout.pipe(wr);
	// on writestream complete; make public/return
	 wr.on("complete", function() {
		gfs.makePublic(dst, {bucket: gfs.Buckets.archiveBucket}, cb);
	});
});

my createWriteStream function looks like this->
GFS.prototype.createWriteStream = function(filename, options) {
	options = options || {};
	return this.getBucket(options.bucket).file(filename).createWriteStream({
		metadata: {contentType: this.getContentType(filename)}
	});
};

stephenplusplus:
Thanks! What happens if you listen for `finish` instead of `complete`? Also, listen for `error` on `wr` as well.

jseeley78:
I don't get the new finished event; I tried listening for finish, complete, end, data and error; I'm only getting back on "error"->
{ [Error: The uploaded data did not match the data from the server. As a precaution, the file has been deleted. To be sure the content is the same, you should try uploading the file again.] code: 'FILE_NO_UPLOAD', errors: [ null ] }

stephenplusplus:
Could you open a new issue so we can keep digging into this? Also please include a link to the library and I'll try to reproduce.

jseeley78:
ok, submitted as ticket #889

Issue #400, opened 2015-02-18
Storage: clean up write streams


ryanseys:
Looking at the file.createWritableStream, validation is intermixed with everything and some internal streams exist solely for validation purposes. We might want to consider having a fast-path for when `validation: false` which by-passes all this logic and extraneous streams. This should help with #397 when `validation: false` but I'll put it in Storage Future because for the default and recommended case, it probably won't help much.

stephenplusplus:
:+1:, though not sure the added complexity this would likely introduce is worth the minimal memory savings.

ryanseys:
I'm not sure of the overhead associated with creating streams in Node.JS but the author of the performance issue seems to suggest that it may play a role so this was an idea for reducing that. I'm with you, I don't think it will play a big difference but worth investigating. Plus createWriteStream is HUGE and might benefit from a cleanup.

stephenplusplus:
I'm going to look into simplifying the logic of our write stream implementation.

stephenplusplus:
Our streams have been cleaned up considerably in some recent PRs and refactors (#694, #820, #831). We now use modules to help us out, like [`gcs-resumable-upload`](https://github.com/stephenplusplus/gcs-resumable-upload) & [`hash-stream-validation`](https://github.com/stephenplusplus/hash-stream-validation). We also found a cleaner way to use Request to make a multi-part upload. I think we can call this one resolved for now, and if we see room for improvements down the road, we'll be sure to jump.

PR #831, opened 2015-08-28
core: refactor util.makeWritableStream

stephenplusplus:
RE: #400

It turns out there's a much easier way to handle multipart uploads with request.


callmehiphop:
:+1:

PR #833, opened 2015-08-28
use gcs-resumable-upload module

### LINKED ISSUE #825: Resumable uploads aren't tested
We kind of are, but not correctly. We need to actually destroy the event mid-stream. Our test should look more like this:

```js
it('should resume an upload after an interruption', function(done) {
  fs.stat(files.big.path, function(err, metadata) {
    assert.ifError(err);

    // Use a random name to force an empty ConfigStore cache.
    var file = bucket.file('LargeFile' + Date.now());
    var fileSize = metadata.size;

    upload({ interrupt: true }, function(err) {
      assert.strictEqual(err.message, 'Interrupted.');

      upload({ interrupt: false }, function(err) {
        assert.ifError(err);

        assert.equal(file.metadata.size, fileSize);
        file.delete(done);
      });
    });

    function upload(opts, callback) {
      var ws = file.createWriteStream();
      var sizeStreamed = 0;

      fs.createReadStream(files.big.path)
        .pipe(through(function(chunk, enc, next) {
          sizeStreamed += chunk.length;

          if (opts.interrupt && sizeStreamed >= fileSize / 2) {
            // stop sending data half way through.
            this.push(chunk);
            this.destroy();
            ws.destroy(new Error('Interrupted.'));
          } else {
            this.push(chunk);
            next();
          }
        }))
        .pipe(ws)
        .on('error', callback)
        .on('finish', callback);
    }
  });
});
```

But more importantly, the file we're using is apparently too small. Either the chunks aren't actually being emitted by request (I think they are, though, cause they're supposed to be drained in 16kb chunks), or GCS doesn't recognize storing any data for a resumable upload until it meets some threshold.

Our test file is 3MB and cuts off half way though. I tried with 10MB, 20MB, 31MB, and 42MB. 42 and higher were the only ones that got a response from the API call to see where the request left off:

```
PUT {resumable_uri}
Content-Length: 0
Content-Range: bytes */*
```

(Note that the amount stored by the API would have been halved (so 21 when 42 worked))

// @jgeewax

stephenplusplus:
Fixes #825

This removes the overhead of supporting resumable uploads in gcloud and uses the (way better tested) [gcs-resumable-upload](http://gitnpm.com/gcs-resumable-upload).

I extracted it not just for that reason, but because by itself, it's a pretty useful module that someone could very easily want without wanting to pull down all of gcloud.


jgeewax:
WRT gcs-resumable-upload (https://github.com/stephenplusplus/gcs-resumable-upload), can you explain a bit more about the need for the configstore file?

My understanding of the resumable upload API was that we can ask GCS "what's the last byte position you got?" and it'll tell us "I have up to byte X" and then we can go "Cool, here's byte X+1 through byte X+1+chunksize" and keep going.

If that API is correct, I think I'm missing where the configstore settings come into the picture. If I have the API wrong, then I need a full refresher course :)

stephenplusplus:
Haha, we need to store the unique URI. As far as I know, you can't ask the API for "Where I left off for this file?", you have to ask, "Where did I leave off for ** this crazy weird unique ID ** upload?"

callmehiphop:
We use ConfigStore to cache the first chunk of an upload, then if the upload is interrupted and resumed, we check that the user is not attempting to upload a different file by comparing the two initial chunks. - https://github.com/stephenplusplus/gcs-resumable-upload/blob/master/index.js#L135

We do this in the current code base as well - https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/storage/file.js#L1453

lib/storage/file.js
```diff
@@ -1314,273 +1309,35 @@ File.prototype.makePublic = function(callback) {
 };
 
 /**
- * `startResumableUpload_` uses the Resumable Upload API: http://goo.gl/jb0e9D.
```
stephenplusplus: @jgeewax here's a breakdown :)

callmehiphop:
@stephenplusplus actually I think you can https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload#resume-upload

jgeewax:
I think those docs say that you should pass the session URI, which is the thing that @stephenplusplus is talking about saving in configstore to persist across the life of the process doing the uploading -- considering the chances that a failed upload would also correspond to a dead process.

In this case, I'm :+1: on storing the session URI for use across processes.

PR #835, opened 2015-08-30
storage: getSigned*: clear up expiration format

### LINKED ISSUE #834: Time unit mismatch in File.prototype.getSignedPolicy
In the [documentation](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/storage/file.js#L905) *expiration* is in seconds, but in [code](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/storage/file.js#L950), it's expected to be in milliseconds.
The given example use seconds.

stephenplusplus:
Fixes #834 

Formerly:

`getSignedPolicy` - options.expiration - Advertised as seconds, handled by our code as milliseconds
`getSignedUrl` - options.expires - Advertised as seconds, handled by the API directly as seconds

Now:

`getSignedPolicy` - options.**expires** - Advertised as **anything Date accepts**
`getSignedUrl` - options.expires - Advertised as **anything Date accepts**

Also added a system test for getSignedPolicy, which we didn't have previously.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -902,14 +902,14 @@ File.prototype.getMetadata = function(callback) {
  *     members.
  *
  * @param {object} options - Configuration object.
- * @param {object} options.expiration - Timestamp (seconds since epoch) when
- *     this policy will expire.
  * @param {array|array[]=} options.equals - Array of request parameters and
  *     their expected value (e.g. [['$<field>', '<value>']]). Values are
  *     translated into equality constraints in the conditions field of the
  *     policy document (e.g. ['eq', '$<field>', '<value>']). If only one
  *     equality condition is to be specified, options.equals can be a one-
  *     dimensional array (e.g. ['$<field>', '<value>']).
+ * @param {*} options.expires - A timestamp when this policy will expire. Any
+ *     value given is passed to `new Date()`.
  * @param {array|array[]=} options.startsWith - Array of request parameters and
  *     their expected prefixes (e.g. [['$<field>', '<value>']). Values are
  *     translated into starts-with constraints in the conditions field of the
@@ -932,22 +932,28 @@ File.prototype.getMetadata = function(callback) {
  * @param {object} callback.policy - The document policy.
  *
  * @example
- * file.getSignedPolicy({
+ * var options = {
  *   equals: ['$Content-Type', 'image/jpeg'],
- *   contentLengthRange: { min: 0, max: 1024 },
- *   expiration: Math.round(Date.now() / 1000) + (60 * 60 * 24 * 14) // 2 weeks.
- * }, function(err, policy) {
+ *   expires: '10-25-2022',
+ *   contentLengthRange: {
+ *     min: 0,
+ *     max: 1024
+ *   }
+ * };
+ *
+ * file.getSignedPolicy(options, function(err, policy) {
  *   // policy.string: the policy document in plain text.
  *   // policy.base64: the policy document in base64.
  *   // policy.signature: the policy signature in base64.
  * });
  */
 File.prototype.getSignedPolicy = function(options, callback) {
-  if (options.expiration < Math.floor(Date.now() / 1000)) {
+  var expires = new Date(options.expires);
+
+  if (expires < Date.now()) {
     throw new Error('An expiration date cannot be in the past.');
   }
 
-  var expirationString = new Date(options.expiration).toISOString();
   var conditions = [
     ['eq', '$key', this.name],
     {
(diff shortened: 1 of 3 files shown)


lib/storage/file.js
```diff
@@ -932,22 +932,29 @@ File.prototype.getMetadata = function(callback) {
  * @param {object} callback.policy - The document policy.
  *
  * @example
- * file.getSignedPolicy({
+ * var TWO_WEEKS_FROM_NOW = Math.round(Date.now() / 1000) + (60 * 60 * 24 * 14);
```
jgeewax: Will this work with something like `moment.duraction(2, 'weeks');` ? (http://momentjs.com/docs/#/durations/creating/)

jgeewax:
LGTM

lib/storage/file.js
stephenplusplus: The better method would be `moment.add(2, 'weeks')` which our library would handle fine. I'll add an example to the docs.

stephenplusplus: My bad, this doesn't work. Maybe we should accept any input type for `expires`. If it's a number, we'll treat it as seconds since epoch. If it's anything else, we can wrap it in a Date object and call toISOString() on that.

That means this will all work:

```js
options.expires = moment().add(2, 'weeks');

options.expires = Math.round(Date.now() / 1000) + (60 * 60 * 24 * 14);
options.expires = new Date(Math.round(Date.now() / 1000) + (60 * 60 * 24 * 14));

options.expires = new Date('2015-09-13')
options.expires = '2015-09-13';

options.expires = '2015-09-13T15:57:33.092Z';
```

lib/storage/file.js
```diff
@@ -902,14 +902,14 @@ File.prototype.getMetadata = function(callback) {
...
  * @param {object} options - Configuration object.
- * @param {object} options.expiration - Timestamp (seconds since epoch) when
- *     this policy will expire.
  * @param {array|array[]=} options.equals - Array of request parameters and
  *     their expected value (e.g. [['$<field>', '<value>']]). Values are
  *     translated into equality constraints in the conditions field of the
  *     policy document (e.g. ['eq', '$<field>', '<value>']). If only one
  *     equality condition is to be specified, options.equals can be a one-
  *     dimensional array (e.g. ['$<field>', '<value>']).
+ * @param {object} options.expires - Timestamp (seconds since epoch) when this
```
callmehiphop: Why are we changing the verbiage? I know we're not `1.0.0` yet, but I'm not sure if this seems like something worth introducing a breaking change for?

stephenplusplus: My stance on breaking our API is the same as always. Breaking changes on pre-1.0 minor versions is why we bump the minor. If we are making things better, then I don't think we should worry about backwards compatibility... at least not for this kind of thing. If it was like we changed "createBucket(name, options)" to "createBucket(options)", then we should probably legacy support that for 1-2 next releases.

But, this is basically an "API fix" - an inconsistency between our docs and code, and between what similar methods accept as input. We need to fix it, and I don't think we need to backwards support it.

lib/storage/file.js
callmehiphop: I'd be in favor of simply accepting anything that the Date object accepts, this is already a breaking change so users will have to update their code anyways.

stephenplusplus:
Updated to accept anything Date accepts.

lib/storage/file.js
```diff
@@ -1115,7 +1119,10 @@ File.prototype.getSignedPolicy = function(options, callback) {
  * });
  */
 File.prototype.getSignedUrl = function(options, callback) {
-  if (options.expires < Math.floor(Date.now() / 1000)) {
+  var expires = new Date(options.expires).getTime();
+  var expiresInSeconds = Math.round(expires / 1000); // The API expects seconds.
+
+  if (expires < Date.now()) {
```
callmehiphop: I don't think you have to convert Dates into Numbers to compare them.

```javascript
var isPast = new Date(options.expires) < new Date;
```

stephenplusplus: Yep, you're right! I left in the `getTime()` because I need to convert it for the places it was concatenated with a string below (instead of calling Number() or `getTime()` later)... but then I remembered it needed to be seconds anyway, and forgot to update the code.

:mag: :tophat:

stephenplusplus: Fixed

callmehiphop: Looks like the documentation unit test busted

stephenplusplus:
The docs test timed out after 2s. I restarted travis, it failed. Restarted again and it passed. Maybe that needs to be bumped up? Wait til it consistently becomes an issue?

callmehiphop:
Sure, that sounds like a good plan.

PR #838, opened 2015-08-31
docs: updates to language links

### LINKED ISSUE #836: Link on landing page to Python sends you to the inner page
1. Go to https://googlecloudplatform.github.io/gcloud-node/#/
1. Click the language switcher
1. Click Python

Expected:
- end up at http://googlecloudplatform.github.io/gcloud-python/

Actual:
- http://googlecloudplatform.github.io/gcloud-python/latest/

(I think even the switcher inside should probably send you to http://googlecloudplatform.github.io/gcloud-python/stable/ (not /latest)

callmehiphop:
closes #836 

In order to provide a consistent user experience with the other `gcloud-*` sites, I've made some modifications to our language switcher directive to route the user to different sections of the other doc sites depending on what section they're currently browsing.

Previously if a user changed languages from either the splash screen or the actual docs, we would **always** route them to the api documentation. Per #836 this did not seem to be the desired functionality.

[Preview](http://callmehiphop.github.io/gcloud-node)

### DIFF
--- a/docs/site/components/language-switcher/language-switcher-directive.js
+++ b/docs/site/components/language-switcher/language-switcher-directive.js
@@ -6,10 +6,29 @@ angular
     return {
       templateUrl: 'site/components/language-switcher/language-switcher.html',
       transclude: true,
-
       link: function($scope, elem, attrs) {
         $scope.extraLinks = $parse(attrs.extraLinks)($scope);
         $scope.showNavDropdown = false;
+      },
+      controller: function($scope, $attrs) {
+        this.useApiDocs = $parse($attrs.useApiDocs)($scope);
+      }
+    };
+  })
+  .directive('apiDocs', function() {
+    'use strict';
+
+    return {
+      require: '^languageSwitcher',
+      restrict: 'A',
+      scope: {
+        href: '@',
+        apiDocs: '@'
+      },
+      link: function (scope, elem, attrs, switcher) {
+        if (switcher.useApiDocs) {
+          elem.attr('href', scope.href + scope.apiDocs);
+        }
       }
     };
   });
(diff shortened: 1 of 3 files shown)


stephenplusplus:
I personally would expect that clicking a link to another site would just take me to its homepage, and not try to put me on the API docs, when I really just might be curious of the high-level "Oh, this is in Python, too?" I'm probably not flipping back and forth between the API docs as if I'm a user of both libraries.

But, @jgeewax's call.

callmehiphop:
I'm going to disagree with you here, since we're making an effort to brand the sites in the same way and switching languages is done via dropdown menu and not an external link, it feels less like the intent is to redirect the user but rather show them the same content for a different language.

If anything the part that I find weird is when viewing a specific api (e.g. storage) it does not link me to the python storage page, but rather the api landing page.

stephenplusplus:
I agree drop downs vs an obviously-external sidebar link give different impressions. But, with *this* dropdown that is highly styled (not like our browser-default version switcher) and *is* an external link, I think the user not only expects to leave the site and go somewhere else, but I think that we should want them to go to their homepage and not an inner page of a place they haven't been before. Beyond that, I still don't think we have many multi-lingual, back-and-forth API doc switchers, so a solution for that problem seems like over-engineering.

Whatever the people want, though!

callmehiphop:
This is where our opinions are different, I don't think it's very obvious to the user that switching languages would navigate to a different site. The theming is the same across each site, which sort of gives the impression that when you switch languages, you haven't gone to different site, but instead a different page on the same site. While the dropdown is indeed highly stylized, nothing about the styling indicates that the links it contains are external, especially when considering we display an external link icon throughout the site for links that are external but not for the links within the dropdown.

If anything, maybe we should raise an issue in gcloud-common, because this is actually the current behavior in both gcloud-ruby and gcloud-python.

stephenplusplus:
¯\_(ツ)_/¯ Code looks good. Merge time!

PR #839, opened 2015-08-31
storage: allow more time for signed urls to expire

stephenplusplus:
In #835, I only allowed 5 milliseconds for a signed link to expires. It used to be interpreted as seconds, so I glossed right over this while changing the code over. It passed locally of course, but since Travis takes his time with everything, he was not pleased.

I changed it back to 5 seconds (5000 ms) and added some more error handling that was previously overlooked.


callmehiphop:
Thanks!

stephenplusplus:
*crossing fingers*

PR #844, opened 2015-09-02
core: preserve prototype methods

stephenplusplus:
Before:

``` js
var gcloud = require('gcloud')({ projectId: 'grape-spaceship-123', keyFilename: 'key.json' });
var dataset = gcloud.datastore.dataset(); // throws "missing projectId" error
```

After:

``` js
var gcloud = require('gcloud')({ projectId: 'grape-spaceship-123', keyFilename: 'key.json' });
var dataset = gcloud.datastore.dataset(); // uses defaults from the gcloud instantiation
```

This used to work, so what happened? Well, [I broke it](https://github.com/GoogleCloudPlatform/gcloud-node/pull/798).

When a user does: 

``` js
var gcloud = require('gcloud')({ projectId: 'grape-spaceship-123', keyFilename: 'key.json' });
```

It prepares instances of the various modules (Storage, PubSub, etc) and then extends them so that static methods are picked up (Datastore.int, Storage.acl, etc). We have methods that are the same name as static methods, eg `Datastore.dataset` and `Datastore.prototype.dataset` so that if a user has an _instance_ of a Datastore, they can access a properly scoped dataset. If they _don't_ have an instance of a Datastore -- if they just did `require('gcloud')` without instantiating it with config options at that time -- `Datastore.dataset` will require configuration details, since you haven't given them anywhere else yet.

So, the fix in this PR preserves the prototype methods, not letting the static ones overwrite them.

// @callmehiphop -- need your help, buddy! Is this just madness?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -260,8 +260,9 @@ function gcloud(config) {
     }
 
     // Extend each API to pick up static members, e.g. Datastore.int and
-    // Storage.acl.
-    extend(gcloudExposedApi[apiName], Class);
+    // Storage.acl. The prototype is last so that static methods by the same
+    // name as an instance method aren't overwritten.
+    extend(gcloudExposedApi[apiName], Class, Class.prototype);
 
     return gcloudExposedApi;
   }, {});
--- a/test/index.js
+++ b/test/index.js
@@ -32,31 +32,43 @@ var fakeUtil = extend({}, util, {
 function FakeBigQuery() {
   this.calledWith_ = [].slice.call(arguments);
 }
+FakeBigQuery.prototype.method = function() { return true; };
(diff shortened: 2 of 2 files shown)


callmehiphop:
It does seem a little odd, but I think that's the cost of preserving the global configurations passed into `gcloud`. `Function#bind` doesn't preserve static members either, so I think this approach is really our only option.

stephenplusplus:
Can you PR that? It sounds less weird!

callmehiphop:
Sure!

callmehiphop:
done - https://github.com/GoogleCloudPlatform/gcloud-node/pull/845

stephenplusplus:
Nullified by #845

PR #845, opened 2015-09-02
core: preserve all class members

callmehiphop:
Closes #844

Unit tests need to be updated to accommodate this change, but I wanted to get some feedback before I attempted to tackle that.

I ran the system tests locally and only received a single failure for an after all hook (might be unrelated) -- so this implementation should work.

// @stephenplusplus

### DIFF
--- a/lib/bigquery/index.js
+++ b/lib/bigquery/index.js
@@ -96,13 +96,8 @@ var SCOPES = ['https://www.googleapis.com/auth/bigquery'];
  */
 function BigQuery(options) {
   if (!(this instanceof BigQuery)) {
+    options = util.normalizeArguments(this, options);
     return new BigQuery(options);
-  }
-
-  options = options || {};
-
-  if (!options.projectId) {
-    throw util.missingProjectIdError;
   }
 
   this.makeAuthorizedRequest_ = util.makeAuthorizedRequestFactory({
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -462,3 +462,26 @@ function decorateRequest(reqOpts) {
 }
 
 util.decorateRequest = decorateRequest;
+
+/**
+ * Merges and validates API configurations
+ *
+ * @throws {Error} If projectId is missing
+ *
+ * @param {?object} globalContext - api level context, this is where the
+ *     gloabl configuration should live
+ * @param {?object} localConfig - api level configurations
+ * @return {object} config - merged and validated configurations
+ */
+function normalizeArguments(globalContext, localConfig) {
+  var globalConfig = globalContext && globalContext.config_ || {};
+  var config = util.extendGlobalConfig(globalConfig, localConfig);
+
+  if (!config.projectId) {
+    throw util.missingProjectIdError;
+  }
+
+  return config;
+}
+
+util.normalizeArguments = normalizeArguments;
--- a/lib/compute/index.js
+++ b/lib/compute/index.js
@@ -106,13 +106,8 @@ var SCOPES = ['https://www.googleapis.com/auth/compute'];
  */
 function Compute(options) {
   if (!(this instanceof Compute)) {
+    options = util.normalizeArguments(this, options);
     return new Compute(options);
-  }
-
-  options = options || {};
-
-  if (!options.projectId) {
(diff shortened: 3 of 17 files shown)


callmehiphop:
Also might be worth mentioning that in order for this to work moving foward all our classes would have to follow the pattern..

```javascript
function Class() {
  if (!(this instanceof Class)) {
    return new Class();
  }
}
```

stephenplusplus:
It looks good to me! Can we maybe wrap consistent instantiation stuff into one method? (Example to follow)

lib/compute/index.js
```diff
@@ -105,6 +105,10 @@ var SCOPES = ['https://www.googleapis.com/auth/compute'];
  * var gce = gcloud.compute();
  */
 function Compute(options) {
+  if (this.config_) {
+    options = util.extendGlobalConfig(this.config_, options);
+  }
+
```
stephenplusplus: I'm thinking this as well as throwing if there is no projectId could go to a single place, so that this just calls "util.normalizeArguments()" or something -- feel free to tell me that's a bad idea.

callmehiphop: I like it!

callmehiphop: Created and implemented `util.normalizeArguments()` - PTAL.

lib/bigquery/index.js
```diff
@@ -96,15 +96,10 @@ var SCOPES = ['https://www.googleapis.com/auth/bigquery'];
  */
 function BigQuery(options) {
   if (!(this instanceof BigQuery)) {
+    options = util.normalizeArguments(this, options);
```
stephenplusplus: Do this outside of the `if`.

lib/common/util.js
```diff
@@ -462,3 +462,26 @@ function decorateRequest(reqOpts) {
 }
 
 util.decorateRequest = decorateRequest;
+
+/**
+ * Merges and validates API configurations
+ *
+ * @throws {Error} If projectId is missing
+ *
+ * @param {object|null} globalContext - api level context, this is where the
```
stephenplusplus: `{?object}`

lib/bigquery/index.js
callmehiphop: I can do that, I just want to point out that doing that will call normalizeArguments twice per object creation.

lib/common/util.js
callmehiphop: Forgot about that one.

lib/bigquery/index.js
stephenplusplus: How does that happen?

callmehiphop: Instead of exposing factory patterns, we just expose the constructors now and the user treats them like a factory. Ultimately what ends up happening is the constructor gets called, sees that it's not instantiated and returns a new instance of itself (aka calls itself a second time)

stephenplusplus: I should have said move the normalize call below the if, so that the `if (!(this ...` is still the first thing to happen.

callmehiphop: That won't work either, the trade off for killing the factory patterns is that we can't store the global config in a closure any longer.

https://github.com/callmehiphop/gcloud-node/blob/config-refactor/lib/index.js#L255

So if you move that line down.. you lose the global config.

stephenplusplus: Gotcha, I was definitely not picking up how this whole thing worked.

lib/index.js
```diff
@@ -254,17 +248,11 @@ function gcloud(config) {
...
+      gcloudExposedApi[apiName] = Class;
     }
 
-    // Extend each API to pick up static members, e.g. Datastore.int and
-    // Storage.acl.
-    extend(gcloudExposedApi[apiName], Class);
-
     return gcloudExposedApi;
-  }, {});
+  }, { config_: config });
```
stephenplusplus: Can you add a `/*! Developer Documentation` block for this function to explain how this part works?

callmehiphop: Done!

stephenplusplus:
Woo! Bravo on this solution :)

PR #846, opened 2015-09-02
Renaming all uses of authorize to authenticate

### LINKED ISSUE #841: README should say "Authentication", not "Authorization"
I'm kind of on the fence if I even believe this to be true or not.

The OAuth 2.0 flows themselves are authorization (i.e. authorizing the clients to act on behalf of the user) but the service account flows are user-independent and really serve to authenticate an entity with a given access.

Also, when we use non-service accounts we just use the Application Default Credentials from the `gcloud` CLI. These will already have a refresh token, so again we perform / participate in no authorization, just authentication.

----

PS Sorry for the esoteric issue, but I was adding this section to the `gcloud-python` README and it just occurred to me.

----

PPS Is there some magic trick to get a link to lines in a MD file on GitHub?

dhermes:
This is a mess, so I tried to make it "reproducible".

Fixes #841 

Some of the links will be broken unless the language and ~~links~~ paths in `gcloud-common` are also updated.

---

To "find" them all I used

``` bash
git grep -i authori
```

and then to replace, it was

``` bash
git grep -l Authorization | xargs sed -i 's/Authorization/Authentication/g'
git grep -l authorization | xargs sed -i 's/authorization/authentication/g'
git grep -l Authorizing | xargs sed -i 's/Authorizing/Authenticating/g'
git grep -l authorizing | xargs sed -i 's/authorizing/authenticating/g'
git grep -l Authorized | xargs sed -i 's/Authorized/Authenticated/g'
git grep -l authorized | xargs sed -i 's/authorized/authenticated/g'
git grep -l authorize | xargs sed -i 's/authorize/authenticate/g'
```

(There is probably a smarter way to make case-insensitive / case-matching replacements with `sed`, I just didn't.)

Then this process touched the license so I had to do

``` bash
git checkout -- COPYING
```

To check this all "quickly" I used 

``` bash
git diff --color-words
```

and finally to do the file renames I did

``` bash
git ls-files | grep -i authori
```


stephenplusplus:
Man... words.

Thanks very much for doing this. I think we can take it over from here to get the tests to pass. (@callmehiphop -  how about renaming all of our `makeAuthenticatedRequest_` to just `request` or something less wordy?)

callmehiphop:
Sure `makeApiRequest_`?

stephenplusplus:
Yeah, that's good enough :+1: I think in a separate issue soon, we should once and for all clear up the multiple "makeRequest-ish" methods we have, and just follow a pattern. Will make an issue for that.

dhermes:
Sorry for being a bad OS citizen, I never ran `npm test`. I recall trying to do it about 12 months ago and having a really bad time (TM).

stephenplusplus:
#861 will take this over. Thanks @dhermes!

dhermes:
Sure thing!

PR #848, opened 2015-09-03
Update repository field in package.json

ryanseys:
Makes http://gitnpm.com/gcloud work properly :smile: @stephenplusplus

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,10 @@
     "CONTRIBUTORS",
     "COPYING"
   ],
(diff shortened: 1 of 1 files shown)


stephenplusplus:
I can't believe I didn't notice that didn't work with gitnpm all this time. I just need to fix gitnpm to handle string repository properties: https://github.com/stephenplusplus/gitnpm/issues/5 as the format we use on gcloud is perfectly fine. Thanks for pointing this out!

ryanseys:
SGTM!

PR #850, opened 2015-09-04
Fixed issue where read streams on files were never ending.

nemo:
Hey folks,

We were having issues with [createReadStream](https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.20.0/storage/file?method=createReadStream) never firing the `end` event.

Looks like you introduced this bug a few days ago after your recent refactor around streams. It was working fine for us until we upgraded.

Cheers.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -537,6 +537,8 @@ File.prototype.createReadStream = function(options) {
         mismatchError.code = 'CONTENT_DOWNLOAD_MISMATCH';
 
         throughStream.destroy(mismatchError);
+      } else {
+        throughStream.end();
       }
     }
 


nemo:
Also, just signed the agreement :)

stephenplusplus:
Thank you for contributing! Just to confirm, you were having issues with master? Could you paste the code that we could use to reproduce? The stream should be ending itself, so it's possible the stream you're piped to isn't fully consuming the data (or a myriad of other problems!)

nemo:
It may have been. It looks like you just recently made a few changes to your package.json – let me re-install and let you know if I'm facing this issue.

nemo:
@stephenplusplus – looks like 0.20 fixes the issue! I was on 0.18 before.

Closing this!

stephenplusplus:
Great, glad it got sorted out. Please come back if you come across any new strangeness!

PR #855, opened 2015-09-09
docs: flagged private methods as private

callmehiphop:
I noticed a couple of methods are on the doc site that shouldn't be!

e.g. http://googlecloudplatform.github.io/gcloud-node/#/docs/v0.20.0/compute/disk?method=formatName_


stephenplusplus:
Hey! Don't look at those!

callmehiphop:
But I can't help it ._.

PR #857, opened 2015-09-09
execute callback with error if signing is not possible

### LINKED ISSUE #211: getSignedUrl requires `private_key`
To generate a signed url, we need a private_key: https://developers.google.com/storage/docs/accesscontrol#Signed-URLs & https://github.com/GoogleCloudPlatform/gcloud-node/blob/v0.6.0/lib/storage/index.js#L267

We get this if a user provides a credentials object or path to a keyfile, but not in GCE or in the future, GAE. Is it going to be possible to get the private key in those environments?

stephenplusplus:
Fixes #211

If the required properties to create a signed URL or policy aren't available, the user's callback will be executed with an error indicating so.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -1061,7 +1061,19 @@ File.prototype.getSignedPolicy = function(options, callback) {
 
   makeAuthorizedRequest_.getCredentials(function(err, credentials) {
     if (err) {
-      callback(err);
+      var signingError = new Error('Signing failed. See `error` property.');
+      signingError.error = err;
+      callback(signingError);
+      return;
+    }
+
+    if (!credentials.private_key) {
+      var errorMessage = [
+        'Signing failed. Could not find a `private_key`.',
+        'Please verify you are authorized with this property available.'
+      ].join(' ');
+
+      callback(new Error(errorMessage));
       return;
     }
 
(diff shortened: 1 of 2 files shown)


lib/storage/file.js
```diff
@@ -1061,7 +1061,19 @@ File.prototype.getSignedPolicy = function(options, callback) {
 
   makeAuthorizedRequest_.getCredentials(function(err, credentials) {
     if (err) {
-      callback(err);
+      var signingError = new Error('Signing failed. See `error` property.');
+      signingError.error = err;
+      callback(signingError);
```
callmehiphop: Since this is happening in more than one place, do you think it would be worth consolidating the logic into something re-usable? like.. `callback(new SigningError(err));`?

stephenplusplus: I think this is a case where that would be more engineering than its worth. Sub-classing an Error is quite a noisy process. However, it would be nice to go over our errors sometime with something like https://github.com/floatdrop/create-error-class and maybe lump them into a "lib/common/errors.js" file or even one for each module: "lib/storage/errors.js". Again though, I'm not even sure *that's* worth the effort.

callmehiphop: IMO if it makes our code base even a little bit cleaner I think it's worth it. I'll open a separate issue

stephenplusplus: SG.

stephenplusplus: We could just undo this change and pass the original error? It's only these situations where many API calls are made that the user doesn't realize. When we get a fail along the way, it would likely make no sense to the user.

callmehiphop: No, I think it's good to be more informative about where it failed.

PR #858, opened 2015-09-11
Fix wrong method in zone.js test

quartzmo:
This appears to be a simple copy/paste error. The enclosing context is `getRecords`, but this test calls `getChanges`.


callmehiphop:
doh! nice catch!

PR #861, opened 2015-09-11
Changing all uses of authorization to authentication.

stephenplusplus:
Following up from #846.
#### To Dos
- [x] Change links in gcloud-common


stephenplusplus:
@dhermes what do we need to change on gcloud-common? (RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/846#issue-104559277)

dhermes:
I think just a similar rename from authorization to authentication.

My merged PR may have already broken a few links by assuming `gcloud-common` also went through the rename.

To find things to be renamed in `gcloud-node` I just did

```
git ls-files | grep -i author
git grep -i author
```

PR #865, opened 2015-09-15
introduce Logging support

### LINKED ISSUE #842: [Feature] gcloud-logging
Is there any plan to include logging api in this package? If so what is the expected timeline?

Link: https://cloud.google.com/logging/docs/api/gcloud-logging

stephenplusplus:
Fixes #842
#### Must Reads
- Introduction to the Cloud Logging API: https://cloud.google.com/logging/docs/api/
- Docs: https://stephenplusplus.github.io/gcloud-node/#/docs/master/logging
#### To Dos
- [x] Make sure ACLs are properly set for Topics: https://github.com/GoogleCloudPlatform/gcloud-node/pull/865#discussion_r47936716
- [x] Enable Cloud Logging API on System Test Account
- [x] Document createSink.config.filter
- [x] Inherit from Service / ServiceObject
- **Support Sink destinations**
  - [x] Storage Bucket
  - [x] BigQuery Dataset
  - [x] Pub/Sub Topic
- **Docs**
  - [x] Readme
  - [x] JSDocs
  - [x] Site
- **Tests**
  - [x] System
  - Unit
    - [x] Index
    - [x] Log
    - [x] Sink
#### Preview

``` js
// Authenticating on a global-basis. You can also authenticate on a per-API-
// basis (see Authentication section above).
  projectId: 'my-project',
});

var logging = gcloud.logging();

// Create a sink using a Bucket as a destination.
var gcs = gcloud.storage();

logging.createSink('my-new-sink', {
  destination: gcs.bucket('my-sink')
}, function(err, sink) {});

// Write a critical entry to a log.
var syslog = logging.log('syslog');

var resource = {
  type: 'gce_instance',
  labels: {
    zone: 'global',
    instance_id: 3
  }
};

var entry = syslog.entry(resource, {
  delegate: process.env.user
});

syslog.critical(entry, function(err) {});

// Get all entries in your project.
logging.getEntries(function(err, entries) {
  if (!err) {
    // `entries` contains all of the entries from the logs in your project.
  }
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ This client supports the following Google Cloud Platform services:
 * [Google Cloud Storage](#google-cloud-storage)
 * [Google Compute Engine](#google-compute-engine)
 * [Google Prediction API](#google-prediction-api)
+* [Google Cloud Logging](#google-cloud-logging-beta) (Beta)
 * [Google Cloud Resource Manager](#google-cloud-resource-manager-beta) (Beta)
 * [Google Cloud Search](#google-cloud-search-alpha) (Alpha)
 
@@ -385,12 +386,62 @@ model.query('Hello', function(err, results) {
     // ]
   }
 });
+
+
+## Google Cloud Logging (Beta)
+
+> **This is a Beta release of Google Cloud Logging.** This API is not covered by any SLA or deprecation policy and may be subject to backward-incompatible changes.
+
+- [API Documentation][gcloud-logging-docs]
+- [Official Documentation][cloud-logging-docs]
+
+```js
+// Authenticating on a global-basis. You can also authenticate on a per-API-
+// basis (see Authentication section above).
+
+var gcloud = require('gcloud')({
+  projectId: 'my-project',
+  keyFilename: '/path/to/keyfile.json'
+});
+
+var logging = gcloud.logging();
+
+// Create a sink using a Bucket as a destination.
+var gcs = gcloud.storage();
+
+logging.createSink('my-new-sink', {
+  destination: gcs.bucket('my-sink')
+}, function(err, sink) {});
+
+// Write a critical entry to a log.
+var syslog = logging.log('syslog');
+
+var resource = {
+  type: 'gce_instance',
+  labels: {
+    zone: 'global',
+    instance_id: 3
+  }
+};
+
+var entry = syslog.entry(resource, {
+  delegate: process.env.user
+});
+
+syslog.critical(entry, function(err) {});
+
+// Get all entries in your project.
+logging.getEntries(function(err, entries) {
+  if (!err) {
+    // `entries` contains all of the entries from the logs in your project.
+  }
+});
 ```
 
 
 ## Google Cloud Resource Manager (Beta)
 
-> This is a *Beta* release of Google Cloud Resource Manager. This feature is not covered by any SLA or deprecation policy and may be subject to backward-incompatible changes.
+> **This is a Beta release of Google Cloud Resource Manager.** This feature is not covered by any SLA or deprecation policy and may be subject to backward-incompatible changes.
 
 - [API Documentation][gcloud-resource-docs]
 - [Official Documentation][cloud-resource-docs]
@@ -426,7 +477,7 @@ project.getMetadata(function(err, metadata) {
 
 ## Google Cloud Search (Alpha)
 
-> This is an *Alpha* release of Google Cloud Search. This feature is not covered by any SLA or deprecation policy and may be subject to backward-incompatible changes.
+> **This is an Alpha release of Google Cloud Search.** This feature is not covered by any SLA or deprecation policy and may be subject to backward-incompatible changes.
 
 - [API Documentation][gcloud-search-docs]
 - [Official Documentation][cloud-search-docs]
@@ -485,6 +536,7 @@ Apache 2.0 - See [COPYING](COPYING) for more information.
 [gcloud-datastore-docs]: https://googlecloudplatform.github.io/gcloud-node/#/docs/datastore
 [gcloud-dns-docs]: https://googlecloudplatform.github.io/gcloud-node/#/docs/dns
 [gcloud-prediction-docs]: https://googlecloudplatform.github.io/gcloud-node/#/docs/prediction
+[gcloud-logging-docs]: https://googlecloudplatform.github.io/gcloud-node/#/docs/logging
 [gcloud-pubsub-docs]: https://googlecloudplatform.github.io/gcloud-node/#/docs/pubsub
 [gcloud-resource-docs]: https://googlecloudplatform.github.io/gcloud-node/#/docs/resource
 [gcloud-search-docs]: https://googlecloudplatform.github.io/gcloud-node/#/docs/search
@@ -513,6 +565,8 @@ Apache 2.0 - See [COPYING](COPYING) for more information.
 [cloud-dns-docs]: https://cloud.google.com/dns/docs
 
 [cloud-prediction-docs]: https://cloud.google.com/prediction/docs
+
+[cloud-logging-docs]: https://cloud.google.com/logging/docs
 
 [cloud-pubsub-docs]: https://cloud.google.com/pubsub/docs
 
--- a/lib/logging/entry.js
+++ b/lib/logging/entry.js
@@ -1,0 +1,140 @@
+/*!
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*!
+ * @module logging/entry
+ */
+
+'use strict';
+
+var extend = require('extend');
+var is = require('is');
+
+/**
+ * Create an entry object to define new data to insert into a log.
+ *
+ * @resource [LogEntry JSON representation]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/LogEntry}
+ *
+ * @param {object=|string=} resource - See a
+ *     [Monitored Resource](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/MonitoredResource).
+ * @param {object|string} data - The data to use as the value for this log
+ *     entry.
+ * @return {module:logging/entry}
+ *
+ * @example
+ * var gcloud = require('gcloud')({
+ *   keyFilename: '/path/to/keyfile.json',
+ *   projectId: 'grape-spaceship-123'
+ * });
+ *
+ * var logging = gcloud.logging();
+ *
+ * var syslog = logging.log('syslog');
+ *
+ * var resource = {
+ *   type: 'gce_instance',
+ *   labels: {
+ *     zone: 'global',
+ *     instance_id: 3
+ *   }
+ * };
+ *
+ * var entry = syslog.entry(resource, {
+ *   delegate: process.env.user
+ * });
+ *
+ * syslog.alert(entry, function(err, apiResponse) {
+ *   if (!error) {
+ *     // Log entry inserted successfully.
+ *   }
+ * });
+ *
+ * //-
+ * // You will also receive `Entry` objects when using
+ * // {module:logging#getEntries} and {module:logging/log#getEntries}.
+ * //-
+ * logging.getEntries(function(err, entries) {
+ *   if (!err) {
+ *     // entries[0].data = The data value from the log entry.
+ *   }
+ * });
+ */
+function Entry(resource, data) {
+  if (!data) {
+    this.data = resource;
+    return;
+  }
+
+  this.resource = resource;
+  this.data = data;
+}
+
+/**
+ * Create an Entry object from an API response, such as `entries:list`.
+ *
+ * @private
+ *
+ * @param {object} entry - An API representation of an entry. See a
+ *     [LogEntry](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/LogEntry).
+ * @return {module:logging/entity}
+ */
+Entry.fromApiResponse_ = function(entry) {
+  // Only one of these are populated.
+  var data = entry.protoPayload || entry.jsonPayload || entry.textPayload;
+  return extend(new Entry(entry.resource, data), entry);
+};
+
+/**
+ * Serialize an entry to the format the API expects.
+ *
+ * @private
+ */
+Entry.prototype.toJSON = function() {
+  var entry = extend(true, {}, this);
+
+  var whitelist = [
+    'logName',
+    'resource',
+    'timestamp',
+    'severity',
+    'insertId',
+    'httpRequest',
+    'labels',
+    'operation'
+  ];
+
+  for (var prop in entry) {
+    if (whitelist.indexOf(prop) === -1) {
+      delete entry[prop];
+    }
+  }
+
+  if (is.string(this.resource)) {
+    entry.resource = {
+      type: this.resource
+    };
+  }
+
+  if (is.object(this.data)) {
+    entry.jsonPayload = this.data;
+  } else if (is.string(this.data)) {
+    entry.textPayload = this.data;
+  }
+
+  return entry;
+};
+
+module.exports = Entry;
--- a/lib/logging/index.js
+++ b/lib/logging/index.js
@@ -1,0 +1,584 @@
+/*!
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*!
+ * @module logging
+ */
+
+'use strict';
+
+var arrify = require('arrify');
+var extend = require('extend');
+var format = require('string-format-obj');
+var is = require('is');
+var nodeutil = require('util');
+
+/**
+ * @type {module:storage/bucket}
+ * @private
+ */
+var Bucket = require('../storage/bucket.js');
+
+/**
+ * @type {module:bigquery/dataset}
+ * @private
+ */
+var Dataset = require('../bigquery/dataset.js');
+
+/**
+ * @type {module:logging/entry}
+ * @private
+ */
+var Entry = require('./entry.js');
+
+/**
+ * @type {module:logging/log}
+ * @private
+ */
+var Log = require('./log.js');
+
+/**
+ * @type {module:common/service}
+ * @private
+ */
+var Service = require('../common/service.js');
+
+/**
+ * @type {module:logging/sink}
+ * @private
+ */
+var Sink = require('./sink.js');
+
+/**
+ * @type {module:common/streamrouter}
+ * @private
+ */
+var streamRouter = require('../common/stream-router.js');
+
+/**
+ * @type {module:pubsub/topic}
+ * @private
+ */
+var Topic = require('../pubsub/topic.js');
+
+/**
+ * @type {module:common/util}
+ * @private
+ */
+var util = require('../common/util.js');
+
+/**
+ * [Google Cloud Logging](https://cloud.google.com/logging/docs) collects and
+ * stores logs from applications and services on the Google Cloud Platform:
+ *
+ *   - Export your logs to Google Cloud Storage, Google BigQuery, or Google
+ *     Cloud Pub/Sub.
+ *   - Integrate third-party logs from your virtual machine instances by
+ *     installing the logging agent, `google-fluentd`.
+ *
+ * @alias module:logging
+ * @constructor
+ *
+ * @resource [What is Google Cloud Logging?]{@link https://cloud.google.com/logging/docs}
+ * @resource [Introduction to the Cloud Logging API]{@link https://cloud.google.com/logging/docs/api}
+ *
+ * @param {object} options - [Configuration object](#/docs/?method=gcloud).
+ *
+ * @example
+ * var gcloud = require('gcloud')({
+ *   keyFilename: '/path/to/keyfile.json',
+ *   projectId: 'grape-spaceship-123'
+ * });
+ *
+ * var logging = gcloud.logging();
+ */
+function Logging(options) {
+  if (!(this instanceof Logging)) {
+    options = util.normalizeArguments(this, options);
+    return new Logging(options);
+  }
+
+  var config = {
+    baseUrl: 'https://logging.googleapis.com/v2beta1',
+    scopes: [
+      'https://www.googleapis.com/auth/cloud-platform'
+    ]
+  };
+
+  Service.call(this, config, options);
+
+  // API requests to `entries` shouldn't contain a project ID in the URI, as it
+  // is passed in the JSON body when used to list or in the logName when used to
+  // write entries.
+  this.interceptors.push({
+    request: function(reqOpts) {
+      if (reqOpts.uri.indexOf('/entries:') > -1) {
+        reqOpts.uri = reqOpts.uri.replace(/projects\/[^\/]*\//, '');
+      }
+
+      return reqOpts;
+    }
+  });
+}
+
+nodeutil.inherits(Logging, Service);
+
+// jscs:disable maximumLineLength
+/**
+ * Create a sink.
+ *
+ * **This method only works if you are authenticated as yourself, e.g. using the
+ * gcloud SDK.**
+ *
+ * @resource [Sink Overview]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/projects.sinks}
+ * @resource [Advanced Logs Filters]{@link https://cloud.google.com/logging/docs/view/advanced_filters}
+ * @resource [projects.sinks.create API Documentation]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/projects.sinks/create}
+ *
+ * @throws {Error} if a name is not provided.
+ * @throws {Error} if a config object is not provided.
+ *
+ * @param {string} name - Name of the sink.
+ * @param {object} config - See a
+ *     [Sink resource](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/projects.sinks#LogSink).
+ * @param {module:storage/bucket|module:bigquery/dataset|module:pubsub/topic} config.destination -
+ *     The destination. The proper ACL scopes will be granted to the provided
+ *     destination.
+ * @param {string=} config.filter - An advanced logs filter. Only log entries
+ *     matching the filter are written.
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request.
+ * @param {module:logging/sink} callback.sink - The created Sink object.
+ * @param {object} callback.apiResponse - The full API response.
+ *
+ * @example
+ * var gcs = gcloud.storage();
+ *
+ * var config = {
+ *   destination: gcs.bucket('logging-bucket'),
+ *   filter: 'severity = ALERT'
+ * };
+ *
+ * function callback(err, sink, apiResponse) {
+ *   // `sink` is a Sink object.
+ * }
+ *
+ * logging.createSink('new-sink-name', config, callback);
+ */
+Logging.prototype.createSink = function(name, config, callback) {
+  // jscs:enable maximumLineLength
+  var self = this;
+
+  if (!is.string(name)) {
+    throw new Error('A sink name must be provided.');
+  }
+
+  if (!is.object(config)) {
+    throw new Error('A sink configuration object must be provided.');
+  }
+
+  if (config.destination instanceof Bucket) {
+    this.setAclForBucket_(name, config, callback);
+    return;
+  }
+
+  if (config.destination instanceof Dataset) {
+    this.setAclForDataset_(name, config, callback);
+    return;
+  }
+
+  if (config.destination instanceof Topic) {
+    this.setAclForTopic_(name, config, callback);
+    return;
+  }
+
+  this.request({
+    method: 'POST',
+    uri: '/sinks',
+    json: extend({}, config, {
+      name: name
+    })
+  }, function(err, resp) {
+    if (err) {
+      callback(err, null, resp);
+      return;
+    }
+
+    var sink = self.sink(resp.name);
+    sink.metadata = resp;
+
+    callback(null, sink, resp);
+  });
+};
+
+/**
+ * Create an entry object.
+ *
+ * Note that using this method will not itself make any API requests. You will
+ * use the object returned in other API calls, such as
+ * {module:logging/log#write}.
+ *
+ * @resource [LogEntry JSON representation]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/LogEntry}
+ *
+ * @param {object=|string=} resource - See a
+ *     [Monitored Resource]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/MonitoredResource}.
+ * @param {object|string} data - The data to use as the value for this log
+ *     entry.
+ * @return {module:logging/entry}
+ *
+ * @example
+ * var resource = {
+ *   type: 'gce_instance',
+ *   labels: {
+ *     zone: 'global',
+ *     instance_id: 3
+ *   }
+ * };
+ *
+ * var entry = logging.entry(resource, {
+ *   delegate: process.env.USER
+ * });
+ *
+ * entry.toJSON();
+ * // {
+ * //   resource: {
+ * //     type: 'gce_instance',
+ * //     labels: {
+ * //       zone: 'global',
+ * //       instance_id: 3
+ * //     }
+ * //   },
+ * //   jsonPayload: {
+ * //     delegate: process.env.USER
+ * //   }
+ * // }
+ */
+Logging.prototype.entry = function(resource, data) {
+  return new Entry(resource, data);
+};
+
+/**
+ * List the entries in your logs.
+ *
+ * @resource [entries.list API Documentation]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/entries/list}
+ *
+ * @param {object=} options - Filtering options.
+ * @param {boolean} options.autoPaginate - Have pagination handled
+ *     automatically. Default: true.
+ * @param {string} options.filter - An
+ *     [advanced logs filter](https://cloud.google.com/logging/docs/view/advanced_filters).
+ *     An empty filter matches all log entries.
+ * @param {string} options.orderBy - How the results should be sorted,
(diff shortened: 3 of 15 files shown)


stephenplusplus:
@VikramTiwari - your thoughts are most welcome if you have the chance to play around with this!

```sh
$ npm install --save stephenplusplus/gcloud-node#spp--logging
```

VikramTiwari:
Hey @stephenplusplus I am at Chrome Dev Summit for today and tomorrow, but I will be testing it by the end of week for sure. :+1:

stephenplusplus:
@VikramTiwari awesome! I'll look for you on the feed :stars:

stephenplusplus:
@jgeewax to prepare for this to merge, can you check that the Logging API is enabled for the Travis project?

stephenplusplus:
Also, any ideas for getting Travis to auth as a user? Many of our system tests require user-level, non-service account auth.

lib/logging/index.js
```diff
@@ -0,0 +1,543 @@
...
+      self.createSink(name, config, callback);
+    });
+
+    return;
+  }
+
+  if (config.destination instanceof Dataset) {
+    var dataset = config.destination;
+
+    config.destination = format('{baseUrl}/projects/{pId}/datasets/{dId}', {
```
filipjs: Datasets and topics (below) also need ACLs set up.
https://cloud.google.com/logging/docs/export/configure_export#setting_product_name_short_permissions_for_writing_exported_logs

lib/logging/index.js
```diff
@@ -0,0 +1,543 @@
...
+      baseUrl: 'bigquery.googleapis.com',
+      pId: dataset.parent.projectId,
+      dId: dataset.id
+    });
+  }
+
+  if (config.destination instanceof Topic) {
+    var topic = config.destination;
+
+    config.destination = format('{baseUrl}/{topicName}', {
```
filipjs: Does topicName include project?
Full destination should be {baseUrl}/projects/{pId}/topics/{tId}'

lib/logging/index.js
```diff
@@ -0,0 +1,543 @@
...
+    uri: '/sinks'
+  }, function(err, resp) {
+    if (err) {
+      callback(err, null, null, resp);
+      return;
+    }
+
+    var nextQuery = null;
+
+    if (resp.nextPageToken) {
```
filipjs: v1beta3 doesn't support ListSinks pagination

lib/logging/log.js
```diff
@@ -0,0 +1,415 @@
...
+var through = require('through2');
+
+/**
+ * @type {module:common/serviceObject}
+ * @private
+ */
+var ServiceObject = require('../common/service-object.js');
+
+/**
+ * A log is a named collection of records or entries, each entry representing a
```
filipjs: remove "records or" ?

lib/logging/log.js
```diff
@@ -0,0 +1,415 @@
...
+    delete: true
+  };
+
+  this.name = name.replace(/^.*logs\//, '');
+  this.unformattedName = name;
+
+  ServiceObject.call(this, {
+    parent: logging,
+    baseUrl: '/logs',
+    id: this.name,
```
filipjs: Shouldn't "unformattedName" be used here? 
That variable is not used otherwise

lib/logging/log.js
```diff
@@ -0,0 +1,415 @@
...
+ *   metadata: {
+ *     serviceName: 'compute.googleapis.com'
+ *   },
+ *   data: {
+ *     delegate: process.env.USER
+ *   }
+ * }, function(err, apiResponse) {});
+ */
+Log.prototype.alert = function(entry, labels, callback) {
+  this.insert(Log.assignSeverityToEntries_(entry, 'ALERT'), labels, callback);
```
filipjs: This doesn't match the example (there are no labels)?
Does "assignSeverityToEntries" work correctly with a single entry? (not sure how "arrify" works)

lib/logging/log.js
```diff
@@ -0,0 +1,415 @@
...
+ *   metadata: {
+ *     serviceName: 'compute.googleapis.com'
+ *   },
+ *   data: {
+ *     delegate: process.env.USER
+ *   }
+ * }, function(err, apiResponse) {});
+ */
+Log.prototype.critical = function(entry, labels, callback) {
+  var entries = Log.assignSeverityToEntries_(entry, 'CRITICAL');
```
filipjs: Same comment as in ".alert" wrapper.
Also, this code can be changed into 1 line like there :)

lib/logging/log.js
```diff
@@ -0,0 +1,415 @@
...
+Log.prototype.info = function(entry, labels, callback) {
+  this.insert(Log.assignSeverityToEntries_(entry, 'INFO'), labels, callback);
+};
+
+/**
+ * Write log entries to Cloud Logging.
+ *
+ * @resource [projects.logs.entries.write API Documentation]{@link https://cloud.google.com/logging/docs/api/ref/rest/v1beta3/projects.logs.entries/write}
+ *
+ * @param {object|object[]} entry - A log entry, or array of entries, to write.
```
filipjs: Maybe it's worth adding that batching entries is highly recommended, as the is a QPS limit.

lib/logging/log.js
```diff
@@ -0,0 +1,415 @@
...
+/**
+ * All entries are passed through here. This will clone the object and properly
+ * format it for the API.
+ *
+ * @private
+ *
+ * @param {object} entry - An entry object.
+ */
+Log.prototype.formatEntry_ = function(entry) {
+  entry = extend(true, {}, entry);
```
filipjs: What does this line do?

lib/logging/sink.js
```diff
@@ -0,0 +1,214 @@
...
+     *   // ...
+     * };
+     *
+     * sink.create(config, function(err, sink, apiResponse) {
+     *   if (!err) {
+     *     // The sink was created successfully.
+     *   }
+     * });
+     */
+    create: true,
```
filipjs: no update method?

lib/logging/sink.js
```diff
@@ -0,0 +1,214 @@
...
+  callback = callback || util.noop;
+
+  this.getMetadata(function(err, currentMetadata, apiResponse) {
+    if (err) {
+      callback(err, apiResponse);
+      return;
+    }
+
+    self.request({
+      uri: '',
```
filipjs: empty uri?

system-test/logging.js
```diff
@@ -0,0 +1,457 @@
...
+          this.skip();
+          return;
+        }
+
+        sink.delete(done);
+      });
+
+      it('should set metadata', function(done) {
+        var metadata = {
+          filter: 'name:*'
```
filipjs: those are unit tests, or against a live system?
if the later, this should fail, as that is an invalid filter
(filter must match an existing LogEntry field).

system-test/logging.js
```diff
@@ -0,0 +1,457 @@
...
+    });
+
+    it('should write from a stream', function(done) {
+      log.createWriteStream()
+        .on('error', done)
+        .on('finish', done)
+        .end(logEntry);
+    });
+  });
+
```
filipjs: A test is missing that exercise the batching path

test/logging/log.js
```diff
@@ -0,0 +1,534 @@
...
+        assert.strictEqual(reqOpts.json.commonLabels, LABELS);
+        assert.strictEqual(reqOpts.json.entries[0], formattedEntry);
+
+        done();
+      };
+
+      log.insert(ENTRY, LABELS, assert.ifError);
+    });
+
+    it('should exec callback with only error and API response', function(done) {
```
filipjs: Can you expand what is this testing?

test/logging/sink.js
```diff
@@ -0,0 +1,200 @@
...
+      assert.strictEqual(calledWith.baseUrl, '/sinks');
+      assert.strictEqual(calledWith.id, SINK_NAME);
+      assert.deepEqual(calledWith.methods, {
+        create: true,
+        delete: true,
+        getMetadata: true
+      });
+    });
+
+    it('should localize the name', function() {
```
filipjs: what is "localize" in this context?

test/logging/sink.js
```diff
@@ -0,0 +1,200 @@
...
+      assert.strictEqual(sink.name, SINK_NAME);
+    });
+  });
+
+  describe('setFilter', function() {
+    var FILTER = 'filter';
+
+    it('should call set metadata', function(done) {
+      sink.setMetadata = function(metadata, callback) {
+        assert.strictEqual(metadata.filter, FILTER);
```
filipjs: Add a check that destination is still there and unchanged

test/logging/sink.js
```diff
@@ -0,0 +1,200 @@
...
+      };
+
+      sink.setFilter(FILTER, done);
+    });
+  });
+
+  describe('setMetadata', function() {
+    var METADATA = { a: 'b', c: 'd' };
+
+    it('should refresh the metadata', function(done) {
```
filipjs: what is "refresh"?

filipjs:
Some initial comments.
One thing I would really suggest is to wait for new version API, to save the headache of updating everything twice (changes shouldn't be that big)

lib/logging/index.js
stephenplusplus: Yep, `topic.name` is its formatted name: https://github.com/GoogleCloudPlatform/gcloud-node/blob/1ac48c3177010f04c72acfcf7f36c76f45cadacc/lib/pubsub/topic.js#L216

lib/logging/index.js
stephenplusplus: Are there plans to add pagination support?

lib/logging/log.js
stephenplusplus: > This doesn't match the example (there are no labels)?

I try to keep the examples as "look how simple this is" as possible, catering to the common use case. Is it more common to include labels with every entry?

> Does "assignSeverityToEntries" work correctly with a single entry? (not sure how "arrify" works)

Yes: https://github.com/sindresorhus/arrify/blob/master/index.js

lib/logging/log.js
stephenplusplus: Yep, so the user's given object isn't affected.

lib/logging/sink.js
stephenplusplus: Our library patches `request`s upwards through the inheritance model. `request.uri` in this context will resolve to `https://logging.googleapis.com/v1beta3/projects/{projectId}/sinks/{thisSinkName}`. Anything specified in `uri` would get appended to that base uri.

system-test/logging.js
stephenplusplus: This is against the real backend. Interesting that it doesn't fail, I confirm that it does in fact update the `filter` on the response object: https://github.com/stephenplusplus/gcloud-node/blob/310017043fba58daad17834e0b95cff9ba615229/system-test/logging.js#L287

test/logging/log.js
stephenplusplus: Sure, specifically, this line: https://github.com/stephenplusplus/gcloud-node/blob/310017043fba58daad17834e0b95cff9ba615229/lib/logging/log.js#L347 -- we want to make sure *only* two total arguments are passed to the user and they are the error and response body from the API request.

test/logging/sink.js
stephenplusplus: Store it in memory as a property on the created object. "sink.name"

test/logging/sink.js
stephenplusplus: This is a unit test, so it's strictly testing this body of code: https://github.com/stephenplusplus/gcloud-node/blob/310017043fba58daad17834e0b95cff9ba615229/lib/logging/sink.js#L158-L160

test/logging/sink.js
stephenplusplus: Because the update (https://cloud.google.com/logging/docs/api/ref/rest/v1beta3/projects.sinks/update) is a PUT vs a PATCH, our API combines the current state of metadata with the input to `setMetadata`, combining them, effectively re-creating PATCH semantics.

Refresh means we get the latest state of the metadata before making the update. This test makes sure we do that.

stephenplusplus:
Thanks for the review! I will be addressing the feedback and making the changes in the near future.

Regarding the new API version, is there a estimate on when that will be publicly available?

lib/logging/index.js
filipjs: Yes, it's added in V2

lib/logging/log.js
filipjs: What I meant with the first question is: shouldn't the function call in example take 3 arguments?
(because alert() takes 3 arguments). 
I guess this works because I was the same in tests

system-test/logging.js
filipjs: Oh yes, this change wasn't released yet. It should start to fail next week :)

lib/logging/log.js
stephenplusplus: Yes, the `labels` argument is marked as optional in the parameter descriptions, so I chose not to show it in the example for the reasons above.

system-test/logging.js
stephenplusplus: Is that part of a new API release?

filipjs: No, this will fail in both API versions.

test/logging/sink.js
```diff
@@ -0,0 +1,200 @@
...
+
+      sink.setMetadata(METADATA, function(err, apiResponse_) {
+        assert.strictEqual(err, error);
+        assert.strictEqual(apiResponse_, apiResponse);
+        done();
+      });
+    });
+
+    it('should make the correct request', function(done) {
+      var currentMetadata = { e: 'f', g: 'h' };
```
filipjs: Thanks for the comments above!
In which case, maybe change currentMetadata to e.g. { 'c': 'e', 'g': 'h'} so the intersection of current and new metadata is not empty

lib/logging/log.js
stephenplusplus: This relates to how requests are patched upwards through our library, continuously appending pieces of the URL. I'm only passing `name` here, as it's already being appended to:

`...com/v1beta3/projects/{projectId}/logs/`

A user can refer to their log only by name ('my-log'), instead of a fully qualified name including the project ID, and it will work the same.

RE: `unformattedName` good catch, will remove.

stephenplusplus:
@filipjs thanks again for the review! I made many of the changes in a follow-up commit: https://github.com/stephenplusplus/gcloud-node/commit/40c860dde83f91ae13d338fee97c45612a92b546, please let me know if you catch anything else I can change.

I'm running into a problem creating sinks now. Hopefully it's a simple mistake on my part, but I'm seeing the "The caller does not have permission" error. Here is a request that's failing:

```
{ method: 'POST',
  uri: 'https://logging.googleapis.com/v1beta3/projects/nth-circlet-705/sinks',
  json:
   { destination: 'storage.googleapis.com/gcloud-logging-test7bb30900-99fb-11e5-ba9b-f7178516428f',
     name: 'gcloud-logging-test7cced940-99fb-11e5-ba9b-f7178516428f' },
  headers:
   { Authorization: 'Bearer {token}',
     'User-Agent': 'gcloud-node/0.25.1' } }
```

Any ideas?

filipjs:
Re: failing test.
Is the auth token refreshed between tests? Maybe it just expired?

stephenplusplus:
I don't believe so, even running the test in isolation fails. We also use [google-auth-library](https://github.com/google/google-auth-library-nodejs) which makes sure the token isn't expired before each request is sent.

lib/logging/index.js
stephenplusplus: Thanks for catching that. I fixed this in https://github.com/GoogleCloudPlatform/gcloud-node/commit/77349e6f6b1daa467a2795ac49a4344cfe8d06b6

filipjs:
V2 is out
https://cloud.google.com/logging/docs/api/introduction_v2

stephenplusplus:
@filipjs congrats on the release and thanks for letting me know. I'll get started on the migration.

stephenplusplus:
Migration complete. PTAL @filipjs. Here's the diff commit: 

I've created a new [`Entry`](https://stephenplusplus.github.io/gcloud-node/#/docs/master/logging/entry) type. Here's an example to write a new entry.

```js
var syslog = logging.log('syslog');

var resource = {
  type: 'gce_instance',
  labels: {
    zone: 'global',
    instance_id: 3
  }
};

var entry = syslog.entry(resource, {
  delegate: process.env.user
});

syslog.alert(entry, function(err, apiResponse) {
  if (!error) {
    // Log entry inserted successfully.
  }
});
```

Entry objects also come back from calling `logging.getEntries()` & `log.getEntries()`. To read the data, you can call `entry.data`.

Looking forward to your review. Thanks!

stephenplusplus:
@callmehiphop PTAL.

README.md
```diff
@@ -13,6 +13,7 @@ This client supports the following Google Cloud Platform services:
 * [Google BigQuery](#google-bigquery)
 * [Google Cloud Datastore](#google-cloud-datastore)
 * [Google Cloud DNS](#google-cloud-dns)
+* [Google Cloud Logging](#google-cloud-logging)
```
callmehiphop: Should we denote that this is API is still in beta? I believe for pubsub we used to add the text *(beta)* after the link.

lib/logging/entry.js
```diff
@@ -0,0 +1,140 @@
...
+ * //-
+ * logging.getEntries(function(err, entries) {
+ *   if (!err) {
+ *     // entries[0].data = The data value from the log entry.
+ *   }
+ * });
+ */
+function Entry(resource, data) {
+  if (!data) {
+    this.data = resource;
```
callmehiphop: Kind of curious, is `resource.type` optional?

lib/logging/entry.js
```diff
@@ -0,0 +1,140 @@
...
+ * var gcloud = require('gcloud')({
+ *   keyFilename: '/path/to/keyfile.json',
+ *   projectId: 'grape-spaceship-123'
+ * });
+ *
+ * var logging = gcloud.logging();
+ *
+ * var syslog = logging.log('syslog');
+ *
+ * var resource = {
```
callmehiphop: I think an additional example of using an entry in its most simple form would be nice here. 

e.g.

```javascript
var entry = syslog.entry('thing1', 'thing2');
```

README.md
stephenplusplus: Yeah, good call!

lib/logging/log.js
```diff
@@ -0,0 +1,531 @@
...
+    projectId: projectId,
+    name: uriSafeName
+  });
+};
+
+/**
+ * Write a log entry with a severity of "ALERT".
+ *
+ * This is a simple wrapper around {module:logging/log#insert}. All arguments
+ * are the same as documented there.
```
callmehiphop: Could we just do a `@mixin` here instead? It might be a little tedious to jump from method to method? Especially if the method has been deep linked into.

lib/logging/log.js
```diff
@@ -0,0 +1,531 @@
...
+ *
+ * log.alert(entry, function(err, apiResponse) {});
+ */
+Log.prototype.alert = function(entry, options, callback) {
+  this.write(Log.assignSeverityToEntries_(entry, 'ALERT'), options, callback);
+};
+
+/**
+ * Write log entries from a streaming object input.
+ *
```
callmehiphop: Maybe a little bit of a nitpick and sort of unnecessary, but could we add a `@return` block to specify a stream is returned?

lib/logging/log.js
```diff
@@ -0,0 +1,531 @@
...
+    ]);
+  });
+
+  return writeStream;
+};
+
+/**
+ * Write a log entry with a severity of "CRITICAL".
+ *
+ * This is a simple wrapper around {module:logging/log#insert}. All arguments
```
callmehiphop: see https://github.com/GoogleCloudPlatform/gcloud-node/pull/865/files#r47702149

README.md
stephenplusplus: Done!

lib/logging/log.js
```diff
@@ -0,0 +1,531 @@
...
+Log.prototype.critical = function(entry, options, callback) {
+  var entries = Log.assignSeverityToEntries_(entry, 'CRITICAL');
+  this.write(entries, options, callback);
+};
+
+/**
+ * Write a log entry with a severity of "DEBUG".
+ *
+ * This is a simple wrapper around {module:logging/log#insert}. All arguments
+ * are the same as documented there.
```
callmehiphop: see https://github.com/GoogleCloudPlatform/gcloud-node/pull/865/files#r47702149

lib/logging/log.js
```diff
@@ -0,0 +1,531 @@
...
+Log.prototype.entry = function(resource, data) {
+  var entryInstance = this.parent.entry(resource, data);
+  entryInstance.logName = this.formattedName_;
+  return entryInstance;
+};
+
+/**
+ * Write a log entry with a severity of "ERROR".
+ *
+ * This is a simple wrapper around {module:logging/log#insert}. All arguments
```
callmehiphop: see https://github.com/GoogleCloudPlatform/gcloud-node/pull/865/files#r47702149

lib/logging/log.js
```diff
@@ -0,0 +1,531 @@
...
+    filter: format('logName="{formattedName_}"', this)
+  }, options);
+
+  return this.parent.getEntries(options, callback);
+};
+
+/**
+ * Write a log entry with a severity of "INFO".
+ *
+ * This is a simple wrapper around {module:logging/log#insert}. All arguments
```
callmehiphop: see https://github.com/GoogleCloudPlatform/gcloud-node/pull/865/files#r47702149

lib/logging/log.js
```diff
@@ -0,0 +1,531 @@
...
+Log.prototype.info = function(entry, options, callback) {
+  this.write(Log.assignSeverityToEntries_(entry, 'INFO'), options, callback);
+};
+
+/**
+ * Write a log entry with a severity of "NOTICE".
+ *
+ * This is a simple wrapper around {module:logging/log#insert}. All arguments
+ * are the same as documented there.
+ *
```
callmehiphop: see https://github.com/GoogleCloudPlatform/gcloud-node/pull/865/files#r47702149

lib/logging/log.js
```diff
@@ -0,0 +1,531 @@
...
+ */
+Log.prototype.notice = function(entry, options, callback) {
+  this.write(Log.assignSeverityToEntries_(entry, 'NOTICE'), options, callback);
+};
+
+/**
+ * Write a log entry with a severity of "WARNING".
+ *
+ * This is a simple wrapper around {module:logging/log#insert}. All arguments
+ * are the same as documented there.
```
callmehiphop: see https://github.com/GoogleCloudPlatform/gcloud-node/pull/865/files#r47702149

lib/logging/log.js
```diff
@@ -0,0 +1,531 @@
...
+ *
+ * While you may write a single entry at a time, batching multiple entries
+ * together is preferred to avoid reaching the queries per second limit.
+ *
+ * @resource [entries.write API Documentation]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/entries/write}
+ *
+ * @param {module:logging/entry|module:logging/entry[]} entry - A log entry, or
+ *     array of entries, to write.
+ * @param {object=} options - Configuration object.
+ * @param {object[]} labels - Labels to set on the log.
```
callmehiphop: Are these supposed to be `options.labels` and `options.resource`?

lib/logging/log.js
stephenplusplus: That's a good idea, but our docs FE doesn't handle mixins on the method sharing method level. It was written for classes to extend classes. I'll change these methods to point to the correct method (these are all pointed to the old method "insert" which is now "write"), but maybe save the docs change for another issue or the common docs overhaul?

lib/logging/log.js
stephenplusplus: Yep!

lib/logging/log.js
stephenplusplus: Done.

lib/logging/entry.js
stephenplusplus: It is not!

stephenplusplus: To elaborate, when you write an entry to a log, you can use a common resource that is shared for all of the entries. That's why it's optional here, so that the user doesn't have to pass in the same resource object to every new entry

callmehiphop: Ok, then correct me if I'm wrong, but if we're allowing for a single argument and setting it to `data`, wouldn't that imply that resource is optional?

callmehiphop: Ah, ok.

lib/logging/entry.js
stephenplusplus: Sadly, I'm not sure I can find one of those. This is pretty much how they're going to look.

callmehiphop:
@stephenplusplus LGTM! Did you want to squash this and/or wait for another review from @filipjs?

stephenplusplus:
Squashed! Let's wait for a little bit to see if we get a thumbs up from @filipjs.

lib/logging/entry.js
callmehiphop: I suppose I'm a little confused then, per https://github.com/GoogleCloudPlatform/gcloud-node/pull/865/files#diff-7d991e86dd4fed14a9532c6a8b99291cR242 it seemed like that was possible.

jgeewax:
Is it worthwhile to create a separate issue to add a bit more color to the overview, defining what the different pieces are and what the expected usage should be like? (My eyes are still fresh so I'm avoiding reading too much about Sinks and Logs and etc, so that I can give useful feedback on whether our docs help new people go "oh, I get it")

stephenplusplus:
Definitely. It might be nice to get some common use cases from a Logging team member as well, so we can center some examples around them.

README.md
```diff
@@ -337,9 +338,61 @@ zone.createVM(name, { os: 'ubuntu' }, function(err, vm, operation) {
...
+  projectId: 'my-project',
+  keyFilename: '/path/to/keyfile.json'
+});
+
+var logging = gcloud.logging();
+
+// Create a sink using a Bucket as a destination.
+var gcs = gcloud.storage();
+
+dns.createSink('my-new-sink', {
```
filipjs: What's "dns" here?

stephenplusplus: A mistake! Thanks for catching that.

README.md
```diff
@@ -337,9 +338,61 @@ zone.createVM(name, { os: 'ubuntu' }, function(err, vm, operation) {
...
+    zone: 'global',
+    instance_id: 3
+  }
+};
+
+var entry = syslog.entry(resource, {
+  delegate: process.env.user
+});
+
+log.critical(entry, function(err) {});
```
filipjs: log -> syslog?

lib/logging/index.js
```diff
@@ -0,0 +1,576 @@
...
+ * @param {?error} callback.err - An error returned while making this request.
+ * @param {module:logging/sink} callback.sink - The created Sink object.
+ * @param {object} callback.apiResponse - The full API response.
+ *
+ * @example
+ * var gcs = gcloud.storage();
+ *
+ * var config = {
+ *   destination: gcs.bucket('logging-bucket'),
+ *   filter: 'metadata.severity = ALERT'
```
filipjs: In V2:
metadata.severity => severity

lib/logging/index.js
```diff
@@ -0,0 +1,576 @@
...
+ * // {
+ * //   resource: {
+ * //     type: 'gce_instance',
+ * //     labels: {
+ * //       zone: 'global',
+ * //       instance_id: 3
+ * //     }
+ * //   },
+ * //   jsonPayload: {
+ * //     delgate: process.env.USER
```
filipjs: delgate => delegate

lib/logging/index.js
```diff
@@ -0,0 +1,576 @@
...
+
+/**
+ * This method is called when creating a sink with a Topic destination. The
+ * topic must first grant proper ACL access to the Cloud Logging account.
+ *
+ * The parameters are the same as what {module:logging#createSink} accepts.
+ *
+ * @private
+ */
+Logging.prototype.setAclForTopic_ = function(name, config, callback) {
```
filipjs: I need to check if this will work, as it differs from our docs.
I'll get back to you.

lib/logging/log.js
```diff
@@ -0,0 +1,549 @@
...
+ * var gcloud = require('gcloud')({
+ *   keyFilename: '/path/to/keyfile.json',
+ *   projectId: 'grape-spaceship-123'
+ * });
+ *
+ * var logging = gcloud.logging();
+ * var log = logging.log('syslog');
+ */
+function Log(logging, name) {
+  var methods = {
```
filipjs: There will be a "list logs" command added

stephenplusplus: Where can I find the API docs? Looking here, I only see delete: https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/projects.logs

lib/logging/log.js
```diff
@@ -0,0 +1,549 @@
...
+    /**
+     * Delete the log.
+     *
+     * @resource [projects.logs.delete API Documentation]{@link https://cloud.google.com/logging/docs/api/ref/rest/v2beta1/projects.logs/delete}
```
filipjs: Unfortunately the docs path is a little different from V1:
https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/projects.logs/delete

lib/logging/log.js
filipjs: id: this.name => name has to be already encoded, which might not be the case here if user forgets to do it.

lib/logging/log.js
```diff
@@ -0,0 +1,549 @@
...
+/**
+ * Format the name of a log. A log's full name is in the format of
+ * 'projects/{projectId}/logs/{logName}'.
+ *
+ * @private
+ *
+ * @return {string}
+ */
+Log.formatName_ = function(projectId, name) {
+  if (name.indexOf('logs/') > -1) {
```
filipjs: This path returns a name which might not be escaped, e.g.:
projects/X/logs/a/b

lib/logging/log.js
```diff
@@ -0,0 +1,549 @@
...
+ * @param {object=} options - Configuration object.
+ * @param {object[]} options.labels - Labels to set on the log.
+ * @param {object} options.resource - A default monitored resource for entries
+ *     where one isn't specified.
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request.
+ * @param {object} callback.apiResponse - The full API response.
+ *
+ * @example
+ * var entry = log.entry('compute.googleapis.com', {
```
filipjs: This code section still uses "services" (here in commends and in the method below).

lib/logging/index.js
```diff
@@ -0,0 +1,576 @@
...
+
+/**
+ * List the entries in your logs.
+ *
+ * @resource [entries.list API Documentation]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/entries/list}
+ *
+ * @param {object=} options - Services filtering options.
+ * @param {boolean} options.autoPaginate - Have pagination handled
+ *     automatically. Default: true.
+ * @param {number} options.pageSize - Maximum number of logs to return.
```
filipjs: What are the default values if not provided in the options?

lib/logging/index.js
```diff
@@ -0,0 +1,576 @@
...
+Logging.prototype.entry = function(resource, data) {
+  return new Entry(resource, data);
+};
+
+/**
+ * List the entries in your logs.
+ *
+ * @resource [entries.list API Documentation]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/entries/list}
+ *
+ * @param {object=} options - Services filtering options.
```
filipjs: The most important option "filter" is missing from the options :)
Also it's rarely a good idea to pass an empty one, as it will return 30 days of logs starting from oldest ones.

lib/logging/entry.js
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Serialize an entry to the format the API expects.
+ *
+ * @private
+ */
+Entry.prototype.toJSON = function() {
+  var entry = extend(true, {}, this);
+
+  var whitelist = [
+    'logName',
```
filipjs: last part of "logName" must be URL escaped, not sure if it's better to do it here, or in the constructor of Entry()

lib/logging/log.js
filipjs: There are no docs yet, but it's going to be pretty much similar to V1
https://cloud.google.com/logging/docs/api/ref/rest/v1beta3/projects.logs/list
with the service replaced with monitored resource

lib/logging/log.js
stephenplusplus: Does this look correct?:

```js
Log.formatName_ = function(projectId, name) {
  var path = 'projects/' + projectId + '/logs/';
  name = name.replace(path, '');

  if (decodeURIComponent(name) === name) {
    // The name has not been encoded yet.
    name = encodeURIComponent(name);
  }

  return path + '/' + name;
};
```

lib/logging/index.js
stephenplusplus: They're not defined, so it lets the API decide.

lib/logging/index.js
stephenplusplus: Good catch, and done! We'll default to `timestamp desc` now.

lib/logging/entry.js
stephenplusplus: For an Entry, it either comes from the API or a user. But when a user creates it or uses it to write to a log, we assign the log name automatically from the `formattedName_` property on the Log that the entry is passed through.

Does the API always return URL escaped names?

lib/logging/log.js
stephenplusplus: I'm getting a 404 on https://logging.googleapis.com/v2beta1/projects/{projectId}/logs. Is this part of the next release for the API?

lib/logging/entry.js
filipjs: Yes it does

lib/logging/log.js
filipjs: Yes. What I meant is that it's not ready yet, but it will be there soon

lib/logging/log.js
filipjs: It would be safer to use a regexp? If name has different projectId, the replace won't match.

lib/logging/log.js
stephenplusplus: I'll make an issue to support it in case this PR lands before then.

lib/logging/log.js
stephenplusplus: All Log objects are scoped to a projectId that's provided up front to the parent Logging instance. Is this a possible scenario?

```js
var logging = gcloud.logging({ projectId: 'project-1' });
var log = logging.log('projects/project-2/logs/log-name');
```

filipjs: Based on the method definition, yes. Nothing stops the user to give invalid data?
(unless the user cannot call this method directly, in which case why not say that name is only the  part after projects/*/logs/ ?

stephenplusplus: The user cannot call this method directly, it's called when they access a Log object:

```js
var logging = gcloud.logging({ projectId: 'project' });
var log = logging.log('log-name'); // Logging.formatName_ is called behind the scenes here
```

> Based on the method definition, yes. 

I'm asking based on use case. Is there a user case to access a log from another project? Is it possible?

> why not say that name is only the part after projects/*/logs/

That's possible, but this caters to an API response which is in the full format. It also allows a user to use whichever version of the name (formatted or short) that they may be working with in other parts of their code.

lib/logging/entry.js
stephenplusplus: Cool, I think we're covered then.

lib/logging/log.js
stephenplusplus: Fixed.

lib/logging/index.js
filipjs: Yes, this will work.

lib/logging/index.js
```diff
@@ -0,0 +1,584 @@
...
+  topic.iam.getPolicy(function(err, policy, apiResp) {
+    if (err) {
+      callback(err, null, apiResp);
+      return;
+    }
+
+    policy.bindings = arrify(policy.bindings);
+
+    policy.bindings.push({
+      role: 'roles/editor',
```
filipjs: roles/pubsub.publisher is enough
https://cloud.google.com/pubsub/access_control?hl=en#tbl_roles

stephenplusplus: Great! Updated.

callmehiphop:
Thanks @stephenplusplus! Also, thanks for the review @filipjs!

PR #870, opened 2015-09-16
Bring test coverage to 100%

stephenplusplus:
... I think. Running `npm cover` doesn't work on my end.

But, I went over all of these < 100% files: https://coveralls.io/builds/3590802 and covered up the holes.

Along the way, I also caught some places we were inconsistent with our `setMetadata` method. The signature is supposed to be:

``` js
object.setMetadata({}, function(err, apiResponse) {});
```

But some places, it was:

``` js
object.setMetadata({}, function(err, metadata, apiResponse) {});
```

Coincidentally, `apiResponse` and `metadata` would have been the same thing-- the apiResponse _is_ what becomes the metadata. Now, however, it's only `function(err, apiResponse) {}` so that if an error occurs, the handler still gets the benefit of receiving the API response.

### DIFF
--- a/lib/bigquery/dataset.js
+++ b/lib/bigquery/dataset.js
@@ -159,7 +159,7 @@ Dataset.prototype.delete = function(options, callback) {
  *
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request
- * @param {object} callback.metadata - The Dataset metadata
+ * @param {object} callback.metadata - The dataset's metadata.
  * @param {object} callback.apiResponse - The full API response.
  *
  * @example
@@ -284,28 +284,27 @@ Dataset.prototype.query = function(options, callback) {
  * @param {object} metadata - Metadata to save on the Dataset.
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request
- * @param {object} callback.metadata - The metadata of the Dataset object.
- * @param {object} callback.apiResponse - The full API response.
+  * @param {object} callback.apiResponse - The full API response.
  *
  * @example
  * var metadata = {
  *   description: 'Information for every institution in the 2013 IPEDS universe'
  * };
  *
- * dataset.setMetadata(metadata, function(err, metadata, apiResponse) {});
+ * dataset.setMetadata(metadata, function(err, apiResponse) {});
  */
 Dataset.prototype.setMetadata = function(metadata, callback) {
   var that = this;
 
   this.makeReq_('PATCH', '', null, metadata, function(err, resp) {
     if (err) {
-      callback(err, null, resp);
+      callback(err, resp);
       return;
     }
 
     that.metadata = resp;
 
-    callback(null, that.metadata, resp);
+    callback(null, resp);
   });
 };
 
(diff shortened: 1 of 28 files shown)


test/bigquery/table.js
```diff
@@ -281,6 +279,18 @@ describe('BigQuery/Table', function() {
       table.createWriteStream(fileType).emit('writing');
     });
 
+    it('should format a schema', function(done) {
+      var expectedSchema = Table.createSchemaFromString_(SCHEMA_STRING);
```
callmehiphop: Should we be stubbing `createSchemaFromString_` to return fake data and assert against said fake data instead?

stephenplusplus: Definitely, but it was such a hassle, I overlooked it...

stephenplusplus: Okay, fine, I'll do it!

callmehiphop: :smile:

stephenplusplus: https://github.com/GoogleCloudPlatform/gcloud-node/pull/870/files#diff-20e2b568923f6824ab7ad5afab9255f0R91

test/bigquery/table.js
```diff
@@ -87,7 +86,23 @@ describe('BigQuery/Table', function() {
...
+    Object.keys(Table).forEach(function(tableMethod) {
+      if (typeof Table[tableMethod] !== 'function') {
+        return;
+      }
+
+      Table[tableMethod] = function() {
+        return (tableOverrides[tableMethod] || tableCached[tableMethod])
+          .apply(this, arguments);
+      };
+    });
```
callmehiphop: :+1: 

Side note (kinda) - do you think it's worth opening an issue to implement spies within our test? I think it would probably clean up the test files some and remove all the `objectOverride` stuff we have, since we should be able to easily clean them up in an `afterEach` hook.

stephenplusplus: Yes, please!

lib/bigquery/dataset.js
```diff
@@ -292,20 +292,20 @@ Dataset.prototype.query = function(options, callback) {
  *   description: 'Information for every institution in the 2013 IPEDS universe'
  * };
  *
- * dataset.setMetadata(metadata, function(err, metadata, apiResponse) {});
+ * dataset.setMetadata(metadata, function(err, apiResponse) {});
```
callmehiphop: We need to remove the metadata param from all the JSDocs now

stephenplusplus: D'oh, good catch.

stephenplusplus: Updated.

test/storage/file.js
```diff
@@ -1489,6 +1530,38 @@ describe('File', function() {
       });
     });
 
+    it('should add success redirect', function(done) {
+      file.getSignedPolicy({
+        expires: Date.now() + 5,
+        successRedirect: 'redirect'
```
callmehiphop: Could we add something like `http://` to this, just to show that it's a url?

stephenplusplus: Done

callmehiphop:
Now we have to work to keep it at 100% forever!

stephenplusplus:
screen shot 2015-09-17 at 1 45 03 pm

Woo!

PR #884, opened 2015-09-22
storage: updated documentation for read / write stream events

phw:
This properly documents the new events for read and write streams introduced in PR #824.


stephenplusplus:
Thanks for catching this!

PR #885, opened 2015-09-22
datastore: throw if an unrecognized commit method is given

### LINKED ISSUE #883: 'auto_insert_id' in documentation should be 'insert_auto_id'
The documentation as well as a comment in the source files list 'auto_insert_id' as last insert method for the datastore, however checking the source revealed that it should be 'insert_auto_id'. 
Also specifying an invalid method (e.g. 'auto_insert_id') does not report any error in the callback and doesn't throw which made finding the error harder than necessary. 

Please fix the documentation and change the lib to report an error or throw when an invalid datastore insert/save method is specified.

stephenplusplus:
Fixes #883

This also did a sweep over the structure of the code, hopefully making it more understandable rather than fancy.

### DIFF
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -227,10 +227,12 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
  *
  * @borrows {module:datastore/transaction#save} as save
  *
+ * @throws {Error} If an unrecognized method is provided.
+ *
  * @param {object|object[]} entities - Datastore key object(s).
  * @param {Key} entities.key - Datastore key object.
  * @param {string=} entities.method - Optional method to explicity use for save.
- *     The choices include 'insert', 'update', 'upsert' and 'auto_insert_id'.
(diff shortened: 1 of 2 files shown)


lib/datastore/request.js
```diff
@@ -382,75 +384,68 @@ DatastoreRequest.prototype.save = function(entities, callback) {
...
+  var mutation = {
+    insert: [],
+    update: [],
+    upsert: [],
+    insert_auto_id: []
+  };
 
-      var ent = {};
-      var method = entityObject.method;
-      delete entityObject.method;
```
callmehiphop: Should we keep this? I realize the overall size difference will be small, but this should make our payload a little smaller, yes?

stephenplusplus: It actually doesn't make its way to the request. `entityObject` is picked apart for pieces by this iterator, but it never is fully dumped into a request.

callmehiphop: Gotcha, sounds good!

PR #886, opened 2015-09-22
core: use custom error types

### LINKED ISSUE #859: Implement custom errors
Related to https://github.com/GoogleCloudPlatform/gcloud-node/pull/857#discussion_r39123168

It would be nice to have custom error types in certain places to improve the error messages we provide to the user - instructing the user to access non standard properties to get the full error message seems painful, IMO errors should be obvious and in the users face.

stephenplusplus:
Fixes #859 

@callmehiphop let me know how I did!

### DIFF
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -1499,9 +1499,8 @@ describe('File', function() {
       file.getSignedPolicy({
         expires: Date.now() + 5
       }, function(err) {
-        var errorMessage = 'Signing failed. See `error` property.';
-        assert.strictEqual(err.message, errorMessage);
-        assert.strictEqual(err.error, error);
+        assert.strictEqual(err.name, 'SigningError');
+        assert.strictEqual(err.message, error.message);
         done();
       });
     });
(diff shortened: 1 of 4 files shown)


test/storage/file.js
```diff
@@ -1516,10 +1515,11 @@ describe('File', function() {
         expires: Date.now() + 5
       }, function(err) {
         var errorMessage = [
-          'Signing failed. Could not find a `private_key`.',
+          'Could not find a `private_key`.',
           'Please verify you are authorized with this property available.'
         ].join(' ');
 
+        assert.strictEqual(err.name, 'SigningError');
```
callmehiphop: Would it be better to assert `err instanceof SigningError`?

callmehiphop:
I love it!

test/storage/file.js
stephenplusplus: I chose not to only because that would mean we would have to export the method, likely as a static member of `File`. I don't like when ease of testing gets in the way of sane code architecture. Sometimes, it's unavoidable, or can be worked around, but in this case, I think it's more effort than its worth.

Perhaps in the future, if we find ourselves needing to create more custom errors, we can put all the errors into their own file(s), and then testing will be easier.

stephenplusplus:
Woo! Looks like I have some rebasing ahead of me!

PR #891, opened 2015-09-23
add more npm keywords

stephenplusplus:


VikramTiwari:
haha did search team came up with these keywords? Looks like one of the lists i would get from my SEO team. :+1:

stephenplusplus:
Nope, all me :) Maybe I was meant to be in SEO and never knew!

callmehiphop:
magical!

PR #892, opened 2015-09-24
storage: do not fail transfers where validation is skipped

### LINKED ISSUE #889: gcloud writestreams no longer working when piping stdout from GraphicsMagick
GraphicsMagick lib:
http://aheckmann.github.io/gm/

ex:
```js
var gm = require('gm').subClass({ imageMagick: true }),
      src = '/src/file.jpg',
      dst = myBucket.file('file.png');

var wr = dst.createWriteStream();

gm(src)
    .setFormat('png')
    .stream(function(err, stdout, stderr) {
    if (err) return cb(err);
    // pipe output to writestream
    stdout.pipe(wr);
    wr.on("error", function(err) {
        return cb(err);
    });
     wr.on("finish", function() {
        return cb();
    });
});
```
works in release 20 and older(using complete event); breaks in release 21, finish doesn't seem to trigger, error is fired with->
{ [Error: The uploaded data did not match the data from the server. As a precaution, the file has been deleted. To be sure the content is the same, you should try uploading the file again.] code: 'FILE_NO_UPLOAD', errors: [ null ] }

stephenplusplus:
Fixes #889 

Instead of assuming a data integrity failure _in all cases_, we only assume a failure **if the user wanted data integrity checks in the first place**.

Definitely should give a patch release for this.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -526,7 +526,10 @@ File.prototype.createReadStream = function(options) {
         hashes[hashType] = hash.substr(hash.indexOf('=') + 1);
       });
 
-      var failed = true;
+      // If we're doing validation, assume the worst-- a data integrity
+      // mismatch. If not, these tests won't be performed, and we can assume the
+      // best.
+      var failed = crc32c || md5;
 
       if (crc32c && hashes.crc32c) {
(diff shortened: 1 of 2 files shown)


callmehiphop:
:+1:

PR #893, opened 2015-09-24
exit early if docs already built

### LINKED ISSUE #862: Switch to verbObject() methods to .object().verb()?
(Using `createBucket` as an example here)

If I want to "get a hold of a bucket", I can do it in two ways:

`gcs.createBucket(name, callback)` or `gcs.bucket(name)`. But I have to know what the state of the world is. Get-or-create semantics don't exist, and if they did, it's two different method names, which is a little annoying.

How would we feel about switching to `.object()`-always syntax, which then has `.verb()` methods on it? That is, `gcs.createBucket(name, callback)` would become `gcs.bucket(name).create(callback)` ? And then for the get-or-create syntax, we could have that as a verb (`.getOrCreate()`)?

So a bunch of examples:

```javascript
var gcs = ...;
bucket.create(function(bucket, err) { ... });
bucket.getOrCreate(function(bucket, err) { ... });
bucket.loadMetadata(function(metadata, err) { ... });
```

Hypothetically, things like `loadMetadata` would update the instance of the bucket as well, right? That is, the `bucket` object in scope of the callback should be identical to the one outside of the scope because we'd store the metadata on `this` ?

### LINKED ISSUE #882: Travis release jobs fail on second pass
Now that we're supporting multiple node versions via CI, we need to update our release scripts. When and if the second pass finishes (0.12), it will fail at creating a new folder for the documentation JSON, because it already exists from the previous pass (4.x).

stephenplusplus:
Fixes #882

This will test if the directory exists (which would mean we already built the docs), and exit 0 to tell Travis "Hey, it's all cool".

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -27,7 +27,7 @@ npm run coveralls
 npm run docs
 git submodule add -f -b gh-pages https://${GH_OAUTH_TOKEN}@github.com/${GH_OWNER}/${GH_PROJECT_NAME} ghpages
 # copy set of json to tag folder
-mkdir ghpages/json/${TRAVIS_TAG}
+test -d "ghpages/json/${TRAVIS_TAG}" && exit 0 || mkdir ghpages/json/${TRAVIS_TAG}
 cp -R docs/json/master/* ghpages/json/${TRAVIS_TAG}
 cd ghpages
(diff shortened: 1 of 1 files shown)


callmehiphop:
:+1:

PR #897, opened 2015-09-26
datastore: update to v1beta3

### LINKED ISSUE #872: Upgrade Datastore to use v1beta3
/cc @pcostell @stephenplusplus

### LINKED ISSUE #1049: Move proto-compatible APIs to grpc
We should migrate our library to use [grpc](https://github.com/grpc/grpc) where possible. These APIs are proto-compatible, thus grpc-ready:

  - [ ] Datastore v1beta3
    - [.proto files](https://github.com/google/googleapis/tree/master/google/datastore/v1beta3)
  - [x] Logging
    - [.proto files](https://github.com/google/googleapis/tree/master/google/logging)
    - [PR](https://github.com/googlecloudplatform/gcloud-node/pull/1134)
  - [x] Pub/Sub v1
    - [.proto file](https://github.com/google/googleapis/tree/master/google/pubsub/v1)
    - [PR](https://github.com/googlecloudplatform/gcloud-node/pull/1070)

### LINKED ISSUE #1181: Datastore can return an empty page and still have a next page
https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/datastore/request.js#L644

I think we need to fix this logic.

stephenplusplus:
Fixes #872
Fixes #1049
Fixes #1181
#### Docs

http://stephenplusplus.github.io/gcloud-node/#/docs/master/datastore
#### Blockers
- [x] #913 -- need it for testing test/datastore/request.js
- [x] New release of gRPC
- [x] Datastore v1beta3 release
#### To Dos
- ~~Update resource links (blocked by v1beta3's release (public docs required to get URLs))~~
  - Resolution: Removed resource links
- [x] Remove entry from Troubleshooting (https://github.com/GoogleCloudPlatform/gcloud-node/pull/1172)
- [x] Convert to gRPC
- [x] Add support for env vars from cli (#1040):
  - [x] `DATASTORE_PROJECT_ID`
  - [x] `DATASTORE_EMULATOR_HOST` (protocol not included)
- [x] Use Service & ServiceObject
- [x] Tests
  - [x] Unit
    - [x] Entity
    - [x] Index (Datastore, formerly Dataset)
    - [x] Query
    - [x] Request
    - [x] Transaction
  - [x] System
- [x] Readme snippet update
- [x] Depend on a new release of protobuf.js ([tracking](https://github.com/dcodeIO/ProtoBuf.js/issues/329#issuecomment-143426391))
## :warning: Big Breaking Changes

Dataset isn't a thing anymore:

``` js
var gcloud = require('gcloud')({/*...*/});
```
### Before

``` js
var datastore = gcloud.datastore;
var dataset = datastore.dataset();

dataset.get(dataset.key(['Kind', 1]), function(err, results) {});
```
### After

``` js
datastore.get(datastore.key(['Kind', 1]), function(err, results) {});

// Use a different project:
var datastore = gcloud.datastore({ projectId: 'grape-spaceship-123' });
```


lib/datastore/dataset.js
```diff
@@ -136,7 +135,7 @@ nodeutil.inherits(Dataset, DatastoreRequest);
  * @param {string=} options.apiEndpoint - Custom API endpoint.
```
pcostell: Dataset is no longer a concept for Cloud Datastore. Perhaps we should migrate this to just be a Client object that is used for interacting with a specific project?

(Side note, I tried to attach this comment to L102, but it seems that isn't possible in github?)

lib/datastore/entity.js
```diff
@@ -142,12 +142,16 @@ module.exports.Double = Double;
...
-    acc[property.name] = propertyToValue(property.value);
-    return acc;
-  }, {});
+  var properties = proto.properties || [];
+
+  var entity = {};
+
+  for (var property in properties.map) {
+    var value = properties.map[property].value;
+    entity[properties.map[property].key] = propertyToValue(value);
```
pcostell: Is properties.map[property].key != property?

lib/datastore/entity.js
```diff
@@ -293,14 +308,17 @@ module.exports.formatArray = formatArray;
  */
 module.exports.isKeyComplete = function(key) {
```
pcostell: This doesn't quite fit Datastore's notion of complete vs incomplete keys. In particular, a key is incomplete if the last element has no id or name. An unset kind or an ancestor path element with no id or name is just an invalid key.

lib/datastore/entity.js
```diff
@@ -314,53 +332,53 @@ module.exports.isKeyComplete = function(key) {
...
  * });
  * // 'Hi'
  *
  * propertyToValue({
- *   blob_value: new Buffer('68656c6c6f')
+ *   blobValue: new Buffer('68656c6c6f')
  * });
  * // <Buffer 68 65 6c 6c 6f>
  */
 function propertyToValue(property) {
```
pcostell: The input to this function is really a proto Value, not a property.

lib/datastore/entity.js
```diff
@@ -378,63 +396,79 @@ module.exports.propertyToValue = propertyToValue;
...
-    return p;
-  }
-  if (v instanceof Int) {
-    p.integer_value = v.get();
-    return p;
-  }
-  if (v instanceof Double) {
-    p.double_value = v.get();
-    return p;
+function valueToProperty(value) {
```
pcostell: Same here, this is really valueToProto

lib/datastore/entity.js
```diff
@@ -378,63 +396,79 @@ module.exports.propertyToValue = propertyToValue;
...
+    propertyObject.doubleValue = value.get();
+    return propertyObject;
   }
-  if (v instanceof Buffer) {
-    p.blob_value = v;
-    return p;
+
+  if (value instanceof Date) {
+    propertyObject.timestampValue = {
+      seconds: value.getTime() / 1000
```
pcostell: I think this should also set nanoseconds with the extra sub-second information.

lib/datastore/entity.js
```diff
@@ -378,63 +396,79 @@ module.exports.propertyToValue = propertyToValue;
...
-      property.push({
-        name: k,
-        value: valueToProperty(v[k])
-      });
+
+  if (is.object(value) && Object.keys(value).length > 0) {
+    var property = Object.keys(value).map(function(key) {
+      return {
+        name: key,
+        value: valueToProperty(value[key])
```
pcostell: Note that is now acceptable for these values to be indexes (i.e. if I put an object:

    {
      a: {
        b: 2
      }
    }

Then I can actually query for this using the property `a.b`.

lib/datastore/request.js
```diff
@@ -617,11 +616,15 @@ DatastoreRequest.prototype.runQuery = function(query, callback) {
...
-      var endCursor = resp.batch.end_cursor.toBase64();
+    if (resp.batch.entityResults) {
+      entities = entity.formatArray(resp.batch.entityResults);
+    }
+
+    if (resp.batch.endCursor && entities.length > 0) {
+      var endCursor = resp.batch.endCursor.toBase64();
 
       nextQuery = extend(true, new Query(), query);
       nextQuery.start(endCursor).offset(0);
```
pcostell: You aren't guaranteed that we'll completely answer the offset. Instead you should compare skipped_results to the initial offset.

lib/datastore/request.js
```diff
@@ -617,11 +616,15 @@ DatastoreRequest.prototype.runQuery = function(query, callback) {
...
+    var entities = [];
     var nextQuery = null;
 
-    if (resp.batch.end_cursor && entities.length > 0) {
-      var endCursor = resp.batch.end_cursor.toBase64();
+    if (resp.batch.entityResults) {
+      entities = entity.formatArray(resp.batch.entityResults);
+    }
+
+    if (resp.batch.endCursor && entities.length > 0) {
```
pcostell: You should use more_results to determine if you need to issue another query (in particular, if it is == NOT_FINISHED.

proto/google/api/annotations.proto
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2015, Google Inc.
```
pcostell: Maybe proto/... should be a git submodule instead from googleapis?

stephenplusplus:
Thanks for the review @pcostell! I'll be making some adjustments throughout the week and will ping if I have any questions.

pcostell:
No worries -- one more note, the url has changed. So for example for runQuery it is now:

datastore.googleapis.com/v1beta3/projects/<project-id>:runQuery
Instead of www.googleapis.com/datastore/v1beta2/datasets/<project-id>/RunQuery

lib/datastore/request.js
```diff
@@ -755,22 +757,23 @@ DatastoreRequest.prototype.makeReq_ = function(method, body, callback) {
...
   var pbKey = method[0].toUpperCase() + method.substr(1);
-  var pbRequest = new pb[pbKey + 'Request'](body).toBuffer();
-  var pbResponse = pb[pbKey + 'Response'];
+  var pbRequest = datastoreProto[pbKey + 'Request'].encode(body).toBuffer();
+  var pbResponse = datastoreProto[pbKey + 'Response'];
 
   var reqOpts = {
     method: 'POST',
-    uri: format('{apiEndpoint}/{path}/{projectId}/{method}', {
+    uri: format('{apiEndpoint}/v1beta3/projects/{projectId}:{method}', {
```
stephenplusplus: @pcostell caught the URL format change here :) How will that affect local datastore servers? We accept an `apiEndpoint` option that is usually "localhost:8080" or something similar I believe. Will the path following the host:port match the v1beta3 format?

pcostell: Ah perfect :-). Right now if you run locally it the url should look like:

http://localhost:8080/datastore/v1beta3/projects/<my-project>:runQuery

proto/google/api/annotations.proto
stephenplusplus: Git submodules + npm is not a great story. It would be great if the proto files could be published to npm, and we could treat them as a normal dependency. Do you think there's any chance for that?

lib/datastore/request.js
stephenplusplus: Great, this will support that perfectly!

proto/google/api/annotations.proto
pcostell: Good to know. Right now there isn't a great story around packaging, but we are working around it by pushing ourselves (I'm hoping that eventually this is done via googleapis). However, I can work today on getting the protos on npm.

lib/datastore/entity.js
stephenplusplus: Does this look more correct?

```js
module.exports.isKeyComplete = function(key) {
  var lastPathElement = keyToKeyProto(key).path.pop();

  return lastPathElement.id || lastPathElement.name;
};

// (same code as previous "isKeyComplete")
module.exports.isKeyValid = function(key) {
  var proto = keyToKeyProto(key);

  for (var i = 0; i < proto.path.length; i++) {
    if (!proto.path[i].kind) {
      return false;
    }

    if (!proto.path[i].id && !proto.path[i].name) {
      return false;
    }
  }

  return true;
};
```

pcostell: Ya that looks right.

pcostell: Just to double check --- id = 0 or name = "" evaluates to false, right?

stephenplusplus: Yeah, good or bad?

pcostell: good.

lib/datastore/request.js
stephenplusplus: I think I understand, but if we set an offset on the `end_cursor` returned, wouldn't it offset from there?

Here's how I'm envisioning the API working:
```
Query:
Get the last 8 things (offset = 2)

10 things exist:
1
2
3 - match! - batch.results.push()
4 - match! - batch.results.push()
5 - match! - batch.results.push()
6 - I need a break! - batch.end_cursor = pointer to this result
7
8
9
10

Response:
{
  "results": [
    3,
    4,
    5
  ],
  end_cursor: "pointer to 6",
  skipped_results: 2,
  more_results: "NOT_FINISHED"
}
```

If we set the next query's offset to anything other than 0, wouldn't it start from "6" + offset, skipping over some of the results that matched the original query?

pcostell: Yes, but consider this case:

    Query:
     SELECT * FROM FOO OFFSET 100
    101 things exist:
    1 -- 50
    51 - I need a break! - batch.end_cursor = pointer to this result.
    52 - 100
    101 - match!

At point 51, we should set `batch.end_cursor = pointer` *and* set `offset = 100 - skipped_results`.

stephenplusplus: *lightbulb turns on* Thanks!

lib/datastore/entity.js
stephenplusplus: Renamed to `decodeValueProto`: https://github.com/GoogleCloudPlatform/gcloud-node/pull/897#discussion_r41195136

lib/datastore/entity.js
stephenplusplus: Renamed to `encodeValue`: https://github.com/GoogleCloudPlatform/gcloud-node/pull/897/files#diff-45459618925a219e865516ffadd74c28R413

lib/datastore/entity.js
stephenplusplus: https://github.com/GoogleCloudPlatform/gcloud-node/pull/897/files#diff-45459618925a219e865516ffadd74c28R442

lib/datastore/entity.js
stephenplusplus: I just removed the former behavior which always set `excludeFromIndexes`.

lib/datastore/request.js
stephenplusplus: https://github.com/GoogleCloudPlatform/gcloud-node/pull/897/files#diff-2df5e45f60c621c6a33adff3b5bd9541R639

lib/datastore/request.js
stephenplusplus: https://github.com/GoogleCloudPlatform/gcloud-node/pull/897/files#diff-2df5e45f60c621c6a33adff3b5bd9541R631

stephenplusplus:
Starting the conversion to gRPC and I ran into an issue:

```
The Cloud Datastore API is not enabled for the project nth-circlet-705
```

I've emailed @pcostell about this, hoping it's just a matter of whitelisting.

jgeewax:
I think you just have to enable it in the APIs page on
cloud.google.com/console

stephenplusplus:
Sorry, I should have mentioned that it's enabled. It's the same project I use with Datastore that works without gRPC.

pcostell:
Have you tried it without gRPC recently? We just enabled the requirement to enable it in the APIs page -- I sent you a direct link as I think it is still unaccessible publicly.

stephenplusplus:
I haven't, sorry the gRPC part wasn't related. For anyone following along, Patrick is going to enable the API himself after becoming an owner of the project.

stephenplusplus:
@pcostell I'm having an issue running a query with a filter on a key. The code executes fine without error, but there aren't any results returned. Does anything look wrong from these:

##### commit
```json
{
  "mutations": [
    {
      "upsert": {
        "key": {
          "path": [
            {
              "kind": "Person",
              "name": "name"
            }
          ]
        },
        "properties": {
          "fullName": {
            "stringValue": "Full name"
          },
          "linkedTo": {
            "keyValue": {
              "path": [
                {
                  "kind": "Person",
                  "name": "name"
                }
              ]
            }
          }
        }
      }
    }
  ],
  "projectId": "nth-circlet-705",
  "mode": "NON_TRANSACTIONAL"
}
```

##### runQuery
```json
{
  "readOptions": {},
  "query": {
    "distinctOn": [],
    "kind": [
      {
        "name": "Person"
      }
    ],
    "order": [],
    "projection": [],
    "filter": {
      "compositeFilter": {
        "filters": [
          {
            "propertyFilter": {
              "property": {
                "name": "linkedTo"
              },
              "op": "EQUAL",
              "value": {
                "keyValue": {
                  "path": [
                    {
                      "kind": "Person",
                      "name": "name"
                    }
                  ]
                }
              }
            }
          }
        ],
        "op": "AND"
      }
    }
  },
  "projectId": "nth-circlet-705"
}
```

pcostell:
Your query isn't strongly consistent. Can you try with a parent and ancestor?

stephenplusplus:
Ah, okay. This code works now:

```js
var personKey = datastore.key(['People', 'US', 'Person', 'name']);

datastore.save({
  key: personKey,
    data: {
      fullName: 'Full name',
      linkedTo: personKey // himself
    }
  }, function(err) {
    assert.ifError(err);
    var query = datastore.createQuery('Person')
      .hasAncestor(personKey.parent);
```

Was it just luck that this test has been working until now?

stephenplusplus:
This fails... should it?

```js
var personKey = datastore.key(['People', 'US', 'Person', 'name']);

      datastore.save({
        key: personKey,
        data: {
          fullName: 'Full name',
          linkedTo: personKey // himself
        }
      }, function(err) {
        assert.ifError(err);

        var query = datastore.createQuery('Person')
          .hasAncestor(personKey.parent)
          .filter('linkedTo', personKey);
```

pcostell:
Hmm I think that should work. What's the proto getting sent?

stephenplusplus:
##### commit
```json
{
  "mutations": [
    {
      "upsert": {
        "key": {
          "path": [
            {
              "kind": "People",
              "name": "US"
            },
            {
              "kind": "Person",
              "name": "name"
            }
          ]
        },
        "properties": {
          "fullName": {
            "stringValue": "Full name"
          },
          "linkedTo": {
            "keyValue": {
              "path": [
                {
                  "kind": "People",
                  "name": "US"
                },
                {
                  "kind": "Person",
                  "name": "name"
                }
              ]
            }
          }
        }
      }
    }
  ],
  "projectId": "nth-circlet-705",
  "mode": "NON_TRANSACTIONAL"
}
```

##### runQuery
```json
{
  "readOptions": {},
  "query": {
    "distinctOn": [],
    "kind": [
      {
        "name": "Person"
      }
    ],
    "order": [],
    "projection": [],
    "filter": {
      "compositeFilter": {
        "filters": [
          {
            "propertyFilter": {
              "property": {
                "name": "__key__"
              },
              "op": "HAS_ANCESTOR",
              "value": {
                "keyValue": {
                  "path": [
                    {
                      "kind": "People",
                      "name": "US"
                    }
                  ]
                }
              }
            }
          },
          {
            "propertyFilter": {
              "property": {
                "name": "linkedTo"
              },
              "op": "EQUAL",
              "value": {
                "keyValue": {
                  "path": [
                    {
                      "kind": "People",
                      "name": "US"
                    },
                    {
                      "kind": "Person",
                      "name": "name"
                    }
                  ]
                }
              }
            }
          }
        ],
        "op": "AND"
      }
    }
  },
  "projectId": "nth-circlet-705"
}
```

pcostell:
@eddavisson Any thoughts here? Could it be related to key normalization?

@stephenplusplus Does this work if you query for the `fullName` instead of `linkedTo`?

eddavisson:
I can't reproduce this. I ran these two commands in our internal gRPC shell, and got the expected single result for the query.

```
Commit mutations:[{upsert:{key:{path:[{kind:"People",name:"US"},{kind:"Person",name:"name"}]},properties:[{key:"fullName",value:{string_value:"Full name"}},{key:"linkedTo",value:{key_value:{path:[{kind:"People",name:"US"},{kind:"Person",name:"name"}]}}}]}}],project_id:"eddavisson-e2e-prod",mode:NON_TRANSACTIONAL

RunQuery query:{kind:[{name:"Person"}],filter:{composite_filter:{filters:[{property_filter:{property:{name:"__key__"},op:HAS_ANCESTOR,value:{key_value:{path:[{kind:"People",name:"US"}]}}}},{property_filter:{property:{name:"linkedTo"},op:EQUAL,value:{key_value:{path:[{kind:"People",name:"US"},{kind:"Person",name:"name"}]}}}}],op:AND}}},project_id:"eddavisson-e2e-prod"
```

stephenplusplus:
Thanks for checking @eddavisson. We have some differences in our commit request. Should `mutations[0].upsert.properties` be an object or array? I tried converting my request to the array format you used and the protobuf message parser didn't agree with that format.

For completeness, here's the runQuery response:

```json
{
  "batch": {
    "skippedResults": 0,
    "skippedCursor": {},
    "entityResultType": "FULL",
    "entityResults": [],
    "endCursor": "CgA=",
    "moreResults": "NO_MORE_RESULTS"
  },
  "query": null
}
```

@pcostell querying for `fullName = "Full name"` doesn't return results either.

pcostell:
@stephenplusplus -- 

Are you sure the data is in Datastore? Can you send us the mutation result and double check the entity is in your viewer?

stephenplusplus:
Yes, here's the result from the commit:

```js
{ mutationResults: [ { key: null } ], indexUpdates: 5 }
```

In an earlier test, I do a commit and then a lookup. The commit returns similar results (a mutationResults array with a single `{ key: null }` object), but the lookup succeeds.

stephenplusplus:
The viewer (which I believe means https://console.cloud.google.com/datastore, lmk if that's wrong) seems to be hit or miss to see an accurate representation of the data. I tested by inserting completely new data `[Kind: Stephen, ID: autogenerated]`, and received a successful response. When I went to the viewer, it appeared instantly. Then I inserted more new data `[Kind: Stephen, name: Two]`. Refreshed and saw it, but `[Kind: Stephen, ID: autogenerated]` was gone. Refreshed again and they both have been missing ever since. I kicked off 15 jobs through the admin panel to remove entities from 15 various kinds (not the newly created "Stephen"). All of those are still showing in the viewer even after the jobs completed.

eddavisson:
The viewer (https://console.cloud.google.com/datastore) uses non-ancestor queries (unless you specifically create one in GQL), so the behavior you describe is possible.

After you run the test, can you see the entity (`{kind:"People",name:"US"},{kind:"Person",name:"name"}`) in the viewer?

stephenplusplus:
No, I actually don't see anything in there. In the Entities section, there appear to be some remnants of the jobs that deleted all of the entities.

screen shot 2016-03-24 at 7 41 04 pm

I'm able to save and do a lookup for the saved data, but it doesn't appear in the viewer or via runQuery.

eddavisson:
Data should generally be visible in the viewer very quickly. Is it possible that the test is deleting the entity before you run the query? Or it's inserting the data into a different project?

A suggestion for isolating this: run just the code that inserts the entity, wait a few minutes, and see if it shows up in the viewer.

stephenplusplus:
I'm embarrassed to say that was the problem with not seeing it in the viewer, which instantly led to me realizing the problem. Feel free to tune out now, 'cause it's the fault of this library. Thank you for your help!

Working with gRPC/protobuf.js means we have to convert camelcase to snakecase in certain places to comply with its API. It's not supposed to seep into user code, but clearly it has, since data the user saves under a property like "linkedTo" is being converted and saved as "linked_to". That's why the query didn't return any results.

Sorry to get you both involved in this, but thank you very much for helping me work through it.

stephenplusplus:
Having some issues with upgrading gRPC. Issue logged here: https://github.com/grpc/grpc/issues/5972

I need to use updates since the last release of gRPC that include https://github.com/grpc/grpc/issues/5005 and https://github.com/grpc/grpc/issues/5006, so it's currently a blocker.

pcostell:
@stephenplusplus Do we need these grpc additions in order to launch v1beta3? If not, are there any blockers?

stephenplusplus:
It's the last thing left aside from updating our docs to link to the v1beta3 docs once they're live. I'll take a look tomorrow if I can work around the grpc issue.

stephenplusplus:
Unsure of [when the next grpc release will be](https://github.com/grpc/grpc/issues/5972#issuecomment-202505716), I published http://npmjs.com/package/@stephenplusplus/grpc (npm install @stephenplusplus/grpc) which is the current state of the grpc library. It has everything we need (https://github.com/grpc/grpc/issues/5005, https://github.com/grpc/grpc/issues/5006, and some other fixes for unwanted logging output).

The PR is now complete.

stephenplusplus:
@callmehiphop ptal. I'll squash this up after you've had a chance to check out the recent commits.

pcostell:
@stephenplusplus We are ready for you to merge this into master.

stephenplusplus:
Is v1beta3 generally available? I've been using the response code of https://cloud.google.com/datastore/docs/apis/v1beta3 to tell me if it's live or not, and it's still a 404.

pcostell:
The docs and blog post won't drop until Monday, however the API can be enabled: https://console.cloud.google.com/apis/api/datastore.googleapis.com/overview

stephenplusplus:
For users that already have Datastore enabled and have been running v1beta2, they won't have to re-enable the API, right?

pcostell:
Any projects active in the last 7 days will not need to re-enable.

stephenplusplus:
Cool, thanks. And more generally, thanks for all of the help along the way! :+1:

pcostell:
Thank you for helping us get this out the door!

As a heads up, the cloud SDK will not have a grpc emulator at launch time. So users that want to use gcloud-node will have to use the gcd.sh tool directly, which they can get from [here](storage.googleapis.com/gcd/tools/gcd-grpc-1.0.0.zip).

pcostell:
Release notes for Datastore available if you need to link to them: https://cloud.google.com/datastore/release-notes

lib/datastore/entity.js
```diff
@@ -22,23 +22,71 @@
...
+});
+
+/**
+ * Build a Datastore Double object.
+ *
+ * @constructor
+ * @param {number} value - The double value.
+ *
+ * @example
+ * var aDouble = new Double(7.3);
```
callmehiphop: This is kind of a general question in regards to this whole file - I'm noticing that in all the examples we're not referencing an entity, is this going to make the doc tests fail since none of these functions exist globally?

stephenplusplus: I think this is excluded from the docs tests. This file isn't exposed to the docs either.

lib/datastore/index.js
```diff
@@ -27,52 +37,49 @@
...
- * The `gcloud.datastore` object gives you some convenience methods, as well as
- * exposes a `dataset` function. This will allow you to create a `dataset`,
- * which is the object from which you will interact with the Google Cloud
+ * The `gcloud.datastore` object allows you to interact with Google Cloud
  * Datastore.
  *
  * To learn more about Datastore, read the
  * [Google Cloud Datastore Concepts Overview](https://cloud.google.com/datastore/docs/concepts/overview)
  *
  * @param {object} options - [Configuration object](#/docs).
```
callmehiphop: Is this optional now?

lib/datastore/index.js
```diff
@@ -152,19 +179,146 @@ Datastore.int = function(value) {
...
+ * Helper to create a Key object, scoped to the instance's namespace by default.
+ *
+ * You may also specify a configuration object to define a namespace and path.
+ *
+ * @param {...*=} options - Key path. To specify or override a namespace,
+ *     you must use an object here to explicitly state it.
+ * @param {object=} options - Configuration object.
+ * @param {...*=} options.path - Key path.
+ * @param {string=} options.namespace - Optional namespace.
+ * @return {Key} A newly created Key from the options given.
```
callmehiphop: Should this be `{module:datastore/key}` ?

callmehiphop: Or rather `{module:datastore#key}`

stephenplusplus: That would be saying the method returns itself. There's not a great way to document this, because a `Key` object is part of the `Entity` class, which is hidden. I went with "Key" because it's at least suggesting it's a custom object type. Should I just say `object`?

stephenplusplus:
@callmehiphop - all squashed.

callmehiphop:
:tada:

stephenplusplus:
@pcostell any API flakiness going on? We are having inconsistent results querying for data:

Commit:

```json
{
  "mutations": [
    {
      "upsert": {
        "key": {
          "path": [
            {
              "kind": "Person",
              "name": "name"
            }
          ]
        },
        "properties": {
          "f": {
            "stringValue": "Full name"
          }
        }
      }
    }
  ],
  "projectId": "nth-circlet-705",
  "mode": "NON_TRANSACTIONAL"
}
```

Query:
```json
{
  "readOptions": {},
  "query": {
    "distinctOn": [],
    "kind": [
      {
        "name": "Person"
      }
    ],
    "order": [],
    "projection": [],
    "filter": {
      "compositeFilter": {
        "filters": [
          {
            "propertyFilter": {
              "property": {
                "name": "f"
              },
              "op": "EQUAL",
              "value": {
                "stringValue": "Full name"
              }
            }
          }
        ],
        "op": "AND"
      }
    }
  },
  "projectId": "nth-circlet-705"
}
```

Sometimes the query returns the entity, other times it doesn't. Is there anything we're doing wrong?

pcostell:
@stephenplusplus That looks like standard Eventual Consistency. For any e2e tests that require a get and a query you should be using ancestors.

stephenplusplus:
I'm getting it with that as well:

Commit:
```json
{
  "mutations": [
    {
      "upsert": {
        "key": {
          "path": [
            {
              "kind": "People",
              "name": "US"
            },
            {
              "kind": "Person",
              "name": "name"
            }
          ]
        },
        "properties": {
          "f": {
            "stringValue": "Full name"
          }
        }
      }
    }
  ],
  "projectId": "nth-circlet-705",
  "mode": "NON_TRANSACTIONAL"
}
```

Query
```json
{
  "readOptions": {},
  "query": {
    "distinctOn": [],
    "kind": [
      {
        "name": "Person"
      }
    ],
    "order": [],
    "projection": [],
    "filter": {
      "compositeFilter": {
        "filters": [
          {
            "propertyFilter": {
              "property": {
                "name": "f"
              },
              "op": "EQUAL",
              "value": {
                "stringValue": "Full name"
              }
            }
          }
        ],
        "op": "AND"
      }
    }
  },
  "projectId": "nth-circlet-705"
}
```

Maybe I'm not understanding how to save and/or query using ancestors?

stephenplusplus:
Sorry, I got it using `hasAncestor`. I think I'm good.

pcostell:
Cool, let me know if there are any other issues

PR #903, opened 2015-09-30
Documentation bug fixes

### LINKED ISSUE #900: Add an example for gcloud.datastore.dataset
Looks like we don't have an example snippet on http://googlecloudplatform.github.io/gcloud-node/#/docs/v0.23.0/datastore?method=dataset

Any chance we could add one?

### LINKED ISSUE #901: dataset.get's example code looks... broken?
Looking at http://googlecloudplatform.github.io/gcloud-node/#/docs/v0.23.0/datastore/dataset?method=get

I see

> Where you see transaction, assume this is the context that's relevant to your use, whether that be a Dataset or Transaction object.
> 
> ```
> (empty area that looks like it should have code)
> ```
> 
> Get a single entity.
> 
> ```javascript
> var key = dataset.key(['Company', 123]);
> 
> transaction.get(key, function(err, entity) {});
> ```

I think there are a few issues

1. The empty space looks like it should have code
1. The second snippet about getting a single entity refers to a variable that doesn't appear to be defined

Any idea what's up here?

### LINKED ISSUE #902: datastore.Query.autoPaginate has no example snippet
Can we add an example?

callmehiphop:
Fixes #900 
Fixes #901 
Fixes #902

### DIFF
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -118,12 +118,15 @@ function DatastoreRequest() {}
  *
  * @example
  * //-
- * // Where you see `transaction`, assume this is the context that's relevant to
- * // your use, whether that be a Dataset or Transaction object.
- * //
  * // Get a single entity.
  * //-
  * var key = dataset.key(['Company', 123]);
+ *
+ * dataset.get(key, function(err, entity) {});
+ *
+ * //-
+ * // Or, if you're using a transaction object.
+ * //-
  *
  * transaction.get(key, function(err, entity) {});
  *
@@ -135,12 +138,12 @@ function DatastoreRequest() {}
  *   dataset.key(['Product', 'Computer'])
  * ];
  *
- * transaction.get(keys, function(err, entities) {});
+ * dataset.get(keys, function(err, entities) {});
  *
  * //-
  * // Or, get the entities as a readable object stream.
  * //-
- * transaction.get(keys)
+ * dataset.get(keys)
  *   .on('error', function(err) {})
  *   .on('data', function(entity) {
  *     // entity is an entity object.
@@ -481,16 +484,18 @@ DatastoreRequest.prototype.save = function(entities, callback) {
  * @param {object} callback.apiResponse - The full API response.
  *
  * @example
+ * dataset.delete(dataset.key(['Company', 123]), function(err, apiResp) {});
+ *
  * //-
- * // Where you see `transaction`, assume this is the context that's relevant to
- * // your use case, whether that be a Dataset or a Transaction object.
+ * // Or, if you're using a transaction object.
  * //-
  *
- * // Delete a single entity.
  * transaction.delete(dataset.key(['Company', 123]), function(err, apiResp) {});
  *
+ * //-
  * // Delete multiple entities at once.
- * transaction.delete([
+ * //-
+ * dataset.delete([
  *   dataset.key(['Company', 123]),
  *   dataset.key(['Product', 'Computer'])
  * ], function(err, apiResponse) {});
@@ -553,11 +558,13 @@ DatastoreRequest.prototype.delete = function(keys, callback) {
  * //-
  * var query = dataset.createQuery('Lion');
  *
- * transaction.runQuery(query, function(err, entities) {
- *   if (!err) {
- *     // Handle entities here.
- *   }
- * });
+ * dataset.runQuery(query, function(err, entities) {});
+ *
+ * //-
+ * // Or, if you're using a transaction object.
+ * //-
+ *
+ * transaction.runQuery(query, function(err, entities) {});
  *
  * //-
  * // To control how many API requests are made and page through the results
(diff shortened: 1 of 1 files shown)


callmehiphop:
I would have thought this had popped up sooner, but did updating `jscs` start causing our `lint` task to fail? It looks like it's parsing one of the system test fixtures.

lib/datastore/query.js
```diff
@@ -75,7 +75,16 @@ function Query(namespace, kinds) {
 }
 
 /**
+ * @param {boolean=} autoPaginateVal - Have pagination handled automatically.
+ *     Default: true.
  * @return {module:datastore/query}
+ *
+ * @example
+ * // Have pagination handled automatically
+ * var autoPaginated = query.autoPaginate();
```
jgeewax: Can we expand this to show what the code would look like when using these? Feels like this is one of those docs for `getFoo` that says "gets foo property".

callmehiphop: Haha, sure. Do you think we should overhaul most of the query examples then? They kind of all resemble that format..

https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/datastore/query.js#L131
https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/datastore/query.js#L149
https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/datastore/query.js#L171

jgeewax: I think a richer example is good, so long as it's representative. That is, we don't want to be building mini apps inside the example snippet, but for example:

```javascript
var ancestoryQuery = query.hasAncestor(dataset.key(['Parent', 123]));
```

Might be better as

```javascript
// Create a couple keys ['Person', 1234] and ['Person', 4567] and ['Person', 1234, 'Person', 'child']
// Create a query for kind=Person
// Set the ancestor to be Person 1234
var ancestoryQuery = query.hasAncestor(dataset.key(['Person', 1234]));
// Mention that the result of the query should be ['Person', 1234, 'Person', 'child']
```

stephenplusplus:
> I would have thought this had popped up sooner, but did updating jscs start causing our lint task to fail? It looks like it's parsing one of the system test fixtures.

Must be some semver fail upstream. Fixed in: https://github.com/GoogleCloudPlatform/gcloud-node/commit/fe2f2e1105562d542a823cc1f0974e1d54fbdd58 Sorry to force a rebase :(

lib/datastore/query.js
callmehiphop: Want to take a look at the updated example? The query is a little more complex, but I think it needs to be in order to properly display `autoPaginate()` in a non-foobar fashion.

callmehiphop:
Some of the methods have large-ish code examples, so rather than duplicating every example, I changed the default object to `dataset` and then add a section that basically says "Or if you're a transaction object" and gave a simple example of the transaction usage.

Does that sound ok for now? I think a better solution would be to use something like JSDocs virtual members and document each method in it's respective file so that we don't have to lump all these examples together (kinda confusing)

stephenplusplus:
I haven't heard of virtual members, but it sounds okay to me. Would we have to re-document the parameters?

lib/datastore/query.js
```diff
@@ -75,7 +75,15 @@ function Query(namespace, kinds) {
...
 /**
+ * @param {boolean=} autoPaginateVal - Have pagination handled automatically.
+ *     Default: true.
  * @return {module:datastore/query}
+ *
+ * @example
+ * // List all people related to person 1234 with a single query
+ * var query = dataset.createQuery('Person')
+ *   .hasAncestor(dataset.key(['Person', 1234]))
+ *   .autoPaginate();
```
stephenplusplus: Since this is the default, we should show running a query as false, since anyone who comes to look at this doc is probably wondering how to shut it off.

callmehiphop:
> I haven't heard of virtual members, but it sounds okay to me. 

The documentation for it is kind of scattered, but you essentially can use `@memberof` and `@type` to document code that lives in a different file or is global, etc.

> Would we have to re-document the parameters?

We would, although it looks like `transaction` overwrites 2 of the `DatastoreRequest` methods, so small price to pay imo.

lib/datastore/query.js

lib/datastore/query.js
```diff
@@ -75,7 +75,15 @@ function Query(namespace, kinds) {
...
 /**
+ * @param {boolean=} autoPaginateVal - Have pagination handled automatically.
+ *     Default: true.
  * @return {module:datastore/query}
+ *
+ * @example
+ * // Retrieve a list of people, but disable pagination
+ * var query = dataset.createQuery('Person')
+ *   .hasAncestor(dataset.key(['Person', 1234]))
+ *   .autoPaginate(false);
```
stephenplusplus: It would be awesome to see a full fledged iterative query example, eg https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/datastore/request.js#L568

callmehiphop: That's a good suggestion!

PR #904, opened 2015-10-01
Implement Service & ServiceObject classes

### LINKED ISSUE #847: Clear up `makeRequest_` naming conventions
We follow a few different patterns throughout the codebase, so I think we should try to sync up.

In some constructors, you might find `this.makeRequest_ = storage.bucket.makeRequest.bind(storage.bucket)`. In other places, we define a `makeRequest_` on the prototype. (props to @callmehiphop for first calling us out on this).

We should also clear up the naming conventions around making requests. Currently:

```js
function Class() {
  this.makeAuthorizedRequest_ = util.makeAuthenticatedRequestFactory();
}

Class.prototype.makeReq_ = function(method, path, qs, body, callback) {
  var reqOpts = {
    method: method,
    uri: path,
    // ...
  };

  this.makeAuthorizedRequest_(reqOpts);
};
```

I think we should go through and have only one style; `Class.prototype.request` instead of constructor-initialized functions.

```js
function Class() {
  this.authClient = util.getAuthClient();
}

Class.prototype.request = function(reqOpts, callback) {
  util.request(this.authClient, reqOpts, callback);
};
```

It should make it easier to follow along with, and it should be possible.

Also note using `(reqOpts, callback)` over `(method, path, query, body, callback)`. It's pretty non-JS to have multiple arguments. We should just use the standard `request` options format, with `reqOpts.uri`, `reqOpts.method`, `reqOpts.qs`, `reqOpts.json`, etc.

LMKWYT!

### LINKED ISSUE #862: Switch to verbObject() methods to .object().verb()?

### LINKED ISSUE #914: `exists` method for storage/bucket#file ?
Currently I can't find a method to check if a file is already existed in the bucket.

stephenplusplus:
# This PR is going to be split into pieces

Core Service & ServiceObject: #928
BigQuery: #931
Compute: #932
Datastore: #897
DNS: #942 
Resource: #943 
Pub/Sub: #944 
Search: #945 
Storage: #946 
## 

Fixes #847 
Fixes #862 
Fixes #914

This will be a big PR, and is still a work in progress. @callmehiphop -- please review at your earliest convenience!
#### To Dos
- [x] Review by @callmehiphop
- Implement across the API
  - [x] BigQuery
  - [x] Compute
  - [ ] Datastore (waiting on #897)
  - [x] DNS
  - [x] Pub/Sub
  - [x] Resource
  - [x] Search
  - [x] Storage
- Docs
  - [x] Make sure all of the `Service.serviceObject = function() {}` docs remove "reference an existing ____"
  - [x] Make sure `create` is shown with a config object where _required_
  - [ ] Breaking changes
    - [ ] `BigQuery.dataset#createTable`: previously `(options, callback)`, now `(id, options, callback)`
- Tests
  - Unit
    - [x] Service
    - [x] ServiceObject
    - [x] BigQuery
      - [x] dataset
      - [x] index
      - [x] job
      - [x] table
    - [x] Compute
      - [x] address
      - [x] disk
      - [x] firewall
      - [x] index
      - [x] network
      - [x] operation
      - [x] region
      - [x] snapshot
      - [x] vm
      - [x] zone
    - [ ] Datastore
      - Waiting on #897
    - [x] DNS
      - [x] change
      - [x] index
      - [x] record
      - [x] zone
    - [x] Pub/Sub
      - [x] iam
      - [x] index
      - [x] subscription
      - [x] topic
    - [x] Resource
      - [x] index
      - [x] project
    - [x] Search
      - [x] document
      - [x] field
      - [x] index-class
      - [x] index
    - [x] Storage
      - [x] acl
      - [x] bucket
      - [x] file
      - [x] index
  - System
    - [x] BigQuery
    - [x] Compute
    - [ ] Datastore
    - [x] DNS
    - [x] Pub/Sub
    - [x] Resource
    - [x] Search
    - [x] Storage
## Goal

Add `create`, `get` (with `autoCreate` behavior), and `exists` methods to all service objects.
## Implementation

Distinctly define the types of classes we have:
1. Service classes (Datastore, Storage, Compute, ...)
2. Service Object classes (Dataset, Bucket, VM, ...)

Define common methods for each:
1. Service classes:
   1. `request` => make an authenticated request
2. Service Object classes:
   1. `create` => create this object
   2. `exists` => does this object exist, returns true or false
   3. `get` => get the metadata for this object -- pass `{autoCreate:true}` to create the object if it doesn't exist
   4. `delete` => delete this object
   5. `getMetadata` => get the metadata for this object
   6. `setMetadata` => set the metadata for this object
   7. `request` => make an authenticated request

We've been re-writing these methods all over our library, which has led to some unwanted side effects (inconsistency / repeated code / repeated tests). This separation will allow us to write once, run everywhere.
## The Problems
1. Not every service object can utilize all of these methods. For example, a Compute Engine Zone can't be created. You can't set the metadata on a Compute Engine Snapshot. You can't delete a Compute Engine Region. This requires configuration so that the unwanted methods are blanked out.
2. Documentation. What we want is to inherit a bunch of methods and be done with it, but we still need to document them matching the context of the service object. We don't want the user to see some generic "To delete this object, call `object.delete()`. And we _have_ to uniquely document `create`, since each method will take different parameters.

### DIFF
--- a/lib/bigquery/dataset.js
+++ b/lib/bigquery/dataset.js
@@ -366,7 +366,7 @@ Dataset.prototype.query = function(options, callback) {
 };
 
 /**
- * Return a new instance of reference to an existing Table object.
+ * Create a Table object.
  *
  * @param {string} id - The ID of the table.
  * @return {module:bigquery/table}
--- a/lib/compute/network.js
+++ b/lib/compute/network.js
@@ -254,7 +254,7 @@ Network.prototype.delete = function(callback) {
  *
  * @resource [Firewalls Overview]{@link https://cloud.google.com/compute/docs/networking#firewalls}
  *
- * @param {string} name - Name of the existing firewall.
+ * @param {string} name - Name of the firewall.
  *
  * @example
  * var firewall = network.firewall('firewall-name');
--- a/lib/compute/region.js
+++ b/lib/compute/region.js
@@ -132,7 +132,7 @@ nodeutil.inherits(Region, ServiceObject);
  *
  * @resource [Instances and Networks]{@link https://cloud.google.com/compute/docs/instances-and-network}
  *
- * @param {string} name - Name of the existing address.
+ * @param {string} name - Name of the address.
  * @return {module:compute/address}
  *
  * @example
--- a/lib/dns/zone.js
+++ b/lib/dns/zone.js
@@ -171,7 +171,7 @@ Zone.prototype.addRecords = function(records, callback) {
 };
 
 /**
- * Create a reference to an existing change object in this zone.
+ * Create a reference to a change object in this zone.
  *
  * @param {string} id - The change id.
  * @return {module:dns/change}
--- a/lib/pubsub/index.js
+++ b/lib/pubsub/index.js
@@ -471,10 +471,9 @@ PubSub.prototype.subscribe = function(topic, subName, options, callback) {
 };
 
 /**
- * Create a Subscription object in reference to an existing subscription. This
- * command by itself will not run any API requests. You will receive a
- * {@linkcode module:pubsub/subscription} object, which will allow you to
- * interact with your subscription.
+ * Create a Subscription object. This command by itself will not run any API
+ * requests. You will receive a {@linkcode module:pubsub/subscription} object,
+ * which will allow you to interact with a subscription.
  *
  * @throws {Error} If a name is not provided.
  *
@@ -487,7 +486,7 @@ PubSub.prototype.subscribe = function(topic, subName, options, callback) {
  * @return {module:pubsub/subscription}
  *
  * @example
- * var subscription = pubsub.subscription('my-existing-subscription');
+ * var subscription = pubsub.subscription('my-subscription');
  *
  * // Register a listener for `message` events.
  * subscription.on('message', function(message) {
@@ -508,8 +507,7 @@ PubSub.prototype.subscription = function(name, options) {
 };
 
 /**
- * Create a Topic object to reference an existing topic. See
- * {module:pubsub/createTopic} to create a topic.
+ * Create a Topic object. See {module:pubsub/createTopic} to create a topic.
  *
  * @throws {Error} If a name is not provided.
  *
@@ -517,7 +515,7 @@ PubSub.prototype.subscription = function(name, options) {
  * @return {module:pubsub/topic}
  *
  * @example
- * var topic = pubsub.topic('my-existing-topic');
+ * var topic = pubsub.topic('my-topic');
  *
  * topic.publish({
  *   data: 'New message!'
--- a/lib/pubsub/subscription.js
+++ b/lib/pubsub/subscription.js
@@ -110,7 +110,7 @@ var util = require('../common/util.js');
  * // From {@linkcode module:pubsub/topic#subscription}:
  * //-
  * var topic = pubsub.topic('my-topic');
- * var subscription = topic.subscription('my-existing-subscription');
+ * var subscription = topic.subscription('my-subscription');
  * // `subscription` is a Subscription object.
  *
  * //-
@@ -167,6 +167,9 @@ function Subscription(pubsub, options) {
      * object with `autoCreate` set to `true`. Any extra configuration that is
      * normally required for the `create` method must be contained within this
      * object as well.
+     *
+     * **`autoCreate` is only available if you accessed this object
+     * through {module:pubsub/topic#subscription}.**
      *
      * @param {options=} options - Configuration object.
      * @param {boolean} options.autoCreate - Automatically create the object if
--- a/lib/pubsub/topic.js
+++ b/lib/pubsub/topic.js
@@ -68,7 +68,7 @@ function Topic(pubsub, name) {
     /**
      * Create a topic.
      *
-     * @param {object} config - See {module:pubsub#createTopic}.
+     * @param {object=} config - See {module:pubsub#createTopic}.
      *
      * @example
      * topic.create(function(err, topic, apiResponse) {
@@ -398,10 +398,9 @@ Topic.prototype.subscribe = function(subName, options, callback) {
 };
 
 /**
- * Create a Subscription object in reference to an existing subscription. This
- * command by itself will not run any API requests. You will receive a
- * {@linkcode module:pubsub/subscription} object, which will allow you to
- * interact with your subscription.
+ * Create a Subscription object. This command by itself will not run any API
+ * requests. You will receive a {@linkcode module:pubsub/subscription} object,
+ * which will allow you to interact with a subscription.
  *
  * @param {string} name - Name of the subscription.
  * @param {object=} options - Configuration object.
@@ -412,7 +411,7 @@ Topic.prototype.subscribe = function(subName, options, callback) {
  * @return {module:pubsub/subscription}
  *
  * @example
- * var subscription = topic.subscription('my-existing-subscription');
+ * var subscription = topic.subscription('my-subscription');
  *
  * // Register a listener for `message` events.
  * subscription.on('message', function(message) {
--- a/lib/resource/index.js
+++ b/lib/resource/index.js
@@ -244,8 +244,8 @@ Resource.prototype.getProjects = function(options, callback) {
 };
 
 /**
- * Create a Project object to reference an existing project. See
- * {module:resoucemanager/createProject} to create a project.
+ * Create a Project object. See {module:resoucemanager/createProject} to create
+ * a project.
  *
  * @throws {Error} If an ID is not provided.
  *
--- a/lib/resource/project.js
+++ b/lib/resource/project.js
@@ -71,7 +71,7 @@ function Project(resource, id) {
     /**
      * Create a project.
      *
-     * @param {object} config - See {module:resource#createProject}.
+     * @param {object=} config - See {module:resource#createProject}.
      *
      * @example
      * project.create(function(err, zone, apiResponse) {
--- a/lib/search/document.js
+++ b/lib/search/document.js
@@ -91,15 +91,6 @@ function Document(index, id) {
     /**
      * Get a document if it exists.
      *
-     * You may optionally use this to "get or create" an object by providing an
-     * object with `autoCreate` set to `true`. Any extra configuration that is
-     * normally required for the `create` method must be contained within this
-     * object as well.
-     *
-     * @param {options=} options - Configuration object.
-     * @param {boolean} options.autoCreate - Automatically create the object if
-     *     it does not exist. Default: `false`
-     *
      * @example
      * document.get(function(err, document, apiResponse) {
      *   // `document.fields` has been populated.
@@ -145,8 +136,6 @@ Document.prototype.addField = function(name) {
 
 /**
  * Create a document.
- *
- * @param {object} config - See {module:search/index#createDocument}.
  *
  * @example
  * document.create(function(err, document, apiResponse) {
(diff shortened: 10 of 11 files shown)


lib/bigquery/dataset.js
```diff
@@ -149,7 +165,11 @@ Dataset.prototype.delete = function(options, callback) {
     deleteContents: !!options.force
   };
 
-  this.makeReq_('DELETE', '', query, null, callback);
+  this.request({
+    method: 'DELETE',
+    uri: '',
```
callmehiphop: Could we just make the default for uri and empty string?

lib/bigquery/dataset.js
```diff
@@ -165,18 +185,8 @@ Dataset.prototype.delete = function(options, callback) {
...
-      callback(err, null, resp);
-      return;
-    }
-
-    that.metadata = resp;
-
-    callback(null, that.metadata, resp);
-  });
+Dataset.prototype.getMetadata = function() {
+  ServiceObject.prototype.getMetadata.apply(this, arguments);
```
callmehiphop: It took me a minute to remember why it's like this way (documentation purposes, yes?) . I believe it's possible to document methods/class/etc. without the code actually being present. Although I'm not sure what kind of effort that would involve to bring into our current docs site.

lib/bigquery/job.js
```diff
@@ -51,11 +58,18 @@ var is = require('is');
  * @constructor
  */
 function Job(bigQuery, id) {
+  ServiceObject.call(this, {
+    parent: bigQuery,
+    baseUrl: '/jobs',
+    id: id,
+    inherit: []
```
callmehiphop: Could we name this something else, like `borrows`? When I initially saw this I thought maybe you passed in an array of constructors and I got confused.

It also might be worth noting that I saw a reference to `exclude` but not `inherit` within the `ServiceObject` constructor.

lib/bigquery/table.js
```diff
@@ -346,8 +364,8 @@ Table.prototype.createWriteStream = function(metadata) {
  * @example
  * table.delete(function(err, apiResponse) {});
  */
-Table.prototype.delete = function(callback) {
-  this.makeReq_('DELETE', '', null, null, callback);
+Table.prototype.delete = function() {
+  return ServiceObject.prototype.delete.apply(this, arguments);
```
callmehiphop: This is the first time I've seen a `return`, I know we don't really need it since everything is async, but could we be consistent with either returning or not returning across the board?

lib/common/service-object.js
```diff
@@ -0,0 +1,261 @@
...
+ * @param {?error} callback.err - An error returned while making this request.
+ * @param {object} callback.apiResponse - The full API response.
+ */
+ServiceObject.prototype.delete = function(callback) {
+  var reqOpts = {
+    method: 'DELETE',
+    uri: ''
+  };
+
+  ServiceObject.prototype.request.call(this, reqOpts, function(err, resp) {
```
callmehiphop: Do we do this because we have to overwrite delete elsewhere for documentation purposes?

lib/common/service-object.js
```diff
@@ -0,0 +1,261 @@
...
+ * @param {?error} callback.err - An error returned while making this request.
+ * @param {object} callback.instance - The instance.
+ * @param {object} callback.apiResponse - The full API response.
+ */
+ServiceObject.prototype.getOrCreate = function(options, callback) {
+  var self = this;
+
+  this.get(function(err, instance, apiResponse) {
+    if (err) {
+      self.create(options, callback);
```
callmehiphop: Is it possible the upstream API will send back an error even though the object does exist?

lib/common/service-object.js
```diff
@@ -0,0 +1,261 @@
...
+/**
+ * Get or create this object. If it doesn't exist, it will be created.
+ *
+ * @param {object=} options - Configuration object.
+ * @param {function=} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request.
+ * @param {object} callback.instance - The instance.
+ * @param {object} callback.apiResponse - The full API response.
+ */
+ServiceObject.prototype.getOrCreate = function(options, callback) {
```
callmehiphop: I know we've talked about this before.. but can we please not name this method `getOrCreate` :)

I don't think it would take very much effort to support some kind of auto create configurations within the `get` method..

```javascript
if (err && config.autoCreate) {
  return this.create(config, callback);
} else if (err) {
  return callback(err, null, apiResponse);
}

// success case
```

lib/bigquery/dataset.js
stephenplusplus: Sounds okay to me.

lib/bigquery/dataset.js
stephenplusplus: Yep, doc purposes. I'm not either, but the doc triplication is a bummer.

lib/bigquery/job.js
stephenplusplus: As you can tell, I went through a few names :) `borrows` is fine.

lib/common/service-object.js
stephenplusplus: Actually, **get this!** It's because Compute has it's own `request` method; Compute API calls generally receive an Operation response, and we return an Operation object. To simplify all of that, we overwrote `request` instead of just using the default behavior from ServiceObject. Calling the ServiceObject.request method insures we don't get caught up in any custom implementations that were layered on top.

lib/common/service-object.js
stephenplusplus: Probably definitely.

lib/common/service-object.js
stephenplusplus: Sure sure.

lib/common/service-object.js
callmehiphop: Got it

stephenplusplus:
@callmehiphop can you take another quick glance over the changes I've made ([latest commit](https://github.com/stephenplusplus/gcloud-node/commit/9a752c2be0a34f30874e8088c043a42484e5490a))? I think I may have come up with an easier way to handle the `inherits/excludes/borrows` mess.

Also, the docs are much simpler to avoid all of the replication. The `create`, `get`, and `getOrCreate` docs are basically shells that don't show anything more than basic usage that will always be true, even if the upstream API arguments change (like `zone#createVM`'s `config`).

callmehiphop:
I like this approach **much** more than the previous. 

However it still kinda feels odd to me, mainly because I don't think we should be adding code for the sake of documentation, even if it's just a little. I think I've suggested this before, but have we considered moving away from dox? JSDoc3 actually has pretty nice support for spitting out plain old JSON (minus custom tags). It also allows you to document "virtual" members, which is just code that exists elsewhere. (e.g. `getOrCreate`).

I realize this would be a large undertaking, so this is obviously not the time or place for a change like that, but it's still something I think we should consider. Aside from that, I think it looks good!

stephenplusplus:
> mainly because I don't think we should be adding code for the sake of documentation, even if it's just a little

Which code does that?

stephenplusplus:
> I think I've suggested this before, but have we considered moving away from dox? JSDoc3 actually has pretty nice support for spitting out plain old JSON (minus custom tags).

I'd be in support of this, but given the scope of the work involved, it's a pretty intimidating task to take on. Maybe switching over should happen when we get to the gcloud-* common docs task.

callmehiphop:
Maybe I misunderstood, but I thought the main reason to switch to an object containing method names flagged as `true` was for documentation purposes?

callmehiphop:
> I'd be in support of this, but given the scope of the work involved, it's a pretty intimidating task to take on. Maybe switching over should happen when we get to the gcloud-* common docs task.

Yeah totally!

stephenplusplus:
It does work for that purpose, but it's also to communicate to `ServiceObject` what methods we're going to inherit.

callmehiphop:
> It does work for that purpose, but it's also to communicate to ServiceObject what methods we're going to inherit.

Gotcha! Do you think it would be better to make a custom `inherit` function to do that? It might be slightly more efficient to tie `ServiceObject` methods to an api's prototype chain

```javascript
inherit(BigQuery, ServiceObject, {
  getOrCreate: null
});
```

Either way, I approve!

stephenplusplus:
RE: the `getOrCreate: null`, I prefer the whitelist over the blacklist because:

  1. each class has to explicitly ask for what methods they want
  1. having all of the methods listed that you want to inherit is a logical place to put the documentation

There definitely might be a better way to do this though, so I'll think about it as I go on. Something worth noting is in some cases, we don't know if a class needs a method or not until it's instantiated.

callmehiphop:
> Something worth noting is in some cases, we don't know if a class needs a method or not until it's instantiated.

Why is that?

lib/compute/snapshot.js
```diff
@@ -50,13 +59,93 @@ var util = require('../common/util.js');
...
+ * //-
+ * var disk = gce.zone('us-central1-a').disk('disk-name');
+ * var snapshot = disk.snapshot('disk-snapshot-name');
  */
-function Snapshot(compute, name) {
-  this.compute = compute;
-  this.name = name;
-  this.metadata = {};
+function Snapshot(scope, name) {
+  var isDisk = scope.constructor.name === 'Disk';
```
stephenplusplus: @callmehiphop here's a place -- a Snapshot can be instantiated from a Disk (`compute().disk().snapshot()`) or the core `compute().snapshot()`. A Snapshot must be paired with a Disk in order to be created. Similar to how an Instance (VM) can't be created without a Zone. To enable these methods only where appropriate, we sniff if this scope is a Disk.

stephenplusplus:
@callmehiphop feel free to take a look at the new implementation of `exists` and `get` with `autoCreate: true`. `getOrCreate` has been removed. I think the pattern I'll be following re: docs, inheritance, ServiceObject is pretty much on lock now. I'll just be carrying over the code to the libraries that haven't been implemented yet.

Then... testing. Dun, dun, dun.

lib/common/service-object.js
```diff
@@ -0,0 +1,295 @@
...
+ */
+ServiceObject.prototype.delete = function(callback) {
+  var reqOpts = {
+    method: 'DELETE',
+    uri: ''
+  };
+
+  // The `request` method may have been overridden to hold any special behavior.
+  // Ensure we call the original `request` method.
+  ServiceObject.prototype.request.call(this, reqOpts, function(err, resp) {
```
callmehiphop: Do we still expect `request` to be overridden?

stephenplusplus: compute still does to handle dealing with operations

callmehiphop:
> getOrCreate has been removed

yay! looks good

callmehiphop:
This is getting pretty huge, GitHub can't even display the entire diff because of the size. Any chance we could break any of this up?

stephenplusplus:
Yes, very good idea. I'll start with Service & ServiceObject.

stephenplusplus:
@callmehiphop: #928

PR #907, opened 2015-10-08
core: emit auto-auth errors

stephenplusplus:
:warning: Opinions needed on this one.

We have been ignoring the error that google-auth-library returns when it can't find the environment's default credentials.

As explained in the comment from the code:

> google-auth-library returns a "Could not load..." error if it can't get an access token. However, it's possible an API request doesn't need to be authenticated, e.g. when downloading a file from a public bucket. We consider this error a warning, and allow the request to go through without authentication, relying on the upstream API to return an error the user would find more helpful, should one occur.

We have a bug currently because of this: https://github.com/GoogleCloudPlatform/nodejs-getting-started/issues/25 - I'll explain this in a comment on the offending line of code.

Considering this instantiation:

``` js

// No credentials provided:
var dataset = gcloud.datastore().dataset({
  projectId: 'grape-spaceship-123'
});
```

Here is what will happen if we merge this PR:

``` js
dataset.runQuery(/*...*/, function(err) {
  // The error comes from google-auth-library:
  // err.message = 'Could not load the default credentials. Browse to https://developers.google.com/accounts/docs/application-default-credentials for more information.'
});
```

Pretending we didn't merge this PR, and just fixed the bug mentioned earlier, here's what would happen:

``` js
dataset.runQuery(/*...*/, function(err) {
  // The error comes from the API:
  // err.message = 'Unauthorized'
});
```

To get the better error (the first example), we have to not allow unauthenticated requests. I'm not sure if/what/how much of our API allows unauthenticated requests, or if we even want to support that in the first place?

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -337,13 +337,7 @@ function makeAuthenticatedRequestFactory(config) {
     }
 
     function onAuthenticated(err, authenticatedReqOpts) {
-      // google-auth-library returns a "Could not load..." error if it can't get
-      // an access token. However, it's possible an API request doesn't need to
-      // be authenticated, e.g. when downloading a file from a public bucket. We
-      // consider this error a warning, and allow the request to go through
-      // without authentication, relying on the upstream API to return an error
-      // the user would find more helpful, should one occur.
-      if (err && err.message.indexOf('Could not load') === -1) {
+      if (err) {
         if (stream) {
           stream.destroy(err);
         } else {
--- a/test/common/util.js
+++ b/test/common/util.js
@@ -535,26 +535,6 @@ describe('common/util', function() {
           };
         });
 
-        it('should not care about "Could not load" errors', function(done) {
-          var error = new Error('Could not load');
-
-          utilOverrides.decorateRequest = function() {};
-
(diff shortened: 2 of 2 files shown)


lib/common/util.js
```diff
@@ -337,13 +337,7 @@ function makeAuthenticatedRequestFactory(config) {
     }
 
     function onAuthenticated(err, authenticatedReqOpts) {
-      // google-auth-library returns a "Could not load..." error if it can't get
-      // an access token. However, it's possible an API request doesn't need to
-      // be authenticated, e.g. when downloading a file from a public bucket. We
-      // consider this error a warning, and allow the request to go through
-      // without authentication, relying on the upstream API to return an error
-      // the user would find more helpful, should one occur.
-      if (err && err.message.indexOf('Could not load') === -1) {
```
stephenplusplus: So here we are, ignoring the "Could not load" error. The whole error handling block is skipped, but the bug is: the `authenticatedReqOpts` argument to this function is `undefined`. To fix this, we would just do: `authenticatedReqOpts = reqOpts` (the original reqOpts that we tried to authenticate. Since it failed, we would just use the unauthenticated ones)

stephenplusplus:
Let's just go with it.

PR #913, opened 2015-10-14
datastore: properly parse HTTP responses from the API

### LINKED ISSUE #910: Problems with Datastore
I am trying to get data from Datastore using the code below. However, it seems like something is not working properly.
```
    projectId: myProjectId
})
var dataset = gcloud.datastore.dataset()
var employeeQuery = dataset.createQuery('Employee')
dataset.runQuery(employeeQuery, function(err, entities) {
    console.log(err)
    console.log(entities)
})
```

And here is the result.
```
{ [Error: Error during request.]
  errors: undefined,
  code: undefined,
  message: 'Error during request.',
  response: undefined }
undefined
```
I am running this on an Compute Engine instance, so there's should not be any problems with the authentication. I have tried using the Python library for Gcloud and there's no problems with it.

Also, I can use the code below to create a new entity but the 'err' in callback still give the same error, although the entity is already created.
```
var dataset = gcloud.datastore.dataset()
var key = dataset.key('Employee')
dataset.save({
  key: key,
  data: {
    name: 'Bob',
    age: 20
  }
}, function(err) {
  console.log(err)
})
```

Am I doing something wrong here ?

stephenplusplus:
Fixes #910 

The bug originally caught [here](https://github.com/GoogleCloudPlatform/gcloud-node/issues/910#issuecomment-147820786) by @electricFeel (thanks!) resulted in a re-factor, also bringing in some of the sane changes from #897 (the datastore v1beta3 upgrade).

PTAL asap, as we should get out a patch release as soon as possible.

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -139,7 +139,12 @@ util.ApiError = function(errorBody) {
 function handleResp(err, resp, body, callback) {
   callback = callback || noop;
 
-  var parsedResp = util.parseApiResp(err, resp, body);
+  var parsedResp = extend(
+    true,
+    { err: err || null },
+    util.parseHttpRespMessage(resp),
+    util.parseHttpRespBody(body)
+  );
 
   callback(parsedResp.err, parsedResp.body, parsedResp.resp);
 }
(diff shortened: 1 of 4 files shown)


lib/common/util.js
```diff
@@ -147,50 +152,64 @@ function handleResp(err, resp, body, callback) {
...
+function parseHttpRespMessage(httpRespMessage) {
+  var parsedHttpRespMessage = {
+    resp: httpRespMessage
   };
 
-  if (resp && (resp.statusCode < 200 || resp.statusCode > 299)) {
+  if (httpRespMessage.statusCode < 200 || httpRespMessage.statusCode > 299) {
     // Unknown error. Format according to ApiError standard.
-    parsedResp.err = new util.ApiError({
+    parsedHttpRespMessage.err = util.ApiError({
```
callmehiphop: should we `new` this up? even if it's just for the sake of clarity?

stephenplusplus: Hmm, yes. Good catch, I didn't mean to remove it!

callmehiphop:
Thank you, sir!

PR #915, opened 2015-10-17
compute: create default firewall rules

stephenplusplus:
Related http://stackoverflow.com/questions/31509722/how-do-i-enable-http-traffic-for-gce-instance-templates
Fixes https://github.com/stephenplusplus/gcloud-deploy/issues/4

We had a "maybe-bug" in our code that could lead to a user thinking they made a VM that accepts HTTP and/or HTTPS connections, but in actuality doesn't.

My theory about how the Dev Console UI works:

If you create an Instance and select "Allow HTTP traffic" and/or "Allow HTTPS traffic", it will create firewall rules in the default network: `default-allow-http` and `default-allow-https`. The VM is created with tags to match `http-server` and `https-server`, which are the target tags on the firewall rules.

When I developed the behavior of `zone.createVM()` that accepts `config.http = true` and `config.https = true`, I assumed _all_ project's networks had these firewall rules by default. But, if my theory is right, I only had them because I used the UI to create an http/https VM before, so my project had the rules.

So, this PR will attempt to create `default-allow-http` and `default-allow-https` (when that functionality is asked for), and if they already exist, it just moves on and creates the VM. If they don't exist, it creates them, then the VM. In other words, this should basically work like the Dev Console UI does (if my theory is correct).

### DIFF
--- a/lib/compute/zone.js
+++ b/lib/compute/zone.js
@@ -20,6 +20,7 @@
 
 'use strict';
 
+var async = require('async');
 var extend = require('extend');
 var format = require('string-format-obj');
 var gceImages = require('gce-images');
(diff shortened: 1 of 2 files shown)


jgeewax:
LGTM

stephenplusplus:
@callmehiphop PTAL

callmehiphop:
Looks good :)

PR #919, opened 2015-10-28
core: [common/utils] fix response error handling.

otolab:
Fixes #918

PR #920, opened 2015-10-28
don't parse an undefined message or body

### LINKED ISSUE #918: Encountered an exception by network error.
My utility has encountered an exception. It's occurred in the "libs/common/utils.js" by network error.

Looks like a regression from https://github.com/GoogleCloudPlatform/gcloud-node/commit/8b2bbf2d5f9da7cceab4d544cf6983e27d2ad534#diff-638107c62532c8d249c9f17a0c7fb355L169

`resp` check has been removed. therefore, It throws an exception when `httpRespMessage` is `undefined`.

I wrote fix patch. please see pull request.

stephenplusplus:
Fixes #918 
Fixes #919 

We previously were trying to parse a non-existent response message and body. No more!

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -142,8 +142,8 @@ function handleResp(err, resp, body, callback) {
   var parsedResp = extend(
     true,
     { err: err || null },
-    util.parseHttpRespMessage(resp),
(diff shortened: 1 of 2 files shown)


callmehiphop:
:dancer:

PR #923, opened 2015-10-30
core: use error message from response body

stephenplusplus:
This will use the error message returned from the API response body instead of "Error during request" -- if there was one. If not, we still fall back to _something_, which is the "Error during request" message.


callmehiphop:
:+1:

PR #928, opened 2015-11-09
Implement Service & ServiceObject classes

stephenplusplus:
Splitting up the massive #904 into pieces. The first is the new classes, `Service` and `ServiceObject`.

### DIFF
--- a/lib/common/service-object.js
+++ b/lib/common/service-object.js
@@ -1,0 +1,307 @@
+/*!
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*!
+ * @module common/serviceObject
+ */
+
+'use strict';
+
+var exec = require('methmeth');
+var is = require('is');
+
+/**
+ * @type {module:common/util}
+ * @private
+ */
+var util = require('./util.js');
+
+/**
+ * ServiceObject is a base class, meant to be inherited from by a "service
+ * object," like a BigQuery dataset or Storage bucket.
+ *
+ * Most of the time, these objects share common functionality; they can be
+ * created or deleted, and you can get or set their metadata.
+ *
+ * By inheriting from this class, a service object will be extended with these
+ * shared behaviors. Note that any method can be overridden when the service
+ * object requires specific behavior.
+ *
+ * @private
+ *
+ * @param {object} config - Configuration object.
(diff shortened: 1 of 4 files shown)


lib/common/service-object.js
```diff
@@ -0,0 +1,303 @@
...
+
+  if (is.object(options)) {
+    args.push(options);
+  }
+
+  // Wrap the callback to return *this* instance of the object, not the newly-
+  // created one.
+  function onCreate(err, instance, apiResponse) {
+    if (err) {
+      callback(err, null, apiResponse);
```
callmehiphop: Per the JSDoc it looks like callbacks are optional, but it doesn't look like we're setting a default anywhere.

stephenplusplus: Will fix. Not optional!

lib/common/service-object.js
```diff
@@ -0,0 +1,303 @@
...
+ServiceObject.prototype.delete = function(callback) {
+  var reqOpts = {
+    method: 'DELETE',
+    uri: ''
+  };
+
+  // The `request` method may have been overridden to hold any special behavior.
+  // Ensure we call the original `request` method.
+  ServiceObject.prototype.request.call(this, reqOpts, function(err, resp) {
+    callback(err, resp);
```
callmehiphop: Same thing with the callback here! Not sure if callback is optional on this one or not.

callmehiphop: Why don't we just say..

```javascript
ServiceObject.prototype.request.call(this, reqOpts, callback);
```

Are there additional arguments we don't want to pass to the callback? Sorry if I asked about this before, it feels familiar.

lib/common/service-object.js
```diff
@@ -0,0 +1,303 @@
...
+      if (err.code === 404) {
+        callback(null, false);
+      } else {
+        callback(err);
+      }
+
+      return;
+    }
+
+    callback(null, false);
```
callmehiphop: Should this be `true`?

lib/common/service-object.js
stephenplusplus: Haha, you did, and it's because this all patches back to `request`, which gives a long request/response object with details of the headers, etc. Instead of documenting yet another argument to all of our methods' callbacks, if there's an error, it's attached to the error object as `response`.

lib/common/service-object.js
stephenplusplus: Yep. I should fix that.

lib/common/service-object.js
```diff
@@ -0,0 +1,303 @@
...
+  var autoCreate = config.autoCreate && is.fn(this.create);
+  delete config.autoCreate;
+
+  this.getMetadata(function(err, metadata) {
+    if (err) {
+      if (err.code === 404 && autoCreate) {
+        if (!is.empty(config)) {
+          self.create(config, callback);
+        } else {
+          self.create(callback);
```
callmehiphop: Warning: nitpick ahead! Feel free to ignore :)

What are your feelings on something like..

```javascript
var args = [callback];

if (!is.empty(config)) {
  args.unshift(config);
}

self.create.apply(self, args);
```

?

stephenplusplus: Fine by me, Mr. Nitpick!

callmehiphop: :)

stephenplusplus:
Fixed the things! PTAL

callmehiphop:
Looks good! Thanks :)

PR #931, opened 2015-11-09
Implement Service & Service Object for BigQuery

stephenplusplus:
For #904 

This hooks BigQuery up with Service & ServiceObject.

### DIFF
--- a/lib/bigquery/dataset.js
+++ b/lib/bigquery/dataset.js
@@ -22,18 +22,25 @@
 
 var extend = require('extend');
 var is = require('is');
+var nodeutil = require('util');
 
 /**
- * @type {module:bigquery/table}
+ * @type {module:common/serviceObject}
  * @private
  */
-var Table = require('./table.js');
+var ServiceObject = require('../common/service-object.js');
 
 /**
  * @type {module:common/streamrouter}
  * @private
  */
 var streamRouter = require('../common/stream-router.js');
+
+/**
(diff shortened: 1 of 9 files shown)


lib/bigquery/dataset.js
```diff
@@ -46,20 +53,121 @@ var streamRouter = require('../common/stream-router.js');
...
+ * });
+ * var dataset = bigquery.dataset('institutions');
  */
 function Dataset(bigQuery, id) {
+  var methods = {
+    /**
+     * Create a dataset.
+     *
+     * @example
+     * dataset.create(function(err, dataset, apiResponse) {
```
callmehiphop: Do we want to document the params here? Or is this something that we won't really be advertising and it's just going to be masked behind `autoCreate`?

callmehiphop: Or does this get mixed in with the `ServiceObject` docs?

stephenplusplus: This method is exposed in the docs, but deliberately undocumented in this case, as `create` takes no arguments and having duplicating the callback argument docs is not ideal.

This is a very unique case, though, as most create methods will accept an `options` or `config` object, and we will show param docs that link to the original method (`parentClass.create____`) so the user can see the full list of docs.

stephenplusplus:
Docs preview: http://stephenplusplus.github.io/gcloud-node/#/docs/master/bigquery

PR #932, opened 2015-11-09
Implement Service & Service Object for Compute

stephenplusplus:
For #904

This hooks Compute up with Service & ServiceObject.

Docs: http://stephenplusplus.github.io/gcloud-node/#/docs/master/compute

### DIFF
--- a/lib/compute/disk.js
+++ b/lib/compute/disk.js
@@ -23,6 +23,19 @@
 var extend = require('extend');
 var format = require('string-format-obj');
 var is = require('is');
+var nodeutil = require('util');
+
+/**
+ * @type {module:common/serviceObject}
+ * @private
+ */
+var ServiceObject = require('../common/service-object.js');
+
+/**
+ * @type {module:compute/snapshot}
+ * @private
+ */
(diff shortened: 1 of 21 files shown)


stephenplusplus:
@callmehiphop rebased :)

stephenplusplus:
Added the docs: http://stephenplusplus.github.io/gcloud-node/#/docs/master/compute

lib/compute/disk.js
```diff
@@ -57,13 +70,92 @@ var util = require('../common/util.js');
...
-  this.zone = zone;
+  var methods = {
+    /**
+     * Create a persistent disk.
+     *
+     * @param {object} config - See {module:compute/zone#createDisk}.
+     *
+     * @example
+     * var config = {
+     *   // ...
```
callmehiphop: Would a user have to go to the upstream docs to see the available configurations?

stephenplusplus: Yeah, they go to `compute/zone#createDisk`, which is linked as the `config` parameter.

callmehiphop: Ok cool

PR #929, opened 2015-11-09
compute: fix detachDisk bug

stephenplusplus:
Fixes #922

PR #933, opened 2015-11-09
compute: detachDisk disk cache

callmehiphop:
#929 introduced a bug where we wouldn't update the disk cache (metadata) before detaching a disk if the metadata already was present. This logic was a problem because it wasn't accounting for metadata being out of sync.

### DIFF
--- a/test/compute/vm.js
+++ b/test/compute/vm.js
@@ -141,14 +141,18 @@ describe('VM', function() {
       name: DISK.formattedName
     });
 
+    var METADATA = {
+      disks: [
+        {
+          source: DEVICE_NAME,
+          deviceName: DEVICE_NAME
+        }
+      ]
+    };
+
     beforeEach(function() {
-      vm.metadata = {
-        disks: [
-          {
-            source: DEVICE_NAME,
-            deviceName: DEVICE_NAME
-          }
-        ]
+      vm.getMetadata = function(callback) {
+        callback(null, METADATA, METADATA);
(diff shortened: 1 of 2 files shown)


test/compute/vm.js
```diff
@@ -141,14 +141,19 @@ describe('VM', function() {
...
     beforeEach(function() {
-      vm.metadata = {
-        disks: [
-          {
-            source: DEVICE_NAME,
-            deviceName: DEVICE_NAME
-          }
-        ]
+      vm.getMetadata = function(callback) {
+        vm.METADATA = METADATA;
```
stephenplusplus: I don't think this is having an affect (`vm.METADATA`).

callmehiphop: Ah, my bad. Fixed!

stephenplusplus:
Gotta update the test that checks the error message.

test/compute/vm.js
stephenplusplus: Can you kill the new line if it doesn't have any affect? I see future Stephen being confused.

callmehiphop:
> Gotta update the test that checks the error message.

Ah shoot, I only changed it to satisfy lint rules. My bad!

test/compute/vm.js
callmehiphop: Done!

stephenplusplus:
Woo, thanks for catching this!

PR #939, opened 2015-11-11
compute: turn operations into event emitters

### LINKED ISSUE #921: compute onComplete example should show "the right way" to attach handlers?
On https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.24.1/compute/operation?method=onComplete

```js
operation.onComplete(function(err, metadata) {
  if (err.code === 'OPERATION_INCOMPLETE') {
    // The operation is not complete yet. You may want to register another
    // `onComplete` listener or queue for later.
  }

  if (!err) {
    // Operation complete!
  }
});
```

This example doesn't really show me how I'm supposed to continue attaching handlers (and I would actually need to redesign the flow of my code to make this work).

Can we update the example to show exactly what "the right way" to do this is?

stephenplusplus:
Fixes #921

### DIFF
--- a/test/compute/operation.js
+++ b/test/compute/operation.js
@@ -17,6 +17,7 @@
 'use strict';
 
 var assert = require('assert');
+var extend = require('extend');
 var mockery = require('mockery');
 var nodeutil = require('util');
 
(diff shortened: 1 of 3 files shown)


test/compute/operation.js
```diff
@@ -179,136 +201,209 @@ describe('Operation', function() {
...
+
+      operation.removeListener('complete', util.noop);
+      assert.strictEqual(operation.hasActiveListeners, true);
+
+      operation.removeListener('complete', util.noop);
+      assert.strictEqual(operation.hasActiveListeners, false);
+    });
+  });
+
+  describe.skip('startPolling_', function() {
```
callmehiphop: Should we not skip this?

stephenplusplus: But it makes getting green builds so much easier!

callmehiphop: :stuck_out_tongue_closed_eyes:

PR #940, opened 2015-11-12
datastore: don't require auth with local datastore server

### LINKED ISSUE #936: Stubbing out authentication
I have the gcloud module talking to a local cloud datastore using the gcd tool for automated testing but am finding that it only works if I provide real credentials from my google console. Is there a way to stub out the authentication process so I can use test credentials?

stephenplusplus:
Fixes #936

When a service object is created (such as a Dataset object), we have the choice of telling our "makeAuthenticatedRequestFactory" function that we don't care about the authentication part via an option called `customEndpoint`. A Dataset is the one place we use this option in the event a user is trying to communicate with their own Datastore server.

We were skipping the auth when:
- A user instantiated a Dataset providing an `apiEndpoint` option:
  
  ``` js
  var dataset = datastore.dataset({ apiEndpoint: 'http://localhost:8080' });
  ```

But, we also need to support when:
- A user has the `DATASTORE_HOST` env var set

We have had half of the puzzle solved; we detect the presence of that env var and send the API requests to it accordingly. But, we forgot to update `customEndpoint`, the option that skips authenticating.

// @brainsiq

### DIFF
--- a/lib/datastore/dataset.js
+++ b/lib/datastore/dataset.js
@@ -110,17 +110,17 @@ function Dataset(options) {
     throw util.missingProjectIdError;
   }
 
+  this.determineApiEndpoint_(options.apiEndpoint);
+  this.namespace = options.namespace;
+  this.projectId = options.projectId;
+
   this.makeAuthenticatedRequest_ = util.makeAuthenticatedRequestFactory({
-    customEndpoint: typeof options.apiEndpoint !== 'undefined',
+    customEndpoint: this.customEndpoint,
     credentials: options.credentials,
     keyFile: options.keyFilename,
     scopes: SCOPES,
     email: options.email
   });
-
-  this.apiEndpoint = Dataset.determineApiEndpoint_(options);
-  this.namespace = options.namespace;
-  this.projectId = options.projectId;
 }
 
 nodeutil.inherits(Dataset, DatastoreRequest);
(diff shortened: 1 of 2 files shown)


lib/datastore/dataset.js
```diff
@@ -132,24 +132,27 @@ nodeutil.inherits(Dataset, DatastoreRequest);
  *
  * @private
  *
- * @param {object} options - Configuration object.
- * @param {string=} options.apiEndpoint - Custom API endpoint.
+ * @param {string} customApiEndpoint - Custom API endpoint.
  */
-Dataset.determineApiEndpoint_ = function(options) {
-  var apiEndpoint = 'https://www.googleapis.com';
+Dataset.prototype.determineApiEndpoint_ = function(customApiEndpoint) {
```
jgeewax: Wouldn't the side-effect style here make this method harder to unit test? Just a thought (we changed from return "the endpoint" to this.apiEndpoint = "the endpoint";)

jgeewax:
LGTM

lib/datastore/dataset.js
stephenplusplus: Testing this method by itself wasn't bad, but testing how it integrates with the constructor wasn't great. I much prefer the no-side effect route, but couldn't find a way to easily fit that in here. We not only need "what is the API endpoint?" answered, but "is it the default one or a custom one?" Ideally, one function can answer both of those questions, so I could either return an object like `{ isCustom: false, apiEndpoint: 'http://...' }` or just treat this method as an extension of the initialization process, giving it full control of setting instance variables. (Note the method moved from static to instance)

I'm not a big fan of either approach, really. But, we have two **major** overhauls to Datastore coming soon (v1beta3 upgrade which is a complete re-write (#897), and after that lands, we will implement the `getOrCreate` base classes, "Service", and "ServiceObject"). Those will probably/definitely force a new solution to this problem fit in with the changes.

callmehiphop:
:+1:

PR #942, opened 2015-11-14
Implement Service & Service Object for DNS

stephenplusplus:
For #904

This hooks DNS up with Service & ServiceObject.

### DIFF
--- a/system-test/dns.js
+++ b/system-test/dns.js
@@ -32,61 +32,64 @@ var DNS_DOMAIN = process.env.GCLOUD_TESTS_DNS_DOMAIN;
 
 // Only run the tests if there is a domain to test with.
 (DNS_DOMAIN ? describe : describe.skip)('dns', function() {
-  var ZONE;
-  var ZONENAME = 'test-zone-' + uuid.v4().substr(0, 18);
+  if (!DNS_DOMAIN) {
+    // The test runner still executes this function, even if it is skipped.
+    return;
+  }
 
-  var records = {};
+  var ZONE_NAME = 'test-zone-' + uuid.v4().substr(0, 18);
+  var ZONE = dns.zone(ZONE_NAME);
 
-  function createRecords() {
-    records.a = ZONE.record('a', {
+  var records = {
+    a: ZONE.record('a', {
       ttl: 86400,
       name: DNS_DOMAIN,
       data: '1.2.3.4'
-    });
+    }),
 
-    records.aaaa = ZONE.record('aaaa', {
+    aaaa: ZONE.record('aaaa', {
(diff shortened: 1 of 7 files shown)


system-test/dns.js
```diff
@@ -142,16 +135,6 @@ var DNS_DOMAIN = process.env.GCLOUD_TESTS_DNS_DOMAIN;
     ZONE.delete({ force: true }, done);
   });
 
-  it('should create a zone', function(done) {
```
callmehiphop: Are we dropping this because we'd get an error in our afterEach/after?

stephenplusplus: I don't think so. I think only because we were creating all over the place, and we usually just use the fact that it was created in the before hook to prove it works.

callmehiphop: I commented on the wrong line! I meant to do it on 172 :((((

stephenplusplus: Same answer!

system-test/dns.js
```diff
@@ -94,26 +92,26 @@ var DNS_DOMAIN = process.env.GCLOUD_TESTS_DNS_DOMAIN;
...
         '1 21600 3600 1209600 300'
       ].join(' ')
-    });
+    }),
 
-    records.spf = ZONE.record('spf', {
+    spf: ZONE.record('spf', {
       ttl: 21600,
       name: DNS_DOMAIN,
       data: 'v=spf1 mx:' + DNS_DOMAIN.replace(/.$/, '') + ' -all'
```
callmehiphop: This threw an error for me locally, I don't have `GCLOUD_TESTS_DNS_DOMAIN` set.

callmehiphop: Probably worth mentioning that master does not throw the same error.

stephenplusplus: Ooh good catch.

system-test/dns.js
callmehiphop: But 172 is for deleting!

system-test/dns.js
stephenplusplus: Yeah it's because I moved the line that throws outside of a mocha hook.  Because it was in a hook of a skipped suite, it didn't run.

system-test/dns.js
stephenplusplus: We delete the zone created in the before in the after! That's a sentence!

callmehiphop: okok! That's what I was asking, my wording was probably just not so awesome :)

system-test/dns.js
stephenplusplus: Fixed this lil guy

PR #943, opened 2015-11-14
Implement Service & Service Object for Resource

stephenplusplus:
For #904

This hooks resource manager up with Service & ServiceObject.

### DIFF
--- a/lib/resource/index.js
+++ b/lib/resource/index.js
@@ -22,12 +22,19 @@
 
 var extend = require('extend');
 var is = require('is');
+var nodeutil = require('util');
 
 /**
  * @type {module:resource/project}
  * @private
  */
 var Project = require('./project.js');
+
+/**
+ * @type {module:common/service}
+ * @private
+ */
+var Service = require('../common/service.js');
 
 /**
  * @type {module:common/streamrouter}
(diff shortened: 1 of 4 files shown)


lib/resource/project.js
```diff
@@ -59,66 +67,130 @@ var util = require('../common/util.js');
...
+     */
+    getMetadata: true,
+
+    /**
+     * Set the project's metadata.
+     *
+     * @resource [projects: update API Documentation]{@link https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/update}
+     * @resource [Project Resource]{@link https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects#Project}
+     *
+     * @private
```
callmehiphop: Is this supposed to be private?

callmehiphop: Or are we just marking it private because the upstream APIs don't allow requests yet?

stephenplusplus: Actually, I think it's because I originally thought that the API was busted, but it's just that you have to be authenticated as yourself (not a service account). I need to just add a line about that and remove the private tag.

lib/resource/index.js
```diff
@@ -88,24 +80,28 @@ function Resource(options) {
...
+  this.defaultProjectId_ = options.projectId;
 }
 
+nodeutil.inherits(Resource, Service);
+
 /**
  * Create a project.
  *
+ * **This method only works if you are authenticated as yourself, e.g. using the
+ * gcloud SDK.**
```
stephenplusplus: RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/943#discussion_r44863079

lib/resource/project.js
callmehiphop: Cool, is there any way we could check for that and dynamically decide whether or not we need to skip the set metadata tests?

stephenplusplus: I don't think we can dynamically check, but we can have an env var maybe that will tell the system tests to run. Could you open a new issue?

callmehiphop: For sure, thanks!

stephenplusplus: And to you. Good catch!

PR #944, opened 2015-11-14
Implement Service & Service Object for PubSub

stephenplusplus:
For #904

This hooks Pub/Sub up with Service & ServiceObject.

### DIFF
--- a/lib/pubsub/iam.js
+++ b/lib/pubsub/iam.js
@@ -20,13 +20,22 @@
 
 'use strict';
 
-var is = require('is');
 var arrify = require('arrify');
+var is = require('is');
+var nodeutil = require('util');
+
+/**
+ * @type {module:common/serviceObject}
+ * @private
+ */
+var ServiceObject = require('../common/service-object.js');
 
 /*! Developer Documentation
  *
- * @param {module:pubsub} pubsub - PubSub Object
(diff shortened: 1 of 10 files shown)


callmehiphop:
Instead of using a multiple inheritance lib, what are your feelings on just having `ServiceObject` extend `EventEmitter`? I realize that this will add event emitter methods to all service classes, however they'll be undocumented and won't really bring any overhead.

stephenplusplus:
That sounds a little crazy. Why would we do that?

callmehiphop:
Well if you just create a prototype chain, then you can have something inherit for multiple "classes". And it seems to me that the only reason we're even including a multiple inheritance lib is because on certain objects we want to use `EventEmitter` but we're already extending a `ServiceObject`.

stephenplusplus:
That is why we're using it. But that's the only place we need it, so I
don't think it's more logical to make all classes event emitters.

callmehiphop:
Well, that's not true - https://github.com/GoogleCloudPlatform/gcloud-node/pull/939 is also using it to extend both an event emitter and service object.

stephenplusplus:
The point is not all. I don't think we want all of our classes to be event
emitters, just the ones that need it.

callmehiphop:
Fair enough!

PR #945, opened 2015-11-14
Implement Service & Service Object for Search

stephenplusplus:
For #904

This hooks Search up with Service & ServiceObject.

### DIFF
--- a/lib/search/document.js
+++ b/lib/search/document.js
@@ -21,12 +21,19 @@
 'use strict';
 
 var is = require('is');
+var nodeutil = require('util');
 
 /**
  * @type {module:search/field}
  * @private
  */
 var Field = require('./field.js');
+
+/**
+ * @type {module:common/serviceObject}
+ * @private
+ */
+var ServiceObject = require('../common/service-object.js');
 
 /**
  * @type {module:common/util}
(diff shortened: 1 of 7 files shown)


lib/search/document.js
```diff
@@ -82,21 +135,17 @@ Document.prototype.addField = function(name) {
...
+ *   if (!err) {
+ *     // The document was created successfully.
+ *   }
+ * });
  */
-Document.prototype.delete = function(callback) {
-  this.makeReq_('DELETE', '', null, null, function(err, resp) {
-    (callback || util.noop)(err, resp);
-  });
+Document.prototype.create = function(callback) {
```
callmehiphop: I thought you could supply the ServiceObject constructor with a `create` method?

stephenplusplus: Actually that's just a list of methods to implement. It might not be a bad idea to accept functions there as well, though I wouldn't prefer the distance/variable placement of the function definitions.

stephenplusplus: And as a "okay, fine, but who cares?" reason, defining it in the prototype allows it the same perf benefits as a prototype method vs a method you define every time inline in a constructor.

callmehiphop: I might have not worded that correctly, but what I'm getting at is that we seem to be providing boolean values for `create` [elsewhere](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/dns/zone.js#L92), so I'm curious what we are not doing that here?

stephenplusplus: In this case, we don't want the default `create` method. When called in the context of a Document, `create` doesn't work with the traditional signature of:

```js
Parent.prototype.createChild = function(name, options, callback) {}
```

Instead, it's:

```js
Parent.prototype.createChild = function(customObject, callback) {}
```

Basically, it's just a one-off that doesn't fit the conventions of the rest of the API.

callmehiphop: Got it! Thanks :)

PR #946, opened 2015-11-14
Implement Service & Service Object for Storage

stephenplusplus:
For #904

This hooks Storage up with Service & ServiceObject.

### DIFF
--- a/lib/storage/acl.js
+++ b/lib/storage/acl.js
@@ -20,6 +20,7 @@
 
 'use strict';
 
+var arrify = require('arrify');
 var is = require('is');
 var nodeutil = require('util');
 
@@ -66,8 +67,8 @@ var nodeutil = require('util');
 function Acl(options) {
   AclRoleAccessorMethods.call(this);
 
-  this.makeReq = options.makeReq;
   this.pathPrefix = options.pathPrefix;
+  this.request_ = options.request;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


lib/storage/acl.js
```diff
@@ -232,28 +233,30 @@ nodeutil.inherits(Acl, AclRoleAccessorMethods);
...
-    role: options.role.toUpperCase()
-  };
-
-  var query = null;
+  var self = this;
 
+  var query = {};
   if (options.generation) {
     query = {
       generation: options.generation
```
callmehiphop: kind of a nitpick here, if we're now setting query as an empty object, wouldn't it be better to just assign the generation property rather than completely blow away the previous object?

stephenplusplus: Yeah!

stephenplusplus: Fixed.

lib/storage/acl.js
```diff
@@ -287,16 +289,20 @@ Acl.prototype.add = function(options, callback) {
  * }, function(err, apiResponse) {});
  */
 Acl.prototype.delete = function(options, callback) {
-  var path = '/' + encodeURIComponent(options.entity);
-  var query = null;
-
+  var query = {};
   if (options.generation) {
     query = {
       generation: options.generation
```
callmehiphop: Same as https://github.com/GoogleCloudPlatform/gcloud-node/pull/946#discussion_r45104160

stephenplusplus: Okay, now fixed.

callmehiphop:
:+1:

PR #958, opened 2015-11-24
core: allow default method options for service objects

stephenplusplus:
This makes a small, but helpful, change to how methods are inherited from a ServiceObject.

Before:

> I'm a new class that needs to inherit these methods:
> - delete
> - getMetadata
> - setMetadata

After:

> I'm a new class that needs to inherit these methods:
> - delete
> - getMetadata
> - setMetadata, but use `PUT` instead of `PATCH`

I've had to work around the last scenario, using PUT over the default PATCH, and similar things before. The old way to fix this was re-implementing the same method, changing a minor detail.

I will implement this behavior in future PRs. It will help out resource manager and also simplify the `File` class which conditionally sets a `generation` property on the query string of a request.

### DIFF
--- a/lib/common/service-object.js
+++ b/lib/common/service-object.js
@@ -21,6 +21,7 @@
 'use strict';
 
 var exec = require('methmeth');
+var extend = require('extend');
 var is = require('is');
 
 /**
@@ -49,6 +50,9 @@ var util = require('./util.js');
  *     name of a Storage bucket or Pub/Sub topic.
  * @param {object=} config.methods - A map of each method name that should be
  *     inherited.
+ * @param {object} config.methods[].reqOpts - Default request options for this
+ *     particular method. A common use case is when `setMetadata` requires a
+ *     `PUT` method to override the default `PATCH`.
  * @param {object} config.parent - The parent service instance. For example, an
  *     instance of Storage if the object is Bucket.
  */
@@ -61,6 +65,7 @@ function ServiceObject(config) {
   this.parent = config.parent; // Parent class.
   this.id = config.id; // Name or ID (e.g. dataset ID, bucket name, etc.)
   this.createMethod = config.createMethod;
+  this.methods = config.methods || {};
 
   if (config.methods) {
     var allMethodNames = Object.keys(ServiceObject.prototype);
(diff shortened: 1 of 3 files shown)


callmehiphop:
My only complaint with this is that we build out the request config every time a method is used, even though it never changes. Why not do it at instantiation time and/or cache it?

stephenplusplus:
> Why

Just for code cleanliness, as opposed to micro-optimization. If you can think of a way to do it that isn't weird/bloated/out of place, I'm all for it.

callmehiphop:
> If you can think of a way to do it that isn't weird/bloated/out of place

I can't really! I think that is probably because of the `ServiceObject` architecture though. We've created an interesting problem for ourselves where we try and have a base class that acts as a catch all for upstream APIs, but they don't all behave exactly the same.

stephenplusplus:
> I think that is probably because of the ServiceObject architecture though

I think it's just because of JavaScript. Regardless of the container the code is in, caching something adds complexity and bloat. Caches have to be stored somewhere.

> where we try and have a base class that acts as a catch all for upstream APIs, but they don't all behave exactly the same.

That's why it's a base class. We have to generalize the 90% use case, and not leave the 10% with nothing. Configuration is how to do it, but I can't argue that "this is the best code ever." If you see a more clear Service/ServiceObject architecture (one that doesn't allow configuration?), lay out some samples in a new issue.

callmehiphop:
> I think it's just because of JavaScript. Regardless of the container the code is in, caching something adds complexity and bloat. Caches have to be stored somewhere.

I'm not saying caching is the ultimate solution, I also wouldn't blame a language since I think this discussion is directly related to our service/service object architecture.

> That's why it's a base class. We have to generalize the 90% use case, and not leave the 10% with nothing. Configuration is how to do it, but I can't argue that "this is the best code ever." If you see a more clear Service/ServiceObject architecture (one that doesn't allow configuration?), lay out some samples in a new issue.

I understand this, but I'm wondering if we generalized a little too much. I don't really have any great ideas like I said earlier, I'm just a little worried about how much configuration we'll need as our API coverage increases.

stephenplusplus:
I can understand the core concern with not wanting to have over-confugurarion. I hope after this, we are in good standing to really not need any. The point of this class was only to not reproduce code. Now that we're there, we can start to refactor and probably improve that code. But for now, I'm glad we were able to generalize enough to affect almost all of the methods, since that was goal number 1.

PR #959, opened 2015-11-24
storage: use default method options in File class

stephenplusplus:
*\* Depends on #958 **

This removes some re-implemented methods that can, after #958 merges, be re-used from the ServiceObject class.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -104,7 +104,43 @@ var STORAGE_UPLOAD_BASE_URL = 'https://www.googleapis.com/upload/storage/v1/b';
  * var file = myBucket.file('my-file');
  */
 function File(bucket, name, options) {
+  options = options || {};
+
+  this.bucket = bucket;
+  this.storage = bucket.parent;
+
+  Object.defineProperty(this, 'name', {
+    enumerable: true,
+    value: name
+  });
+
+  var requestQueryObject = {};
+
+  if (is.number(options.generation)) {
+    requestQueryObject.generation = options.generation;
+    this.generation = options.generation;
+  }
+
   var methods = {
+    /**
+     * Delete the file.
+     *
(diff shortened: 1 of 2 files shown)


lib/storage/file.js
```diff
@@ -104,8 +104,36 @@ var STORAGE_UPLOAD_BASE_URL = 'https://www.googleapis.com/upload/storage/v1/b';
  * var file = myBucket.file('my-file');
  */
 function File(bucket, name, options) {
+  options = options || {};
+
+  var requestQueryObject = {};
+  var generation = parseInt(options.generation, 10);
+
+  if (generation) {
```
callmehiphop: Should we use `isNaN` for clarity here?

callmehiphop:
Is the don't merge label here because of the previous PR dependency? Or is there still stuff to be done?

lib/storage/file.js
stephenplusplus: Relevant: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/isNaN#Description

I'll switch it to `is.number(options.generation)`

stephenplusplus:
It was indeed! Switched to a strict number check. PTAL

lib/storage/file.js
```diff
@@ -104,8 +104,44 @@ var STORAGE_UPLOAD_BASE_URL = 'https://www.googleapis.com/upload/storage/v1/b';
...
+  this.storage = bucket.parent;
+
+  Object.defineProperty(this, 'name', {
+    enumerable: true,
+    value: name
+  });
+
+  var requestQueryObject = {};
+
+  if (is.number(options.generation)) {
```
callmehiphop: This works? I was looking at [`is.number`](https://github.com/enricomarino/is/blob/master/index.js#L412) and I thought `NaN` would be considered a number by that logic.

PR #960, opened 2015-11-24
resource: use default method options & run system tests conditionally

### LINKED ISSUE #947: Conditionally run resource manager system tests
As it is today, certain resource manager API calls require you to be authenticated as yourself and not through a service account, because of this we're not able to properly test certain endpoints (e.g. `setMetadata`). If possible it would be nice to have an environment variable available to conditionally run said tests.

stephenplusplus:
*\* Depends on #958 **

Fixes #947 

This removes some re-implemented methods that can, after #958 merges, be re-used from the ServiceObject class.

### DIFF
--- a/system-test/resource.js
+++ b/system-test/resource.js
@@ -17,6 +17,7 @@
 'use strict';
 
 var assert = require('assert');
+var googleAuth = require('google-auto-auth');
 
 var env = require('./env.js');
 var Resource = require('../lib/resource/index.js');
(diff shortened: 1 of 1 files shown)


system-test/resource.js
```diff
@@ -100,15 +58,73 @@ describe('Resource', function() {
...
+      }
+
+      project.delete(done);
+    });
+
+    it('should have created the project', function() {
+      assert.strictEqual(project.metadata.projectId, PROJECT_ID);
+    });
+
+    it('should set metadata', function(done) {
```
callmehiphop: This is consistently failing for me locally with an `Uncaught ApiError: Not Found`

stephenplusplus: Is it based off the latest master?

callmehiphop: Ah, sure wasn't, my bad.

PR #961, opened 2015-11-25
storage: generations are returned as strings

stephenplusplus:
Apparently, the representation of a `long` is returned as a string in the API response from a `get` of an Object in the Storage API. This reverts the behavior where we force a file generation to be a number and tries to parse a string instead.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -114,11 +114,12 @@ function File(bucket, name, options) {
     value: name
   });
 
+  var generation = parseInt(options.generation, 10);
   var requestQueryObject = {};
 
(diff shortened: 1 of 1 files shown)


lib/storage/file.js
```diff
@@ -372,7 +373,7 @@ File.prototype.copy = function(destination, callback) {
   }
 
   var query = {};
-  if (is.number(this.generation)) {
+  if (is.defined(this.generation)) {
```
stephenplusplus: `is.defined` because "is this not a not a number that is not `NaN`" is, um, yeah.

PR #965, opened 2015-11-25
storage: system tests: manage created objects

stephenplusplus:
The main change here is forcing all created files and buckets to share a prefix. That way, they can be deleted all at once after all tests have completed.

### DIFF
--- a/system-test/storage.js
+++ b/system-test/storage.js
@@ -18,9 +18,7 @@
 
 var assert = require('assert');
 var async = require('async');
-var Bucket = require('../lib/storage/bucket.js');
 var crypto = require('crypto');
-var File = require('../lib/storage/file.js');
 var fs = require('fs');
 var prop = require('propprop');
 var request = require('request');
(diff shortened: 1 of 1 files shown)


system-test/storage.js
```diff
@@ -28,63 +26,57 @@ var through = require('through2');
...
+    gzip: {
+      path: 'system-test/data/long-html-file.html.gz'
+    }
+  };
 
   before(function(done) {
     bucket.create(done);
   });
 
   after(function(done) {
```
callmehiphop: Is this working for you locally? Mine keeps failing saying the bucket I'm trying to delete is not empty. Logic looks sound though, so maybe this is an upstream error?

stephenplusplus: I had to run this in the terminal to clean up my old buckets. For some reason, they were un-deletable, unless I went through gsutil:

```sh
$ gsutil -m rm -r gs://gcloud-test-bucket-*
```

callmehiphop: This worked for me, what happens if Travis runs into a similar issue?

stephenplusplus: We'll have to have @jgeewax run it for us.

stephenplusplus: I really can't speak to what the cause of this issue is (see https://github.com/GoogleCloudPlatform/gcloud-node/issues/968#issuecomment-159722959).

callmehiphop: @jgeewax we've run into an issue with Travis, any chance you could fix that for us? Or just add us to the project so we don't have to ask you to fix it for us :stuck_out_tongue_winking_eye:

jgeewax: Sent - sorry about the delay.

PR #967, opened 2015-11-25
storage: implement channels

### LINKED ISSUE #403: Storage: implement channels
Channels: "A channel represents the means by which a notification message is sent to an application watching a bucket." https://cloud.google.com/storage/docs/json_api/v1/channels

Objects -> watchAll: "Watch for changes on all objects in a bucket" https://cloud.google.com/storage/docs/json_api/v1/objects/watchAll

Hopefully I dropped this in the right milestone.

stephenplusplus:
Fixes #403

### DIFF
--- a/system-test/storage.js
+++ b/system-test/storage.js
@@ -696,6 +696,25 @@ describe('storage', function() {
     });
   });
 
+  describe('channels', function() {
+    it('should create a channel', function(done) {
+      var config = {
+        address: 'https://yahoo.com'
+      };
+
+      bucket.createChannel('new-channel', config, function(err) {
+        // Actually creating a channel is pretty complicated. This will at least
+        // let us know we hit the right endpoint and it received "yahoo.com".
+        assert.strictEqual(
+          err.message,
+          'Unauthorized WebHook callback channel: ' + config.address
+        );
+
+        done();
+      });
+    });
+  });
+
   describe('combine files', function() {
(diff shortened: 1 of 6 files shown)


system-test/storage.js
```diff
@@ -696,6 +696,25 @@ describe('storage', function() {
     });
   });
 
+  describe('channels', function() {
```
callmehiphop: Should we have a test for `stop` as well?

callmehiphop: Or is that not happening because of the difficulty involved in creating a channel..?

stephenplusplus: That would be good, but I can't create a channel without an HTTPS server. Look into how many things are required, I'm not sure I'll be able to do that for the tests :\

callmehiphop: Ah ok, can we use gcloud-node.com?

stephenplusplus: Maybe someday, but I don't think that's meant to turn into a site. Plus, our tests only passing if some dummy server is up is weird. I think we just have to trust the unit tests for these methods :\

callmehiphop: Sad times, what if we just created a local server?

stephenplusplus: Well, for the API call to work, it would have to be able to ping the created server. Running within a localhost environment, that would be problematic. Slash face.

callmehiphop: Fair enough!

PR #979, opened 2015-11-30
core: introduce interceptors

### LINKED ISSUE #916: Allow users to configure forwarded headers?
The Cloud Trace team wants to make it easy for users to send trace IDs along with various API calls, effectively just adding a special header (`X-Cloud-Trace-Context`) with an opaque ID.

The issue is that this ID will vary across requests, which presents a bit of an issue in that whole concurrency thing...

For example, here's how I'd want to do this in my app: https://gist.github.com/jgeewax/3b351d3522fdf061bdea

Note that I have zero idea if this is at all reasonable given the isolation model for Node. (I think, for example, that this won't work for Express, and that we should be doing something with `req.locals`....? see http://stackoverflow.com/questions/33310324/using-express-middleware-for-per-request-module-configuration)

Anyway -- we need to find a way to allow users to...

1. Set custom headers to be forwarded along with each API request to Google
1. Configure those headers in one place (ie, not as arguments to each method as people don't want to change *all* their code for this)
1. Make sure that it's obvious how to do this *the right way* (ie, avoid giving people a gun and aiming it at their feet -- which `setTraceContext` in my example might do...)

stephenplusplus:
Fixes #916 
#### To Dos
- [x] Support gcloud-level interceptors.

Doc preview:

screen shot 2015-11-30 at 3 19 00 pm

_... further down ..._

screen shot 2015-11-30 at 8 19 53 pm

### DIFF
--- a/lib/common/service.js
+++ b/lib/common/service.js
@@ -20,6 +20,8 @@
 
 'use strict';
 
+var arrify = require('arrify');
+
 /**
  * @type {module:common/util}
  * @private
@@ -48,6 +50,8 @@ function Service(config, options) {
   this.authClient = this.makeAuthenticatedRequest.authClient;
   this.baseUrl = config.baseUrl;
   this.getCredentials = this.makeAuthenticatedRequest.getCredentials;
+  this.globalInterceptors = arrify(options.interceptors_);
+  this.interceptors = [];
   this.projectId = options.projectId;
   this.projectIdRequired = config.projectIdRequired !== false;
 }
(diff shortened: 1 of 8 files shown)


callmehiphop:
Do interceptors cascade from services to service objects? I kind of expect that they would because other properties (`projectId`) do.

stephenplusplus:
Yep, all service object requests go through the parent service. So if you set an interceptor on a service (Storage), all requests made from child service objects (Bucket, File) will have be affected.

Maybe we should support one level up as well? Interceptors on the "require('gcloud')" level, so all Services are affected?

callmehiphop:
> Maybe we should support one level up as well? Interceptors on the "require('gcloud')" level, so all Services are affected?

Yes! I love that

stephenplusplus:
@callmehiphop updated the heck out of this one. There was some rethinking involved. PTAL.

lib/common/service.js
```diff
@@ -48,6 +50,8 @@ function Service(config, options) {
   this.authClient = this.makeAuthenticatedRequest.authClient;
   this.baseUrl = config.baseUrl;
   this.getCredentials = this.makeAuthenticatedRequest.getCredentials;
+  this.globalInterceptors = arrify(options.interceptors_);
```
callmehiphop: Out of curiosity, what's the benefit of storing global and service level interceptors separately?

stephenplusplus: We need to be able to reference any later-assigned global interceptors, so a pointer to the original global interceptors object can't end up combined with this instance's ones. Likewise, we can't have something assigned from a Service instance's get muddied up with the global interceptors.

Did I say interceptors enough?

callmehiphop: Not by my count!

PR #649, opened 2015-06-09
Subscription#new add option maxInProgress

leibale:
Subscription#new add option maxInProgress (to make sure the application working on {Number} messages in parallel)

PR #981, opened 2015-11-30
pubsub: allow setting a pull HTTP timeout

### LINKED ISSUE #980: Pub/Sub set timeout of pull()
It would be nice to be able to set the timeout of a pull(). Right now it can either ```returnImmediately=true``` or it will long poll for ~90sec. Is it possible to configure the pull to timeout after a specific amount of time? See http://stackoverflow.com/questions/33976227/google-pub-sub-how-to-set-read-timeout-on-pull

stephenplusplus:
Fixes #980 

> What I would like to do is be able to send a timeout param in the pull request to tell subpub to send back a response after a 30000ms if no messages are published during that time.

We can't control the server's timeout setting through the API, but I tried to effectively recreate it. The user can cancel an outstanding pull request by breaking the HTTP connection. So this PR lets them set a timeout on the HTTP request. When that amount of time is reached, the request will be broken.

If we don't get a response from the server in that amount of time, we handle the "ETIMEDOUT" error by assuming because we haven't heard anything, there aren't any messages to return. Thus, we return an empty array of messages to the user.

### DIFF
--- a/lib/pubsub/index.js
+++ b/lib/pubsub/index.js
@@ -483,11 +483,16 @@ PubSub.prototype.subscribe = function(topic, subName, options, callback) {
  * @param {string} name - The name of the subscription.
  * @param {object=} options - Configuration object.
  * @param {boolean} options.autoAck - Automatically acknowledge the message once
- *     it's pulled.
+ *     it's pulled. (default: false)
  * @param {string} options.encoding - When pulling for messages, this type is
  *     used when converting a message's data to a string. (default: 'utf-8')
  * @param {number} options.interval - Interval in milliseconds to check for new
- *     messages.
+ *     messages. (default: 10)
+ * @param {number} options.maxInProgress - Maximum messages to consume
+ *     simultaneously.
+ * @param {number} options.timeout - Set a maximum amount of time in
+ *     milliseconds on an HTTP request to pull new messages to wait for a
+ *     response before the connection is broken.
  * @return {module:pubsub/subscription}
  *
  * @example
--- a/lib/pubsub/subscription.js
+++ b/lib/pubsub/subscription.js
@@ -48,8 +48,8 @@ var util = require('../common/util.js');
  *
  * @param {module:pubsub} pubsub - PubSub object.
  * @param {object} options - Configuration object.
- * @param {boolean} options.autoAck - Automatically acknowledge the message
- *     once it's pulled. (default: false)
+ * @param {boolean} options.autoAck - Automatically acknowledge the message once
+ *     it's pulled. (default: false)
  * @param {string} options.encoding - When pulling for messages, this type is
  *     used when converting a message's data to a string. (default: 'utf-8')
  * @param {number} options.interval - Interval in milliseconds to check for new
@@ -57,6 +57,9 @@ var util = require('../common/util.js');
  * @param {string} options.name - Name of the subscription.
  * @param {number} options.maxInProgress - Maximum messages to consume
  *     simultaneously.
+ * @param {number} options.timeout - Set a maximum amount of time in
+ *     milliseconds on an HTTP request to pull new messages to wait for a
+ *     response before the connection is broken. (default: 90000)
  */
 /**
  * A Subscription object will give you access to your Google Cloud Pub/Sub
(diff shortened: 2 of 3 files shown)


lib/pubsub/subscription.js
```diff
@@ -267,12 +270,13 @@ function Subscription(pubsub, options) {
...
   this.autoAck = is.boolean(options.autoAck) ? options.autoAck : false;
   this.closed = true;
-  this.interval = is.number(options.interval) ? options.interval : 10;
   this.inProgressAckIds = {};
+  this.interval = is.number(options.interval) ? options.interval : 10;
   this.maxInProgress =
     is.number(options.maxInProgress) ? options.maxInProgress : Infinity;
   this.messageListeners = 0;
   this.paused = false;
+  this.timeout = options.timeout;
```
callmehiphop: Do we need to provide a default? Or if it's null/undefined a timeout won't occur?

stephenplusplus: I looked at the request code, and it looks like it'll skip right by it: https://github.com/request/request/blob/c4d3c9c3058bf1b5a85039c6e428ad6de99bfe1b/request.js#L752

stephenplusplus:
@callmehiphop @mattraby -- I updated the PR. See the opening post for details. Let me know what you think!

stephenplusplus:
@tmatsuo read over the issue #980 and the originating [StackOverflow issue](http://stackoverflow.com/questions/33976227/google-pub-sub-how-to-set-read-timeout-on-pull) that inspired this PR -- is this a good idea or not so much? Thanks!

callmehiphop:
With the addition of #979, is this still necessary?

stephenplusplus:
If we consider this feature useful (still TBD -- need help with that decision), then I think we're just making it easy with this. It would be a bit awkward for the user to do it themselves.

callmehiphop:
It sounds like we can't resolve the original issue via the client, however imo this is still a nice feature and it's one that could potentially benefit all of the other services.

stephenplusplus:
Oh, gotcha. Supporting easily configurable timeouts is a good idea and request interceptors would be the right way to do that on a global basis. But I still consider that an advanced need and a bit of a learning curve to understand how to do it with our library. In this case, *only* the HTTP ':pull' request should be affected, so I could see both making sense.

Related: https://github.com/GoogleCloudPlatform/gcloud-node/issues/446#issuecomment-145724558

stephenplusplus:
@jgeewax @tmatsuo what do you think about this PR?

lib/pubsub/index.js
```diff
@@ -483,11 +483,16 @@ PubSub.prototype.subscribe = function(topic, subName, options, callback) {
...
  * @param {object=} options - Configuration object.
  * @param {boolean} options.autoAck - Automatically acknowledge the message once
- *     it's pulled.
+ *     it's pulled. (default: false)
  * @param {string} options.encoding - When pulling for messages, this type is
  *     used when converting a message's data to a string. (default: 'utf-8')
  * @param {number} options.interval - Interval in milliseconds to check for new
- *     messages.
+ *     messages. (default: 10)
+ * @param {number} options.maxInProgress - Maximum messages to consume
```
jgeewax: is this the right name? isnt this really more like page-size ?

stephenplusplus: `maxInProgress` is a client-side feature implemented in #649/#650.

jgeewax: Yea , I mean, does it convey the right meaning... ? Doesn't this actually "mean" a page size of sorts?

stephenplusplus: In case you missed it, the last word might be the most important, which is on the next line. The full sentence is "Maximum messages to consume simultaneously."

jgeewax: OK then this LGTM. Sorry :)

stephenplusplus: Np :)

lib/pubsub/subscription.js
```diff
@@ -269,13 +272,16 @@ function Subscription(pubsub, options) {
...
   this.inProgressAckIds = {};
+  this.interval = is.number(options.interval) ? options.interval : 10;
   this.maxInProgress =
     is.number(options.maxInProgress) ? options.maxInProgress : Infinity;
   this.messageListeners = 0;
   this.paused = false;
-  this.encoding = options.encoding || 'utf-8';
+  // The default timeout used in this library is 60s, but a pull request times
+  // out around 90 seconds.
+  this.timeout = is.number(options.timeout) ? options.timeout : 90000;
```
tmatsuo: Probably it's safer to make it 88000 instead of 90000. Also Avoid magic number?

tmatsuo:
@stephenplusplus The PR makes sense to me.

lib/pubsub/subscription.js
stephenplusplus: Good idea for some padding on the timeout. I just pushed an update, but went on the side of allowing more time (92s) rather than less (88s). I figured its best to give the API more time to fulfill the request before terminating. Here's the code: https://github.com/GoogleCloudPlatform/gcloud-node/commit/a78b77d6e377e0737f549d60259669a86cbab135#diff-9c43bce0f7f5594b8154b1b8a747af5aR286

LMK if you think 88 is still the right way to go.

stephenplusplus:
@tmatsuo @jgeewax Thanks for the review!

tmatsuo:
LGTM

PR #983, opened 2015-12-01
datastore: add Key helpers

### LINKED ISSUE #574: Provide helper methods on entity.Key for Kind, parent, name, and ID
(In reference to https://github.com/GoogleCloudPlatform/gcloud-node/blob/412598cad200198caaac5dbe83700fd1665b2102/lib/datastore/entity.js#L54)

It'd be really nice if we had some helper methods on the Key class, so that you could do the following...

```javascript
var dataset = gcloud.datastore.dataset({...});
```

### Auto-generated ID example:

```javascript
entity = {
  key: dataset.key('Person').
  data: {name: 'Jim'}
};

dataset.save(entity, function(err, entity) {
  var key = entity.key;
  console.log('kind is', key.kind); // 'Person'
  console.log('id is', key.id);  // 1234567 (auto-generated ID in this case)
  console.log('name is', key.name); // undefined or null
  console.log('parent is', key.parent); // undefined or null
});
```

### Specific name and parent example

```javascript
entity = {
  key: dataset.key(['Person', 'mom', 'Person', 'jim']).
  data: {name: 'Jim'}
};

dataset.save(entity, function(err, entity) {
  var key = entity.key;
  console.log('kind is', key.kind); // 'Person'
  console.log('id is', key.id);  // undefined or null
  console.log('name is', key.name); // 'jim'
  console.log('parent is', key.parent); // {key object}
  console.log('parent kind is', key.parent.kind); // 'Person'
  console.log('parent id is', key.parent.id); // undefined or null
  console.log('parent name is', key.parent.name); // 'mom'
  console.log('parent parent is', key.parent.parent); // undefined or null
});
```

Is this a reasonable request?

stephenplusplus:
Fixes #574

Alas, key helpers!

### DIFF
--- a/lib/datastore/dataset.js
+++ b/lib/datastore/dataset.js
@@ -168,23 +168,60 @@ Dataset.prototype.determineApiEndpoint_ = function(customApiEndpoint) {
  * @return {Key} A newly created Key from the options given.
  *
  * @example
- * var key;
+ * //-
+ * // Create an incomplete key with a kind value of `Company`.
+ * //-
+ * var key = dataset.key('Company');
  *
- * // Create an incomplete key from the dataset namespace, kind='Company'
- * key = dataset.key('Company');
+ * //-
+ * // Create a complete key with a kind value of `Company` and id `123`.
+ * //-
+ * var key = dataset.key(['Company', 123]);
  *
- * // A complete key from the dataset namespace, kind='Company', id=123
- * key = dataset.key(['Company', 123]);
+ * //-
+ * // Create a complete key with a kind value of `Company` and name `Google`.
+ * // Note: `id` is used for numeric identifiers and `name` is used otherwise.
+ * //-
+ * var key = dataset.key(['Company', 'Google']);
  *
- * // A complete key from the dataset namespace, kind='Company', name='Google'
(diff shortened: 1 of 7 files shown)


lib/datastore/dataset.js
```diff
@@ -168,23 +168,60 @@ Dataset.prototype.determineApiEndpoint_ = function(customApiEndpoint) {
...
  *   path: ['Company', 123]
  * });
+ *
+ * //-
+ * // All keys have helper properties that you can access or overwrite
+ * // directly.
+ * //-
+ * var key = dataset.key(['Company', 123]);
+ *
+ * // key.path: ['Company', 123]
```
callmehiphop: Instead of trying to document all of these inline, would it make more sense to make a Key section within the docs, then we can document them JSDoc style?

stephenplusplus: I thought of that, but since we don't have methods and just properties, they would all end up described in the constructor, just like this. It would end up being an extra click to go from `dataset.key` to the `Key` page for what would end up being the same thing. Does that change your mind or do you still think it's a good idea?

callmehiphop: I'm cool with it, I think it's more appropriately placed here, but maybe feels inconsistent with the rest of our documentation.

PR #988, opened 2015-12-03
Add pubsub encoding option

roberttod:
As discussed in https://github.com/GoogleCloudPlatform/gcloud-node/issues/987
- UTF-8 conversion of binary data is not reversable
- Use UTF-8 by default in order maintain current functionality

### DIFF
--- a/lib/pubsub/subscription.js
+++ b/lib/pubsub/subscription.js
@@ -273,6 +273,7 @@ function Subscription(pubsub, options) {
     is.number(options.maxInProgress) ? options.maxInProgress : Infinity;
   this.messageListeners = 0;
   this.paused = false;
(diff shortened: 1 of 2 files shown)


roberttod:
I signed it!

stephenplusplus:
LGTM. Thanks very much! :+1: I'm hoping to have a release out soon. I'll give you a ping when it's out.

roberttod:
Thanks for merging!

tseaver:
@jgeewax `gcloud-python` base64-encodes the message payload, per https://cloud.google.com/pubsub/publisher#publish.

blowmage:
We will investigate if this is needed in gcloud-ruby. Ruby string have encoding attributes, so a non-UTF-8 string shouldn't be converted to UTF-8 before/after encoding to base64. But we haven't tested any binary so we will make sure it works.

PR #989, opened 2015-12-03
introduce Prediction API

### LINKED ISSUE #976: Add support for the Prediction API
https://cloud.google.com/prediction/docs

API Docs: https://cloud.google.com/prediction/docs/reference/v1.6/

stephenplusplus:
Fixes #976 

Hello, world for Prediction API.
#### To Dos
- [x] Enable API for the Travis account
- [x] Add analyze method
- [x] Add to side navigation in docs
- [x] Docs
  - [x] Show how to use a file when creating a model
  - [x] Document/alias `modelType` to `type` when creating a model to distinguish between classification and regression
- [x] Tests
  - [x] System
  - [x] Unit
    - [x] common/util
    - [x] prediction/index
    - [x] prediction/model
#### Preview

``` js

// Authenticating on a per-API-basis. You don't need to do this if you auth on a
// global basis (see Authentication section above).

var prediction = gcloud.prediction({
  projectId: 'my-project',
});

// Get all of the trained models in your project.
prediction.getModels(function(err, models) {
  if (!err) {
    // `models` is an array of Model objects.
  }
});

// Reference an existing trained model.
var model = prediction.model('my-existing-model');

// Train a model.
model.train('english', 'Hello from your friends at Google!', function(err) {});

// Query a model.
model.query('Hello', function(err, results) {
  if (!err) {
    // results.winner == 'english'
    // results.scores == [
    //   {
    //     label: 'english',
    //     score: 1
    //   },
    //   {
    //     label: 'spanish',
    //     score: 0
    //   }
    // ]
  }
});
```


stephenplusplus:
@jgeewax

jgeewax:
Basic usage here LGTM. How about bulk-loading via the CSV trainedmodels.update method ? (https://cloud.google.com/prediction/docs/reference/v1.6/trainedmodels/update)

I'll ask someone from the Prediction team to take a look.

stephenplusplus:
I could be misunderstanding the API, but I think the only place it accepts a CSV file is when you create the model and point it to a GCS file (which is supported here):

```js
prediction.createModel('new-model-name', {
  data: gcs.file('input-data.csv')
}, function() {});
```

For the update API, we have three ways of hitting that endpoint:

```js
model.setMetadata({
  output: 'label',
  csvInstance: ['data']
}, function() {});
```
```js
model.train('label', 'data', function() {});
```
```js
var writeStream = model.createWriteStream('label');
writeStream.write('data');
// ... any number of writes
writeStream.end();
```

*Added todos for better docs that show how to do these things*

stephenplusplus:
@callmehiphop feel free to pick this one apart now :)

callmehiphop:
Still reading through a lot of the documentation, but are we not providing support for hostedmodels?

lib/prediction/model.js
```diff
@@ -0,0 +1,370 @@
...
+ *     //   {
+ *     //     label: 'english',
+ *     //     score: 1
+ *     //   },
+ *     //   ...other results and their scores are listed here.
+ *     // ]
+ *   }
+ * });
+ */
+Model.prototype.query = function(input, callback) {
```
callmehiphop: Are we naming this method `query` to better align with the rest of our APIs? I kind of think this would be better named `predict` after it's upstream counterpart.

lib/prediction/index.js
```diff
@@ -0,0 +1,270 @@
...
+  if (is.fn(options)) {
+    callback = options;
+    options = {};
+  }
+
+  var body = extend({}, options, {
+    id: id
+  });
+
+  var file = body.data;
```
callmehiphop: should we be deleting `body.data`? I didn't see it in the upstream request body docs.

lib/prediction/index.js
```diff
@@ -0,0 +1,270 @@
...
+ *
+ * @resource [Trainedmodels: insert API Documentation]{@link https://cloud.google.com/prediction/docs/reference/v1.6/trainedmodels/insert}
+ *
+ * @throws {error} If a model ID is not provided.
+ *
+ * @param {string} id - Unique name for the model. E.g. "my-model"
+ * @param {object=} options - Configuration object.
+ * @param {module:storage/file} options.data - A CSV file to use as training
+ *     data.
+ * @param {string} options.type - The type of model, `classification`
```
callmehiphop: Seems like we're not documenting a few options, are we trying to avoid stale docs? Or are we only documenting common use case options..?

lib/prediction/model.js
stephenplusplus: Just a case where I went with the logical action name. It felt strange using the action "predict", since the user isn't predicting anything, they're asking for a prediction. 

Throughout the docs, they refer to it as "querying": https://cloud.google.com/prediction/docs/hello_world

lib/prediction/index.js
stephenplusplus: Yes! Good catch.

lib/prediction/index.js
stephenplusplus: Common use case + aliases for things that aren't as simply named upstream. I should change the description for `options` though to link directly to the upstream docs.

lib/prediction/index.js
```diff
@@ -0,0 +1,270 @@
...
+ * @param {number} query.maxResults - Maximum number of results to return.
+ * @param {string} query.pageToken - Page token.
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An API error.
+ * @param {?module:prediction/model[]} callback.models - An array of
+ *     {module:prediction/model} objects.
+ * @param {object} callback.apiResponse - Raw API response.
+ *
+ * @example
+ * prediction.getModels(function(err, models, apiResponse) {});
```
callmehiphop: I think we should include examples of using the streaming API!

callmehiphop: Also it looks like we're missing a `@param` for `nextQuery`

stephenplusplus:
Changes made!

> are we not providing support for hostedmodels?

I intended to originally. But when I [read more](https://cloud.google.com/prediction/docs/gallery) about them, I found there are only 3, and they have a pretty serious limit:

> Hosted models have a usage limit of 100 predictions/day/user across **_all_ demonstration models**. Developers can set higher paid usage limits for models **they create**.

So, it sounded more like a "getting to know Prediction API" learning tool to me at this point. If and when it handles more queries, and the gallery grows larger, supporting this would be great.

@jgeewax any thoughts on supported hostedmodels?

stephenplusplus:
@jgeewax should we wait for a review from someone from the Prediction team?

:watch: *This PR will self-merge in 24 hours*

PR #992, opened 2015-12-04
pubsub: prevent maxInProgress from being sent to the API

### LINKED ISSUE #985: cant subscribe with maxInProgress option
it's return ApiError to callback
```javascript
Invalid JSON payload received. Unknown name "max_in_progress": Cannot find field.
```

stephenplusplus:
Fixes #985

### DIFF
--- a/lib/pubsub/index.js
+++ b/lib/pubsub/index.js
@@ -389,6 +389,8 @@ PubSub.prototype.getTopics = function(query, callback) {
  *     it's pulled. (default: false)
  * @param {number} options.interval - Interval in milliseconds to check for new
  *     messages. (default: 10)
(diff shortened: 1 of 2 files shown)


stephenplusplus:
@leibale can you give this a shot?

```sh
$ npm install --save stephenplusplus/gcloud-node#spp--pubsub-985
```

leibale:
you fixed the #985 bug, but breaking something else (it might not be you, but it does not matter).. :(
at line `471` in `pubsub/index.js` you should use `resp.name || subName` (if the status is 409 and `reuseExisting` is true i got `The name of a subscription is required.` error).

stephenplusplus:
That's definitely a new bug, thanks for catching that. Opening a new issue now. (PR welcome as always :))

PR #1004, opened 2015-12-07
tests: bail after first test failure

### LINKED ISSUE #1002: system tests: bail after first failure?
Mocha has an option `-b, --bail | bail after first test failure`. Travis runs can go on forever, using real resources, only to result in a failed build if a single test fails. Should we use bail here?

stephenplusplus:
Fixes #1002

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
     "docs": "./scripts/docs.sh",
     "lint": "jshint lib/ system-test/ test/ && jscs lib/ system-test/ test/",
     "test": "npm run docs && mocha test/*/*.js test/index.js test/docs.js",
-    "system-test": "mocha system-test/* --timeout 60s",
+    "system-test": "mocha system-test/* --timeout 60s --bail",
     "cover": "istanbul cover -x 'system-test/*' _mocha -- --timeout 60s test/*/*.js test/docs.js system-test/*",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -128,9 +128,9 @@
     "docs": "./scripts/docs.sh",
     "lint": "jshint lib/ system-test/ test/ && jscs lib/ system-test/ test/",
     "test": "npm run docs && mocha test/*/*.js test/index.js test/docs.js",
-    "system-test": "mocha system-test/* --timeout 60s",
-    "cover": "istanbul cover -x 'system-test/*' _mocha -- --timeout 60s test/*/*.js test/docs.js system-test/*",
-    "coveralls": "istanbul cover -x 'system-test/*' _mocha --report lcovonly -- --timeout 60s test/*/*.js test/docs.js system-test/* -R spec && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage"
+    "system-test": "mocha system-test/* --timeout 60s --bail",
+    "cover": "istanbul cover -x 'system-test/*' _mocha -- --timeout 60s --bail test/*/*.js test/docs.js system-test/*",
```
callmehiphop: Do we want to bail for coverage reports? Even if a test fails in coverage, I believe you're still able to determine whether or not a certain piece of code is covered.

stephenplusplus: Makes sense. Removed.

PR #1010, opened 2015-12-09
core: address early return rule per latest jscs release

callmehiphop:
jscs released a [new version](https://github.com/jscs-dev/node-jscs/releases/tag/v2.7.0) that implemented a new rule which [broke our build](https://travis-ci.org/GoogleCloudPlatform/gcloud-node/jobs/95700307)


stephenplusplus:
You're not allowed to release again.

callmehiphop:
._.

PR #1026, opened 2015-12-17
tests: cleaned up unused test code

callmehiphop:
We ran across an issue where the [documentation tests were breaking](https://travis-ci.org/GoogleCloudPlatform/gcloud-node/jobs/97504347#L2541). Not 100% on what the cause is, but it appears that some unneeded `mockery` code has been causing the issue.


stephenplusplus:
Good catch, sir!

PR #1028, opened 2015-12-18
core: fix unit test stream errors

stephenplusplus:
Our tests have been bombing lately after a transient dependency made a simple perf change: https://github.com/nodejs/readable-stream/commit/670ddf3a27cb0e88f62cbf10b9b0ebf44ce45053

It's loaded through [`duplexify`](http://gitnpm.com/duplexify), but I'm not sure why it's causing our tests to fail. It's definitely related to being combined with [`mockery`](http://gitnpm.com/mockery), but I can't pin down exactly... how.

But, this PR fixes it.

### DIFF
--- a/test/common/util.js
+++ b/test/common/util.js
@@ -17,7 +17,7 @@
 'use strict';
 
 var assert = require('assert');
-var duplexify = require('duplexify');
+var duplexify;
 var extend = require('extend');
 var googleAuth = require('google-auto-auth');
 var mockery = require('mockery');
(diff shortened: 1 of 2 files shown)


callmehiphop:
So just to clarify, if a spec file requires the use of duplexify, we need to wait to require it after mockery has been enabled?

stephenplusplus:
Actually, after the file using it has been required.

callmehiphop:
Got it, thanks! That should be a fun "gotcha" for the future, maybe we should look into a mockery alternative.

PR #1034, opened 2015-12-21
big query jobs can be cancelled

peterwillis:
I have only included one test, I wasn't sure what else to include. In particular I couldn't figure out how to check the complete URL from a test, which I thought would be a good test to check the interceptor built the right url ('project' instead of 'projects'). If the Big Query REST api is updated, the interceptor could be removed.

### DIFF
--- a/lib/bigquery/job.js
+++ b/lib/bigquery/job.js
@@ -110,10 +110,42 @@ function Job(bigQuery, id) {
     methods: methods
   });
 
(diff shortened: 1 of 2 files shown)


stephenplusplus:
Thank you for doing this! I just opened #1035 which adds some more tests and docs. I'll close this one out, but would you mind checking it out and making sure it still looks okay? Thanks!

PR #1035, opened 2015-12-21
big query jobs can be cancelled (cont'd)

### LINKED ISSUE #1027: Not able to implement Big Query cancel job
The cancel job function is not implemented for big query so I went about putting together a pull request. It should be straight forward but it's not because the url structure for all the other job api calls are slightly different...

To cancel a job:
```POST https://www.googleapis.com/bigquery/v2/project/projectId/jobs/jobId/cancel```

This is the correct URL because it works on the documentation site through the try it tester.

All the others have a pluralised 'projects' rather than 'project':
```GET https://www.googleapis.com/bigquery/v2/projects/projectId/jobs/jobId```
```POST https://www.googleapis.com/bigquery/v2/projects/projectId/queries```
```GET https://www.googleapis.com/bigquery/v2/projects/projectId/jobs```
...etc

I don't think this could be neatly worked around for the purposes of this library because that part of the url is baked into the service class. Is this a mistake in the Big Query REST api? If so where should I raise that issue?

stephenplusplus:
Fixes #1027
Evolves #1034

### DIFF
--- a/lib/bigquery/job.js
+++ b/lib/bigquery/job.js
@@ -29,6 +29,12 @@ var nodeutil = require('util');
  */
 var ServiceObject = require('../common/service-object.js');
 
+/**
+ * @type {module:common/util}
+ * @private
+ */
+var util = require('../common/util.js');
+
 /*! Developer Documentation
  *
  * @param {module:bigquery} bigQuery - BigQuery instance.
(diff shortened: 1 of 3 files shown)


peterwillis:
I authored the commits in #1034

peterwillis:
I checked this out and it still looks ok.

stephenplusplus:
Cool, thanks.

PR #1043, opened 2016-01-04
datastore: recognize DATASTORE_DATASET env var

### LINKED ISSUE #1040: Support Datastore default dataset env var
See https://github.com/GoogleCloudPlatform/gcloud-common/issues/24

We currently support `DATASTORE_HOST`, but not `DATASTORE_DATASET`.

stephenplusplus:
Fixes #1040

### DIFF
--- a/lib/datastore/dataset.js
+++ b/lib/datastore/dataset.js
@@ -68,6 +68,9 @@ var SCOPES = [
  * Interact with a dataset from the
  * [Google Cloud Datastore](https://developers.google.com/datastore/).
  *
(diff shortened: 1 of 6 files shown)


callmehiphop:
I didn't narrow down what it is, but regardless of whether I set the env var all the transaction system-tests fail, where they appear ok in master.

callmehiphop:
@stephenplusplus I think you need to change `projectId` to `datasetId` for [Transactions](https://github.com/stephenplusplus/gcloud-node/blob/spp--1040/lib/datastore/transaction.js#L76) as well.

stephenplusplus:
Fixed!

PR #1045, opened 2016-01-04
bigquery: allow duplicate entries

### LINKED ISSUE #1041: bigquery table.insert fails for identical rows
It seems to be failing sometimes when you insert identical rows in parallel. I think this is because the insertId is being generated from a stringified md5 hash of the row. Would you be open to a pull request to pass in your own insert id in an optional metadata object? It could fallback to how it operates now.

https://github.com/GoogleCloudPlatform/gcloud-node/blob/fbf1ecc36c45c48aeee5469e9729a529d94ee610/lib/bigquery/table.js#L904

stephenplusplus:
Fixes #1041

### DIFF
--- a/lib/bigquery/table.js
+++ b/lib/bigquery/table.js
@@ -849,6 +849,9 @@ Table.prototype.import = function(source, metadata, callback) {
  * @resource [Tabledata: insertAll API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/v2/tabledata/insertAll}
  *
  * @param {object|object[]} rows - The rows to insert into the table.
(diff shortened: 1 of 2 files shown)


callmehiphop:
@stephenplusplus do you see value in a system test for this?

stephenplusplus:
Sure!

stephenplusplus:
Trashing this idea, and going with making insertId an opt-in feature as [suggested](https://github.com/GoogleCloudPlatform/gcloud-node/issues/1041#issuecomment-168941079) by @palamccc.

PR #1055, opened 2016-01-07
core: filter out duplicate error messages

callmehiphop:
In certain cases we are seeing [duplicate error messages](https://travis-ci.org/GoogleCloudPlatform/gcloud-node/jobs/100434361#L5239). This change will filter out any duplicates to avoid error messages like `Not Found - Not Found`.

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -33,6 +33,7 @@ var request = require('request').defaults({
 });
 var retryRequest = require('retry-request');
 var through = require('through2');
+var uniq = require('array-uniq');
 
 /** @const {object} gcloud-node's package.json file. */
 var PKG = require('../../package.json');
@@ -81,13 +82,13 @@ var ApiError = createErrorClass('ApiError', function(errorBody) {
 
   if (errorBody.errors && errorBody.errors.length === 1) {
     messages.push(errorBody.errors[0].message);
-  } else if (errorBody.response.body) {
+  } else if (errorBody.response && errorBody.response.body) {
     messages.push(errorBody.response.body.toString());
   } else if (!errorBody.message) {
     messages.push('Error during request.');
   }
 
-  this.message = messages.join(' - ');
+  this.message = uniq(messages).join(' - ');
 });
 
 /**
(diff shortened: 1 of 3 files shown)


stephenplusplus:
To answer your earlier question, I guess this is why this whole idea was kinda weird. Maybe we should just revert the change and only special-case the error messages for Datastore? That's where we saw the vagueness originally.

callmehiphop:
I think it's still useful, it was really only behaving oddly in Storage, but if you're rather special-case Datastore, that's cool with me too.

lib/common/util.js
```diff
@@ -81,13 +81,19 @@ var ApiError = createErrorClass('ApiError', function(errorBody) {
 
   if (errorBody.errors && errorBody.errors.length === 1) {
     messages.push(errorBody.errors[0].message);
-  } else if (errorBody.response.body) {
+  } else if (errorBody.response && errorBody.response.body) {
```
stephenplusplus: Do we not always get a response?

callmehiphop: Nope, PubSub was throwing an error because there wasn't one.
https://travis-ci.org/GoogleCloudPlatform/gcloud-node/jobs/100434361#L5191

lib/common/util.js
```diff
@@ -81,13 +81,19 @@ var ApiError = createErrorClass('ApiError', function(errorBody) {
...
     messages.push(errorBody.response.body.toString());
   } else if (!errorBody.message) {
     messages.push('Error during request.');
   }
 
-  this.message = messages.join(' - ');
+  this.message = messages.reduce(function(errorMessages, message) {
```
stephenplusplus: We should just use https://github.com/sindresorhus/array-uniq:

```js
this.message = uniq(messages).join(' - ');
```

callmehiphop: Sure, I take it that means we're not special casing Datastore?

stephenplusplus:
I'm cool with keeping the feature if you see it's useful. Let's see what happens!

lib/common/util.js
callmehiphop: Done!

stephenplusplus:
Cool. Pretty sure the answer is yes, but did you confirm the system tests are happy now?

callmehiphop:
Sure did!

PR #1066, opened 2016-01-18
Accept insertID for BigQuery API

vladmiller:
Sometimes you would like to have control over insertID, in my case for each request I have _processedTimestamp_ which defaults to current timestamp thus making md5 always unique and therefore making insertId inefficient.

### DIFF
--- a/lib/bigquery/table.js
+++ b/lib/bigquery/table.js
@@ -850,9 +850,18 @@ Table.prototype.import = function(source, metadata, callback) {
  *
  * @param {object|object[]} rows - The rows to insert into the table.
  * @param {?object} opts - The config
- * @param {?boolean} opts.skipInvalidRows - Insert all valid rows of a request, even if invalid rows exist. The default value is false, which causes the entire request to fail if any invalid rows exist.
- * @param {?boolean} opts.ignoreUnknownValues - Accept rows that contain values that do not match the schema. The unknown values are ignored. Default is false, which treats unknown values as errors.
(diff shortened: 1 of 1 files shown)


vladmiller:
I signed it!

TrejGun:
:+1:

stephenplusplus:
Thanks for this @vladmiller! Did you see #1041? I think we're interested in a solution that allows specifying a `raw` option, as outlined here: https://github.com/GoogleCloudPlatform/gcloud-node/issues/1041#issuecomment-170376959 -- what are your thoughts on that approach?

vladmiller:
@stephenplusplus It doesn't seems like there aren't any other fields except insertID, but things could change. I'll see if I can easily implement solution with raw option

stephenplusplus:
That's true, here's the spec: https://cloud.google.com/bigquery/docs/reference/v2/tabledata/insertAll

- rows[]
- rows[].insertId
- rows[].json
- skipInvalidRows
- ignoreUnknownValues
- templateSuffix

So, maybe the options object should also accept the `skipInvalidRows`, `ignoreUnknownValues`, and `templateSuffix` options:

```js
var opts = {
  raw: true,
  ignoreUnknownValues: true
};

var row = {
  insertId: 'my-id',
  json: {
    // row data
  }
};

table.insert([row, row], opts, function() {});
```

vladmiller:
@stephenplusplus okay, let's see if it's easy

stephenplusplus:
Thanks!

vladmiller:
@stephenplusplus pushed new commit with tests and fix. Please check and let me know what you think.

PR #1068, opened 2016-01-18
bigquery: remove automatic `insertId` generation & allow specifying raw format

stephenplusplus:
Closes #1066
### Breaking change included!

`insertId` is no longer defaulted to a value.

@vladmiller - Please take a look. I just wanted to change a few things (mostly style things) and also remove the default `insertId` generation. The CLA bot will say something about confirming you are the original author of this code. If you're okay with it, just leave a note that it's okay.

### DIFF
--- a/lib/bigquery/table.js
+++ b/lib/bigquery/table.js
@@ -21,7 +21,6 @@
 'use strict';
 
 var arrify = require('arrify');
-var crypto = require('crypto');
 var duplexify = require('duplexify');
 var extend = require('extend');
 var format = require('string-format-obj');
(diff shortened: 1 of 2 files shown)


vladmiller:
@stephenplusplus I think that we have to leave autogenerated insertID untouched s.t. old apps won't break.

vladmiller:
Okay Google, contribute my commits.

vladmiller:
Okay Google, I agree to contribute my commits.

vladmiller:
Google set cla yes

stephenplusplus:
We will just bump the minor version to follow semver rules for the breaking change (pre-1.0, a minor bump is equivalent to a post-1.0 major bump).

stephenplusplus:
Thanks @vladmiller!

vladmiller:
@stephenplusplus any estimate on when package becomes available in npm?

stephenplusplus:
We're due for a release soon. Maybe I can get one out this week. In the meantime, you can use master:

```sh
$ npm install --save googlecloudplatform/gcloud-node
```

vladmiller:
@stephenplusplus Thank you!

vladmiller:
@stephenplusplus do you know when this PR will be merged?

stephenplusplus:
Just needs a review from @callmehiphop.

This should work if you want to install from my branch for now:
```sh
$ npm install --save stephenplusplus/gcloud-node#vlad--patch-1
```

callmehiphop:
@stephenplusplus Looks good to me!

jgeewax:
@stephenplusplus I'm confused by this one. Why can't we allow control of the insert ID via overriding without keeping the auto-generation?

The request in #1066 was about being able to manually specify an insert ID, so adding the `raw` thing works, but taking out the other part kind of screws over the people who don't necessarily care about what value the ID is so long as it is unique (effectively saying "if an API request is one I've already sent, treat it as a duplicate")

The request in #1041 is about adding multiple of the same rows in parallel (aka, same timestamp, same data), so I can see how a hash wouldn't make sense there.

What we're trying to control for is:

- Insert rows `r`
- Request times out (maybe network cable was yanked, maybe GCP had an error)
- It's impossible to know for sure whether this batch of rows was added
- Retry inserting rows `r`

I'd argue that this is *not* for de-duping your data, and it is *not* for recovering after a client-side failure. If you want uniqueness or transactions like that, you should use a different storage system or de-duplicate, save to GCS, and bulk load the data.

Can we re-open the discussion about using a UUID-1 (or similar) to generate insert ID values when none are provided (and which are only used when we automatically retry a failed request) ?

stephenplusplus:
Sure, let's move the convo back to the now re-opened #1041.

PR #1069, opened 2016-01-19
datastore: separated property names from operators

### LINKED ISSUE #1061: Datastore Query order should use an explicit ascending/descending parameter instead of "-" in a string.
Currently:

    // Sort by size ascendingly.
    var companiesAscending = companyQuery.order('size');

    // Sort by size descendingly.
    var companiesDescending = companyQuery.order('-size');

This is bad.  It is mixing the properties of a string based query language with a json based query API.

Instead, consider just adding a boolean ascending / descending property that defaults to ascending.

    var companiesAscendingImplicit = companyQuery.order('size');
    var companiesAscendingExplicit = companyQuery.order('size', true);
    var companiesDescending = companyQuery.order('-size', false);

### LINKED ISSUE #1062: Datastore Query filter should use a separate parameter for operator instead of combining field and operator in a single string.
Currently:

    var companyQuery = query
      .filter('name =', 'Google')
      .filter('size <', 400);

Instead, use a separate parameter for operator.  Combining them is weird.  It is more like a text based query language, instead of a JSON based API.

    var companyQuery = query
      .filter('name', '=', 'Google')
      .filter('size', '<', 400);

Later, if desired, this could easily be extended to use declared values for valid operators.

    var companyQuery = query
      .filter('name', query.operator.EQUALS, 'Google')
      .filter('size', query.operator.GREATER_THAN, 400);

I think that final example might be overkill, but at the least, the name and operator should not be combined in a single string.

callmehiphop:
### Breaking changes ahead!

closes #1062 
closes #1061 

There are two methods within datastore that combine property names with operators when building queries, this PR moves the operators to a separate parameter.

@krisnye - If you don't mind, please take a look! :)

### DIFF
--- a/lib/datastore/query.js
+++ b/lib/datastore/query.js
@@ -109,29 +109,39 @@ Query.prototype.autoPaginate = function(autoPaginateVal) {
  *
  * @resource [Datastore Filters]{@link https://cloud.google.com/datastore/docs/concepts/queries#Datastore_Filters}
  *
- * @param {string} filter - Property + Operator (=, <, >, <=, >=).
+ * @param {string} property - The field name.
+ * @param {string=} operator - Operator (=, <, >, <=, >=). Default: `=`.
  * @param {*} value - Value to compare property to.
  * @return {module:datastore/query}
  *
  * @example
+ * //-
+ * // List all companies that are located in California.
+ * //-
+ * var caliQuery = query.filter('state', 'CA');
+ *
+ * //-
  * // List all companies named Google that have less than 400 employees.
+ * //-
  * var companyQuery = query
- *   .filter('name =', 'Google')
- *   .filter('size <', 400);
+ *   .filter('name', '=', 'Google')
+ *   .filter('size', '<', 400);
  *
  * // To filter by key, use `__key__` for the property name. Filter on keys
  * // stored as properties is not currently supported.
- * var keyQuery = query.filter('__key__ =', dataset.key(['Company', 'Google']));
+ * var key = dataset.key(['Company', 'Google']);
+ * var keyQuery = query.filter('__key__', '=', key);
  */
-Query.prototype.filter = function(filter, value) {
+Query.prototype.filter = function(property, operator, value) {
+  if (arguments.length === 2) {
+    value = operator;
+    operator = '=';
+  }
+
   // TODO: Add filter validation.
-  var operatorRegex = /[><=]/g;
-  filter = filter.trim();
-  var fieldName = filter.replace(operatorRegex, '').trim();
-  var op = filter.substr(fieldName.length, filter.length).trim();
   this.filters.push({
-    name: fieldName,
-    op: op,
+    name: property.trim(),
+    op: operator.trim(),
     val: value
   });
   return this;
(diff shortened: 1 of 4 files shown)


lib/datastore/query.js
```diff
@@ -159,7 +157,10 @@ Query.prototype.hasAncestor = function(key) {
  *
  * @resource [Datastore Sort Orders]{@link https://cloud.google.com/datastore/docs/concepts/queries#Datastore_Sort_orders}
  *
- * @param {string} property - Optional operator (+, -) and property to order by.
+ * @param {string} property - The property to order by.
+ * @param {object} options - Options object.
```
stephenplusplus: `object=`

lib/datastore/query.js
```diff
@@ -109,29 +109,27 @@ Query.prototype.autoPaginate = function(autoPaginateVal) {
...
- *   .filter('name =', 'Google')
- *   .filter('size <', 400);
+ *   .filter('name', '=', 'Google')
+ *   .filter('size', '<', 400);
  *
  * // To filter by key, use `__key__` for the property name. Filter on keys
  * // stored as properties is not currently supported.
- * var keyQuery = query.filter('__key__ =', dataset.key(['Company', 'Google']));
+ * var key = dataset.key(['Company', 'Google']);
+ * var keyQuery = query.filter('__key__', '=', key);
```
stephenplusplus: Should we default to `=`?

```js
query.filter('__key__', key);
```

lib/datastore/query.js
callmehiphop: Ahh, good call.

lib/datastore/query.js
callmehiphop: I think that's a great idea!

callmehiphop: Not 100% if this is the way to go.. since a value could potentially be anything it's hard to tell when a user is omitting an operator, is checking to see if the value is undefined a bad move?

stephenplusplus: Check the arguments.length!

callmehiphop: Derp, good call!

lib/datastore/query.js
```diff
@@ -109,29 +109,32 @@ Query.prototype.autoPaginate = function(autoPaginateVal) {
...
+ * @param {string=} operator - Operator (=, <, >, <=, >=). Default: `=`.
  * @param {*} value - Value to compare property to.
  * @return {module:datastore/query}
  *
  * @example
  * // List all companies named Google that have less than 400 employees.
  * var companyQuery = query
- *   .filter('name =', 'Google')
- *   .filter('size <', 400);
+ *   .filter('name', '=', 'Google')
```
stephenplusplus: Can you squeeze in an example showing the simplest use possible, something like...

```js
 * //-
 * // List all companies that are located in California.
 * //-
 * var caliQuery = query.filter('state', 'CA');
 *
 * //-
 * // List all companies named Google that have less than 400 employees.
 * ...
```

stephenplusplus:
LGTM!

krisnye:
Looks good.  Impressed by the quick response and fix.

Issue #1081, opened 2016-01-21
Transaction.save does not allow a NULL value.


krisnye:
Null is a perfectly valid value to insert into the datastore and should be supported by this API.  Null indicates that the value is not present, but that it still needs to be indexed, and so it is different from simply not having a property which means the entity will not be indexed for that property.

When attempting to save a record with a null value this exception is thrown:

      throw new Error('Unsupported field value, ' + v + ', is provided.');
            ^
    Error: Unsupported field value, null, is provided.
        at Object.valueToProperty (gcloud/lib/datastore/entity.js:437:9)
        at gcloud/lib/datastore/request.js:405:29
        at gcloud/lib/datastore/request.js:404:48
        at Transaction.DatastoreRequest.save (gcloud/lib/datastore/request.js:397:12)
        at gcloud/lib/datastore/transaction.js:229:42
        at Transaction.commit_ (gcloud/lib/datastore/transaction.js:225:6)
        at /Users/krisnye/Projects/glass/lib/googlecloud/Dataset.js:343:21

krisnye:
Ok, I found a fix for this, it's incredibly simple.  In datastore/entity.js, simply encode a null value as {stringValue:null}

Old code

  if (v instanceof String || typeof v === 'string') {
    p.string_value = v;
    return p;
  }

New code

  if (v == null || v instanceof String || typeof v === 'string') {
    p.string_value = v;
    return p;
  }

stephenplusplus:
Duplicate of #1022. Sadly, we're a bit stuck until the next version of Datastore is released, v1beta3.

Issue #1022, opened 2015-12-14
Datastore fails when saving entity with null attribute value


toboid:
When trying to save an object that has null property values `gcloud-node` errors:

``` javascript
dataset.insert({
  key: dataset.key('Test'),
  data: { foo: 'bar', baz: null }
}, err => {
  if (err) return console.log(err);
});
```
prints:
>Error: Unsupported field value, null, is provided.

I can see that the definition for `Value` in the [Datastore proto buffers api](https://github.com/GoogleCloudPlatform/google-cloud-datastore/blob/v1beta2-rev1-3.0.2/proto/datastore_v1.proto#L99) does not provide for null values. The [Ruby](https://github.com/blowmage/gcloud-ruby/blob/master/lib/gcloud/datastore/proto.rb#L67) and [Python](https://github.com/lucemia/gcloud-python/blob/master/gcloud/datastore/helpers.py#L261) clients create a message with no value attribute, is this an option for `gcloud-node`? I'd be really pleased to take a look at this if you'll take a pull request?

stephenplusplus:
Definitely, that would be very welcome!

// @pcostell - makes sense, right?

pcostell:
Makes sense to me. Can we make sure this gets updated for the v1beta3 branch as well? In particular the next version of the API includes NullValue as an explicit field: https://github.com/google/googleapis/blob/master/google/datastore/v1beta3/entity.proto#L119

toboid:
OK great. Correctly saving the Null to datastore is straightforward, however the value comes back out as an empty array because [this](propertyToValue) function isn't able to distinguish between a property representing an empty array is currently indistinguishable from a property representing a null value since they have the exact same structure:

``` javascript
{
  "key": {
    ...
  },
  "property": [
    {
      "name": "i-am-a-null",
      "value": {
        "boolean_value": null,
        "integer_value": null,
        "double_value": null,
        "timestamp_microseconds_value": null,
        "key_value": null,
        "blob_key_value": null,
        "string_value": null,
        "blob_value": null,
        "entity_value": null,
        "list_value": [],
        "meaning": null,
        "indexed": true
      }
    },
    {
      "name": "i-am-an-empty-array",
      "value": {
        "boolean_value": null,
        "integer_value": null,
        "double_value": null,
        "timestamp_microseconds_value": null,
        "key_value": null,
        "blob_key_value": null,
        "string_value": null,
        "blob_value": null,
        "entity_value": null,
[...]

PR #1082, opened 2016-01-21
Allow NULL to be used as a property value

krisnye:
Fix for bug #1081

### DIFF
--- a/lib/datastore/entity.js
+++ b/lib/datastore/entity.js
@@ -422,7 +422,7 @@ function valueToProperty(v) {
     p.timestamp_microseconds_value = v.getTime() * 1000;
     return p;
   }
-  if (v instanceof String || typeof v === 'string') {
+  if (v instanceof String || typeof v === 'string' || v === null) {
     p.string_value = v;
     return p;
   }


krisnye:
I signed it!

jgeewax:
@stephenplusplus : Would this be better as a separate if-statement line for nulls?

stephenplusplus:
This doesn't actually work. The problem isn't with saving null data, but detecting a null value when it's returned from the API. We talked about this in #1022, and the easiest thing to do is wait for native support for a null value in the next version of Datastore, v1beta3... which hopefully is coming soon!

krisnye:
Yes, reading the value back is a separate issue though.  At least being able to put a null value into the datastore can be supported immediately.

I, for one have no problem reading back the values, because I use a separate field which stores the entire JSON object I am interested in.  Other people could do a work-around if they know the intended type of the field.  There really is no work-around possible for not being able to put null into the datastore.

stephenplusplus:
Since v1beta3 is around the corner, waiting for native support without side effects seems like the best solution to me. Supporting setting but not getting feels irresponsible for our library. However, feel free to share your examples in #1022, which will involve people who have already considered this problem. Maybe with all of our heads combined, we can find a solid way to support this now.

krisnye:
Ok.  How soon is v1beta3 coming?

PR #1093, opened 2016-01-27
datastore: normalize arguments in Dataset constructor

### LINKED ISSUE #1092: Datastore client not getting ProjectID from env
It looks like Datastore client is not getting the projectID from the env automatically.

```javascript
var pubsub = gcloud.pubsub();
```
This works fine, but:

```javascript
```
returns:
```
Error: Sorry, we cannot connect to Google Cloud Services without a project ID. You may specify one with an environment variable named "GCLOUD_PROJECT". See https://googlecloudplatform.github.io/gcloud-node/#/authentication for a detailed guide on creating an authenticated connection.
```
This fixes the problem
```javascript
var dataset = gcloud.datastore.dataset({projectId: process.env.GCLOUD_PROJECT});
```

stephenplusplus:
Fixes #1092 

It looks like we forgot to normalize the arguments for a Dataset.

### DIFF
--- a/lib/datastore/dataset.js
+++ b/lib/datastore/dataset.js
@@ -104,12 +104,17 @@ var SCOPES = [
  */
 function Dataset(options) {
   if (!(this instanceof Dataset)) {
+    options = util.normalizeArguments(this, options);
     return new Dataset(options);
   }
 
   options = options || {};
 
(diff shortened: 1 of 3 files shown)


callmehiphop:
Good catch!

Issue #1094, opened 2016-01-28
[Docs] External links should open in a new tab


VikramTiwari:
Most of the links in the documentation are going to the API docs on [cloud.google.com](https://cloud.google.com) or in general out of the gcloud-node docs website. Its always a better choice to open all such external links in a new tab.

Even the icon next to such link indicates to open in new tab.
![screen shot 2016-01-28 at 11 40 41 am](https://cloud.githubusercontent.com/assets/1330677/12656617/1942561a-c5b4-11e5-8847-307144a8d307.png)

Material Design Ref:
![screen shot 2016-01-28 at 11 40 28 am](https://cloud.githubusercontent.com/assets/1330677/12656614/1796533e-c5b4-11e5-9756-d089f41f611e.png)

stephenplusplus:
I think the steps to complete this are:

- [ ] Overload `<a>` from `gcloud-common`. It should look for links that begin with `http` and assign a `target="_blank"`.
- [ ] Search for links in this repo that are external, and manually add `target="_blank"`

@callmehiphop does that sound correct? Do we still need to do no. 2 if no. 1 is implemented?

callmehiphop:
I don't think so, but it should be mentioned that overloading <a> is slightly more complicated than it sounds. I tried this and ran into some issues where anchors dropped into the page via `bind-html` wouldn't get overloaded. My guess had something to do with the compiling phase ending and not traversing HTML appended in that fashion

stephenplusplus:
Makes sense. Do we get any events when compiling is done that we can hook onto and re-compile? And if a loop is unavoidable there, maybe we can use that same event to just run some raw JS that assigns the target onto links with a shared class, e.g. `ext`.

callmehiphop:
I don't believe there is an event for that.

Not a very graceful solution, but maybe we should just make a service and pass it to resource links, descriptions and markdown files.

eoogbe:
This issue may no longer be relevant due to its age. Feel free to re-open.

eoogbe:
I am no longer on the team

PR #1095, opened 2016-01-28
open external links in new tab

VikramTiwari:
I think this is all, but there might be more such links/

Regarding: https://github.com/GoogleCloudPlatform/gcloud-node/issues/1094

### DIFF
--- a/lib/logging/entry.js
+++ b/lib/logging/entry.js
@@ -48,7 +48,7 @@ var is = require('is');
  *   type: 'gce_instance',
  *   labels: {
  *     zone: 'global',
- *     instance_id: `3`
+ *     instance_id: '3'
  *   }
  * };
  *


stephenplusplus:
Thanks for sending this, @VikramTiwari!

@callmehiphop is this the best way to open external links in new windows, or can we use some fancy JS to handle this for us?

docs/site/components/language-switcher/language-switcher.html
```diff
@@ -74,25 +74,25 @@ <h4 class="list-item--heading">External Resources</h4>
       <h4 class="list-item--heading">gcloud Libraries</h4>
     </li>
     <li>
-      <a href="http://googlecloudplatform.github.io/gcloud-java" title="gcloud-java Documentation" class="skip-external-link">
+      <a href="http://googlecloudplatform.github.io/gcloud-java" target="_blank" title="gcloud-java Documentation" class="skip-external-link">
```
callmehiphop: These links are supposed to behave more like a dropdown than anchor tags.. so I don't think we want new window here.

callmehiphop:
> is this the best way to open external links in new windows, or can we use some fancy JS to handle this for us?

Well, if we wanted to simplify the whole thing we could create an angular directive to overload the `<a>` tag and set the target dynamically.

stephenplusplus:
I'm cool with that, since a lot of our links are going to be coming from Markdown in the docs.

callmehiphop:
Cool!

@VikramTiwari did you want to take that on? Otherwise I can get to it later :)

VikramTiwari:
I can take a stab at it, but what would be the logic that we would like to use here? Not all absolute links are external links.

callmehiphop:
We use a CSS class called `skip-external-link`, so I would think any absolute URL missing that class would qualify? 

@stephenplusplus thoughts?

stephenplusplus:
Sounds sane to me.

stephenplusplus:
@callmehiphop since all of the docs changes we've pushed, will @VikramTiwari rebasing be enough, or is there anything else to it?

callmehiphop:
Nope :(

We'll probably have to mirror these changes in https://github.com/GoogleCloudPlatform/gcloud-common

stephenplusplus:
@VikramTiwari we moved most of our docs to a centralized repo, so that the other languages (gcloud-ruby, gcloud-php, etc) can set up their docs as well.

The new location is https://github.com/GoogleCloudPlatform/gcloud-common/tree/master/site/src -- I believe the task now is finding links across gcloud-common that need to be updated, as well as some changes still inside this repo. Are you still up for it? It's totally understandable if you'd rather have us do it. Thanks for your help so far and sorry for the curveball!

// @callmehiphop is all of that ^ correct?

callmehiphop:
@stephenplusplus yep!

VikramTiwari:
Hey @stephenplusplus Good move for centralized repo. I will try to do the changes at that repo by this weekend. Will keep you all updated.

stephenplusplus:
Awesome, thanks!

stephenplusplus:
I'll close this up and let #1094 track our progress.

PR #1096, opened 2016-01-29
docs: overhauled documentation

callmehiphop:
### _Big Update Ahead!_

> This PR is one (of several) to move gcloud-node to the new gcloud-common docs app.
#### Notable changes:
- Completely removed old docs site
- Created a manifest.json file for the new site
- Updated all `@example` tags to follow JSDoc standard
- Created a new script to generate documentation JSON
### TODOS
- [x] Remove `@linkcode` tags in favor for custom type syntax (`{module:service/object#method}`)
- [x] Correct improper use of `@mixes` tag for ACL and IAM properties
- [x] Update script to parse example comment syntax
- [x] Move overviews to @classdesc tags
- [x] Re-generate JSON for previous versions (separate PR to gh-pages) (https://github.com/GoogleCloudPlatform/gcloud-node/pull/1098)
- [x] version markdown files
- [x] Wait for new doc app to be merged into gcloud-common (https://github.com/GoogleCloudPlatform/gcloud-common/pull/34)

@stephenplusplus let me know if there's anything I can do to make this easier to go through.. :)


stephenplusplus:
> Correct improper use of @mixes tag for ACL and IAM properties

What is the improper use?

callmehiphop:
> What is the improper use?

I think `@mixin` implies that we're inheriting either IAM or ACL, but we're actually just storing an instance as a property.. little bit different.

stephenplusplus:
@mixin: A mixin provides functionality that is intended to be added to other objects. If desired, you can use the @mixin tag to indicate that an object is a mixin. You can then add the @mixes tag to objects that use the mixin.

JSDocs is pretty wild. What's a better way to describe (and document) the connection?

callmehiphop:
> What's a better way to describe (and document) the connection?

Still looking into that.. worst case scenario I'll just make `@mixin` work somehow!

stephenplusplus:
Cool cool.

callmehiphop:
But if I can make it work without it, one thing I'd strive to do is kill off all the [empty variables](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/storage/bucket.js#L283)..

stephenplusplus:
Yes, please!

lib/index.js
```diff
@@ -317,9 +317,6 @@ var scopedApis = {
  * 2. Provide them at the time of instantiation of sub-modules, e.g. a Datastore
  * dataset, a Cloud Storage bucket, etc.
  *
- * See our [Authentication Guide](#/authentication) for how to obtain the
```
callmehiphop: I took this out because all the guides are now tied to a version

lib/datastore/transaction.js
```diff
@@ -353,20 +375,25 @@ Transaction.prototype.delete = function(entities) {
...
- *     data: {
- *       HQ: 'Dallas, TX'
- *     }
- *   },
- *   {
- *     key: productKey,
- *     data: {
- *       vendor: 'Dell'
+ *
+ * dataset.runInTransaction(function(transaction, done) {
```
callmehiphop: An unfortunate side effect to generating the new JSON schema is that we filter out all private examples, so we'll need to test transactions via `dataset.runInTransaction` now :(((

docs/manifest.json
```diff
@@ -0,0 +1,260 @@
+{
+  "lang": "node",
+  "friendlyLang": "Node.js",
+  "markdown": "javascript",
+  "versions": [
```
callmehiphop: @stephenplusplus any thoughts on how we should go about programmatically updating this?

stephenplusplus: Maybe we should have a release script that:

1. Runs linter
2. Builds docs
3. Runs unit tests
4. Runs system tests
5. Makes release tag via `npm version`
6. Updates `manifest.json` & commits changes
7. Pushes (with tags) to master
8. Runs `npm publish`

--

Or, could the FE just ping GH's API for the versions? Maybe like how it currently gets the latest version for the homepage? `(Latest Release v0.27.0 Dec 22, 2015)`

callmehiphop: > Or, could the FE just ping GH's API for the versions? Maybe like how it currently gets the latest version for the homepage?

Oddly enough I tried this and ran into rate limiting pretty quick, so I abandoned the idea.

scripts/docs.js
```diff
@@ -0,0 +1,274 @@
...
+}
+
+function createMethod(file, block) {
+  return {
+    metadata: {
+      constructor: !!block.isConstructor,
+      name: detectName(block),
+      source: file.replace(/^\./, '') + '#L' + block.codeStart,
+      description: formatHtml(block.description.full),
+      // examples: getByType(block, 'example').map(createExample),
```
stephenplusplus: Unneeded?

scripts/docs.js
```diff
@@ -0,0 +1,274 @@
...
+
+      var docs = parseDocs(file, contents, 0);
+      var outputFile = path.join(
+        './docs/json/master',
+        file.replace('/lib', '').replace('.js', '.json')
+      );
+
+      fs.writeFile(outputFile, JSON.stringify(docs), callback);
+    });
+  });
```
stephenplusplus: I might be reading this wrong, but should we add a callback here:

```js
async.each(files, function(file, callback) {/*...*/}, function(err) {
  if (err) {
    throw err;
  }
});
```

scripts/docs.sh
```diff
@@ -1,73 +0,0 @@
-#!/bin/bash
```
stephenplusplus: Bye, weird file!

scripts/docs.js
```diff
@@ -0,0 +1,274 @@
...
+ */
+
+'use strict';
+
+var globby = require('globby');
+var async = require('async');
+var fs = require('fs');
+var dox = require('dox');
+var path = require('path');
+var exec = require('propprop');
```
stephenplusplus: We usually call this `prop`. `exec` is for `methmeth`. The difference is `prop` just returns the property, where `exec` will execute the method.

scripts/docs.js
callmehiphop: Ahh, yeah. I ended up backing out the JSDoc `<caption>` stuff. Thanks!

scripts/docs.js
callmehiphop: Gotcha, good call!

quartzmo:
@callmehiphop Can you point me to where your build waits "for new doc app to be merged into gcloud-common"? I saw that got checked off but can't find the commit. I'm currently working on our gcloud-ruby json output using the docs app & assets that I manually copied from your gh-pages... But of course I need to automate.

callmehiphop:
@quartzmo I was just waiting for the main app to get merged into gcloud-common, which it did yesterday via https://github.com/GoogleCloudPlatform/gcloud-common/pull/34. We have a hook that will generate JSON, etc. and didn't want to merge these changes in until all the other parts were good to go.

PR #1098, opened 2016-02-01
docs: updated doc site and json schema

callmehiphop:
[Demo](http://callmehiphop.github.io/gcloud-node/#/)
- Latest doc site from https://github.com/GoogleCloudPlatform/gcloud-common/pull/34
- Latest JSON from https://github.com/GoogleCloudPlatform/gcloud-node/pull/1096


jgeewax:
Looks like there are some conflicts, which isn't hard to believe when there are over 1000 changes files! Demo looks awesome (as awesome as always).

I just noticed that http://callmehiphop.github.io/gcloud-node/#/docs/v0.27.0/datastore/transaction has a `@linkcode{` snippet that doesn't seem to fit in. Did we forget something in our parsing situation?

stephenplusplus:
:clap: x 1,094

Found some hopefully small things!

Some `{@linkcode`s are here as well: http://callmehiphop.github.io/gcloud-node/#/docs/v0.27.0/pubsub/subscription

--

The Subscription constructor has some malformed examples. See the `Once you have obtained a subscription object, you may begin to register listeners` section.

--

Links like these: http://callmehiphop.github.io/gcloud-node/#/docs/v0.27.0/dns
screen shot 2016-02-01 at 8 08 40 am

Seem to be pointing to a dead page now (http://callmehiphop.github.io/gcloud-node/#/docs/?method=gcloud).

I think they just need the `?method=gcloud` to be removed?

--

http://callmehiphop.github.io/gcloud-node/#/docs/v0.27.0/storage/bucket?method=acl.acl.delete

`acl.acl.delete`? I think there's an extra `acl.` in there.

--

Any reason `initial-width=1` was removed? It makes my phone (Note 5) start scaled down (.8):

screen shot 2016-02-01 at 7 55 30 am
*this is using the Note 3 emulator*

I tried with others using the emulator, and they seem to be affected as well.

The inner docs pages work for my physical device, but the simulator shows a different story for example:

Note 3:
screen shot 2016-02-01 at 8 02 16 am

iPhone 5:
screen shot 2016-02-01 at 8 04 14 am

--

Small one!: 

new site:
screen shot 2016-02-01 at 7 59 22 am

old site:
screen shot 2016-02-01 at 8 00 50 am

stephenplusplus:
https://github.com/callmehiphop/gcloud-node/tree/08afcbb5a135677d66e3cffd8aed19c23b5b94c5/scripts seems to have extra files in there. Also, can you point me to where the script is that gives them the unique hashes?

callmehiphop:
> I just noticed that http://callmehiphop.github.io/gcloud-node/#/docs/v0.27.0/datastore/transaction has a @linkcode{ snippet that doesn't seem to fit in. Did we forget something in our parsing situation?

Yeah, slipped my mind when generating the JSON for previous versions, should be fixed now though!

> I think they just need the ?method=gcloud to be removed?

Good catch! Fixed!

> acl.acl.delete? I think there's an extra acl. in there.

This might have been a side effect from all the `@alias` tags, I ended up removing them from master.. which is probably why I missed it!

> Any reason initial-width=1 was removed?

Accident! Went with a yeoman scaffold and didn't realize their meta tag would be different.

> Small one!

Should be fixed!

> seems to have extra files in there. Also, can you point me to where the script is that gives them the unique hashes

Ah, yeah, I suppose when we update the site we'll need to do a little clean up. The hashes are made via [gulp script](https://github.com/callmehiphop/gcloud-common/blob/shared-docs/site/gulp/build.js#L54) when running `gulp build`.

> The Subscription constructor has some malformed examples. See the Once you have obtained a subscription object, you may begin to register listeners section.

Eek, going to look into this one, must be the example parsing code.

callmehiphop:
Ran into an issue where certain versions of dox changed where the source line no. was, (I had to rollback versions to avoid parse errors in previous releases), so I need to go through and tidy up those.

stephenplusplus:
Sorry :(

callmehiphop:
Don't be!

callmehiphop:
@stephenplusplus if you want to take another look, I think I've tackled everything

stephenplusplus:
LG :+1:, but the `Small one!` is still there:

screen shot 2016-02-01 at 3 48 47 pm

Just to clarify, I mean the space between "Version" and the dropdown. Previously, they had some breathing room, but now they're best friends.

callmehiphop:
> Small one! is still there:

Ahh, I think that might have resurfaced from minifying the HTML. I added a `&nbsp;`

Issue #1104, opened 2016-02-03
Abort outstanding HTTP requests when subscription is closed


Moezalez:
Hi!

After a certain event I just want my app to shut down. It seems, though, that gcloud (where I only use pubsub) keeps my App alive. Is there a way to kill pubsub so my App can shut down?

stephenplusplus:
It is probably a subscription object that is still pulling for new messages. You can quickly stop that process by calling subscription.removeAllListeners()

Moezalez:
Hey @stephenplusplus, thanks for your answer.
That did not work for me.
Here a little example of what I mean. This will stay alive:

```JavaScript
  let pubsub = gcloud.pubsub({
    projectId: projectId,
    keyFilename: keyPath
  });

  let update = pubsub.topic('someTopic');

  update.subscribe('update', {reuseExisting: true}, function (error, updateSub) {
    if (error) {
      return console.log(error);
    }

    updateSub.on('error', function (error) {
      console.log(error);
    });

    updateSub.on('message', function (message) {
      console.log(message);
    });

    setTimeout(function () {
      updateSub.removeAllListeners();
    }, 500);
  });
```

stephenplusplus:
Got to the bottom of this! Unfortunately, there are a few pieces involved...

1) Removing listeners is the right way to do this :+1:, but our code doesn't abort the HTTP request that remains open while waiting for new messages. We have to fix that (pr incoming soon)
2) We need a way to abort requests, but we're limited by one of the libraries we use, `retry-request`. I just fixed that: https://github.com/stephenplusplus/retry-request/commit/ecb30d01251f14079e4530f32291c5f050af9285
3) We need a way for the underlying library, `request` to allow us to abort a request that has a `timeout` set. PR sent for that here: https://github.com/request/request/pull/2052

stephenplusplus:
This is actually going to be resolved by #1070, since we're switching to using [`gRPC`](https://github.com/grpc/grpc). If all goes well, a release that has this completely resolved will be put out on Monday.

PR #1105, opened 2016-02-03
pubsub: allow aborting an active HTTP request + auto-abort when subsc…

stephenplusplus:
…ription closed

RE: #1104

This modifies our core request functionality to expose an `abort` function which cancels an active HTTP request (via `request`).

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -323,6 +323,7 @@ function makeAuthenticatedRequestFactory(config) {
   function makeAuthenticatedRequest(reqOpts, options) {
     var stream;
     var reqConfig = extend({}, config);
+    var activeRequest_;
 
     if (!options) {
       stream = duplexify();
@@ -345,7 +346,8 @@ function makeAuthenticatedRequestFactory(config) {
       if (options && options.onAuthenticated) {
         options.onAuthenticated(null, authenticatedReqOpts);
       } else {
-        util.makeRequest(authenticatedReqOpts, reqConfig, options);
+        activeRequest_ =
+          util.makeRequest(authenticatedReqOpts, reqConfig, options);
       }
     }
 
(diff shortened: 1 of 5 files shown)


lib/common/util.js
```diff
@@ -360,6 +362,14 @@ function makeAuthenticatedRequestFactory(config) {
     if (stream) {
       return stream;
     }
+
+    return {
+      abort: function() {
+        if (activeRequest_) {
+          activeRequest_.abort();
```
callmehiphop: Should we set `activeRequest_` to `null` after aborting to prevent multiple `abort()` calls?

stephenplusplus: Yes, good thinking!

stephenplusplus: Done!

callmehiphop: hmm, would it actually make more sense to just return the `activeRequest_` instead of just wrapping `abort`? Or are there things we'd prefer not to expose?

stephenplusplus: I would like to, but `activeRequest_` comes to exist asynchronously via `onAuthenticated`

callmehiphop: doh'

test/common/util.js
```diff
@@ -789,6 +789,35 @@ describe('common/util', function() {
...
+
+          utilOverrides.makeRequest = function() {
+            return retryRequest;
+          };
+
+          var makeAuthenticatedRequest = util.makeAuthenticatedRequestFactory();
+          makeAuthenticatedRequest(reqOpts, assert.ifError).abort();
+        });
+
+        it.only('should only abort() once', function(done) {
```
callmehiphop: Don't forget about this little guy!

lib/common/util.js
callmehiphop: Can we document that? If it's async, I could see people running into a scenario where they try and abort but nothing happens.

stephenplusplus: Well, let's walk through the scenarios, and keep in mind these things:

- only we see this API, so no users are involved
- `abort()`, which is ultimately from request (the library), doesn't take a callback so it's always just "assumed to have worked"

Our use case is only calling abort when we know an existing HTTP connection is open and we want to kill it; like for this PR, because a subscription object wants to be discarded, but we're still polling for messages.

In the future if we call it erroneously, like we call it 2ms after the connection was closed or something, the abort will just quietly do nothing. But it's all the same to us, because either way, the connection is closed.

PR #1111, opened 2016-02-04
compute: create progress event

### LINKED ISSUE #1109: Could Compute operation emit 'running' event?
It seems that the operation object can only emit 'complete' and 'error' events. I think it would be great if the object could also emit 'running' or similar event that indicate the 'RUNNING' status in all Operation resources.  
A use case example for that feature is to show users that the instance is in the process of creation. It can be done manually by the method similar to[`Operation.prototype.startPolling_`](https://github.com/GoogleCloudPlatform/gcloud-node/blob/v0.27.0/lib/compute/operation.js#252) but since there already is the function for that job, it should not be re-implemented.

callmehiphop:
Closes #1109 

This adds support for `progress` and `running` events on compute operations.

``` javascript
operation.on('progress', function(progress, metadata) {
  // the operation has progressed
  console.log(progress);
});

operation.on('running', function(metadata) {
  // the operation started running
});
```

Should be noted that it does not fire the event if progress is equal to 0 or 100. 0 implies it hasn't started (and no progress has been made) and 100 implies that it is complete.

### DIFF
--- a/lib/compute/operation.js
+++ b/lib/compute/operation.js
@@ -88,6 +88,13 @@ var util = require('../common/util.js');
  * });
  *
  * //-
+ * // You can register a listener to monitor when the operation begins running.
+ * //-
+ * operation.on('running', function(metadata) {
+ *   // The operation is running.
+ * });
+ *
+ * //-
  * // Be sure to register an error handler as well to catch any issues which
  * // impeded the operation.
  * //-
@@ -267,11 +274,17 @@ Operation.prototype.startPolling_ = function() {
       return;
     }
 
+    if (metadata.status === 'RUNNING' && !self.status) {
+      self.status = metadata.status;
+      self.emit('running', metadata);
+    }
+
     if (metadata.status !== 'DONE') {
       setTimeout(self.startPolling_.bind(self), 500);
       return;
     }
 
+    self.status = metadata.status;
     self.emit('complete', metadata);
   });
 };
--- a/test/compute/operation.js
+++ b/test/compute/operation.js
@@ -406,5 +406,50 @@ describe('Operation', function() {
         operation.startPolling_();
       });
     });
+
+    describe('operation status', function() {
+      var apiResponse;
+
+      beforeEach(function() {
+        apiResponse = { status: 'RUNNING' };
+
+        operation.getMetadata = function(callback) {
+          callback(null, apiResponse, apiResponse);
+        };
+      });
+
+      it('should emit the running event when running', function(done) {
+        operation.on('running', function(metadata) {
+          assert.strictEqual(metadata, apiResponse);
+          done();
+        });
+
+        operation.startPolling_();
+      });
+
+      it('should only emit the running event once', function(done) {
+        var statusSteps = ['PENDING', 'RUNNING', 'RUNNING', 'DONE'];
+        var metadataCallCount = 0;
+        var runningCallCount = 0;
+
+        this.timeout(2000);
+
+        operation.getMetadata = function(callback) {
+          apiResponse.status = statusSteps[metadataCallCount++];
+          callback(null, apiResponse, apiResponse);
+        };
+
+        operation.on('running', function() {
+          runningCallCount++;
+        });
+
+        operation.on('complete', function() {
+          assert.strictEqual(runningCallCount, 1);
+          done();
+        });
+
+        operation.startPolling_();
+      });
+    });
   });
 });


lib/compute/operation.js
```diff
@@ -160,6 +166,7 @@ function Operation(scope, name) {
   this.completeListeners = 0;
   this.hasActiveListeners = false;
   this.name = name;
+  this.progress = 0;
```
stephenplusplus: Maybe we shouldn't initialize this to 0, as one could get a list of operations from `getOperations()` and printing them out would show 0, even if they were completed (while metadata.progress would be 100). We should probably only set progress every time we do a `getMetadata`.

callmehiphop: Good call!

lib/compute/operation.js
```diff
@@ -160,6 +166,7 @@ function Operation(scope, name) {
   this.completeListeners = 0;
   this.hasActiveListeners = false;
   this.name = name;
+  this.progress = null;
```
stephenplusplus: Time for one of these:



`null` is typically used to reference a placeholder for an object that doesn't exist yet. In this case, it's only going to be a number, so the "correct" substitute is simply `undefined`. Do you hate me?

callmehiphop: Haha, nope! Should I just remove it from the constructor altogether then?

callmehiphop:
@tgknight if you don't mind taking a look, I'd appreciate any feedback!

stephenplusplus:
Wdyt about emitting just the progress?

```js
operation.on('progress', function(progress) {
  // progress = 80
  // operation.metadata is already updated
})
```

lib/compute/operation.js
```diff
@@ -228,6 +234,7 @@ Operation.prototype.listenForEvents_ = function() {
 
       if (!self.hasActiveListeners) {
         self.hasActiveListeners = true;
+        self.progress = 0;
```
stephenplusplus: I think this has to go away too

lib/compute/operation.js
```diff
@@ -268,6 +275,11 @@ Operation.prototype.startPolling_ = function() {
     }
 
     if (metadata.status !== 'DONE') {
+      if (metadata.progress > self.progress) {
```
stephenplusplus: This needs testing

callmehiphop:
> Wdyt about emitting just the progress?

I can dig it, I originally decided to emit the metadata to follow suit with `complete`, but if you think it's better without it, that's fine with me!

stephenplusplus:
We could also do `on('progress', function(progress, metadata) {})`. I think just the number is fine, since if they want progress updates, the most important thing is probably just the numeric indicator of progress. If they want metadata, they can still do `operation.metadata` to get whatever else they might need...

So, I'm cool with `.on('progress', function(progress) {})` or `.on('progress', function(progress, metadata) {})` -- you pick :D :ballot_box_with_check:

callmehiphop:
@stephenplusplus PTAL!

lib/compute/operation.js
```diff
@@ -268,6 +274,11 @@ Operation.prototype.startPolling_ = function() {
     }
 
     if (metadata.status !== 'DONE') {
+      if (self.progress !== metadata.progress) {
```
stephenplusplus: Test to make sure it doesn't re-emit the same progress.

lib/compute/operation.js
```diff
@@ -88,6 +88,12 @@ var util = require('../common/util.js');
  * });
  *
  * //-
+ * // You can monitor the progress of an operation by listening for a "progress"
+ * // event.
+ * //-
+ * operation.on('progress', function(metadata) {});
```
stephenplusplus: `function(progress, metadata) {}`

lib/compute/operation.js
callmehiphop: I've written an additional test for that - I think it could be cleaner but I'm not sure how to go about doing that.

test/compute/operation.js
```diff
@@ -406,5 +406,38 @@ describe('Operation', function() {
...
+
+      it('should only emit when progress has updated', function(done) {
+        operation.on('progress', function() {
+          // if done is called multiple times the test will fail
+          done();
+        });
+
+        operation.startPolling_();
+        operation.startPolling_();
+      });
```
stephenplusplus: While this is absolutely not as cool, I would rather see something more explicit:

```js
this.timeout(5000);

var polledCount = 0;
var timesToPoll = 4;
var progressesEmitted = [];

operation.on('progress', function(progress) {
  progressesEmitted.push(progress);
});

operation.getMetadata = function () {
  polledCount++;

  if (polledCount === timesToPoll) {
    callback(null, { state: 'DONE' });

    setImmediate(function() {
      assert.deepEqual(progressesEmitted, [33, 66]);
      done();
    });

    return;
  }

  if (polledCount < timesToPoll / 2) {
    callback(null, { progress: 33 });
  } else if {
    callback(null, { progress: 66 });
  }
};
```

(un-tested, hope it (or something like it) works!)

stephenplusplus: PS: I just edited that post about 100 times

tgknight:
I've updated my project dependencies to be the latest commit, and found that the 'progress' event was emitted only once. Below is a snippet from my code:  
```javascript
zone.createVM(instanceName, instanceObj, function (err, newInstance, operation) { 
  operation.on('progress', function (progress, metadata) {
    console.log('progress', progress) // log as "progress 0"
    console.log('metadata', metadata) // log as "metadata {kind: 'compute#operation',..."
   })
})
```
Then I proceeded to investigate in startPolling_ method. I tried to log new metadata for each polling, and this is what I found:

> status PENDING progress 0
status PENDING progress 0
...
status PENDING progress 0
status RUNNING progress 0
...
status RUNNING progress 0
status RUNNING progress 0
status DONE progress 100

Which explains why the 'progress' event was only emitted once. Emitting the 'progress' event when the operation status changes is a nice workaround for this problem, but I think how the numeric value of metadata.progress is handled should be checked.

test/compute/operation.js
callmehiphop: Alright, updated!

stephenplusplus: It's a beaut.

callmehiphop:
@tgknight, I created a second event that will trigger once the operation starts running. (I've updated the PR description with an example)

@jgeewax the progress field not being updated looks like a potential upstream error, can we confirm?

tgknight:
@callmehiphop The 'running' event is nice and does its job. When the progress filed issue is solved (I still get 0 progress for each polling), 'progress' event should be enough since 'running' is actually a part of the progress. Thank you!

stephenplusplus:
I think we should ditch the `progress` event and just use `running`. Apparently, the operations we allow through our API just don't qualify for progress updates from the upstream API. We can add `progress` support later, if when we fill out the API, we notice it being more informative for the new additions. But for now, supporting `progress` would probably mislead users into thinking they can render UI elements based on frequent updates which just won't come.

tgknight:
I do agree with @stephenplusplus on ditching the ```progress``` event. What is more important for now is knowing the status of operation (whether it is ```RUNNING``` or ```DONE```). I have to admit that my purpose of creating the issue was about UI rendering, but it turned out that the event status is enough (end-users do not need to know the exact percentage of progress anyway).

callmehiphop:
@stephenplusplus PTAL

lib/compute/operation.js
```diff
@@ -88,6 +88,13 @@ var util = require('../common/util.js');
  * });
  *
  * //-
+ * // You can register a listener to monitor when the operation begins running.
+ * //-
+ * operation.on('running', function(metadata) {
+ *   // The operation is running
```
stephenplusplus: `The operation is running.`

lib/compute/operation.js
```diff
@@ -268,10 +275,16 @@ Operation.prototype.startPolling_ = function() {
     }
 
     if (metadata.status !== 'DONE') {
+      if (metadata.status === 'RUNNING' && !self.status) {
```
stephenplusplus: Could you move this above the "if !== DONE" block? It's a little easier to follow along when it's not nested.

stephenplusplus:
Tiny comments, but LGTM!

callmehiphop:
Updated.

PR #1118, opened 2016-02-05
Add Translate API to manifest

stephenplusplus:
I also sorted the service objects under a service to match the hierarchy (if any) between them. Some of our sections did this previously, while others didn't. Another option is just to alphabetize. I'm not particularly attached to either.

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -80,13 +80,13 @@
       "id": "dataset",
       "contents": "bigquery/dataset.json"
     }, {
-      "title": "Job",
-      "id": "job",
-      "contents": "bigquery/job.json"
-    }, {
       "title": "Table",
       "id": "table",
       "contents": "bigquery/table.json"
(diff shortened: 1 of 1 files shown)


docs/manifest.json
```diff
@@ -246,16 +246,20 @@
...
+    }, {
+      "title": "Channel",
+      "id": "channel",
+      "implemented": ">=0.26.0",
+      "contents": "storage/channel.json"
     }]
+  }, {
+    "title": "Translate",
+    "id": "translate",
+    "contents": "translate/index.json"
```
callmehiphop: You're going to want to add an implemented param, otherwise it will show up in every version.

stephenplusplus: Oh my.

stephenplusplus: Fixed in #1119 because I'm a noob.

PR #1123, opened 2016-02-07
docs: filtered out overwritten mixin methods

callmehiphop:
Relates to https://github.com/GoogleCloudPlatform/gcloud-node/issues/1120#issuecomment-180658251

Docs script wasn't filtering out overwritten mixin methods.
Separate PR coming to fix previously generated JSON in gh-pages.


stephenplusplus:
Nice!

PR #1124, opened 2016-02-07
Removing duplicate method entries

callmehiphop:
Relates to https://github.com/GoogleCloudPlatform/gcloud-node/issues/1120


authentication.md
```diff
@@ -1,30 +0,0 @@
-## With `gcloud-node`
```
stephenplusplus: Where did this file end up?

callmehiphop: These are actually versioned, I think they ended up in the root directory due to a build script error that has since been resolved.

https://github.com/GoogleCloudPlatform/gcloud-node/blob/gh-pages/json/master/authentication.md

stephenplusplus: Ah, okay. I thought they weren't versioned... Anyways, merge!

PR #1130, opened 2016-02-16
core: parse error body for JSON

### LINKED ISSUE #1128: ApiError errors field is alway [] during a failed save to datastore
On preforming a duplicate insert the datastore normally returns
 
`{
  "error": {
    "errors": [
     {
       "domain": "global",
       "reason": "INVALID_ARGUMENT",
       "message": "entity already exists"
     }
    ],
    "code": 400,
    "message": "entity already exists"
  }
}`

but instead 'errors' in the ApiError are an []
**Is this intentional?**

From some digging around, the problem is in `lib/datastore/request.js` 
line 813 creates the request 

`813: ...  request(authenticatedReqOpts, function(err, resp, body)`

The value of `body` is already the expected json error message. 
On 'line 819' `util.parseHttpRespMessage(resp)` sets `parsedResp` with ApiError with [] for errors. 
The callback is called and `body` param with the nicely formatted error is lost.

The 'message' field is filled but it has the stringified error which would require re-parsing
`Bad Request - {
  "error": {
    "errors": [
     {
       "domain": "global",
       "reason": "INVALID_ARGUMENT",
       "message": "entity already exists"
     }
    ],
    "code": 400,
    "message": "entity already exists"
  }
}`

stephenplusplus:
Fixes #1128
#### To Dos
- [x] Tests

The response body might be parse-able JSON. If it is, pluck the `error` property out of it.

### DIFF
--- a/lib/common/util.js
+++ b/lib/common/util.js
@@ -22,6 +22,7 @@
 
 var createErrorClass = require('create-error-class');
 var duplexify = require('duplexify');
+var ent = require('ent');
 var extend = require('extend');
 var googleAuth = require('google-auto-auth');
 var is = require('is');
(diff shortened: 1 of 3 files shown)


callmehiphop:
@stephenplusplus the code looks solid to me, do you think we could add some unit tests to cover the try/catch block as well as the character decoding bit?

stephenplusplus:
Done

PR #1135, opened 2016-02-18
docs: schema updates

callmehiphop:
Relates to https://github.com/GoogleCloudPlatform/gcloud-common/pull/102

This PR aims to decouple the site navigation from the <kbd>manifest.json</kbd>, it also includes adjustments to the service JSON schema required for features the gcloud-ruby team has proposed.

---
#### Links
- [New schema gist](https://gist.github.com/blowmage/13be352f40ee7ac5ee48)
- [Demo](http://callmehiphop.github.io/gcloud-node/#/docs/master/gcloud)

---
#### TODOS:
- [x] Split Manifest JSON - https://github.com/GoogleCloudPlatform/gcloud-common/issues/77
  - [x] Create Navigation JSON - https://github.com/GoogleCloudPlatform/gcloud-common/issues/78
- [x] Update Service JSON schema
  - [x] Remove `metadata` keys
  - [x] Make `id` keys unique - https://github.com/GoogleCloudPlatform/gcloud-common/issues/79
  - [x] Create `method.type` - https://github.com/GoogleCloudPlatform/gcloud-common/issues/80
  - [x] Create `type` key - https://github.com/GoogleCloudPlatform/gcloud-common/issues/81
  - [x] Create `title` key - https://github.com/GoogleCloudPlatform/gcloud-common/issues/82
  - [x] Create `parent` & `children` keys - https://github.com/GoogleCloudPlatform/gcloud-common/issues/83
- [x] Move `types` to separate file - https://github.com/GoogleCloudPlatform/gcloud-common/pull/102#issuecomment-195078002
- [x] Re-generate JSON for older versions - https://github.com/GoogleCloudPlatform/gcloud-node/pull/1167

/cc @blowmage @quartzmo


blowmage:
:metal:

quartzmo:
As this progresses, can you publish the output to your gh-pages (or anywhere) so that we can follow the action?

docs/toc.json
```diff
@@ -0,0 +1,346 @@
...
+    "title": "BigQuery",
+    "contents": "bigquery/index.json"
+  }, {
+    "id": "bigquery/dataset",
+    "title": "BigQuery » Dataset",
+    "contents": "bigquery/dataset.json"
+  }, {
+    "id": "bigquery/table",
+    "title": "BigQuery » Table",
+    "contents": "bigquery/table.json"
```
blowmage: I really like this step. I think `types` is a great name for this data. I'm disappointed I didn't think of it. :)

I do have a concern with the data as-is. We have unique IDs for guides and types, but not for methods. I don't see how we are able to link to a method in the `data-custom-type` attribute if we don't list the methods in the version manifest.

Currently, each type is defined as an ID, title, and a URL to the contents.

```json
}, {
  "id": "bigquery/table",
  "title": "BigQuery » Table",
  "contents": "bigquery/table.json"
}, {
```

This structure doesn't list methods and the method IDs. I'd like to see them listed on the same JSON object, so the angular app knows which file to load to get the information. Here is an example of what that might look like, with each method having an ID and a name:

```json
}, {
  "id": "bigquery/table",
  "title": "BigQuery » Table",
  "methods": [{
    "id": "bigquery/table.copy",
    "name": "copy"
  },{
    "id": "bigquery/table.create",
    "name": "create"
  },{
    "id": "bigquery/table.createReadStream",
    "name": "createReadStream"
  }],
  "contents": "bigquery/table.json"
}, {
```

If you have reservations about specifying both an ID and a name, only the ID could be provided, but I think that would hurt future uses, like for search:

```json
}, {
  "id": "bigquery/table",
  "title": "BigQuery » Table",
  "methods": ["bigquery/table.copy", "bigquery/table.create", "bigquery/table.createReadStream"],
  "contents": "bigquery/table.json"
}, {
```

callmehiphop: I apologize for not calling this out in the PR description, but I left them out intentionally. I'm not opposed to adding methods at all, but I am a little lost on how the deeplinking will work and felt that it might be better placed in a separate PR that came after this (Perhaps @quartzmo could help out there). I think that this PR is a necessary first step that will (hopefully) make it easier to implement.

blowmage: Totes understand. This is a terrific step regardless. :+1:

PR #1142, opened 2016-02-29
core: add support for promises

### LINKED ISSUE #551: Why don't we use promises in our API?
#### Promises coming soon. Track progress here: #1142

---

Seems like we're passing callbacks in rather than getting promises back -- was this an explicit choice?

/cc @idosela

callmehiphop:
Fixes #551

This PR aims to add support for [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).

This will also introduce a breaking change moving all stream related APIs to a separate method.

Because this will be such an enormous change and GitHub probably won't be able to display everything, logical commits will be used to make reviewing the changes a little easier.
#### TODO:
- [x] Core support
- [x] Service support - this includes unit tests, system tests and documentation.
  - [x] BigQuery
  - [x] Bigtable
  - [x] Compute
  - [x] Datastore
  - [x] DNS
  - [x] Logging
  - [x] Prediction
  - [x] PubSub
  - [x] Resource Manager
  - [x] Speech
  - [x] Storage
  - [x] Translate
  - [x] Vision
- [x] ~~Documentation site support~~ - for now we're going to rely on examples.

---
#### Testing Locally

To test locally you need to link the `@google-cloud/common` package to all other modules. To do so you can run

``` sh
$ npm run link-common
```

Several modules also have a dependency on `@google-cloud/storage` so additional links may be required.

``` sh
$ cd packages/storage
$ npm link
$ cd ../bigquery
$ npm link @google-cloud/storage
```

---
#### Rollout Plan
- [x] Merge `@google-cloud/common` streaming and promise updates. (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1693)
- [x] Release `@google-cloud/common` `0.7.0`
- [x] Merge `@google-cloud/storage` streaming and promise updates. (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1697)
- [x] Release `@google-cloud/storage` `0.3.0`
- [x] Merge `@google-cloud/pubsub` and `@google-cloud/bigquery` updates.
- [x] Release `@google-cloud/pubsub` and `@google-cloud/bigquery` updates
- [x] Merge everything else!
  - [x] Bigtable (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1713)
  - [x] Compute (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1703)
  - [x]  Datastore (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1704)
  - [x] DNS (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1705)
  - [x]  Language (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1706)
  - [x] Logging (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1707)
  - [x] Prediction (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1708)
  - [x] Resource (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1709)
  - [x] Speech (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1710)
  - [x] Translate (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1711)
  - [x] Vision (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1712)
- [x] Release everything else!
  - [x] Bigtable
  - [x] Compute
  - [x]  Datastore
  - [x] DNS
  - [x]  Language
  - [x] Logging
  - [x] Prediction
  - [x] Resource
  - [x] Speech
  - [x] Translate
  - [x] Vision
- [x] Merge umbrella package updates (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1714)
- [x] Release umbrella package

### DIFF
--- a/packages/bigquery/src/job.js
+++ b/packages/bigquery/src/job.js
@@ -313,6 +313,31 @@ Job.prototype.getQueryResultsStream = function(options) {
 };
 
 /**
+ * Convenience method that wraps the `complete` and `error` events in a
+ * Promise.
+ *
+ * @return {promise}
+ *
+ * @example
+ * job.promise().then(function(metadata) {
+ *   // The job is complete.
+ * }, function(err) {
+ *   // An error occurred during the job.
+ * });
+ */
+Job.prototype.promise = function() {
+  var self = this;
+
+  return new self.Promise(function(resolve, reject) {
+    self
+      .on('error', reject)
+      .on('complete', function(metadata) {
+        resolve([metadata]);
+      });
+  });
+};
+
+/**
  * Begin listening for events on the job. This method keeps track of how many
  * "complete" listeners are registered and removed, making sure polling is
  * handled automatically.
--- a/packages/bigquery/test/job.js
+++ b/packages/bigquery/test/job.js
@@ -50,7 +50,8 @@ var fakeUtil = Object.keys(util).reduce(function(fakeUtil, methodName) {
 
 describe('BigQuery/Job', function() {
   var BIGQUERY = {
-    projectId: 'my-project'
+    projectId: 'my-project',
+    Promise: Promise
   };
   var JOB_ID = 'job_XYrk_3z';
   var Job;
@@ -271,6 +272,45 @@ describe('BigQuery/Job', function() {
       };
 
       job.getQueryResultsStream().done();
+    });
+  });
+
+  describe('promise', function() {
+    beforeEach(function() {
+      job.startPolling_ = util.noop;
+    });
+
+    it('should return an instance of the localized Promise', function() {
+      var FakePromise = job.Promise = function() {};
+      var promise = job.promise();
+
+      assert(promise instanceof FakePromise);
+    });
+
+    it('should reject the promise if an error occurs', function() {
+      var error = new Error('err');
+
+      setImmediate(function() {
+        job.emit('error', error);
+      });
+
+      return job.promise().then(function() {
+        throw new Error('Promise should have been rejected.');
+      }, function(err) {
+        assert.strictEqual(err, error);
+      });
+    });
+
+    it('should resolve the promise on complete', function() {
+      var metadata = {};
+
+      setImmediate(function() {
+        job.emit('complete', metadata);
+      });
+
+      return job.promise().then(function(data) {
+        assert.deepEqual(data, [metadata]);
+      });
     });
   });
 
--- a/packages/common/src/grpc-operation.js
+++ b/packages/common/src/grpc-operation.js
@@ -134,6 +134,23 @@ GrpcOperation.prototype.cancel = function(callback) {
 };
 
 /**
+ * Wraps the `complete` and `error` events in a Promise.
+ *
+ * @return {promise}
+ */
+GrpcOperation.prototype.promise = function() {
+  var self = this;
+
+  return new self.Promise(function(resolve, reject) {
+    self
+      .on('error', reject)
+      .on('complete', function(metadata) {
+        resolve([metadata]);
+      });
+  });
+};
+
+/**
  * Begin listening for events on the operation. This method keeps track of how
  * many "complete" listeners are registered and removed, making sure polling is
  * handled automatically.
--- a/packages/common/src/service-object.js
+++ b/packages/common/src/service-object.js
@@ -71,6 +71,7 @@ function ServiceObject(config) {
   this.createMethod = config.createMethod;
   this.methods = config.methods || {};
   this.interceptors = [];
+  this.Promise = this.parent.Promise;
 
   if (config.methods) {
     var allMethodNames = Object.keys(ServiceObject.prototype);
--- a/packages/common/src/service.js
+++ b/packages/common/src/service.js
@@ -60,10 +60,7 @@ function Service(config, options) {
   this.packageJson = config.packageJson;
   this.projectId = options.projectId;
   this.projectIdRequired = config.projectIdRequired !== false;
-
-  if (options.promise) {
-    util.setPromiseOverride(options.promise);
-  }
+  this.Promise = options.promise || Promise;
 }
 
 /**
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -51,8 +51,6 @@ var errorMessage = format([
   path: '/docs/guides/authentication'
 });
 
-var PromiseOverride;
-
 var missingProjectIdError = new Error(errorMessage);
 
 util.missingProjectIdError = missingProjectIdError;
@@ -656,7 +654,14 @@ function promisify(originalMethod) {
       return originalMethod.apply(context, args);
     }
 
-    var PromiseCtor = PromiseOverride || Promise;
+    var PromiseCtor = Promise;
+
+    // Because dedupe will likely create a single install of
+    // @google-cloud/common to be shared amongst all modules, we need to
+    // localize it at the Service level.
+    if (context && context.Promise) {
+      PromiseCtor = context.Promise;
+    }
 
     return new PromiseCtor(function(resolve, reject) {
       args.push(function() {
@@ -708,15 +713,3 @@ function promisifyAll(Class, options) {
 }
 
 util.promisifyAll = promisifyAll;
-
-/**
- * Allows user to override the Promise constructor without the need to touch
- * globals. Override should be ES6 Promise compliant.
- *
- * @param {promise} override
- */
-function setPromiseOverride(override) {
-  PromiseOverride = override;
-}
-
-module.exports.setPromiseOverride = setPromiseOverride;
--- a/packages/common/test/grpc-operation.js
+++ b/packages/common/test/grpc-operation.js
@@ -47,7 +47,9 @@ var FakeGrpcServiceObject = createFake(GrpcServiceObject);
 var FakeGrpcService = createFake(GrpcService);
 
 describe('GrpcOperation', function() {
-  var FAKE_SERVICE = {};
+  var FAKE_SERVICE = {
+    Promise: Promise
+  };
   var OPERATION_ID = '/a/b/c/d';
 
   var GrpcOperation;
@@ -146,6 +148,45 @@ describe('GrpcOperation', function() {
       };
 
       grpcOperation.cancel();
+    });
+  });
+
+  describe('promise', function() {
+    beforeEach(function() {
+      grpcOperation.startPolling_ = util.noop;
+    });
+
+    it('should return an instance of the localized Promise', function() {
+      var FakePromise = grpcOperation.Promise = function() {};
+      var promise = grpcOperation.promise();
+
+      assert(promise instanceof FakePromise);
+    });
+
+    it('should reject the promise if an error occurs', function() {
+      var error = new Error('err');
+
+      setImmediate(function() {
+        grpcOperation.emit('error', error);
+      });
+
+      return grpcOperation.promise().then(function() {
+        throw new Error('Promise should have been rejected.');
+      }, function(err) {
+        assert.strictEqual(err, error);
+      });
+    });
+
+    it('should resolve the promise on complete', function() {
+      var metadata = {};
+
+      setImmediate(function() {
+        grpcOperation.emit('complete', metadata);
+      });
+
+      return grpcOperation.promise().then(function(data) {
+        assert.deepEqual(data, [metadata]);
+      });
     });
   });
 
--- a/packages/common/test/service-object.js
+++ b/packages/common/test/service-object.js
@@ -109,6 +109,19 @@ describe('ServiceObject', function() {
       assert.strictEqual(typeof serviceObject.create, 'function');
       assert.strictEqual(serviceObject.delete, undefined);
     });
+
+    it('should localize the Promise object', function() {
+      var FakePromise = function() {};
+      var config = extend({}, CONFIG, {
+        parent: {
+          Promise: FakePromise
+        }
+      });
+
+      var serviceObject = new ServiceObject(config);
+
+      assert.strictEqual(serviceObject.Promise, FakePromise);
+    });
   });
 
   describe('create', function() {
--- a/packages/common/test/service.js
+++ b/packages/common/test/service.js
@@ -150,15 +150,15 @@ describe('Service', function() {
       assert.strictEqual(service.projectIdRequired, true);
     });
 
-    it('should call setPromiseOverride when promise is set', function(done) {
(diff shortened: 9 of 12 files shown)


stephenplusplus:
Go team promises, go!

README.md
```diff
@@ -1,5 +1,5 @@
 # Google Cloud Node.js Client
-> Node.js idiomatic client for [Google Cloud Platform](https://cloud.google.com/) services.
+> Node.js idiomatic client for [Google Cloud Platform](https://cloud.google.com/) services, now with more promises!
```
jgeewax: :)

jasonswearingen:
re:
> "This PR aims to remove callbacks and replace them with Promises."

if possible you can allow [optional] callbacks and return Promises, as that will help users with migration.

stephenplusplus:
@jasonswearingen -- we talked about this starting here: https://github.com/GoogleCloudPlatform/gcloud-node/issues/551#issuecomment-153175390 and landed on just promises.

stephenplusplus:
@callmehiphop I think we should update this with the state of the world now. We are no longer removing callback support, right?

callmehiphop:
@stephenplusplus that is correct, the plan is now to move streams behind a `stream` property which will then allow us to safely return a promise if the callback is omitted.

```js
// stream mode
bucket.getFiles({ stream: true })
  .on('data', function() {});

// promise mode
bucket.getFiles()
  .then(function() {});

// callback mode
bucket.getFiles(function() {});
```

I'll update the PR description accordingly.

zixia:
It's so great to see this PR after read the long ISSUE #551 

Do we have a date of when the gcloud-node with promise support will be published on NPM? I can't waiting for it. :+1:

jgeewax:
@callmehiphop , I'm a bit worried about the `{ stream: true }` syntax. Is there anyway we can make this a bit more readable, maybe...

```js
// Promise mode
bucket.getFiles()
  .then(function(files) { });

// Maybe we can let people create streams from promises (?)
bucket.getFiles().createStream()
  .on('data', function(file) { });

// Or if we can make it even shorter, promises also act like streams?
bucket.getFiles()
  .on('data', function(file) { });
```

callmehiphop:
@zixia I share your enthusiasm! Hopefully we'll get Promises published within the next couple of months.

@jgeewax I can't help but agree that `{ stream: true }` isn't the most elegant solution. I originally wanted to do something similar to your second example, but it meant that we had to streamify our methods at object creation time, which I think both @stephenplusplus and I didn't feel was very practical. 

The third example is pretty interesting, I'm not aware of any objects that inherit from both Promises and Streams. However, it wouldn't be very difficult to concat the stream data together and resolve via Promises - there are already several packages that do this.

@stephenplusplus what are your feelings on decorating the stream object we return with a `then` method that essentially returns a promise and calls `concat-stream` under the hood?

jgeewax:
@callmehiphop : The only issue I can think of here is what happens when you consume stuff with a stream and then decide you want a promise?

```js
bucket.getFiles().on('data', function(file) { }).then(function(files) { });
```

What should that do? Error somewhere? or is it technically valid?

callmehiphop:
I just tested that, technically it's valid and does not throw an error. We could unbind events if a user adds a promise, assuming that's a concern.

stephenplusplus:
> what are your feelings on decorating the stream object we return with a then method that essentially returns a promise and calls concat-stream under the hood?

This?

```js
typeof bucket.getFiles(callback) // undefined
typeof bucket.getFiles() // Stream
typeof bucket.getFiles().then() // Promise
```

Aren't there more Promise methods, like `catch` (maybe others?). Are those available as well via `bucket.getFiles().catch()`?

```js
bucket.getFiles()
  .on('data', function(file) {})
  .then(function(files) {});
```

I think this could definitely lead to confusion, but where there is flexibility, that's a known side effect. This is up to @callmehiphop's intended implementation, but it could either start two requests to fetch fresh data, or fight with each other for which stream (the one the user is holding or the internal `concat` stream we run when it's a promise) to receive each result from the API.

callmehiphop:
> Aren't there more Promise methods, like catch (maybe others?). Are those available as well via bucket.getFiles().catch()?

Currently the only methods on the Promise prototype chain are `then` and `catch`. However `catch` is pretty much just a convenience method for calling `then` with a null value for the success callback.

That does raise a valid concern that this approach would not be very future proof if additional methods were introduced.

callmehiphop:
> but it could either start two requests to fetch fresh data

That would be bad, the idea would be to re-use the already existing stream.

> or fight with each other for which stream (the one the user is holding or the internal concat stream we run when it's a promise) to receive each result from the API.

Technically the results would be emitted to both, so I'm not sure I understand how they would be fighting with each other?

If it helps to illustrate how it works, locally (for testing) I added this to the `runAsStream_` method.

```js
  stream.then = function(success, failure) {
    return new Promise(function(resolve, reject) {
      stream.on('error', reject).pipe(concat(resolve));
    }).then(success, failure);
  };
```

stephenplusplus:
Sure it would be an extra request, but since the user is asking for a stream and a promise, it might be expected.

I believe it would conflict when more complicated stream use cases are considered, like the user isn't just draining a stream with on(data) listeners, but manipulating or throttling the incoming messages.

callmehiphop:
I think trying to accommodate multiple interfaces in this fashion is bad idea. Which unfortunately doesn't make any of those examples very practical for us. An alternative that we haven't really discussed is making a separate method for the streaming versions

```js
// callback
bucket.getFiles(function() {});

// promise
bucket.getFiles().then(function() {});

// stream
bucket.getFilesStream().on('data', function(file) {});
```

@stephenplusplus @jgeewax not sure what your feelings on an approach like this are - IMO the only real downside here is that the documentation would not be consolidated like it is today, but that might actually be a selling point since the user would no longer have to search through several examples to see what the streaming syntax looks like.

stephenplusplus:
I do like the docs simplification part. What about declaring up front, at
the time of requiring gcloud, that we want the streams api or the promise
of? Then the methods don't have to support both. If the user needs both in
the same app (really, though?), they could re-instantiate gcloud or at the
service level, override their global preference.

jgeewax:
I was going to suggest this one also, but had trouble with the `sS`, and figured that since we combine callbacks and streams with the same API call, it seems pretty reasonable to do the same with promises and streams. I also assumed that if we were going to do this, it seemed kind of reasonable to offer the `.createStream()` method...

@stephenplusplus : I'm not a huge fan of forcing folks to determine their usage at the start... that seems... scary.

What if we tried doing this the same that fs does something like this?

```js
// callback
bucket.getFiles(function() {});

// promise
bucket.getFiles().then(function() {});

// stream
gcloud.createReadStream(bucket.getFiles())
  .on('data', function(file) { });
```

callmehiphop:
> since we combine callbacks and streams with the same API call, it seems pretty reasonable to do the same with promises and streams. 

When dealing with callbacks and Streams it's really obvious what a user wants, since they have to supply us with the actual callback. With Promises and Streams however, we don't have anything that allows us to distinguish one from the other - hence why we were suggesting `{ stream: true }`.

> I also assumed that if we were going to do this, it seemed kind of reasonable to offer the .createStream() method...

This is a bit more feasible than trying to combine Streams with Promises, however I think we might come across some of the same pitfalls. This does have the added benefit of not stubbing Promise methods though.

From a technical standpoint, I see two options to achieve this:

1. Create a request and return a promise, if `createStream()` is called, abort the previous request made and create a new request with a stream. I think the issue here is that we would potentially be making multiple requests, which is more $ for our users.

2. Create the stream up front and return a promise that resolves via `stream-concat`ing under the hood. When `createStream()` is called, we simply return the pre-existing stream. I think this might have the potential to get weird in the same ways that combining Streams and Promises got weird, in that the user could still use both APIs simultaneously.

e.g.

```js
var promise = bucket.getFiles();
var stream = promise.createStream();

promise.then(function(files) {});
stream.on('data', function(file) {});
```

We could `unpipe` the stream (I think) so that the promise never resolves, but I think that might be even more confusing.

> What if we tried doing this the same that fs does

IMO this would have to be implemented identically to `createStream()`, with some kind of Promise/Stream hybrid.

As for `getFilesStream` we could rename it to be something more intuitive `createFileStream` and simply reference it in `getFiles` documentation. This has the added benefit of being able to live on the prototype chain and giving us a clear indication of what the developer is trying to accomplish.

stephenplusplus:
@jgeewax what's scary about the `streams: true` option?

I think mixing streams and promises is confusing... even if we figure out how to solve it, the user has to figure out what we figured out, y'know what I mean?

Here's my personal preference of what is the least intrusive on the user, and the most predictable:

  1. require('gcloud')({ streams: true }) (or some other name for basically "I don't need promises") -- promises will be the default
  2. bucket.getFiles({ stream: true }) -- @jgeewax why don't you like this option?
  3. bucket.getFilesStream() -- regarding this naming, I'm totally fine with smushing s's. It's comforting for a user to know that "all I have to do is append 'Stream' and it'll work", without having to relearn our API with a new method or combinator

callmehiphop:
> require('gcloud')({ streams: true }) (or some other name for basically "I don't need promises") -- promises will be the default

I'm not a big fan of this. Lets assume a dev would want to use both streams and promises in the same app, they would either have to go back and forth between setting that flag or manage multiple objects that are more or less the same. I think this might actually be more confusing than trying to make a Stream/Promise hybrid.

stephenplusplus:
Maybe it can be treated as a global setting of `stream: true`, where the most local setting is on the method itself. So it's like saying, "I'm not using promises this time... or next time... or the time after that... so let me just tell you one time I'm not going to use the promise option... ever."

jasonswearingen:
I skimmed the recent posts, and I think  an explicit ```somemethodStream()``` is the way to go if someone needs streams.  It's a bad idea to inject streaming features/functions into Promises.   Separation of concerns!

also, I can imagine a scenario where you mostly use promises, but in some cases need to use streams.  For example:  I use promises with datastore right now (a wrapper I made) and it works great, except if I have  a datastore query that returns tons of results.   for that I'd use streams.   

regarding some global ```stream:true``` flag, if I'm understanding it correctly,  that sounds like a bad idea to me as it effectively changes the function signatures based on a runtime variable.  I know javascript isn't explicitly typed, but this seems like it's abusing that "feature".

stephenplusplus:
Thanks, @jasonswearingen. I think we should sit back until we have an implentation to review and let @callmehiphop get started on his ideal vision. This is his baby, and I trust him fully to deliver.

c0b:
for how long can we see a Promise API?

callmehiphop:
@c0b I don't think this is our top priority at the moment, so it's hard to nail down an accurate estimate, but I would guess sometime within the next several months.

jmdobry:
Looking at the implementation of `gax-nodejs`, I'm thinking there would be a rather unfortunate loss of information/loss of functionality if the hand-written methods were to return a Promise, as opposed to returning the event emitter. For example:

```js
var Language = require('@google-cloud/language');

var language = Language();
var api = Language.v1beta1();

var document = {
  type: 'PLAIN_TEXT',
  content: 'Hello, world!'
};

// The hand-written method returns just a promise
var promise = language.detectSentiment(document.content);

// gax-nodejs returns an event emitter
var apiRequest = api.analyzeSentiment(document);

try {
  // uh oh, how to I cancel the request
  promise.cancel();
} catch (err) {
  // Uncaught TypeError: promise.cancel is not a function
}

// Nice, this works
apiRequest.cancel();

promise
  .then((sentiment) => {
    // ...
  })
  .catch((err) => {
    // ...
  });

// This works today
apiRequest.result
  .then((response) => {
    // ...
  })
  .catch((err) => {
    // ...
  });
```

I'm thinking the hand-written methods should just return the event emitter returned by `gax-nodejs`:

```js
var languageRequest = language.detectSentiment(document.content);
var apiRequest = api.analyzeSentiment(document);

// Yay, I can cancel the request!
languageRequest.cancel();

// Nice, this works
apiRequest.cancel();

// Access the promise on the "result" property
languageRequest.result
  .then((sentiment) => {
    // ...
  })
  .catch((err) => {
    // ...
  });

// This works today
apiRequest.result
  .then((response) => {
    // ...
  })
  .catch((err) => {
    // ...
  });
```

Use the event emitter in combination with callback-style:

```js
var languageRequest = language.detectSentiment(document.content, function (err, sentiment) {
  // ...
});
var apiRequest = api.analyzeSentiment(document, function (err, response) {
  // ...
});

// Yay, I can cancel the request!
languageRequest.cancel();

// Nice, this works
apiRequest.cancel();
```

For those methods that optionally return a stream, the method will have to be split into two methods:

`Table#getRows` => `Table#getRows` and `Table#getRowsStream`:

```js
// callback-style
table.getRows(function (err, rows) {
  // ...
});

// promise-style
table.getRows().result.then(function (rows) {
  // ...
}).catch(function (err) {
  // ...
});

// stream
table.getRowsStream()
  .on('error', console.error)
  .on('data', function (row) {})
  .on('end', function () {
    // All rows have been retrieved.
  });
```

stephenplusplus:
@callmehiphop can you update the TODOs to outline our roll out plan?

Also, just throwing these out:

  - Where and how do we explain a user has multiple choices to run this? Yes, the method examples show it, but we might want to explain more globally that our library supports promises and callbacks. Maybe the README? An Overview doc?
  - Should we update each API's README examples to show promises in use?

callmehiphop:
> can you update the TODOs to outline our roll out plan?

Done!

> Where and how do we explain a user has multiple choices to run this?

I think obviously we should emphasize this in the release notes, otherwise I'm not too sure.. the main README is fine. I've updated the param list on the umbrella package as well showing how to integrate Bluebird, so it should also be available on the default docs page. Otherwise I'm not sure where else it fits?

> Should we update each API's README examples to show promises in use?

Yar, I'll update the READMEs as I create PRs for each service.

callmehiphop:
Promises have been released!

sebelga:
great job guys! thanks!

kevhill:
This is wonderful, just what I was looking for. I bet there are others who were looking for this but never found it because 'Promise' or '.then' are nowhere to be found in the readme.md and all of those examples use the callback style.

stephenplusplus:
We'd like to fix this as well: #1699, #2167 -- PRs are very welcome :)

kevhill:
Well, let me make sure I can use it correctly, then I'll give it a shot ;)

PR #1147, opened 2016-03-02
storage: delete versioned files

### LINKED ISSUE #968: Lingering buckets
Occasionally I'll create a bucket for testing purposes and when I go to delete it, I'll get an error stating that the bucket can't be deleted because it contains files, if I attempt to list the files within it there aren't any. I've also seen this occur without modifying any of the ACLs.

I'm pretty sure this is an upstream error, however I'm opening this issue to verify that it isn't our client and to also raise visibility on this issue.

stephenplusplus:
Fixes #968 (possibly)

I can't positively say all of the ghost buckets had versioning enabled... but if they did, the explanation for not being able to delete them is we didn't delete old versions of the files. That makes sense to me!

In the after hook for the Storage system tests, this will now get all versions of a file from each bucket, so they will be deleted.

### DIFF
--- a/system-test/storage.js
+++ b/system-test/storage.js
@@ -71,22 +71,16 @@ describe('storage', function() {
         // anymore.
         var CONSISTENCY_DELAY_MS = 250;
 
-        bucket.deleteFiles(function(err) {
+        bucket.deleteFiles({
+          versions: true
+        }, function(err) {
           if (err) {
             callback(err);
             return;
           }
 
           setTimeout(function() {
-            bucket.delete(function() {
-              // Ignoring this error:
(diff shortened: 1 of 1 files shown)


system-test/storage.js
```diff
@@ -71,22 +71,16 @@ describe('storage', function() {
...
-            bucket.delete(function() {
-              // Ignoring this error:
-              // https://github.com/GoogleCloudPlatform/gcloud-node/issues/968
-              // if (err) {
-              //   callback(err);
-              //   return;
-              // }
-              callback();
-            });
+            bucket.delete(done);
```
stephenplusplus: Shoot, this was supposed to be `callback`, not `done` :(

PR #1150, opened 2016-03-02
Revert "storage: delete versioned files"

callmehiphop:
Reverts GoogleCloudPlatform/gcloud-node#1147


stephenplusplus:
I think it's probably easier just to change `done` to `callback` :)

callmehiphop:
@stephenplusplus All I had to do is hit `revert` and it did all the work for me :)

PR #1151, opened 2016-03-02
Promisified Prediction API.

jmdobry:
The purpose of this PR is to show how to promisify the `gcloud-node` API while still maintaining support for callbacks. See #551 

`npm test` succeeds for me. I did not try the system tests (but I can, depending on how serious we are about this pull request).
# Summary of changes
## Breaking API changes
- `Service#request`, the Prediction API methods, and the ServiceObject methods now only return a single errback value (which is more idiomatic, as opposed to also passing random positional success arguments to the callback). The positional argument (e.g. `apiResponse`) which was previously passed to the callback as a separate argument is now attached as a property of the single `err` argument which is now the only value returned in the event of an error/rejection. Now, in the event of an error/rejection, `err` is all you need to inspect.
## Backwards compatible API changes
- `Service#request` now returns a promise, but makes use of `.asCallback` to maintain support for callbacks.
- The Prediction API methods now return promises, which make use of `.asCallack` to maintain support for callbacks.
- The `ServiceObject` methods now return promises, which make use of `.asCallback` to maintain support for callbacks.
- Added dependency on `bluebird`
## Other
- Refactored Prediction API and ServiceObject unit tests, which frequently mock `Service#request`, to take into account the new promisified methods.
- Made a second version of a number of the refactored tests. One version interacts with the API via promises, and the other via callbacks, to show that the API now supports both styles.

**Note:** Many methods return multiple values. If you use callbacks, the multiple values will still be passed to your callback as multiple positional arguments. But if you use Promises, `.then` will receive an array containing the return values, so make use of `.spread` instead.

**Note 2:** My first commit did not do anything about streams, which still need to be addressed.

**Note 3:** I'm not sure I completely promisified `ServiceObject`, as I haven't spent enough time studying the code. In fact I'm sure I didn't promisify it completely, because I would have expected to have needed to refactor more tests than I did.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     "array-uniq": "^1.0.2",
     "arrify": "^1.0.0",
     "async": "^1.4.2",
+    "bluebird": "^3.3.3",
     "camelize": "^1.0.0",
     "concat-stream": "^1.5.0",
     "create-error-class": "^2.0.1",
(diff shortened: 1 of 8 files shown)


stephenplusplus:
Awesome, thanks for doing this!

callmehiphop:
@jmdobry thanks for taking the time to put this together! I started to implement Promises via https://github.com/GoogleCloudPlatform/gcloud-node/pull/1142, but with the recent decision to continue callback support I was a little thrown off/set back.

@stephenplusplus and I have been having a lot of discussions about promises and how we should go about implementing them into gcloud-node. One thing we decided on was not to use Bluebird, but instead a lightweight polyfill, although `.spread()` does make Bluebird very appealing.

package.json
```diff
@@ -86,6 +86,7 @@
     "array-uniq": "^1.0.2",
     "arrify": "^1.0.0",
     "async": "^1.4.2",
+    "bluebird": "^3.3.3",
```
Glenn230593: hetLeave a 

 note

stephenplusplus:
Thanks again for the PR @jmdobry. As @callmehiphop said, he's working on adding support in #1142 -- we'll be sure to come back and reference the changes here as we go.

PR #1157, opened 2016-03-08
Patch gRPC leaks in unit test

### LINKED ISSUE #1140: Testing failures in Travis

stephenplusplus:
Fixes #1140 

Not sure this works... Just seeing what Travis thinks. UPDATE: :smile:


stephenplusplus:
I'm feeling pretty good about this one. @callmehiphop wdyt?

Issue #1073, opened 2016-01-19
compute: missing resources


callmehiphop:
RE: #1071 

As it is today, we're missing the following resource types:
- [x] [Autoscalers](https://cloud.google.com/compute/docs/reference/latest/#Autoscalers) (@stephenplusplus) PR: #1137
- [x] [BackendServices](https://cloud.google.com/compute/docs/reference/latest/#BackendServices) (@stephenplusplus) PR: #1159
- [x] [ForwardingRules](https://cloud.google.com/compute/docs/reference/latest/#ForwardingRules) (@stephenplusplus) PR: #1191
- [x] [Http(s)HealthChecks](https://cloud.google.com/compute/docs/reference/latest/#HttpHealthChecks) (@stephenplusplus) PR: #1276
- [ ] [Images](https://cloud.google.com/compute/docs/reference/latest/#Images)
- [ ] [InstanceGroupManagers](https://cloud.google.com/compute/docs/reference/latest/#InstanceGroupManagers)
- [x] [InstanceGroups](https://cloud.google.com/compute/docs/reference/latest/#InstanceGroups) - #1071 (@stephenplusplus) PR: #1320
- [ ] [InstanceTemplates](https://cloud.google.com/compute/docs/reference/latest/#InstanceTemplates) - #1071 
- [ ] [Instances](https://cloud.google.com/compute/docs/reference/latest/#Instances)
- [ ] [Licenses](https://cloud.google.com/compute/docs/reference/latest/#Licenses)
- [x] [MachineTypes](https://cloud.google.com/compute/docs/reference/latest/#MachineTypes)
- [x] [Projects](https://cloud.google.com/compute/docs/reference/latest/#Projects) PR: #2506
- [ ] [Routes](https://cloud.google.com/compute/docs/reference/latest/#Routes)
- [ ] [SslCertificates](https://cloud.google.com/compute/docs/reference/latest/#SslCertificates)
- [x] [Subnetworks](https://cloud.google.com/compute/docs/reference/latest/#Subnetworks)
- [ ] [TargetHttp(s)Proxies](https://cloud.google.com/compute/docs/reference/latest/#SslCertificates)
- [ ] [TargetPools](https://cloud.google.com/compute/docs/reference/latest/#TargetPools)
- [ ] [TargetVpnGateways](https://cloud.google.com/compute/docs/reference/latest/#TargetVpnGateways)
[...]

PR #1159, opened 2016-03-09
compute: support BackendServices

stephenplusplus:
For #1073 

This implements Backend Services, known here as "services".
#### To Dos
- [x] Tests
  - [x] System
  - [x] Unit
    - [x] compute/index
    - [x] compute/service

``` js
var service = compute.service('service-name');

service.getHealth({
  name: 'instance-group-name',
  zone: 'us-central1-a'
}, function(err, status) {
  status = [
    {
      ipAddress: '...',
      instance: '...',
      healthState: '...',
      port: '...'
    }
  ]
});
```

### DIFF
--- a/docs/toc.json
+++ b/docs/toc.json
@@ -71,6 +71,9 @@
       "title": "Region",
       "type": "compute/region"
     }, {
+      "title": "Service",
+      "type": "compute/service"
+    }, {
       "title": "Snapshot",
       "type": "compute/snapshot"
     }, {
--- a/lib/compute/index.js
+++ b/lib/compute/index.js
@@ -56,6 +56,12 @@ var Region = require('./region.js');
 var Service = require('../common/service.js');
 
 /**
+ * @type {module:compute/service}
+ * @private
+ */
+var ServiceClass = require('./service.js');
+
+/**
  * @type {module:compute/snapshot}
  * @private
  */
@@ -302,6 +308,70 @@ Compute.prototype.createNetwork = function(name, config, callback) {
     operation.metadata = resp;
 
     callback(null, network, operation, resp);
+  });
+};
+
+/**
+ * Create a backend service.
+ *
+ * @resource [Backend Services Overview]{@link https://cloud.google.com/compute/docs/load-balancing/http/backend-service}
+ * @resource [BackendServices: insert API Documentation]{@link https://cloud.google.com/compute/docs/reference/v1/backendServices/insert}
+ *
+ * @param {string} name - Name of the backend service.
+ * @param {object} config - See a
+ *     [BackendService resource](https://cloud.google.com/compute/docs/reference/v1/backendServices#resource).
+ * @param {function=} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request.
+ * @param {module:compute/service} callback.service - The created Service
+ *     object.
+ * @param {module:compute/operation} callback.operation - An operation object
+ *     that can be used to check the status of the request.
+ * @param {object} callback.apiResponse - The full API response.
+ *
+ * @example
+ * var config = {
+ *   backends: [
+ *     {
+ *       group: 'URL of an Instance Group resource'
+ *     }
+ *   ],
+ *   healthChecks: [
+ *     'URL of an HTTP/HTTPS health check resource'
+ *   ]
+ * };
+ *
+ * function callback(err, service, operation, apiResponse) {
+ *   // `service` is a Service object.
+ *
+ *   // `operation` is an Operation object that can be used to check the status
+ *   // of network creation.
+ * }
+ *
+ * gce.createService('new-service', config, callback);
+ */
+Compute.prototype.createService = function(name, config, callback) {
+  var self = this;
+
+  var body = extend({}, config, {
+    name: name
+  });
+
+  this.request({
+    method: 'POST',
+    uri: '/global/backendServices',
+    json: body
+  }, function(err, resp) {
+    if (err) {
+      callback(err, null, null, resp);
+      return;
+    }
+
+    var service = self.service(name);
+
+    var operation = self.operation(resp.name);
+    operation.metadata = resp;
+
+    callback(null, service, operation, resp);
   });
 };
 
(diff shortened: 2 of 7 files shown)


stephenplusplus:
@callmehiphop I'm going to start on the unit tests, but let me know if you catch anything weird with the code.

stephenplusplus:
<kbd>don't merge</kbd> lifted!

lib/compute/index.js
```diff
@@ -306,6 +312,70 @@ Compute.prototype.createNetwork = function(name, config, callback) {
...
+ *
+ * @resource [Backend Services Overview]{@link https://cloud.google.com/compute/docs/load-balancing/http/backend-service}
+ * @resource [BackendServices: insert API Documentation]{@link https://cloud.google.com/compute/docs/reference/v1/backendServices/insert}
+ *
+ * @param {string} name - Name of the backend service.
+ * @param {object} config - See a
+ *     [BackendService resource](https://cloud.google.com/compute/docs/reference/v1/backendServices#resource).
+ * @param {function=} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request.
+ * @param {module:compute/network} callback.network - The created Network
```
callmehiphop: Is this correct? I'm not familiar with the upstream counterpart, but if it does create a network I would think this method should be titled `createNetwork`.

stephenplusplus: Copy/paste error. Fixed :white_check_mark:

lib/compute/index.js
```diff
@@ -1090,6 +1160,111 @@ Compute.prototype.getRegions = function(options, callback) {
...
+ *     - **`comparison`**: the comparison operator, `eq` (equal) or `ne`
+ *       (not equal)
+ *     - **`filterString`**: the string to filter to. For string fields, this
+ *       can be a regular expression.
+ * @param {number} options.maxResults - Maximum number of snapshots to return.
+ * @param {string} options.pageToken - A previously-returned page token
+ *     representing part of the larger set of results to view.
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request.
+ * @param {module:compute/snapshot} callback.snapshots - Snapshot objects from
```
callmehiphop: I think this may be another copypasta error!

lib/compute/service.js
```diff
@@ -0,0 +1,291 @@
...
+ */
+var ServiceObject = require('../common/service-object.js');
+
+/**
+ * @type {module:common/util}
+ * @private
+ */
+var util = require('../common/util.js');
+
+/*! Developer Documentation
```
callmehiphop: Looks like we forgot some documentation here :)

lib/compute/service.js
```diff
@@ -0,0 +1,291 @@
...
+ * var gce = gcloud.compute();
+ *
+ * var service = gce.service('service-name');
+ */
+function Service(compute, name) {
+  var methods = {
+    /**
+     * Create a backend service.
+     *
+     * @param {object} config - See {module:compute#createService}.
```
callmehiphop: Looks like we're missing the callback param stuff

lib/compute/service.js
```diff
@@ -0,0 +1,291 @@
...
+    /**
+     * Get a Service object if it exists.
+     *
+     * You may optionally use this to "get or create" an object by providing an
+     * object with `autoCreate` set to `true`. Any extra configuration that is
+     * normally required for the `create` method must be contained within this
+     * object as well.
+     *
+     * @param {options=} options - Configuration object.
+     * @param {boolean} options.autoCreate - Automatically create the object if
```
callmehiphop: Missing callback param here as well

callmehiphop:
Is there a docs preview for this?

stephenplusplus:
Docs preview up: http://stephenplusplus.github.io/gcloud-node/#/docs/master/compute/service

RE: the missing callback param docs, that was an intentional decision back when we introduced the famous "get or create" overhaul. While having multiple methods to get at the same data can be convenient, the consequence was having to write the *exact same documentation* in multiple places. That has some hazardous side effects if the code is updated with a docs change local to the source method, the branched methods that reference it would go out of date. The chosen path was simply to have a link back to the source method from the branched methods and let the example do a quick and dirty job of showing what will happen in your callback. If you have any ideas how we can improve on this, open an issue so we can :hash: it out!

callmehiphop:
Thanks for the preview! I think we need to add BackendServices to the site navigation via <kbd>toc.json</kbd>!

docs/toc.json
```diff
@@ -71,6 +71,9 @@
       "title": "Region",
       "type": "compute/region"
     }, {
+      "title": "Service",
+      "type": "compute/service"
+    }, {
```
stephenplusplus: @callmehiphop it's in here! Did I do something wrong? Also, is `implemented` still a thing?

callmehiphop: Oh weird, I don't see Service in the navigation. `implemented` went away. We version <kbd>toc.json</kbd> in the same fashion as the rest of the JSON/markdown.

callmehiphop:
This might not be the place for it, but are there are any JSDoc tags (aside from link) that would allow us to inherit the documentation from one method to another?

lib/compute/service.js
```diff
@@ -0,0 +1,300 @@
...
+};
+
+/**
+ * Get the most recent health check results.
+ *
+ * @resource [BackendServices: getHealth API Documentation]{@link https://cloud.google.com/compute/docs/reference/latest/backendServices/getHealth}
+ *
+ * @param {string|object} group - The fully-qualified URL of an Instance Group
+ *     resource.
+ * @param {string} - group.name - The name of the Instance Group resource.
```
callmehiphop: The first hyphen causes a formatting error - http://stephenplusplus.github.io/gcloud-node/#/docs/master/compute/service?method=getHealth

docs/toc.json
stephenplusplus: I like it! My local "deploy preview docs" script was wrong. It's in the sidebar in the preview now: http://stephenplusplus.github.io/gcloud-node/#/docs/master/compute/service

stephenplusplus:
Not that I know of. Possibly the callback can be a "type" that we can then reference? I'm not sure if that's real / a hack usage.

lib/compute/service.js
stephenplusplus: Phew, fixed.

PR #1165, opened 2016-03-13
Fixed: prediction/model#query does not work with Regression models

savethefails:
The `prediction#output` response from calling Trainedmodels:predict on a Regression model does not include an `outputMulti` attribute. This attribute is used to calculate the `results.score` return value from `prediction/model#query` but this calculation causes an error when predicting using Regression models.

The test for `prediction/model#query` "query success should return the outputValue as the winner" was not set up correctly so this bug wasn't caught. When the `outputMulti` attribute is removed from mock apiResponse the error seen when using gcloud-node with Regression models becomes apparent: `TypeError: Cannot read property 'sort' of undefined
      at lib/prediction/model.js:351:9
`
I've included the steps to update the test, and a solution to the problem (only calculating query result scores if `response.outputMulti` is present).

Thanks!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,3 +8,4 @@
 
 Google Inc.
 Anand Suresh
+Brett Bergmann
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -12,6 +12,7 @@
 # Keep the list alphabetically sorted.
 
 Burcu Dogan <jbd@google.com>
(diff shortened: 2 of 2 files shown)


stephenplusplus:
Thanks @savethefails! There were a couple linting issues, so I pushed manually (https://github.com/GoogleCloudPlatform/gcloud-node/commit/63f658619e884ddbeb36bb7f2896372fcd5cb9e4)

PR #1166, opened 2016-03-14
testing: add spellchecker

### LINKED ISSUE #1127: Add a spell checker to our lint task
Cont'd from https://github.com/GoogleCloudPlatform/gcloud-node/pull/1126#issuecomment-183034508

We have words written all over the place that might not be correct grammatically.

@callmehiphop found a tool called [write-good](https://github.com/btford/write-good) which we could add to our link task to catch such non-good writings. We would likely need a spell-checker as well, because it doesn't seem to do that.

We'd have to be precise in steering it only towards words and away from code. Sometimes they are mixed, such as our `@example`s. @callmehiphop, is there a good way to filter out only the text?

stephenplusplus:
Fixes #1127 

~~The [`spellchecker`](https://github.com/atom/node-spellchecker) dependency only works Node with >= v4, so it is an optional dependency.~~

Apparently versions >= 3.2 can crash on Macs, so Travis wasn't affected. I've downgraded to 3.1 and problem resolved.
#### To Dos
- [x] Spell check docs
- [ ] Spell check markdown files
- [x] Figure out Travis


jgeewax:
OK, so this does require correct spelling (or addition to the dictionary)?

jgeewax:
Is there any way to make this a common thing to be shared across libraries? Or not worth it?

stephenplusplus:
I think that's a good idea, but I'm not sure how difficult that would be. @callmehiphop wdyt?

jgeewax:
No problem - we can punt on that for later :)

stephenplusplus:
@callmehiphop Travis is getting an error that I can't get locally. Just wanted to make a note that I saw that. So whenever you're looking at this PR, please try locally as well and let me know if it's just me.

scripts/docs.js
```diff
@@ -275,12 +283,74 @@ globby('./lib/*{,/*}.js', { ignore: IGNORE }).then(function(files) {
       }
 
       var docs = parseDocs(file, contents, 0);
-      var outputFile = path.join(
-        './docs/json/master',
-        file.replace('/lib', '').replace('.js', '.json')
-      );
 
-      fs.writeFile(outputFile, JSON.stringify(docs), callback);
+      function writeFile() {
```
callmehiphop: Rather than lumping this all in at the end, could we just create a `spellcheck` function and reference it within the `formatHtml` function? Everything that we would want to spell check (classdesc, decription, example comments, etc.) is going to get passed through there. I think this would simplify the code a bit and allow us to avoid traversing the JSON.

stephenplusplus: I think we lose the context of what file is producing the words in that chain. Is there another place I can plug this in that would have that information available?

callmehiphop:
@stephenplusplus locally I'm not getting errors for 0.12.x or 0.4.x. One thing that kind of concerns me is that on Travis, spellchecker is running on 0.12, which it should not be.

callmehiphop:
Also, is this something we should be running our markdown files through as well?

stephenplusplus:
Taking a break on this one. The dictionary is getting quite huge and ridiculous.

stephenplusplus:
We merged the speller errors caught by this PR. I don't think there's a solid path forward for managing the custom dictionary, so I'm going to say this PR likely isn't going to make the cut. We can take it back to the drawing board in #1127, if not call it quits on the task altogether.

PR #1171, opened 2016-03-18
compute: add VM#setMetadata

stephenplusplus:
This implements `setMetadata` on a GCE VM.

### DIFF
--- a/lib/compute/vm.js
+++ b/lib/compute/vm.js
@@ -415,6 +415,59 @@ VM.prototype.reset = function(callback) {
 };
 
 /**
+ * Set the metadata for this instance.
+ *
+ * @resource [Instances: setMetadata API Documentation]{@link https://cloud.google.com/compute/docs/reference/v1/instances/setMetadata}
+ *
+ * @param {object} metadata - New metadata.
+ * @param {function=} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request.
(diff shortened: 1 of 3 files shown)


lib/compute/vm.js
```diff
@@ -415,6 +415,59 @@ VM.prototype.reset = function(callback) {
...
+ *   // `operation` is an Operation object that can be used to check the status
+ *   // of the request.
+ * });
+ */
+VM.prototype.setMetadata = function(metadata, callback) {
+  var self = this;
+
+  callback = callback || util.noop;
+
+  this.getMetadata(function(err, currentMetadata, apiResponse) {
```
callmehiphop: Do you think it would be worthwhile to add an `overwrite` option (or similar) to this? My rationale being that in some cases we might want to completely overwrite ALL the metadata.

stephenplusplus: That's actually what this does (overwrites, instead of extends). The upstream API is designed that way, so I thought that might be the most expected use case. We have to get the metadata to use the "fingerprint", so the setMetadata operation only succeeds if the fingerprint represents the current state of metadata.

callmehiphop: Ahh, ok, I misread that part. Thanks for the explanation :+1:

stephenplusplus: No problem, my bad for not explaining in the first place!

PR #1182, opened 2016-03-24
core: cache grpc services

### LINKED ISSUE #1164: Cache gRPC connections
Hello guys,

I use PubSub to pulish thousands of messages and a memory leak crashes my server.

Here is a Passenger screenshot using gcloud:
gcloud

Using request:
gcloud

Here is a simple repo to reproduce it: https://github.com/pierre-b/test-leak

Environnement:
- Ubuntu 14.04
- node 4.4.0
- Passenger 5.0.26
- gcloud 0.29.0

Did I miss something?
thanks for your help

stephenplusplus:
Fixes #1164 

@murgatroid99 when you have a chance, could you see if this is started off in the right direction? This makes a huge difference in memory usage, but I'm wondering if there are any gotchas I need to be aware of. Thanks!

### DIFF
--- a/lib/common/grpc-service.js
+++ b/lib/common/grpc-service.js
@@ -157,6 +157,7 @@ function GrpcService(config, options) {
   var service = this.service = config.service;
   var rootDir = googleProtoFiles('..');
 
+  this.activeServiceMap_ = new Map();
   this.protos = {};
 
   var protoServices = config.protoServices;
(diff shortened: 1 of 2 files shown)


stephenplusplus:
@callmehiphop this is safe for review & merge.

lib/common/grpc-service.js
```diff
@@ -217,21 +218,26 @@ GrpcService.prototype.request = function(protoOpts, reqOpts, callback) {
...
 
+  // Clean up gcloud-specific options.
+  delete reqOpts.autoPaginate;
+  delete reqOpts.autoPaginateVal;
+
+  var serviceMapKey = JSON.stringify(protoOpts);
+  var service = this.activeServiceMap_.get(serviceMapKey);
+
+  if (!service) {
+    service = new proto[protoOpts.service](
```
callmehiphop: Should we cache this once we've created it?

stephenplusplus: Yes :)

PR #1191, opened 2016-03-29
compute: support Forwarding Rules

stephenplusplus:
For #1073

This adds support for forwarding rules (regional and global).
#### To Dos
- [x] Docs
- [x] Tests
  - [x] System
  - [x] Unit
    - [x] index.js
    - [x] region.js
    - [x] rule.js

### DIFF
--- a/lib/compute/index.js
+++ b/lib/compute/index.js
@@ -50,6 +50,12 @@ var Operation = require('./operation.js');
 var Region = require('./region.js');
 
 /**
+ * @type {module:compute/rule}
+ * @private
+ */
+var Rule = require('./rule.js');
+
+/**
  * @type {module:common/service}
  * @private
  */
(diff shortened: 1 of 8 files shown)


stephenplusplus:
@callmehiphop ptal! I had some issues with the docs tests not having `rule` defined. That's just a heads-up for when Travis finishes, he'll likely be complaining about the same thing... could you see if I missed anything obvious that would fix that?

lib/compute/rule.js
```diff
@@ -0,0 +1,247 @@
...
+ * @type {module:common/util}
+ * @private
+ */
+var util = require('../common/util.js');
+
+/*! Developer Documentation
+ *
+ * @param {module:compute|module:compute/region} scope - The parent scope this
+ *     firewall rule belongs to.
+ * @param {string} name - Rule name.
```
callmehiphop: Looks like you forgot a `*/` here! I think adding that should fix the doc tests.

lib/compute/index.js
```diff
@@ -312,6 +318,96 @@ Compute.prototype.createNetwork = function(name, config, callback) {
 };
 
 /**
+ * Create a global forwarding rule.
+ *
+ * @resource [GlobalForwardingRule Resource]{@link https://cloud.google.com/compute/docs/reference/v1/globalForwardingRules#resource}
+ * @resource [GlobalForwardingRules: insert API Documentation]{@link https://cloud.google.com/compute/docs/reference/v1/globalForwardingRules/insert}
+ *
+ * @param {string} name - Name of the rule.
+ * @param {object} config - See a
```
callmehiphop: Should this say *"See the GlobalForwardingRule resource"*?

lib/compute/index.js
```diff
@@ -312,6 +318,96 @@ Compute.prototype.createNetwork = function(name, config, callback) {
...
+ *     Default: `TCP`.
+ * @param {string=} config.range - A single port or single contiguous port
+ *     range, ranging from low to high for which this forwarding rule matches.
+ *     Packets of the specified protocol sent to these ports will be forwarded
+ *     on to the appropriate target pool or target instance. If this field is
+ *     left empty, then the forwarding matches traffic for all ports for the
+ *     specified protocol.
+ * @param {string} config.target - The full or valid partial URL of the target
+ *     resource to receive the matched traffic. This target must be a global
+ *     `TargetHttpProxy` or `TargetHttpsProxy` resource.
```
callmehiphop: Are there any docs that we can link to for these resources?

lib/compute/rule.js
stephenplusplus: D'oh!

lib/compute/index.js
stephenplusplus: As opposed to "See *a* GlobalForwardingRule resource"?

lib/compute/index.js
stephenplusplus: Yes!

system-test/compute.js
```diff
@@ -24,7 +24,7 @@ var is = require('is');
 var env = require('./env.js');
 var Compute = require('../lib/compute/index.js');
 
-describe('Compute', function() {
+describe.only('Compute', function() {
```
callmehiphop: Don't forget about this lil' guy!

stephenplusplus: I can't believe it's 2016 and I'm still leaving `only`s in my code.

lib/compute/index.js
callmehiphop: Yar.. since we are redirecting the user to upstream docs, when I see *a*, it almost makes me assume there are several upstream docs called the same thing (which doesn't really make sense). Opposed to *the* which would kind of imply it's the definitive resource for GlobalForwardingRules.

Mostly a nitpick, so feel free to ignore if you disagree!

stephenplusplus: I think there's context confusion! It's not "resource" as in "See the resource for explaining what this is", it's "See the definition of this resource". It's really just borrowed lingo from the official docs (see https://cloud.google.com/compute/docs/reference/latest/globalForwardingRules#resource: "A ForwardingRule resource.")

If it was "The ForwardingRule resource", that would mean I'm speaking of a particular, existing ForwardingRule.

Issue #1192, opened 2016-03-30
Implement IAM API


stephenplusplus:
[IAM](https://cloud.google.com/iam/docs) is currently buried inside of our Pub/Sub code, but should be abstracted to be used with any API that supports it:
- [ ] [Compute](https://cloud.google.com/compute/docs/access/iam)
- [ ] ~~[Logging](https://cloud.google.com/logging/docs/access-control)~~ (not supported)
- [x] [Pub/Sub](https://cloud.google.com/pubsub/access_control)
- [ ] ~~[Resource Manager](https://cloud.google.com/resource-manager/docs/access-control-proj)~~ (not supported)
- [x] [Storage](https://cloud.google.com/storage/docs/iam)
  - [x] Bucket
  - [x] Object

stephenplusplus:
I haven't been able to get through to IAM with gRPC: https://github.com/stephenplusplus/grpc-usage-question/issues/1

// @jgeewax

stephenplusplus:
@jgeewax for the libraries that use gRPC, there are no IAM endpoints defined within each service's proto files. For an HTTP API like Compute, can you give me an example of a URL where you think I can find an IAM endpoint?

I've tried on an autoscaler to test, `https://www.googleapis.com/compute/v1/projects/nth-circlet-705/zones/us-central1-a/autoscalers/gcloud-tests-autoscaler-1468433795482/getIamPolicy` -- 404, but I'm not sure if that was a good guess for where I could find an IAM endpoint.

jgeewax:
I'm pretty sure that you use the iam_policy.proto file to send the request
with the right resource name, but to the pubsub endpoint... (
https://cloud.google.com/pubsub/reference/rpc/google.iam.v1 and
https://github.com/googleapis/googleapis/blob/master/google/iam/v1/iam_policy.proto
)

No?

stephenplusplus:
Pub/Sub is supported using IAM: https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/pubsub/iam.js. It's Logging that's been an issue.

[...]

PR #1193, opened 2016-03-30
iam: Implement support

stephenplusplus:
RE: #1192 

This pulls the IAM object outside of the Pub/Sub code it's previously been buried under.
#### Blockers
- [ ] https://github.com/stephenplusplus/grpc-usage-question/issues/1
#### To Dos
- [ ] Tests
  - [ ] System
    - [ ] IAM
    - [x] IAM via Pub/Sub
  - [ ] Unit
    - [ ] iam/index.js

### DIFF
--- a/lib/pubsub/subscription.js
+++ b/lib/pubsub/subscription.js
@@ -296,7 +296,7 @@ function Subscription(pubsub, options) {
    */
   var iam;
   Object.defineProperty(this, 'iam', {
-    get() {
+    get: function() {
       if (!iam) {
         iam = new IAM(this);
       }
--- a/lib/pubsub/topic.js
+++ b/lib/pubsub/topic.js
@@ -189,7 +189,7 @@ function Topic(pubsub, name) {
    */
   var iam;
   Object.defineProperty(this, 'iam', {
-    get() {
+    get: function() {
       if (!iam) {
         iam = new IAM(this);
       }


lib/pubsub/subscription.js
```diff
@@ -293,11 +293,17 @@ function Subscription(pubsub, options) {
    * //-
    * // Get the IAM policy for your subscription.
    * //-
-   * subscription.iam.getPolicy(function(err, policy) {
-   *   console.log(policy);
-   * });
    */
-  this.iam = new IAM(pubsub, this.name);
+  var iam;
+  Object.defineProperty(this, 'iam', {
```
stephenplusplus: This is so gRPC is lazily instantiated (for memory perf).

stephenplusplus:
@callmehiphop when you have a chance, can you try to track down what I might be doing wrong that causes [this error](https://travis-ci.org/stephenplusplus/gcloud-node/jobs/119604571#L3040) during the build?

lib/pubsub/subscription.js
```diff
@@ -293,11 +293,17 @@ function Subscription(pubsub, options) {
...
+    get() {
```
callmehiphop: This throws an error for me running tests against `v0.12.x`. Should we use the `get: function() {` syntax instead? :/

lib/pubsub/topic.js
```diff
@@ -187,7 +187,16 @@ function Topic(pubsub, name) {
    *   console.log(policy);
    * });
    */
-  this.iam = new IAM(pubsub, this.name);
+  var iam;
+  Object.defineProperty(this, 'iam', {
+    get() {
```
callmehiphop: RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1193/files#r58237170

lib/pubsub/subscription.js
stephenplusplus: Fixed.

stephenplusplus:
This PR has been somewhat nullified until IAM support is better supported. It may be worth pulling some code out of this PR when that happens.

PR #1198, opened 2016-04-04
datastore: fix broken system test

stephenplusplus:
We need to use an ancestor query to avoid unsuccessful queries due to eventual consistency.

@pcostell does this look right?


pcostell:
LGTM

PR #1201, opened 2016-04-04
datastore: fix runQuery end logic

### LINKED ISSUE #1200: Unable to query more than 300 entities after updating to 0.30.0
After updating from 0.29.0 to 0.30.0 one of my queries is now capped at returning 300 entities with no cursor. When I revert back to 0.29.0 it properly returns the limit of 1000. Any ideas why this might be happening?

```
var query = datastore.createQuery('Payment')
                        .filter('datemade', '>=', new Date(dateStart))
                        .filter('datemade', '<', new Date(dateEnd))
                        .limit(1000);

datastore.runQuery(
    query,
    function(error, entities, cursor) {
        if (error) {
            console.error(error);
        } else {
            console.log(entities.length);
            console.log(cursor);
        }
    }
);
```

stephenplusplus:
Fixes #1200
#### To Dos
- [x] Unit Tests

### DIFF
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -424,6 +424,8 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
  * });
  */
 DatastoreRequest.prototype.runQuery = function(query, callback) {
+  var self = this;
+
   var protoOpts = {
     service: 'Datastore',
     method: 'runQuery'
@@ -440,7 +442,7 @@ DatastoreRequest.prototype.runQuery = function(query, callback) {
     };
   }
 
-  this.request_(protoOpts, reqOpts, function(err, resp) {
+  function onResponse(err, resp) {
     if (err) {
       callback(err, null, null, resp);
       return;
@@ -453,19 +455,29 @@ DatastoreRequest.prototype.runQuery = function(query, callback) {
       entities = entity.formatArray(resp.batch.entityResults);
     }
 
-    var NOT_FINISHED_CODE = 'MORE_RESULTS_AFTER_LIMIT';
+    var notFinished = resp.batch.moreResults === 'NOT_FINISHED';
+    var moreResults = resp.batch.moreResults === 'MORE_RESULTS_AFTER_LIMIT';
 
-    if (resp.batch.moreResults === NOT_FINISHED_CODE) {
+    if (notFinished || moreResults) {
       var endCursor = resp.batch.endCursor;
       var offset = query.offsetVal === -1 ? 0 : query.offsetVal;
       var nextOffset = offset - resp.batch.skippedResults;
-
       nextQuery = extend(true, new Query(), query);
       nextQuery.start(endCursor).offset(nextOffset);
     }
 
+    if (notFinished) {
+      // Run the query again to make sure all of the requested entities are
+      // returned.
+      reqOpts.query = entity.queryToQueryProto(nextQuery);
+      self.request_(protoOpts, reqOpts, onResponse);
+      return;
+    }
+
     callback(null, entities, nextQuery, resp);
-  });
+  }
+
+  this.request_(protoOpts, reqOpts, onResponse);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


stephenplusplus:
@pcostell this breaks our system test: https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/system-test/datastore.js#L430

When I log the response after the first query, the `resp.batch.moreResults` value is "MORE_RESULTS_AFTER_LIMIT". The change from this PR prevents the block that creates a `nextQuery` from running. Should I check for both "FINISHED" and "MORE_RESULTS_AFTER_LIMIT"?

pcostell:
@stephenplusplus Based on your code, I think that if you set auto-pagination=False, you can then issue another request if `MORE_RESULTS_AFTER_LIMIT`. However, I'm not really convinced that this is a great API here (If you want 8 results, why are you only asking for a limit of 5)?

stephenplusplus:
autoPaginate is true by default, which means if you just want every result, `runQuery` will recursively call itself until the results are fully collected. The user's callback is called after with all of the results.

When it's set to `false`, that's to handle cases where users don't want to deal with the unknown and only process them in batches of a set amount. Isn't this a use case you outlined before for UI pagination (apologies if this was someone else)?

When our API is used with `autoPaginate` set to false, the user's callback receives the `nextQuery` object, which is a pre-built query with the proper startValue to let the user continue querying for results when they are ready.

stephenplusplus:
Specifically, that test uses a limit of 5 as a way to ensure there will be a nextQuery returned that allows us ("the user") a way to get the next set of results. Is there a better approach?

pcostell:
@stephenplusplus There are two fundamentally different issues here:

1) Normal query continuation -- In this case, the user asks for N results but Datastore returned less than that for resource reasons. `more_results` will be `NOT_FINISHED`, and the client library should request the next batch.

2) UI pagination -- The user asks for a page of N. The clients follow normal procedure in (1) to receive all 5 entities. Then, we tell the app either "Yes, there are more results after this cursor X (`more_results=MORE_RESULTS_AFTER_LIMIT`)" or "No, you're at the end (`more_results=NO_MORE_RESULTS`)". The app can then display to their user a page of N entities, and can decide to show a "Next page" icon based on that information.

pcostell:
The reason this API feels a little awkward is that in general I'd expect the "nextQuery" to never actually be executed. Instead, I would expect it to be serialized and executed later.

However -- I think you can keep the existing API and make this work -- it's just that setting `autoPaginate(false)` really indicates that the user wants you to generate the "nextQuery" for them. Running with `autoPaginate(true)` vs `autoPaginate(false)` shouldn't have an impact on how many results are returned -- it should always be MIN(limit, # total results).

stephenplusplus:
I think I got it, thanks for explaining the difference. I agree all results that the user requests should be returned; I wasn't aware of the differences in response codes. I was under the impression the "MORE_RESULTS" code meant keep going, not appreciating the "_AFTER_LIMIT" nuance. I updated the code, let me know if this is the correct approach.

lib/datastore/request.js
```diff
@@ -464,8 +467,18 @@ DatastoreRequest.prototype.runQuery = function(query, callback) {
       nextQuery.start(endCursor).offset(nextOffset);
     }
 
+    if (notFinished) {
+      // Run the query again to make sure all of the requested entities are
+      // returned.
+      reqOpts.query = entity.queryToQueryProto(query);
```
pcostell: Should this be nextQuery?

stephenplusplus: Yep, good catch!

pcostell:
I don't know the test code will enough to understand the change there, but the rest LGTM.

pcostell:
Thanks for getting this fixed so quickly!

stephenplusplus:
No problem, I'm still tidying that up. Thanks for today's Datastore lesson :)

pcostell:
No worries -- the API is unfortunately complex right now. I'm hoping a gRPC streaming-based API will make this much simpler.

Issue #1208, opened 2016-04-06
How are query start cursorTokens used?


mcfarljw:
My application is using express and I'm having some troubles getting the latest `~0.30.0` releases to return a cursor value I can pass back to the client for usage. If I recall correctly the previous releases were using a base64 string. What kind of cursorToken should I be passing a start query now?

https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.30.2/datastore/query?method=start

```
datastore.runQuery(
    datastore.createQuery('User').start(???).limit(1),
    function(error, entities, cursor) {
        // cursor used to return base64 startVal
    }
);
```

Maybe I should manually convert it myself now?

`new Buffer(cursor.startVal).toString('base64')`

mcfarljw:
I've just spent a bit of time looking into the issue and did get it working. Here is what I came up with:

```
var cursorToken = req.cursorToken ? new Buffer(req.cursorToken, 'base64') : null;
datastore.runQuery(
    datastore.createQuery('User').start(cursorToken).limit(1),
    function(error, entities, nextQuery) {
        var result = {};
        result.cursorToken = new Buffer(nextQuery.startVal).toString('base64');
        res.status(200).json(result);
);
```

It might make things a bit more straightforward and easier to work with if the start method would accept both base64 and buffer parameters. It might also be convenient for the `nextQuery` to include a base64 version called endCursorToken and startCursorToken so that it matches up a bit more with the documentation.

stephenplusplus:
I think we should just go back to only accepting base64 strings and skip buffers altogether. Any interest in putting a PR together?

mcfarljw:
I'll take a shot at making a PR for this and see what happens.

stephenplusplus:
Thanks! Feel free to send it in even if it's not perfect. I can always help with the grunt work (code style, tests, etc).

stephenplusplus:
Fixed via a80ff9b0b98dc5e8d2844dfcf00b65cf3ce87420

PR #1209, opened 2016-04-06
datastore: base64 cursors

mcfarljw:
Fixes: #1208 

Not sure if this is the right way to handle it, but either way it could probably also use a test to prevent regressions in formatting going forward.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.idea/
 node_modules/*
 coverage/*
 docs/json/**/*.json
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -465,7 +465,7 @@ DatastoreRequest.prototype.runQuery = function(query, callback) {
       var offset = query.offsetVal === -1 ? 0 : query.offsetVal;
       var nextOffset = offset - resp.batch.skippedResults;
       nextQuery = extend(true, new Query(), query);
-      nextQuery.start(endCursor).offset(nextOffset);
+      nextQuery.start(endCursor.toString('base64')).offset(nextOffset);
     }
 
     if (notFinished) {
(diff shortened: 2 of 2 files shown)


lib/datastore/request.js
```diff
@@ -487,6 +487,16 @@ DatastoreRequest.prototype.runQuery = function(query, callback) {
...
 
+    if (nextQuery && nextQuery.startVal) {
+      // Convert startVal to base64 format
+      nextQuery.startVal =  new Buffer(nextQuery.startVal).toString('base64');
+    }
+
+    if (nextQuery && nextQuery.endVal) {
+      // Convert endVal to base64 format
+      nextQuery.endVal =  new Buffer(nextQuery.endVal).toString('base64');
+    }
```
stephenplusplus: I think we can simplify this by changing line 468 to:

```js
nextQuery.start(endCursor.toString('base64')).offset(nextOffset);
```

For `endVal` to be set on `nextQuery`, it has to have first been set on the query the user passed in originally. It would be a string already if that's what they had set it as.

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
stephenplusplus: We try to keep IDE/environment-specific config out of this file, and instead recommend using [global .gitignore files](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) for this type of stuff.

stephenplusplus:
Thanks for sending this in, you have been a huge help getting our Datastore support into good shape. :muscle:

mcfarljw:
No problem. Thanks for pointing out the global `.gitignore` thing, it certainly makes more sense. I'm not sure why I wasn't using it before! I'll get things cleaned up a bit and then make another commit. Looking forward to getting more of our server side code on node.

stephenplusplus:
Merged in a80ff9b0b98dc5e8d2844dfcf00b65cf3ce87420 -- thanks again for helping us out :)

PR #1211, opened 2016-04-07
tests: fix datastore improper override

### LINKED ISSUE #1199: Datastore + gRPC Travis errors
We're getting a new error in Travis, but it cannot be reproduced locally:

```
  1) Datastore create, retrieve and delete should save keys as a part of entity and query by key:
     Uncaught Error: one of fields Query.query and Query.gql_query must be set
      at node_modules/grpc/src/node/src/client.js:411:17
```
https://travis-ci.org/GoogleCloudPlatform/gcloud-node/jobs/120723899#L5784

This is an error from gRPC saying that the object we're giving it is missing a required property from the proto file's Query message schema. The object that it's talking about goes through a few phases before it gets to gRPC, but it is first defined [here](https://github.com/GoogleCloudPlatform/gcloud-node/blob/cd0c3e98505c5e29004d063b37a8605b7230a504/lib/datastore/request.js#L432) ... eventually being passed to gRPC [here](https://github.com/GoogleCloudPlatform/gcloud-node/blob/cd0c3e98505c5e29004d063b37a8605b7230a504/lib/common/grpc-service.js#L244).

As far as I can see, we are not deleting the `query` property, yet *only on Travis*, gRPC says that we are. At the moment, I'm not sure where to look, so I'm putting this out for extra eyes.

stephenplusplus:
Fixes #1199

Just spent about 9 hours with Travis. It was great.

One of our datastore unit tests was overriding an actual method, instead of properly mocking it. Also, caught a file our `cover` and `coveralls` tasks weren't... covering.

I'm insta-merging this because seeing Travis green is going to make me feel happy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,8 +137,8 @@
     "lint": "jshint lib/ system-test/ test/ && jscs lib/ system-test/ test/",
     "test": "npm run docs && mocha test/*/*.js test/index.js test/docs.js",
     "system-test": "mocha system-test/* --no-timeouts --bail",
-    "cover": "istanbul cover -x 'system-test/*' _mocha -- --no-timeouts --bail test/*/*.js test/docs.js system-test/*",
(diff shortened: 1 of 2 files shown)


jgeewax:
:( and :)

PR #1215, opened 2016-04-07
run coveralls on PRs

stephenplusplus:
This gets coveralls to run from a PR, then leave a comment letting us know if coverage increased or decreased.

Until now, our coverage reported to coveralls has included both unit and system tests. It has only run after a merge to master. This would create a mismatch in coverage reports between a PR (that doesn't include system test coverage) and the coverage status of the master branch of our repo (which included system test coverage). I've changed our coveralls task to only ever run unit tests-- I think our concerns of line coverage should come from our unit tests, anyway. Let me know if you can think of a better way to handle this!


stephenplusplus:
Hey, buddy!

PR #1216, opened 2016-04-08
testing: bump coverage to 100

### LINKED ISSUE #1219: Bigquery BOOLEAN data type conversion bug
I'm running a query on a BigQuery table. The table includes a BOOLEAN field and has records where the field is `false`. Running `bigQuery.query(query)`with this library results in every record having its boolean variable set to `true`. The culprit is [this line](https://github.com/GoogleCloudPlatform/gcloud-node/blob/c38c548240257625b6a62342eeb3606479cd04a5/lib/bigquery/table.js#L225):

```javascript
value = Boolean(value).valueOf();
```

`value` comes in as the string `'false'`.

From the [Javascript docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean):

> The value passed as the first parameter is converted to a boolean value, if necessary. If value is omitted or is 0, -0, null, false, NaN, undefined, or the empty string (""), the object has an initial value of false. All other values, including any object or the string "false", create an object with an initial value of true.

stephenplusplus:
Fixes #1219

Mission: Getting our coverage back up to 100.

### DIFF
--- a/lib/bigquery/table.js
+++ b/lib/bigquery/table.js
@@ -193,7 +193,7 @@ Table.createSchemaFromString_ = function(str) {
   return str.split(/\s*,\s*/).reduce(function(acc, pair) {
     acc.fields.push({
       name: pair.split(':')[0],
-      type: pair.split(':')[1] || 'string'
+      type: (pair.split(':')[1] || 'STRING').toUpperCase()
     });
 
     return acc;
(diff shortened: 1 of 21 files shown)


stephenplusplus:
I think I got them all. @coveralls wdyt?

stephenplusplus:
Good catch. PTAL.

stephenplusplus:
Dude.

stephenplusplus:
HEY!! :100: Have a good weekend.

@callmehiphop, my non-robot friend, ptal.

sindresorhus:
🍾🎉

PR #1227, opened 2016-04-13
README: Consistent use of GCLOUD_PROJECT

ofrobots:
Elsewhere we in the codebase we use GCLOUD_PROJECT but in README.md there were some examples with GCLOUD_PROJECT_ID.

/cc @JustinBeckwith


stephenplusplus:
Thank you!

PR #1233, opened 2016-04-15
storage: add file#save

### LINKED ISSUE #1179: Allow byte uploads in upload, not just file uploads.
Currently the `bucket.upload()` API provides uploads from the local file system, and any streaming requests have to happen by opening a writable stream. We should provide an API to allow upload to consume either a path on the file system, or a byte array. Like so:

```javascript
var options = {
  destination: 'hello.txt'
}

var data = ['h', 'e', 'l', 'l', 'o', '!']

bucket.upload(data, options, function(err, file) {
  // File hello.txt contains 'hello!'
});
```

Also, I get that byte uploads allow people to do terrible, unscalable things, so we should preface the docs by saying that for large files they should stream the file using a `writableStream` or upload from a file on disk. That said, I think this is a nice convenience method that allows devs to get up and running quickly from files they already have in memory.

stephenplusplus:
Fixes #1179

This adds `file#save`, which is a simple wrapper around `file#createWriteStream`:

``` js
file.save(new Buffer('buffer'), function(err) {});
file.save('string', function(err) {});
```

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -1428,6 +1428,37 @@ File.prototype.move = function(destination, callback) {
 };
 
 /**
+ * Write arbitrary data to a file.
+ *
+ * *This is a convenience method which wraps
+ * {module:storage/file#createWriteStream}.*
+ *
+ * @param {*} data - The data to write to a file.
+ * @param {object=} options - See {module:storage/file#createWriteStream}'s
+ *     `options` parameter.
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request
+ *
+ * @example
+ * file.save('This is the contents of the file.', function(err) {
+ *   if (!err) {
+ *     // File written successfully.
+ *   }
+ * });
+ */
+File.prototype.save = function(data, options, callback) {
+  if (is.fn(options)) {
+    callback = options;
+    options = {};
+  }
+
+  this.createWriteStream(options)
+    .on('error', callback)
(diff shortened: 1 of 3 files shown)


test/storage/file.js
```diff
@@ -2072,6 +2072,68 @@ describe('File', function() {
     });
   });
 
+  describe.only('save', function() {
```
stephenplusplus: Woops.

lib/storage/file.js
```diff
@@ -1428,6 +1428,37 @@ File.prototype.move = function(destination, callback) {
 };
 
 /**
+ * Write arbitrary data to a file.
+ *
+ * *This is a convenience method which wraps
+ * {module:storage/file#createWriteStream}.*
```
callmehiphop: Do we need the stray asterisk?

stephenplusplus: There are two, to make it italicized.

callmehiphop: Ahh, sorry, just noticed that!

callmehiphop:
@stephenplusplus I realize this is just a convenience method, but what are your feelings on adding a system-test?

stephenplusplus:
Added!

mcdonamp:
Thanks for doing this!

stephenplusplus:
Thank you for the idea :)

existeundelta:
where can I find minimal Cloud Function using this? 
thanks

PR #1240, opened 2016-04-18
vision: support image URLs

### LINKED ISSUE #1239: Allow recognition of image URL in Vision API
Now, it's support two type that local file and gcsImageUri. 
`eg: 
/data/images/aa.jpg, {"name": aa.jpg, "bucket": {"name": "demo"}}`
Can you add the support of image url.
`eg:http://mydomain/images/aa.jpg`

stephenplusplus:
Fixes #1239 

Docs: http://stephenplusplus.github.io/gcloud-node/#/docs/master/vision?method=detect
#### To Dos
- [x] Docs
- [x] Tests
  - [x] System
  - [x] Unit

This change will allow our Vision API to accept image URLs.

``` js
vision.detectLogos('http://google.com/logo.png', function(err, logos) {
  logos = ['Google']
});
```

### DIFF
--- a/lib/vision/index.js
+++ b/lib/vision/index.js
@@ -29,6 +29,7 @@ var fs = require('fs');
 var is = require('is');
 var nodeutil = require('util');
 var prop = require('propprop');
+var request = require('request');
 
 /**
  * @type {module:storage/file}
@@ -183,7 +184,7 @@ Vision.prototype.annotate = function(requests, callback) {
  *
  * @param {string|string[]|module:storage/file|module:storage/file[]} images -
  *     The source image(s) to run the detection on. It can be either a local
- *     image path or a gcloud File object.
+ *     image path, a remote image URL, or a gcloud File object.
  * @param {string[]|object=} options - An array of types or a configuration
  *     object.
  * @param {number} options.maxResults - The maximum number of results, per type,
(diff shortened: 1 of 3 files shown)


jgeewax:
Is this supported? I thought it was just GCS URLs?

stephenplusplus:
It's not supported by the Vision API. We just do a `GET` on the URL, get a base64 string of the contents, and pass that to Vision to be annotated.

jgeewax:
Gotcha - can we make sure that's documented?

stephenplusplus:
Definitely!

stephenplusplus:
@callmehiphop ready for your review, sir.

lib/vision/index.js
```diff
@@ -1178,6 +1190,24 @@ Vision.findImages_ = function(images, callback) {
...
     }
 
+    // File is a URL.
+    if (image.indexOf('http') === 0) {
+      request({
+        method: 'GET',
+        uri: image,
+        encoding: 'base64'
+      }, function(err, resp, body) {
+        if (err) {
```
jgeewax: Wouldn't this be better as an if / else?

```js
if (err) {
  callback(err);
} else {
  callback(null, {content: body});
}
```

?

stephenplusplus: It's just a style thing. Throughout our API, we always handle errors in the topmost portion of the code and return. That leaves the rest of the block available for un-indented handling of a successful response. Usually, that's much more code than just executing the callback in a slightly different way, but nonetheless, it would be breaking the pattern. IMO, it's better to be consistent than to have a super huge style guide that tries to define when to use what style.

jgeewax: SGTM

lib/vision/index.js
```diff
@@ -1178,6 +1190,24 @@ Vision.findImages_ = function(images, callback) {
       return;
     }
 
+    // File is a URL.
+    if (image.indexOf('http') === 0) {
```
callmehiphop: Probably splitting hairs, but would this be better as a regex to verify the placement of http? (e.g. `^http`)

stephenplusplus: Done!

PR #1241, opened 2016-04-18
Fix documentation that labels for logging entry has to be STRING

faithseed:
The doc says string for instance_id

```
var resource = {
  type: 'gce_instance',
  labels: {
    zone: 'global',
    instance_id: 3
  }
};
```

yet in practice, you will hit the following if you pass an int. You will get the following error. This fixes the doc to reflect the correct usage of this API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,7 +475,7 @@ var resource = {
   type: 'gce_instance',
   labels: {
     zone: 'global',
-    instance_id: 3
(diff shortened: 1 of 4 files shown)


faithseed:
@stephenplusplus please review this change.

stephenplusplus:
Thanks!

PR #1251, opened 2016-04-19
datastore: fixing a query limit bug

### LINKED ISSUE #1249: Query without Limit throws exception when too many results
Hi,

I'm using Datastore with gcloud 0.30.3 and when executing a simple Query to find all entities in a Kind that has 1.000 entities, it fails if the query isn't provided a Limit value:

### **Without Limit**
```
{"namespace":null,"kinds":["postdatastore"],"filters":[],"orders":[],"groupByVal":[],"selectVal":[],"autoPaginateVal":true,"startVal":null,"endVal":null,"limitVal":-1,"offsetVal":-1}
```

##### **Output**
```
/Users/jpdiazvaz/Documents/github/loopback-connector-gcloud-metrics/node_modules/gcloud/lib/datastore/request.js:475
      var limit = reqOpts.query.limit.value;
                                     ^
TypeError: Cannot read property 'value' of undefined
    at onResponse (/Users/jpdiazvaz/Documents/github/loopback-connector-gcloud-metrics/node_modules/gcloud/lib/datastore/request.js:475:38)
```

### **With Limit**
```
{"namespace":null,"kinds":["postdatastore"],"filters":[],"orders":[],"groupByVal":[],"selectVal":[],"autoPaginateVal":true,"startVal":null,"endVal":null,"limitVal":100000,"offsetVal":-1}
```

##### **Output**
```
[{"key":{"id":4503908059709440,"kind":"postdatastore","path":["postdatastore",4503908059709440]},"data":{"body":"This is the body of a test POST","title":"This is a test POST"}},...]
```

callmehiphop:
Closes #1249

Since migrating to v1beta3 we've changed the way we store the `limit` value, this causes a bug if the user did not set a query limit when retrieving a large number of results.

### DIFF
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -471,8 +471,8 @@ DatastoreRequest.prototype.runQuery = function(query, callback) {
     if (notFinished) {
       // Run the query again to make sure all of the requested entities are
       // returned.
-      var limit = reqOpts.query.limit.value;
-      if (limit > -1) {
+      var limit = reqOpts.query.limit && reqOpts.query.limit.value;
+      if (limit && limit > -1) {
         // Update the limit on the nextQuery to return only the amount of
         // results originally asked for.
         nextQuery.limit(limit - resp.batch.entityResults.length);
--- a/test/datastore/request.js
+++ b/test/datastore/request.js
@@ -696,6 +696,46 @@ describe('Request', function() {
           done();
         });
       });
+
+      it('should handle large limitless queries', function(done) {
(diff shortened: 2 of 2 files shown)


stephenplusplus:
LGTM. Is there a test case we can add to cover the situation caused in #1249?

callmehiphop:
@stephenplusplus I think this would obviously be difficult for us to reproduce via e2e tests, so I'll write a unit test.

callmehiphop:
@stephenplusplus PTAL!

PR #1253, opened 2016-04-19
Remove insertAutoId documentation.

pcostell:
insertAutoId went away with `v1beta3` and was merged with insert.


stephenplusplus:
Thanks!

PR #1259, opened 2016-04-23
datastore: accept read consistencies

### LINKED ISSUE #1218: Read consistency support for Datastore
It would be great if we could specifiy the read consitency like the java sdk can for google datastore.

stephenplusplus:
Fixes #1218 
#### To Dos
- [x] Docs
- [x] Unit tests

@pcostell would you mind taking a look? Also, is there a reliable way to run tests against the API for eventual & strong consistency, or should we just trust our unit tests that the correct read options are making their way to the API?

### DIFF
--- a/lib/datastore/request.js
+++ b/lib/datastore/request.js
@@ -52,6 +52,17 @@ var streamRouter = require('../common/stream-router.js');
  */
 var util = require('../common/util.js');
 
+/**
+ * A map of read consistency values to proto codes.
+ *
+ * @type {object}
+ * @private
+ */
+var CONSISTENCY_PROTO_CODE = {
+  eventual: 2,
+  strong: 1
+};
+
 /*! Developer Documentation
  *
  * Handles request logic for Datastore.
@@ -212,6 +223,11 @@ DatastoreRequest.prototype.delete = function(keys, callback) {
  * @throws {Error} If at least one Key object is not provided.
  *
  * @param {Key|Key[]} keys - Datastore key object(s).
+ * @param {object=} options - Optional configuration.
+ * @param {string} options.consistency - Specify either `strong` or `eventual`.
+ *     If not specified, default values are chosen by Datastore for the
+ *     operation. Learn more about strong and eventual consistency
+ *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-google-cloud-datastore).
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request
  * @param {object|object[]} callback.entity - The entity object(s) which match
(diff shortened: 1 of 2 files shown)


lib/datastore/request.js
```diff
@@ -423,19 +444,31 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
...
   };
 
   var reqOpts = {
-    readOptions: {},
     query: entity.queryToQueryProto(query)
   };
 
+  if (is.defined(options.strongConsistency)) {
+    reqOpts.readOptions = {
+      readConsistency: options.strongConsistency ? 1 : 2 // (strong : eventual)
```
pcostell: Should this use the string values "STRONG" and "EVENTUAL"?

lib/datastore/request.js
```diff
@@ -821,8 +854,9 @@ DatastoreRequest.prototype.request_ = function(protoOpts, reqOpts, callback) {
   }
 
   if (isTransaction && (method === 'lookup' || method === 'runQuery')) {
```
pcostell: This will overwrite the read options. Maybe you should make it an error if isTransaction and readOptions are both set?

lib/datastore/request.js
```diff
@@ -351,6 +368,10 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
  * See below for examples of both approaches.
  *
  * @param {module:datastore/query} q - Query object.
+ * @param {object=} options - Optional configuration.
+ * @param {boolean} options.strongConsistency - Use strong read consistency. For
+ *     eventual consistency, use `false` for this value. If not specified,
+ *     default values are chosen by Datastore for the operation.
```
pcostell: It may be nice to add a comment here that you can't just set strongConsistency to strong to make an operation strongly consistent. You can only use this to make a strongly consistent operation eventual.

lib/datastore/request.js
stephenplusplus: Sure, good idea.

lib/datastore/request.js
stephenplusplus: Definitely. I will add links to the official docs.

lib/datastore/request.js
stephenplusplus: That is probably a better plan. 👍

stephenplusplus:
@pcostell thanks for taking a look. Can you think of any test cases synopses that would work for this change against the real API?

pcostell:
Probably no good test cases. This is primarily for efficiency at the sake of consistency, so any tests would have to be eventually consistent (and thus flaky).

lib/datastore/request.js
```diff
@@ -432,19 +460,36 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
...
     method: 'runQuery'
   };
 
   var reqOpts = {
-    readOptions: {},
     query: entity.queryToQueryProto(query)
   };
 
+  if (options.consistency) {
+    var consistencyMap = {
```
callmehiphop: Looks like this map/logic is duplicated. Could we just make the `consistencyMap` global to this file? Or put it in a static(ish) method that retrieved the proper value?

stephenplusplus: Good call.

stephenplusplus: Done!

PR #1263, opened 2016-04-25
docs: add documentation for ending streams early

### LINKED ISSUE #1261: Datastore.runQuery with .on('data') doesn't demonstrate how to "abort"?
In a for-loop, I can call `break;` to exit. When I'm in a runQuery stream, I'm unsure how to exit early. Is there a way to document that?

Ie, it might be...

```js
datastore.runQuery(query)
  .on('data', function(entity) {
    if (my exit criteria) this.exit(); // ?
  });
```

callmehiphop:
Closes #1261 

There were several spots in our documentation where we failed to give an example of how to end a stream early.

### DIFF
--- a/lib/bigquery/index.js
+++ b/lib/bigquery/index.js
@@ -457,6 +457,15 @@ BigQuery.prototype.job = function(id) {
  *   .on('end', function() {
  *     // All rows retrieved.
  *   });
+ *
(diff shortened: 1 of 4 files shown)


stephenplusplus:
Nice!

PR #1264, opened 2016-04-25
search: removing search api

### LINKED ISSUE #1262: Remove search API
Looks like we're changing things around and search won't be it's own API surface.

Can we chop from gcloud-node? (It was a private alpha, so the impact to gcloud-node users should be relatively minimal). If there are concerns about timing, let's discuss here.

callmehiphop:
Closes #1262


callmehiphop:
@stephenplusplus I think I got everything search related, PTAL!

stephenplusplus:
package.json keywords! Do we also remove it from toc.json?

callmehiphop:
@stephenplusplus good call! Thanks :)

stephenplusplus:
I think there might be an exception for 'index-class' in the doc building file that we can probably remove.

callmehiphop:
done and done!

lostpebble:
I used the App Engine Search API extensively in the previous version of my application. Now that I'm moving over to NodeJS it would be awesome if there was a way I could use it directly from Node. For now, I'm running a whole other Java instance just so I can access it, which isn't very practical.

Is there any way for us to access this API from NodeJS?

For reference, just in case I'm at the wrong place, [this is the API I'm talking about.](https://cloud.google.com/appengine/docs/java/search/)

PR #1272, opened 2016-04-28
core: implement retry strategy for grpc requests

### LINKED ISSUE #1265: 503 unavailable response from datastore after app has idled for a few minutes
I've been getting this response in both production and locally (using the production datastore):

```
{ [Error]
  code: 503,
  metadata: Metadata { _internal_repr: {} },
  response: '' }
```

The error is thrown from here - https://github.com/grpc/grpc/blob/release-0_13_1/src/node/src/client.js#L411

Steps to reproduce:
1. Make a http request "/" that hits datastore
2. Wait a few minutes and don't make an http request that hits the datastore
3. Make a http request "/" that hits datastore

Expect: Datastore to return results
Actual: 503 unavailable

4. Immediately make a http request "/" that hits datastore 

The datastore will work and continue to work for another few minutes.
I've deployed the bookstore sample app and it seems to have the same behavior saying "Something broke!".

Here is the code:

```
    projectId: 'app'
});

var app = express();
app.disable('etag');
app.set('trust proxy', true);


app.get('/', function(req, res, next) {
    var query = datastore.createQuery('Post').filter("Status", "staged");

    datastore.runQuery(query, function(err, posts) {
        if (err) {
            console.log(err.message)
            console.log(err.code)
            console.log(err.metadata)
            next(err)
        } else {
            res.status(200).send("Posts staged = " + posts.length)
        }
    })

});
```

I'm using the default app credentials and Flexible VMs. Let me know if there is any other info I could provide.

callmehiphop:
Closes #1265 

This adds a retry with backoff strategy for gRPC requests. Currently it will only attempt to retry for gRPC error code 14 (service unavailable).

@heidmotron @leonardpunt @jmdobry if any of you would like to test if this resolves your issues, that would be awesome :)

### DIFF
--- a/lib/common/grpc-service.js
+++ b/lib/common/grpc-service.js
@@ -25,6 +25,7 @@ var grpc = require('grpc');
 var is = require('is');
 var nodeutil = require('util');
 var path = require('path');
+var retryRequest = require('retry-request');
 
 /**
  * @type {module:common/service}
@@ -109,7 +110,8 @@ var HTTP_ERROR_CODE_MAP = {
 
   14: {
     code: 503,
-    message: 'Service Unavailable'
+    message: 'Service Unavailable',
+    shouldRetry: true
   },
 
   15: {
@@ -122,6 +124,12 @@ var HTTP_ERROR_CODE_MAP = {
     message: 'Unauthorized'
   }
 };
+
+/**
+ * @const {number} - Allowed number of a request retries
+ * @private
+ */
+var MAX_RETRIES = 2;
 
 /**
  * Service is a base class, meant to be inherited from by a "service," like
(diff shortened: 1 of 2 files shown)


stephenplusplus:
Is it possible to use retry request directly? It accepts a custom request function. If it's close, but kind of a hack, maybe it just needs a pr to generalize it.

callmehiphop:
I initially did go down this route but it started to get pretty hacky, so I gave up on it. However if you think a PR there is the way to go, I'd be more than happy to try and tackle that.

stephenplusplus:
If it's possible without getting into hack territory, I think that would be awesome.

heidmotron:
@callmehiphop this works, thanks!

callmehiphop:
Closing in favor of #1273

PR #1273, opened 2016-04-30
core (grpc) use retry strategy

### LINKED ISSUE #1265: 503 unavailable response from datastore after app has idled for a few minutes

stephenplusplus:
Fixes #1265 
#### To Dos
- [x] Unit tests

// @callmehiphop

### DIFF
--- a/lib/common/grpc-service.js
+++ b/lib/common/grpc-service.js
@@ -20,11 +20,13 @@
 
 'use strict';
 
+var extend = require('extend');
 var googleProtoFiles = require('google-proto-files');
 var grpc = require('grpc');
 var is = require('is');
 var nodeutil = require('util');
 var path = require('path');
+var retryRequest = require('retry-request');
 
 /**
  * @type {module:common/service}
@@ -151,6 +153,8 @@ function GrpcService(config, options) {
     this.grpcCredentials = grpc.credentials.createInsecure();
   }
 
+  this.maxRetries = config.maxRetries;
+
   var apiVersion = config.apiVersion;
   var service = this.service = config.service;
   var rootDir = googleProtoFiles('..');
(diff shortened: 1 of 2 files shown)


stephenplusplus:
@callmehiphop this is just another option, didn't mean for it to auto-invalidate your work. If you don't think this approach is the right way to go, just let me know! I'll wait before I get the tests ready.

callmehiphop:
It's cool, this was the first response I'd seen from you in regards to this issue in several days, so I figured you weren't a fan of what I was propsing. I originally did start going down this route, but I felt like the nesting/wrapping increased the cyclomatic complexity, which is why I decided to abandon it. If you think this is the best way to go, that's fine with me.

stephenplusplus:
It actually wasn't that I wasn't a fan, it was just hard to review from the 🚗 😢 👶 situation. Like you, I'm okay with all, but not a super-fan of any. I like this overall because it didn't force changes to retry-request (which I thought would be good originally, but in practice, turns out awkward [as you predicted]). I also like this because it doesn't duplicate any retry logic which already exists in retry-request. I'll get going on some tests.

PR #1277, opened 2016-05-02
tests: run tests against node v6

stephenplusplus:
[Version 6.0.0 is out!](https://nodejs.org/en)


callmehiphop:
Do we still care about v5? It looks like only v4 and v6 are LTS.

stephenplusplus:
Later, 5!

PR #1278, opened 2016-05-02
storage: support setting public/private at time of upload

### LINKED ISSUE #1178: Be able to set default object ACL on GCS upload
Currently, there's no way to set the predefined object ACL on an object that is being uploaded (though there is in the [REST API](https://cloud.google.com/storage/docs/json_api/v1/objects/insert)). One can overwrite the metadata with a new ACL, but that is a hard overwrite, and doesn't preserve existing ACLs, unlike add default ACL. It would be neat if we had an API that worked like:

```javascript
var options = {
  predefinedAcl:'publicRead',
  validation: 'crc32',
  metadata: {/* add metadata here */}
};

bucket.upload('local/path/to/bar.txt', options, function(err, file) {
  // File is publicly readable!
});
```

stephenplusplus:
Fixes #1178 (whoa, exactly 100 issues/PRs ago!)

Set the `predefinedAcl` at the time of upload:

``` js
bucket.upload(filepath, { predefinedAcl: 'authenticatedRead' }, function(err, file) {})

var writeStream = file.createWriteStream({ predefinedAcl: 'authenticatedRead' })
```

Also with built-in `public` (predefinedAcl: publicRead) and `private` (predefinedAcl: private) options:

``` js
bucket.upload(filepath, { public: true }, function(err, file) {})

var writeStream = file.createWriteStream({ private: true })
```

### DIFF
--- a/lib/storage/bucket.js
+++ b/lib/storage/bucket.js
@@ -936,9 +936,33 @@ Bucket.prototype.makePublic = function(options, callback) {
  *     bucket using the name of the local file.
  * @param {boolean} options.gzip - Automatically gzip the file. This will set
  *     `options.metadata.contentEncoding` to `gzip`.
- * @param {object=} options.metadata - See an
+ * @param {object} options.metadata - See an
  *     [Objects: insert request body](https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request_properties_JSON).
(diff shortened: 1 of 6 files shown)


mcdonamp:
Thanks for this--I like the quick `public` `private` ones as well :)

PR #1280, opened 2016-05-03
datastore: add query.run

### LINKED ISSUE #1266: Add query.run() ?
We create a query from a datastore instance (which has the auth stuff) but then don't allow people to blindly "run" the query. Is there a reason?

Code today:

```js
var query = datastore.createQuery('Kind').filter('key', 'value');
datastore.runQuery(query).on('data', function(entity) {
  // do something with the entity
});
```

Would be nice if....

```js
var query = datastore.query('Kind').filter('key', 'value');
query.run().on('data', function(entity) {
  // Do something with the entity
});

// And if we wanted the same query, but with different credentials/project ID/etc:

otherDatastore.runQuery(query).on(....)
```

stephenplusplus:
Fixes #1266  
#### To Dos
- [x] Docs
- [x] Tests
  - [x] System
  - [x] Unit

Previously, a query has only been created through `{module:datastore#createQuery}`. But since adding `run` on the Query object, we need a reference to the parent object. The only thing that made this a bit tricky was running a query inside of a transaction. That's why this PR has to introduce `{module:transaction#createQuery}`.

### DIFF
--- a/lib/datastore/query.js
+++ b/lib/datastore/query.js
@@ -307,18 +307,15 @@ Query.prototype.offset = function(n) {
 /**
  * Run the query.
  *
+ * @param {object=} options - Optional configuration.
+ * @param {string} options.consistency - Specify either `strong` or `eventual`.
+ *     If not specified, default values are chosen by Datastore for the
+ *     operation. Learn more about strong and eventual consistency
+ *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-google-cloud-datastore).
  * @param {function=} callback - The callback function. If omitted, a readable
  *     stream instance is returned.
  * @param {?error} callback.err - An error returned while making this request
- *     (may be null).
- * @param {array} callback.entities - The list of entities returned by this
- *     query. Note that this is a single page of entities, not necessarily
- *     all of the entities.
- * @param {?module:datastore/query} callback.nextQuery - If present, run another
- *     query with this object to check for more results.
- * @param {object} callback.apiResponse - The full API response.
- * @param {?error} callback.err - An error returned while making this request
- * @param {module:datastore/entity[]} callback.entities - A list of Entities
+ * @param {object[]} callback.entities - A list of entities.
  * @param {?object} callback.nextQuery - If present, query with this object to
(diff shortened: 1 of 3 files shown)


stephenplusplus:
@callmehiphop ptal!

lib/datastore/query.js
```diff
@@ -73,15 +80,17 @@ function Query(namespace, kinds) {
...
  * @example
- * // Retrieve a list of people related to person "1234",
- * // disabling auto pagination
+ * //-
+ * // Retrieve a list of people related to "Dave", with auto-pagination
+ * // disabled.
+ * //-
  * var query = datastore.createQuery('Person')
- *   .hasAncestor(datastore.key(['Person', 1234]))
+ *   .hasAncestor(datastore.key(['Person', 'Dave']))
```
callmehiphop: woot

lib/datastore/query.js
```diff
@@ -295,4 +304,76 @@ Query.prototype.offset = function(n) {
...
+ *     stream instance is returned.
+ * @param {?error} callback.err - An error returned while making this request
+ *     (may be null).
+ * @param {array} callback.entities - The list of entities returned by this
+ *     query. Note that this is a single page of entities, not necessarily
+ *     all of the entities.
+ * @param {?module:datastore/query} callback.nextQuery - If present, run another
+ *     query with this object to check for more results.
+ * @param {object} callback.apiResponse - The full API response.
+ * @param {?error} callback.err - An error returned while making this request
```
callmehiphop: Looks like we got some copy pasta duplication here?

stephenplusplus: It is copied from `runQuery` but what did I miss?

callmehiphop: `callback.err` is listed twice? As well as `entities` and `apiResponse`

stephenplusplus: Woops, it's apparently been messed up over there for a while as well!

stephenplusplus:
Fixed the docs issue + realized query.run didn't take an `options`, which is required for setting the consistency.

callmehiphop:
LGTM!

PR #1294, opened 2016-05-06
bigquery: implement Job as an event emitter

### LINKED ISSUE #1285: Bigquery import does not populate err in callback when error occurs
In trying to track down why data was not appearing in Bigquery even though the import completed "successfully", I noted that the `err` argument was not populated even though an error had occurred.

code:

```javascript

table.import(gcsFile, function(err, job, apiResponse) {
  if(err) {
    callback(err);
    return;
  } 

  console.log('Data import job created');
  console.log(apiResponse);
}
```

The `err` argument is `undefined` (or perhaps `null`), however the apiResponse reports (verbatim from logs):

```javascript
{ kind: 'bigquery#job',
  etag: '"oGMCLvGjZO7RB3oFjn17umMEDU4/v1SSKa6IBg_jGNKysjFGzeC_xYA"',
  id: 'gcf-tests:job_19Q8iGmmqYjyV0V7rj5fiSwiQFw',
  selfLink: 'https://www.googleapis.com/bigquery/v2/projects/gcf-tests/jobs/job_19Q8iGmmqYjyV0V7rj5fiSwiQFw',
  jobReference:
   { projectId: 'gcf-tests',
     jobId: 'job_19Q8iGmmqYjyV0V7rj5fiSwiQFw' },
  configuration:
   { load:
      { sourceUris: [Object],
        destinationTable: [Object],
        sourceFormat: 'CSV' } },
  status:
   { state: 'DONE',
     errorResult:
      { reason: 'invalid',
        message: 'No schema specified on job or table.' },
     errors: [ [Object] ] },
  statistics:
   { creationTime: '1462423012789',
     startTime: '1462423012849',
     endTime: '1462423012849' },
  user_email: '709128455856-compute@developer.gserviceaccount.com' }
```

It *might* be reasonable to have developers inspect the status/errorResult property of the apiResponse, but it seems like it would make more sense to report this as an error

stephenplusplus:
Fixes #1285 
#### To Dos
- [x] Docs
- [x] Tests
  - [x] System
  - [x] Unit

Like Compute Engine Operation objects, BigQuery Job objects are now event emitters. Only after registering a `complete` listener, we start polling the job for updates. Errors are parsed and emitted to any `error` listeners.

// cc @jasonpolites

### DIFF
--- a/test/bigquery/job.js
+++ b/test/bigquery/job.js
@@ -31,6 +31,15 @@ function FakeServiceObject() {
 
 nodeutil.inherits(FakeServiceObject, ServiceObject);
 
+var utilOverrides = {};
+var fakeUtil = Object.keys(util).reduce(function(fakeUtil, methodName) {
+  fakeUtil[methodName] = function() {
+    var method = utilOverrides[methodName] || util[methodName];
+    return method.apply(this, arguments);
+  };
+  return fakeUtil;
+}, {});
+
 describe('BigQuery/Job', function() {
   var BIGQUERY = {
     projectId: 'my-project'
@@ -44,6 +53,8 @@ describe('BigQuery/Job', function() {
       '../../lib/common/service-object.js',
       FakeServiceObject
     );
+    mockery.registerMock('../../lib/common/util.js', fakeUtil);
+
     mockery.enable({
       useCleanCache: true,
       warnOnUnregistered: false
@@ -58,6 +69,7 @@ describe('BigQuery/Job', function() {
   });
 
   beforeEach(function() {
+    utilOverrides = {};
     job = new Job(BIGQUERY, JOB_ID);
   });
 
@@ -79,6 +91,11 @@ describe('BigQuery/Job', function() {
         get: true,
         getMetadata: true
       });
+    });
+
+    it('should correctly initialize variables', function() {
+      assert.strictEqual(job.completeListeners, 0);
+      assert.strictEqual(job.hasActiveListeners, false);
     });
 
     describe('request interceptor', function() {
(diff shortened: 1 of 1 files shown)


stephenplusplus:
@callmehiphop tests added, ptal! (don't forget to squash and merge 😄 )

lib/bigquery/job.js
```diff
@@ -156,35 +188,9 @@ nodeutil.inherits(Job, ServiceObject);
  * @example
  * job.cancel(function(err, apiResponse) {
  *   // Check to see if the job completes successfully.
- *   onJobComplete(function(err) {
- *     if (!err) {
- *       // Job cancelled successfully.
- *     }
- *   });
+ *   job.on('error', function(err) {});
+ *   job.on('complete', function(metadata) {});
```
callmehiphop: Pretty sure I've asked this question in regards to operations before, but any reason we don't just add the event listeners internally so the usage is

```js
job.cancel(function(err, metadata) {});
```

callmehiphop: Or perhaps return an the emitter?

```js
job.cancel()
  .on('error', console.log)
  .on('complete', function(metadata) {});
```

stephenplusplus: I think since it can take a long time to complete the operation/job, we have the separate type. All of the GCE operations are good examples of long-running operations (minutes). I think it's just a judgement call, however, if you think we should just wait to call the callback? // @jgeewax

callmehiphop: I think the usage patten of mixing callbacks with emitters just throws me off. What if instead of a callback we emitted a `running` (or similar) event

```js
job.cancel()
  .on('error', console.log)
  .on('running', function() {apiResponse}) // job has started (optional)
  .on('complete', function() {metadata}); // job is complete
  
```

stephenplusplus: Returning an emitter is an awesome idea. I will open a new issue for that, since implementing it will also affect GCE at the same time.

stephenplusplus: Issue opened: #1306

Issue #1260, opened 2016-04-24
nextQuery always returns a value


susanlinsfu:
I have 42 items which should be returned:

```
/*
query originally looks like this:
datastore.createQuery('Messages').filter('sender',username).limit(10).autoPaginate(false);
*/

  var query = nextQuery_var_passed_in_from_client;
  getmessages(query,function(err,data){
        if(err){
          console.log('error: ' + err);
        }
    });

});

//callback function
function getrecmessages(query,callback){
    datastore.runQuery(query, function(err, entities, nextQuery, apiResponse){
        if (err) {
            return(callback(err));
        }

        if (nextQuery) {
          console.log('nextQuery = ' + JSON.stringify(nextQuery) );
            if(entities){
                console.log(entities);
            }else{
               console.log('no entities');
	     }
        } else {
            // No more results exist.
            console.log('NextQuery IS NULL' );
        }
    });
}
```

The query first gets the first 10 items, then the next 10, then the next 10, then the remaining 2(42 items). The problem then is when I make the last query to get the last 2 items it returns a nextQuery. As a result it looks as if there is more items even though there are none.  I have gone over my code and I can't see anything that is causing this so I am thinking it has something to do with gcloud. I am expecting that when the last query is made no nextQuery should be returned by the datastore. Is this correct?

I am using gcd-rpc emulator and gcloud-node 0.31.0.

stephenplusplus:
I believe there are issues with the emulator. @pcostell Do you think is related to the issue we found before?

pcostell:
Yes I believe this is the same issue as before. However, even if there are no more results Cloud Datastore will still return an end cursor which may be useful. So I'm re-echoing my concern that nextQuery isn't the right API for when `more_results` is `MORE_RESULTS_AFTER_LIMIT` or `MORE_RESULTS_AFTER_CURSOR`. Importantly, unlike other APIs the Cloud Datastore `end_cursor` is not a traditional `next_page_token`.

stephenplusplus:
What would be the ideal way of handling pagination? Sorry if I missed your explanation before, feel free to link me.

stephenplusplus:
// just dropping a link to #1237 as that was the issue we discovered the API/emulator mismatch

pcostell:
Coming back to this:

I think we need the following:

On more_results = NOT_FINISHED, set next_query. Properly track offset/limit so that this next_query has these set.
On any other more_results, do not set next_query.
[...]

PR #1295, opened 2016-05-06
datastore: pagination changes

### LINKED ISSUE #1293: Show a manual pagination workflow in Datastore docs
There should be a more complete/realistic example on https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.32.0/datastore of manually paginating through the results of a query.

stephenplusplus:
Fixes #1293
RE: #1260 
RE: https://github.com/GoogleCloudPlatform/gcloud-node/issues/1260#issuecomment-217513314
#### To Dos
- [x] @pcostell sign off
- [x] Docs
- [x] Tests
  - [x] System
  - [x] Unit

Changes:
- `autoPaginate` is gone
- `apiResponse` from `run` and `runQuery` is gone
- We only continue running a query automatically for a user if Datastore returns `NOT_FINISHED`
- In place of a `nextQuery` object, we return an `info` object, which contains
  - `moreResults`
  - `endCursor` (if provided by Datastore)

### DIFF
--- a/test/docs.js
+++ b/test/docs.js
@@ -111,9 +111,12 @@ describe('documentation', function() {
       global: global
     };
 
+    // For {module:datastore} docs.
     function FakeExpress() {
       return {
-        get: function() {}
+        get: function(route, callback) {
+          callback({ query: {} }, {});
+        }
       };
     }
 


pcostell:
This seems like a reasonable approach to me.

Wouldn't this fit it with all the existing pagination + streaming infrastructure if we return a `next_query` when `more_results == NOT_FINISHED` (and then let the system handle everything for us)?

In the case of other `more_results` values, just return the info object and no `next_query`.

stephenplusplus:
> Wouldn't this fit it with all the existing pagination + streaming infrastructure if we return a next_query when more_results == NOT_FINISHED (and then let the system handle everything for us)?

Yes, we can do that, but I would still remove `autoPaginate` as an option. We should always just run `NOT_FINISHED` automatically (right?). The only downside is the pagination/streaming infrastructure ignores further arguments, such as `apiResponse`, because if there are multiple API calls, a single `apiResponse` is somewhat confusing. I would just do a manual override and implement the streaming behavior inside of the `runQuery` method itself-- not a huge deal.

> In the case of other more_results values, just return the info object and no next_query.

I think I'm still seeing the value in returning `info` in place of `nextQuery`. An example of where we're introducing complexity is one of our tests which went from:

```js
var q = datastore.createQuery('Character')
  .hasAncestor(ancestor)
  .limit(5);

datastore.runQuery(q, function(err, firstEntities, nextQuery) {
  // ... get the next results ...
  datastore.runQuery(nextQuery, // ...
```

to:

```js
var q = datastore.createQuery('Character')
  .hasAncestor(ancestor)
  .limit(5);

datastore.runQuery(q, function(err, firstEntities, nextQuery) {
  // ... get the next results ...
  q.start(info.endCursor).limit(q.limitVal - firstEntities.length);
  datastore.runQuery(q, ...
```

If the user only needs to know the start cursor of the next query for returning to the FE:

```js
var q = datastore.createQuery('Character')
  .hasAncestor(ancestor)
  .limit(5);

datastore.runQuery(q, function(err, firstEntities, nextQuery) {
  hypotheticalHttpResponse.send({ startCursor: nextQuery.startVal });
```

There's always the option of getting it from the raw response as well:

```js
datastore.runQuery(q, function(err, firstEntities, nextQuery, apiResponse) {
  hypotheticalHttpResponse.send({ startCursor: apiResponse.batch.endCursor });
```

pcostell:
I'm confused about that test change:
```

var q = datastore.createQuery('Character')
  .hasAncestor(ancestor)
  .limit(5);

datastore.runQuery(q, function(err, firstEntities, nextQuery) {
  // ... get the next results ...
  q.start(info.endCursor).limit(q.limitVal - firstEntities.length);
  datastore.runQuery(q, ...
```

If we always continue the query when the backend returns `NOT_FINISHED`, the callback should never get a query where q.limitVal > firstEntities.length.

stephenplusplus:
Doesn't it get "MORE_RESULTS_AFTER_LIMIT"?

stephenplusplus:
Sorry if I misunderstood. The test checks that we only get 5 results first, then after the second query, we get the remaining. I was just duplicating the logic that we do (currently) when preparing the nextQuery:

```js
nextQuery.limit(limit - resp.batch.entityResults.length);
```

pcostell:
Yes in that test `info.more_results == MORE_RESULTS_AFTER_LIMIT`. But that means the limit should equal the number of entities returned. So `q.limitVal - firstEntities.length` would always be `<= 0`.

stephenplusplus:
So it's basically just a weird way of saying "remove the limit" then, right? The previous test didn't do that because it was done by `runQuery` automatically. But without `nextQuery` being built for the user already, the user can either make an entirely new query object, or remove the limit themselves from their previous query, which is what this test is doing.

pcostell:
I don't think so in this case. If you want to use this for paging, you would always set the same limit ("I want pages of size 50").

The logic for updating the limit and offset should only happen inside the clients and only when `more_results == NOT_FINISHED`. 

Clearing the limit after the first query is more like saying "I want the first 50 results, then after that I want all the results.". I don't think the client should make this easy -- if a user wants all the results, they should just not set a limit and ask for all the results.

stephenplusplus:
My mistake. It's just a fluke in this test case that it works. Ignoring the part where we always run `nextQuery` automatically, this is the logic I think we should go back to, in regards to how `nextQuery` is built: https://github.com/GoogleCloudPlatform/gcloud-node/blob/339cda2b14bb116c619a92788ebabdcd1c837e77/lib/datastore/request.js#L510 -- is there anything wrong with how we're setting the offset and limit in this case?

pcostell:
That logic makes sense. I think though that it gets confusing that basically your first query handles an offset, but then we generate more queries that don't. I think we shouldn't try to handle so much in the client and just expose the cursors -- The user can make their own decision about how to treat the first page and 2-n pages differently.

stephenplusplus:
>  I think though that it gets confusing that basically your first query handles an offset, but then we generate more queries that don't.

Can you elaborate on that?

pcostell:
It makes sense in the case of `NOT_FINISHED` -- basically the user says "I want 50 entities, starting with the 100th". This is **exactly** what we should give them. Then we should just give them a cursor and say "here's where to start if you want to continue." Providing a nextQuery assumes that they want another 50 entities starting at the 150th. Giving them a cursor makes the user be explicit about what they are doing.

stephenplusplus:
Thanks. So at this point, you prefer this:

```js
var q = datastore.createQuery('Character')
  .hasAncestor(ancestor)
  .limit(5);

datastore.runQuery(q, function(err, firstEntities, info) {
  // ... get the next results ...
  var nextQuery = datastore.createQuery('Character')
    .hasAncestor(ancestor)
    .start(info.endCursor);

  datastore.runQuery(q, ...
```

I'm also a fan of forcing users to be explicit, but I do like providing sane defaults and making things as easy as possible. Isn't it a relatively safe assumption that if a user asked for 50, that they expect the "nextQuery" would be prepared to get the next 50? The user will always have the choice of creating a new query or simply overriding the limit if it's wrong.

pcostell:
Possibly, but we are also assuming that they don't want an offset either. Since we are expecting them to not actually call this in the same context (given the httpresponse to the user and back), they'll likely need to recreate the query anyways.

I would consider making it easy to issue the same query multiple times in the same request a non-goal of the client. It will encourage users to try managing batching themselves. Instead we should let the server handle it's own resource management. It will also make it easier to switch to a streaming setup in the future without changing how the customer interacts with the client. In particular, if we have a streaming RPC it would be less efficient to fetch entities in multiple batches as we'll have to create and tear down these connections, rather than streaming as many results as we can.

stephenplusplus:
Alright, I'll go forward with the info object and manual stream override of `runQuery`. Will ping again for another review soon!

stephenplusplus:
@pcostell ptal!

lib/datastore/index.js
```diff
@@ -192,19 +192,41 @@ var util = require('../common/util.js');
...
+ *     }
+ *
+ *     // Respond to the front end with the contacts and the cursoring token
+ *     // from the query we just ran.
+ *     var frontEndResponse = {
+ *       contacts: entities
+ *     };
+ *
+ *     if (info.moreResults !== 'NO_MORE_RESULTS') {
+ *       frontEndResponse.startCursor = info.endCursor;
```
stephenplusplus: Not sure this is the best naming convention for this to make the point clear. Should I just return `frontEndResponse.endCursor` and then the reader assumes that's being passed in as `req.query.startCursor` (found above)?

Additionally, maybe there's a better name than simply `info`?

pcostell: Maybe nextPageCursor?

lib/datastore/request.js
```diff
@@ -400,7 +401,7 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
  * //-
  * var query = datastore.createQuery('Lion');
  *
- * datastore.runQuery(query, function(err, entities) {});
+ * datastore.runQuery(query, function(err, entities, info, apiResponse) {});
```
pcostell: In this invocation, isn't apiResponse only ever the last API response?

stephenplusplus: It is. I'm not sure if it's better to include all responses, the last response, or no response at all. Returning all would mean returning an array, which might only have a single element most of the time. Any thoughts on which way we should go?

pcostell: Since we force the autobatching, I would probably vote for none at all until we get a FR with a case asking for it (then we'll know what people are looking for).

stephenplusplus:
@pcostell feel free to take another look. I believe I've covered everything we're going for.

pcostell:
LGTM

stephenplusplus:
Great! I'm glad we are finally doing things the Datastore-approved way. Thanks for being patient with my questions!

pcostell:
Haha thank you for being patient with my answers :-). Hopefully it makes user's lives easier. Let me know what the feedback is and we can evaluate if the server API should change at all to accommodate.

lib/datastore/index.js
```diff
@@ -192,19 +192,41 @@ var util = require('../common/util.js');
...
+ *       return;
+ *     }
+ *
+ *     // Respond to the front end with the contacts and the cursoring token
+ *     // from the query we just ran.
+ *     var frontEndResponse = {
+ *       contacts: entities
+ *     };
+ *
+ *     if (info.moreResults !== 'NO_MORE_RESULTS') {
```
callmehiphop: WDYT about moving this to some kind of static-y constant-y value (as well as any other values here)?

```js
info.moreResults === datastore.NO_MORE_RESULTS
``

File: lib/datastore/query.js [hunk 4]
```diff
@@ -316,35 +287,27 @@ Query.prototype.offset = function(n) {
...
  * //-
- * // If you omit the callback, `run` will automatically call subsequent queries
- * // until no results remain. Entity objects will be pushed as they are found.
+ * // If you omit the callback, you will get the matching entities in a readable
+ * // object stream.
  * //-
  * query.run()
  *   .on('error', console.error)
  *   .on('data', function (entity) {})
+ *   .on('info', function(info) {})
```
callmehiphop: Any reason we don't just emit with this with `end`? Or is it assumed that if `info` is emitted the query is done running?

stephenplusplus: We can't override `end` since it's an event coming from the stream's plumbing.

lib/datastore/index.js
stephenplusplus: I will do that, but I can't help myself from seeing this before and after:

```diff
-if (info.moreResults !== 'NO_MORE_RESULTS') {
+if (info.moreResults !== datastore.NO_MORE_RESULTS) {
```

stephenplusplus:
@callmehiphop updated with the constant.

lib/datastore/index.js
```diff
@@ -374,6 +398,8 @@ Datastore.int = function(value) {
   return new entity.Int(value);
 };
 
+Datastore.NO_MORE_RESULTS = 'NO_MORE_RESULTS';
```
callmehiphop: Aren't there more options than just this one?

stephenplusplus: There are these 4: https://github.com/GoogleCloudPlatform/gcloud-node/issues/1260#issuecomment-217508895

We won't ever return `NOT_FINISHED`, since we always run it. I'll add the others.

stephenplusplus:
Added the other constants!

lib/datastore/query.js
```diff
@@ -316,35 +287,27 @@ Query.prototype.offset = function(n) {
  *     stream instance is returned.
  * @param {?error} callback.err - An error returned while making this request
  * @param {object[]} callback.entities - A list of entities.
- * @param {?object} callback.nextQuery - If present, query with this object to
- *     check for more results.
- * @param {object} callback.apiResponse - The full API response.
+ * @param {object} callback.info - An object useful for pagination.
+ * @param {?string} callback.info.endCursor - Use this in a follow-up query to
+ *     begin from where these results ended.
+ * @param {string} callback.info.moreResults - Datastore responds with one of:
```
callmehiphop: Should we document the constants here?

lib/datastore/query.js
callmehiphop: Ah, derp.

lib/datastore/request.js
```diff
@@ -389,9 +383,15 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
  *     stream instance is returned.
  * @param {?error} callback.err - An error returned while making this request
  * @param {object[]} callback.entities - A list of entities.
- * @param {?object} callback.nextQuery - If present, query with this object to
- *     check for more results.
- * @param {object} callback.apiResponse - The full API response.
+ * @param {object} callback.info - An object useful for pagination.
+ * @param {?string} callback.info.endCursor - Use this in a follow-up query to
+ *     begin from where these results ended.
+ * @param {string} callback.info.moreResults - Datastore responds with one of:
```
callmehiphop: Re: constants

package.json
```diff
@@ -127,6 +127,7 @@
     "coveralls": "^2.11.2",
     "deep-strict-equal": "^0.1.0",
     "dox": "^0.8.0",
+    "express": "^4.13.4",
```
callmehiphop: Is this just for testing docs?

stephenplusplus: Yeah :(

lib/datastore/query.js
stephenplusplus: Done!

package.json
callmehiphop: Dang, could we just stub it and add that to our sandbox object?

stephenplusplus: Good idea, done!

test/docs.js
```diff
@@ -111,10 +111,17 @@ describe('documentation', function() {
       global: global
     };
 
+    function FakeExpress() {
+      return {
+        get: function() {}
```
callmehiphop: I think we need execute the callback in order to verify the code is valid, otherwise we'll get false positives.. unfortunately that requires us to stub parts of the request and response objects.

stephenplusplus: That's a good point, although I think we should just skip it in this case. Testing the docs code is always a sync-only smoke test. No matter how we slice this, it will leave some holes (or over-complicate the example, or over-complicate the code).

callmehiphop: I'm not trying to be a stickler, but we've always strictly followed the "no example code untested" rule. That being said, this particular piece of code would be very easy to smoke test, so I don't see why we don't.

e.g.

```js
get: function(route, callback) {
  callback({ query: {} }, {});
}
```

stephenplusplus: We don't really though, any code that runs inside of a callback isn't tested. That's been a known limitation of the docs testing since it was implemented. Just using this an example since I worked on it today-- everything starting here isn't tested: https://github.com/GoogleCloudPlatform/gcloud-node/blob/f52d1a8268af90b6132b03ada3bee19b5cd139a0/lib/storage/file.js#L1200

I've added your code, but this still doesn't save us... it just gets us one layer deeper in the execution path.

callmehiphop: Ah, I never realized that before, that kinda stinks, but I suppose it can't be helped.

callmehiphop: This discussion may be better suited for a new issue, but could we mock some kind of general response (at least for non-grpc calls) via `mitm` that would allow us to traverse all the layers?

stephenplusplus: If mitm did allow such a thing (it probably does), a response that works for all requests wouldn't be possible. And there's still the error handling that would get skipped. We would get to a point of managing a response for each method, that we've crossed over into the territory of what our unit tests do.

If you think going down that road is worth it, go ahead and open up an issue!

PR #1317, opened 2016-05-14
Append newline to end of extensionHeaders if necessary

### LINKED ISSUE #1316: storage/file#getSignedUrl: It shouldn't be necessary to add newline to end of extensionHeaders
The way storage/file#getSignedUrl computes the signature, it is necessary for the API client to end `extensionHeaders` with a newline, otherwise it won't correspond to the server's calculated signature. This should be handled automatically by gcloud-node.

aknuds1:
Fixes #1316.

### DIFF
--- a/lib/storage/file.js
+++ b/lib/storage/file.js
@@ -1252,9 +1252,14 @@ File.prototype.getSignedUrl = function(options, callback) {
       return;
     }
 
-    options.extensionHeaders = options.extensionHeaders || '';
-    if (options.extensionHeaders != '' && !options.extensionHeaders.endsWith('\n')) {
-      options.extensionHeaders = options.extensionHeaders + '\n';
+    var extensionHeadersString = '';
+    if (options.extensionHeaders != null) {
+      for (var headerName in options.extensionHeaders) {
+        extensionHeadersString += format('{name}:{value}\n', {
+          name: headerName,
+          value: options.extensionHeaders[headerName],
+        })
+      }
     }
     var sign = crypto.createSign('RSA-SHA256');
     sign.update([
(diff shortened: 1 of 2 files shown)


stephenplusplus:
Thank you for this! What do you think about accepting an object for headers, e.g.

```js
file.getSignedUrl({
  extensionHeaders: {
    'x-goog-acl': 'public-read',
    'x-goog-meta-foo': 'bar'
  },
  // ...

File.prototype.getSignedUrl = function(options, callback) {
  ...

  var extensionHeadersString;

  if (options.extensionHeaders) {
    for (var headerName in options.extensionHeaders) {
      extensionHeadersString += format('{name}:{value}\n', {
        name: headerName,
        value: options.extensionHeaders[headerName]
      });
    }
  }

  ...
};
```

aknuds1:
@stephenplusplus I also thought that an object would be better, but I thought about maintaining backwards compatibility and that it would be easier to fix the existing interface (extension headers represented by a string) if it's supposed to be supported in the future.

stephenplusplus:
Thank you for thinking about the backwards compatibility issue, but it's okay to break it if we've discovered an easier API. We will just bump the minor to follow along with semver.

aknuds1:
Do you want me to update the PR to represent extension headers as an object then?

stephenplusplus:
If you're up for it, that would be great 👍

aknuds1:
Will do then.

aknuds1:
@stephenplusplus Done!

stephenplusplus:
Merged in via https://github.com/GoogleCloudPlatform/gcloud-node/commit/17e510b32b0b2dfb1901c5d4b918df10daf36cc9. Thanks again @aknuds1!

PR #1322, opened 2016-05-16
Storage write streams should be able to use a pre-existing uri

### LINKED ISSUE #1321: Storage write streams should be able to use a pre-existing uri
Allow users to pass through an already existing uri when creating a write stream.

```js
// create resumable url
file.createResumableUpload((err, uri) =>  {
    // hang on to uri somehow
});

// create a new stream with the same uri
const config = { uri };
bucket.file('zebra.jpg').createWriteStream(config);
```

PR with the enhancement coming shortly!

bhstahl:
Allow users to pass through an already existing uri when creating a write stream.

``` js
// create resumable url
file.createResumableUpload((err, uri) =>  {
    // hang on to uri somehow
});

// create a new stream with the same uri
const config = { uri };
bucket.file('zebra.jpg').createWriteStream(config);
```

Closes #1321

### DIFF
--- a/test/storage/file.js
+++ b/test/storage/file.js
@@ -1073,6 +1073,21 @@ describe('File', function() {
       writable.end('data');
     });
 
+    it('should pass through a provided uri', function(done) {
+      var options = {
+        uri: 'https://www.googleapis.com/afakeuri'
+      };
+
+      var writable = file.createWriteStream(options);
+
+      file.startResumableUpload_ = function(stream, options_) {
+        assert.deepEqual(options_.uri, options.uri);
+        done();
+      };
+
+      writable.write('data');
+    });
+
     describe('validation', function() {
       var data = 'test';
 


lib/storage/file.js
```diff
@@ -1528,7 +1528,7 @@ File.prototype.startResumableUpload_ = function(dup, options) {
     metadata: {}
   }, options);
 
-  var uploadStream = resumableUpload({
+  var config = {
```
stephenplusplus: Feel free to just add the uri from the `options` object. It will be ignored if falsy from gcs-resumable-upload.

```js
  ...
  uri: options.uri
});
```

stephenplusplus:
Thank you! Could you add a line in the jsdocs for:

- Bucket#upload
- File#createWriteStream

Something like:

```js
 * @param {string} options.uri - The URI for an already-created resumable
 *     upload. See {module:storage/file#createResumableUpload}.
```

bhstahl:
Ah yes, good call @stephenplusplus. Updated!

stephenplusplus:
Coveralls is going crazy again. Sorry I missed this the first time, could you add a test to check that the uri option is passed through to gcs-resumable-upload?

stephenplusplus:
Also, just released gcs-resumable-upload@0.6.0 with URI support -- can you upgrade the package.json to use this new one?

bhstahl:
👍 on it @stephenplusplus

stephenplusplus:
Merged in via https://github.com/GoogleCloudPlatform/gcloud-node/commit/3d6fa8a164c7307c6d62197e69ac3adde37037bf. Thanks @bhstahl!

PR #1332, opened 2016-05-19
core (stream-router): allow setting maxApiCalls

### LINKED ISSUE #1281: Allow limiting API calls across all methods
RE: https://github.com/GoogleCloudPlatform/gcloud-node/issues/1260#issuecomment-216542762

> Let's figure out how to do maxApiCalls across the entire library so that people know how to limit things. (In many APIs, the number of calls is directly related to money.)
> \- @jgeewax

stephenplusplus:
Fixes #1281 
#### To Dos
- [x] Docs: http://stephenplusplus.github.io/gcloud-node/#/docs/master/gcloud
- Tests
  - Unit
    - [x] stream-router
    - [x] datastore/request#get
    - [x] datastore/request#runQuery
  - [x] System - bigquery

This adds a setting for `maxApiCalls` on all methods that use the `nextQuery` style pagination, as well as Datastore's `get` and `runQuery`. Let me know if I missed any methods!

``` js
datastore.runQuery(query, { maxApiCalls: 1 }, function(err, results) {
  // 1 request was made.
  // If a 2nd request needed to be made, it wasn't.
  // This callback receives the results from 1 request.
});
```

Given the example above, is there a creative way to let the user know that we had to interrupt making API calls?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -314,6 +314,10 @@ var apis = {
  * of now, we support a request interceptor, allowing you to tweak all of the
  * API request options before the HTTP request is sent.
  *
+ * Note: If you are using the `maxApiCalls` option with a method, your request
+ * interceptor will still be called, even if the `maxApiCalls` limit was
+ * reached. The request to the API will still be prevented.
+ *
  * See the example below for more.
  *
  * @alias module:gcloud


stephenplusplus:
After a quick @callmehiphop lint, I'll get going on tests.

callmehiphop:
Can we consolidate the retry code into a centralized place? The code looks good, my concern being that we've essentially duplicated it across files.

stephenplusplus:
Good idea! Pushed an update.

lib/common/util.js
```diff
@@ -545,3 +545,43 @@ function normalizeArguments(globalContext, localConfig, options) {
...
+ * @param {number} options.maxApiCalls - The maximum number of API calls to
+ *     make.
+ */
+function iterator(makeRequestFn, options) {
+  var stream = through.obj();
+
+  var requestsMade = 0;
+  var requestsToMake = -1;
+
+  options = options || {};
```
callmehiphop: Could we add an option for `onreading` or something similar? I feel like we could consolidate all the `stream.on('reading', function() { ... });` lines.

lib/common/util.js
```diff
@@ -545,3 +545,43 @@ function normalizeArguments(globalContext, localConfig, options) {
...
+
+/**
+ * Throttle requests according to a `maxApiCalls` limit.
+ *
+ * @param {function} makeRequestFn - The function that will be called.
+ * @param {object=} options - Configuration object.
+ * @param {number} options.maxApiCalls - The maximum number of API calls to
+ *     make.
+ */
+function iterator(makeRequestFn, options) {
```
callmehiphop: Not sure how I feel about calling this an iterator, my initial reaction was, *"like a generator?"*. How about something like `createLimiter`?

```js
var limiter = uitl.createLimiter(makeRequest, { ... });
```

stephenplusplus: Sure, changed. The idea was that, like a generator/iterator, it requires an explicit "next()" that triggers "doing the next thing". In this case, that's "makeRequest". Just words, though. Whatever makes sense to you.

lib/common/util.js
stephenplusplus: I did initially try to fit that in. My issue was that it took away too much control from the originating method. Having it called automatically is nice, but (when I was thinking along the lines of this being a generator), it blanketed the call, and made the next "iterator.makeRequest()" feel unexpected.

There's also the issue of where the original options the function is called with come from. If util.createLimiter handles it, we would have to provide the initial objects to `createLimiter`.

I just made a change that made the stream a "stream-events"-ified one inside of `util.createLimiter` itself. Now we don't need to worry about that logic wherever we use a limiter.

package.json
```diff
@@ -144,7 +144,7 @@
   "scripts": {
     "docs": "node ./scripts/docs.js",
     "lint": "jshint lib/ system-test/ test/ && jscs lib/ system-test/ test/",
-    "test": "npm run docs && mocha test/*/*.js test/index.js test/docs.js",
+    "test": "npm run docs && mocha test/docs.js test/index.js test/*/*.js",
```
stephenplusplus: I believe Mockery is not properly releasing a stub, or messing with the dependency tree altogether. Letting the docs run first is required to get around this.

stephenplusplus:
@callmehiphop I'm currently struggling with how to approach system tests for this change. Since this is a modification to streamrouter, it affects many APIs. So, that means we would have to test every method using stream router again that it honors the `maxApiCalls` property.

I added request interceptors to increment a `numRequestsMade` counter. This worked for a while, but then I ran into another hurdle; not all requests need to make multiple API calls, so limiting it to 1 and verifying only 1 request was sent isn't exactly testing what we think it is. If we get clever and set `maxApiCalls` to 0, we could technically get passing tests, though that seems like we're not really writing system tests at that point, because we're blocking all interactions with the "system".

callmehiphop:
@stephenplusplus maybe it's a more appropriate place for unit tests then? I think as long as we have a couple system-tests, and really solid unit tests it should be ok. WDYT?

stephenplusplus:
Sure, lmk if you think the unit tests I have written already could be improved. I'll add some systems.

stephenplusplus:
System test for BigQuery added (other APIs aren't very compatible), docs published: http://stephenplusplus.github.io/gcloud-node/#/docs/master/gcloud

system-test/bigquery.js
```diff
@@ -98,6 +98,31 @@ describe('BigQuery', function() {
...
+        return reqOpts;
+      }
+    });
+
+    bigquery.getDatasets({ maxApiCalls: maxApiCalls }, function(err) {
+      assert.ifError(err);
+
+      // Even though the request interceptor is called, the request can still be
+      // prevented if the `maxApiCalls` limit was reached.
+      numRequestsMade -= 1;
```
callmehiphop: This seems like an interesting quirk, would it be too difficult to try and prevent interceptors from executing in the event that the `maxApiCalls` was hit?

callmehiphop: Assuming it is kind of a pain to prevent, at the very least could we document this in the interceptors section? It probably won't affect anyone, but I think on the off chance some one is using this for capturing some kind of metric/etc., it might be good to have it documented somewhere.

stephenplusplus: Added!

stephenplusplus: ... to the docs.

PR #1334, opened 2016-05-20
vision: return errors

### LINKED ISSUE #1333: formatMethodMap[type] is not a function in vision lib
Just got this error calling `vision.detectText`

```
Debug: internal, implementation, error 
    TypeError: Uncaught error: formatMethodMap[type] is not a function
    at /home/uli/Projetos/projeto-a/server/node_modules/gcloud/lib/vision/index.js:345:39
    at Array.map (native)
    at /home/uli/Projetos/projeto-a/server/node_modules/gcloud/lib/vision/index.js:388:18
    at Array.map (native)
    at /home/uli/Projetos/projeto-a/server/node_modules/gcloud/lib/vision/index.js:377:10
    at /home/uli/Projetos/projeto-a/server/node_modules/gcloud/lib/vision/index.js:158:5
    at Object.handleResp (/home/uli/Projetos/projeto-a/server/node_modules/gcloud/lib/common/util.js:126:3)
    at /home/uli/Projetos/projeto-a/server/node_modules/gcloud/lib/common/util.js:446:12
    at Request.onResponse [as _callback] (/home/uli/Projetos/projeto-a/server/node_modules/retry-request/index.js:117:7)
    at Request.self.callback (/home/uli/Projetos/projeto-a/server/node_modules/gcloud/node_modules/request/request.js:200:22)
```

I am using node v4.3.2 with hapi framework

stephenplusplus:
Fixes #1333
#### To Dos
- [x] Docs
- Tests
  - [x] Unit
  - [x] System

If the Vision API fails to annotate an image, it returns an `error` (see https://cloud.google.com/vision/reference/rest/v1/images/annotate#annotateimageresponse) for each failed annotation. However, it will still annotate what it can.

``` js
vision.detect('malformed-image.jpg', ['faces', 'labels'], function(err, detections) {
  err === null
  detections === {
    faces: [/* successful annotation */],
    labels: [], // nothing, this failed
    errors: [/* error(s) from the request */]
  }
});
```

To add on errors to a single feature was a bit more complex:

``` js
vision.detect('malformed-image.jpg', ['faces'], function(err, detections) {
  err === null
  detections === [] // nothing, this failed
  detections.errors = [/* error(s) from the request */]
});
```

### DIFF
--- a/lib/vision/index.js
+++ b/lib/vision/index.js
@@ -271,7 +271,10 @@ Vision.prototype.annotate = function(requests, callback) {
  * });
  *
  * //-
- * // To handle errors that occurred, check the length of the `errors` array.
+ * // It's possible for part of your request to be completed successfully, while
+ * // a single feature request was not successful. Each returned detection will
+ * // have an `errors` array, including any of these errors which may have
+ * // occurred.
  * //-
  * vision.detect('malformed-image.jpg', types, function(err, detections) {
  *   if (detections.faces.errors.length > 0) {


stephenplusplus:
@callmehiphop honest talk time. The code in Vision.prototype.detect_ is kind of out of control. Considering this latest refactoring: https://gist.github.com/stephenplusplus/aa0c82b4939e5901680e41c9a89e8c03 -- can you think of a way I can trim down the insanity?

callmehiphop:
What are your feelings on moving things outside of the function? We have several maps and inner functions used to massage data that could easily live as a static/private members.

stephenplusplus:
I made an attempt at moving them outside as static members. The issue I ran into was losing access to variables that help the original response, the types the user originally requested, and the shorthand->formalApiName lookup tables. I'll just leave things like they are in the gist for now, and move on with this PR. If I can't come up with a good refactoring plan before we merge, I'll be sure to open an issue about it to hit in the future.

stephenplusplus:
@callmehiphop ready to go. See the updated first post for the `.errors` possible weirdness.

callmehiphop:
I think overall the code looks good, I'm not sure how else you would solve this problem. My only real critique is that the documentation for this seems like it would be easily overlooked. I know we generally don't highlight error handling in the doc examples, but what are your feelings on creating an additional example showing how to handle errors when dealing with a malformed image?

stephenplusplus:
Example added!

lib/vision/index.js
```diff
@@ -269,6 +269,15 @@ Vision.prototype.annotate = function(requests, callback) {
  *   //   }
  *   // ]
  * });
+ *
+ * //-
+ * // To handle errors that occurred, check the length of the `errors` array.
```
callmehiphop: Not trying to be a stickler, but can we elaborate here a little more? Like if we explained how in the event of a malformed image, certain detection types may fail while others will succeed, I think a little more context would be helpful.

stephenplusplus: Updated, ptal!

PR #1341, opened 2016-05-25
datastore: concat transactional mutations

### LINKED ISSUE #1339: Can't save and delete in a transaction
I asked this question on stackoverflow, but am also posting here as an issue (http://stackoverflow.com/questions/37401751/cant-save-and-delete-in-a-transaction).

```
var cachedTransaction;
datastore.runInTransaction(function(transaction, done) {
            cachedTransaction = transaction;

            var keys = [
                datastore.key([PARENT_TABLE, p1]),
                datastore.key([PARENT_TABLE, p2]),
                datastore.key([PARENT_TABLE, p1, CHILD_TABLE, p2])
            ];

            transaction.get(keys, function(err, entities) {

                    if (err) {
                        transaction.rollback();
                    }

                    if (entities) {
                        transaction.save([{
                            key: datastore.key([PARENT_TABLE, p1]),
                            method: 'update',
                            data: {
                                p: 100
                            }
                        }, {
                            key: datastore.key([config.PARENT_TABLE, p2]),
                            method: 'update',
                            data: {
                                p: 100
                            }
                        }]);

                        transaction.delete([

                            datastore.key([PARENT_TABLE, p1, CHILD_TABLE, p2]),
                            datastore.key([PARENT_TABLE, p2, CHILD_TABLE, p1])

                        ]);

                    } else {
                        console.log('Entities are empty');
                    }
                    console.log('commiting ... ');
                    done();
                    console.log('commited ');

                } else {
                    transaction.rollback();
                }
            });

    },
    function(err) {
        console.log('commited complete. Was it successful ? ');
        if (err) {
            console.log('err : ' + err);
            cachedTransaction.rollback();
        }
    });
```
With the following code I am getting two entities from a parent table and one entity from a child table (I print some stuff with this child entity not shown in code). I then update the two parent entities and delete two child entities (one which I did not get get from my original response). The result is that the parent entities do get updated, but the two children entities do NOT get deleted. I get no error messages.

If I comment out the save portion of the code which updated the parents, then it will delete the child entities, but I need to update the parent and delete the children. How can I do this? I am using gcloud-node 0.34.0 and running off the actual Google Cloud Datastore.

Thanks!

stephenplusplus:
Fixes #1339

### DIFF
--- a/lib/datastore/transaction.js
+++ b/lib/datastore/transaction.js
@@ -376,9 +376,11 @@ Transaction.prototype.commit_ = function(callback) {
       }
     })
 
-    // Group entities together by action (delete or save).
+    // Group entities together by method: `save` mutations, then `delete`. Note:
+    // `save` mutations being first is required to maintain order when assigning
+    // IDs to incomplete keys.
     .sort(function(a, b) {
-      return a.method > b.method ? 1 : a.method < b.method ? -1 : 0;
+      return a.method < b.method ? 1 : a.method > b.method ? -1 : 0;
     })
 
     // Group arguments together so that we only make one call to each method.
--- a/system-test/datastore.js
+++ b/system-test/datastore.js
@@ -690,48 +690,52 @@ describe('Datastore', function() {
       var key = datastore.key(['Company', 'Google']);
       var incompleteKey = datastore.key('Company');
 
-      datastore.runInTransaction(function(t, tDone) {
-        t.delete(deleteKey);
-
-        t.save([
-          {
-            key: key,
-            data: { rating: 10 }
-          },
-          {
-            key: incompleteKey,
-            data: { rating: 100 }
-          }
-        ]);
-
-        tDone();
+      datastore.save({
+        key: deleteKey,
+        data: {}
       }, function(err) {
         assert.ifError(err);
 
-        // Incomplete key should have been given an ID.
-        assert.strictEqual(incompleteKey.path.length, 2);
+        datastore.runInTransaction(function(t, tDone) {
+          t.delete(deleteKey);
 
-        async.parallel([
-          // The key queued for deletion should have been deleted.
-          function(done) {
-            datastore.get(deleteKey, function(err, entity) {
-              assert.ifError(err);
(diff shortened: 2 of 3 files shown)


lib/datastore/transaction.js
```diff
@@ -418,10 +417,12 @@ Transaction.prototype.commit_ = function(callback) {
 
   // Take the `req` array built previously, and merge them into one request to
   // send as the final transactional commit.
-  var reqOpts = this.requests_.reduce(function(acc, req) {
-    return extend(true, acc, req);
-  }, {});
-
+  var reqOpts = {
+    mutations: this.requests_.reduce(function(acc, req) {
```
callmehiphop: Is there a module we can use for this (or just add it to util?) I find myself doing this for BigTable in places, as well.

callmehiphop:
Any reason why we can't add a e2e test for this?

lib/datastore/transaction.js
stephenplusplus: 👍 Added a commit.

lib/datastore/transaction.js
```diff
@@ -375,9 +376,11 @@ Transaction.prototype.commit_ = function(callback) {
       }
     })
 
-    // Group entities together by action (delete or save).
+    // Group entities together by method: `save` mutations, then `delete`. Note:
+    // `save` mutations being first is required to maintain order when assigning
+    // IDs to incomplete keys.
     .sort(function(a, b) {
-      return a.method > b.method ? 1 : a.method < b.method ? -1 : 0;
+      return a.method < b.method ? 1 : a.method > b.method ? -1 : 0;
```
stephenplusplus: Found a bug here. This groups methods by their name (`save` or `delete`). Previously, it was just going alphabetical, but I switched it so that `saves` come first. Important for the reasons explained in the comment above.

stephenplusplus:
> Any reason why we can't add a e2e test for this?

In place of this, I just fixed the system test. It previously was deleting something that never existed, which was giving false positives. Now, we create that entity before deleting it.

PR #1358, opened 2016-06-06
core: correctly locate maxRetries option

### LINKED ISSUE #1357: maxRetries not passed through to GrpcService
Caught here: https://github.com/GoogleCloudPlatform/gcloud-node/issues/1356#issuecomment-223667333

stephenplusplus:
Fixes #1357

We were trying to find the user's `maxRetries` option on the `config` argument, instead of `options`.

Additionally, `retry-request` expects an argument named `retries`, where we were calling it `maxRetries`.

### DIFF
--- a/lib/common/grpc-service.js
+++ b/lib/common/grpc-service.js
@@ -156,7 +156,7 @@ function GrpcService(config, options) {
     this.grpcCredentials = grpc.credentials.createInsecure();
   }
 
-  this.maxRetries = config.maxRetries;
+  this.maxRetries = options.maxRetries;
 
   var apiVersion = config.apiVersion;
   var service = this.service = config.service;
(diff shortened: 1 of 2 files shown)


stephenplusplus:
@coveralls what can we do to make you happy?

callmehiphop:
I know we decided to not add system tests for every service in regards to this, but maybe one grpc system test test would benefit us here. WDYT?

stephenplusplus:
What would the system test look like?

callmehiphop:
I was going to suggest `mitm` but I now remember that doesn't work with grpc requests. Is this basically untestable (outside of unit tests)?

stephenplusplus:
Yeah, we are just going on trust that retry-request works. Which sounds worse than it is. That's essentially what we do with all of our dependencies... we scope them out, see if the code looks good, that it's tested... then if we trust it, we let its own test suite worry about covering features that it says it offers.

But yeah, testing that we do retry logic when we get a retry-able response from gRPC is quite untestable. I don't think we can control when gRPC will give us, so it's just meant to unit testing-land.

PR #1359, opened 2016-06-06
docs: recommend lambda-packager

### LINKED ISSUE #1297: Fail gracefully if gRPC isn't available
Re #1287 , seems like there are cases where gRPC won't be able to be initialized. I understand if a service that *requires* gRPC (ie, Bigtable) failed in these cases, but if an unrelated module fails because of the unused gRPC import, that makes me sad....

Can we try to fail gracefully if gRPC isn't installed in the environment for whatever reason?

stephenplusplus:
Fixes #1297

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -10,7 +10,7 @@ If you are not running this client on Google Compute Engine, you need a Google D
 
 ## How do I use `gcloud` with AWS Lambda?
 
-Our library uses some native libraries, which must be recompiled against the platform running your application. For AWS Lambda, we recommend using [lambda-pacakger](https://github.com/tomdale/lambda-packager) prior to deploying your code.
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -2,10 +2,19 @@
...
 
+
 ## I'm not using Compute Engine. What do I need to do?
 
 If you are not running this client on Google Compute Engine, you need a Google Developers service account. To create a service account, please see our [Authentication][auth-guide] guide.
 
+
+## How do I use `gcloud` with AWS Lambda?
+
+Our library uses some native libraries, which must be recompiled against the platform running your application. For AWS Lambda, we recommend using [lambda-pacakger](https://github.com/tomdale/lambda-packager) prior to deploying your code.
```
callmehiphop: `lambda-packager`*

stephenplusplus: Woops. Done.

PR #1364, opened 2016-06-07
storage: allow custom file encryption

### LINKED ISSUE #1361: Support custom encryption keys for storage
Can we support custom encryption keys for files uploaded?

Spec is here: https://cloud.google.com/storage/docs/encryption#customer-supplied

stephenplusplus:
Fixes #1361
#### To Dos
- [x] Support a custom key (instead of generating one from a passphrase)
- [x] Support `options.encryption` in gcs-resumable-upload
- [x] Tests
  - [x] System
  - [x] Unit

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "ent": "^2.2.0",
     "extend": "^3.0.0",
     "gce-images": "^0.2.0",
-    "gcs-resumable-upload": "^0.7.0",
+    "gcs-resumable-upload": "^0.7.1",
     "google-auto-auth": "^0.2.4",
     "google-proto-files": "^0.2.1",
     "grpc": "^0.14.1",


lib/storage/file.js
```diff
@@ -277,6 +277,28 @@ function File(bucket, name, options) {
...
+    code += ' ';
+  }
+
+  var key = cipher.update(code, 'utf8', 'base64') + cipher.final('base64');
+
+  return {
+    key: key,
+    hash: crypto.createHash('sha256').update(key, 'base64').digest('base64')
+  };
+};
```
stephenplusplus: @jgeewax not sure if this is what you had in mind, PTAL.

daspecster: Possible silly question, but couldn't you just hash `code` if it's less than 32 bytes? That could potentially upgrade their key a little bit as well.

jgeewax: When you upload a file, you should be able to provide a custom key. Is that doable here?

stephenplusplus: The whole File object will be "encrypted" when you call `encrypt`:

```js
var unencryptedFile = bucket.file('my-file')
var file = bucket.file('my-file').encrypt('passphrase') // encrypted

// encrypt an unencrypted file reference
unencryptedFile.encrypt('passphrase')
```

Here are some ways to encrypt as a one liner:

```js
// way #1
file.encrypt('passphrase').save('new data', function(err) {})

// way #2
fs.createReadStream('local-file.txt')
  .pipe(file.encrypt('passphrase').createWriteStream())

// way #3
var file = bucket.file('new-file').encrypt('passphrase')
bucket.upload('local-file.txt', { destination: file }, function(err) {})
```

If way #&shy;3 is too complex, we can add an option to `bucket.upload`:

```js
bucket.upload('local-file.txt', { encrypt: 'passphrase' }, function(err) {})
```

jgeewax:
Just to confirm, the goal here is to be able to set a key when uploading a file. Eventually we might want a key-derivation library involved, but for now this is just about setting a key on a per-file basis.

stephenplusplus:
Let me know how far off https://github.com/GoogleCloudPlatform/gcloud-node/pull/1364#discussion_r66319409 is from what you're looking for.

stephenplusplus:
If we only support uploading a file, the user would be blocked out from other operations, like downloading it. The way this PR works is:

- the user provides a passphrase to `file.encrypt()` 
- every API request made with a `file` object (file.download, .copy, .createWriteStream, etc) will pass the generated key so that these operations don't fail

lib/storage/file.js
```diff
@@ -980,6 +1002,50 @@ File.prototype.download = function(options, callback) {
...
+ *     // err: Bad Request
+ *
+ *     // Try again with the encrypted file:
+ *     encryptedFile.download(function(err, contents) {
+ *       // contents.toString() === 'data'
+ *     });
+ *   });
+ * });
+ */
+File.prototype.encrypt = function(code) {
```
jgeewax: @stephenplusplus I'm worried about calling this "encrypt" and hinting to people that we're encrypting this before sending it to GCS, which we aren't. We're sending the file in plaintext, with an instruction to Google saying "please encrypt with this key for me...".

Can we make this a "setting" of sorts rather than the verb "encrypt" ?

stephenplusplus: Totally.

```js
var file = bucket.file('my-file', {
  key: 'passphrase'
})

file.setKey('passphrase')
```

?

stephenplusplus:
@callmehiphop PTAL!

callmehiphop:
@stephenplusplus I think npm might have put your most recent release of `gcs-resumable-upload` to bed, both travis and my local are failing to install `v0.7.x`, however `v0.6.x` still seems to be installing fine in master.

stephenplusplus:
Fixed... v0.7.1.

callmehiphop:
This might be a slightly off-topic question, but the upstream documentation for sending encryption keys also mentions a separate set of headers for `rewrite` operations, which we don't appear to support - any reason why that is?

stephenplusplus:
`rewrite` is new! At least, since we had a Storage deep dive. Mind making an issue to remind us to add support?

callmehiphop:
Done! #1373

PR #1367, opened 2016-06-07
storage: use fast-crc32c conditionally

### LINKED ISSUE #1299: Is the optional dependency really worth it?
Re https://github.com/pkgcloud/pkgcloud/issues/452

The issue is that optional deps show a scary warning message, often confusing our users (hell, even I've been confused at first until I read the error message and see "warning").

The CRC32 stuff is performance optimization only, right? If so, what's stopping us from wrapping the require statement in a try/catch ? And defaulting to the required dependency?

Then when people go "this is too slow! make it faster!" we say "no problem, just run `npm install <whatever>`" ? Sounds like something to put in the FAQ or a "performance" section of the docs when people go looking for that?

stephenplusplus:
Fixes #1299 

I published `hash-stream-validation@0.2.0` which try/catches fast-crc32c. This uses it, and leaves a note to users about how to get faster speeds.

### DIFF
--- a/lib/storage/bucket.js
+++ b/lib/storage/bucket.js
@@ -923,6 +923,11 @@ Bucket.prototype.makePublic = function(options, callback) {
  * `options.resumable`. *Resumable uploads are enabled by default if your input
  * file is larger than 5 MB.*
  *
+ * For faster crc32c computation, you must manually install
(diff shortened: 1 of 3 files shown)


stephenplusplus:
@jgeewax -- can you ping the firebase team who might be interested in this change?

PR #1369, opened 2016-06-08
datastore: use transaction object

### LINKED ISSUE #633: Inaccessible `transaction.rollback`
From https://github.com/GoogleCloudPlatform/gcloud-node/pull/627/files#r31352472

##### Question 1

A rollback can only be done on a transaction that has been committed (I think?), but that's a problem:

```js
dataset.runInTransaction(function(transaction, commit) {
  commit();
}, function(err) {
  // no `transaction` here to call `rollback` on.
});
```

An option would be to remove the last callback.

```js
dataset.runInTransaction(function(transaction, commit) {
  commit(function(err) {
    if (err) {
      transaction.rollback();
    }
  });
});
```

##### Question 2
Is there a use case for rolling back a transaction that wasn't just created? We currently don't support this.

stephenplusplus:
Fixes #633 

The goal of this change is to give more visibility into the operations being performed when using transactions.
#### Before

``` js
datastore.runInTransaction(function(transaction, done) {
  transaction.save({
    key: key,
    data: { value: true }
  });

  done();
}, function(err) {
  // Where is this error coming from?
});
```
#### After

``` js
var transaction = datastore.transaction();

transaction.run(function(err) {
  if (err) {
    // Error handling omitted.
  }

  transaction.save({
    key: key,
    data: { value: true }
  });

  transaction.commit(function(err) {
    if (!err) {
      // Transaction committed successfully.
    }
  });
});
```

The diff scariness doesn't reflect reality... I organized method names alphabetically. The code bodies themselves didn't change, just their location in the file & doc blocks.

@pcostell is this a good change in terms of usability?

### DIFF
--- a/lib/datastore/transaction.js
+++ b/lib/datastore/transaction.js
@@ -201,8 +201,8 @@ Transaction.prototype.commit = function(callback) {
     if (err) {
       // Rollback automatically for the user.
       self.rollback(function() {
-        // Provide the response items from the failed commit to the user. Even
-        // a failed rollback should be transparent.
+        // Provide the error & API response from the failed commit to the user.
+        // Even a failed rollback should be transparent.
         // RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1369#discussion_r66833976
         callback(err, resp);
       });
@@ -330,7 +330,7 @@ Transaction.prototype.rollback = function(callback) {
     method: 'rollback'
   };
 
-  this.request_(protoOpts, {}, function(err, resp) {
+  this.request_(protoOpts, function(err, resp) {
     self.skipCommit = true;
 
     callback(err || null, resp);
@@ -344,10 +344,12 @@ Transaction.prototype.rollback = function(callback) {
  * @param {function} callback - The function to execute within the context of
  *     a transaction.
  * @param {?error} callback.err - An error returned while making this request.
+ * @param {module:datastore/transaction} callback.transaction - This transaction
+ *     instance.
  * @param {object} callback.apiResponse - The full API response.
  *
  * @example
- * transaction.run(function(err) {
+ * transaction.run(function(err, transaction) {
  *   // Perform Datastore transactional operations.
  *   var key = datastore.key(['Company', 123]);
  *
@@ -377,15 +379,15 @@ Transaction.prototype.run = function(callback) {
     method: 'beginTransaction'
   };
 
-  this.request_(protoOpts, {}, function(err, resp) {
+  this.request_(protoOpts, function(err, resp) {
     if (err) {
-      callback(err, resp);
+      callback(err, null, resp);
       return;
     }
 
     self.id = resp.transaction;
 
-    callback(null, resp);
+    callback(null, self, resp);
   });
 };
 
--- a/test/datastore/transaction.js
+++ b/test/datastore/transaction.js
@@ -119,25 +119,6 @@ describe('Transaction', function() {
     });
   });
 
-  describe('createQuery', function() {
-    it('should return query from datastore.createQuery', function() {
-      var args = [0, 1, 2, 3];
-      var createQueryReturnValue = {};
-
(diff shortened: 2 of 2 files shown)


pcostell:
What happens if there is an error? Do we rollback automatically? Does the user have to do this manually?

stephenplusplus:
Nope, and I'll add a to do item to make sure we rollback if the `commit` call fails. So, how does this case work:

```js
transaction.commit(function(err) {
  // Behind the scenes, we see that this call fails.
  // So before executing this callback, we should automatically rollback.

  // What happens if the user didn't catch that part in the docs, and tries
  // to rollback themselves?
  if (err) {
    transaction.rollback(...);
  }
});
```

pcostell:
We just need to make sure to swallow any errors on the automatic rollback -- it will get an error:

```
INVALID_ARGUMENT: The referenced transaction has expired or is no longer valid.
```

stephenplusplus:
Added a commit to automatically rollback after a failed commit: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1369/commits/41eec4740f9fc68190690e5fac071f8ea089b850

lib/datastore/transaction.js
```diff
@@ -195,14 +199,29 @@ Transaction.prototype.commit = function(callback) {
 
   this.request_(protoOpts, reqOpts, function(err, resp) {
     if (err) {
-      callback(err, resp);
+      // Rollback automatically for the user.
+      self.rollback(function(err) {
+        if (err) {
+          err.message = [
```
pcostell: I think we can just eat this error -- right now rollbacks are fine as best effort.

stephenplusplus: Done 👍

pcostell:
To verify -- if an error happens during the BeginTransaction call, all attempted calls as part of that transaction will fail, correct?

pcostell:
Are we worried users may try to do something like:

`datastore.transaction().run(....)` and then use datastore instead of transaction inside that run method?

stephenplusplus:
> if an error happens during the BeginTransaction call, all attempted calls as part of that transaction will fail, correct?

```js
var transaction = datastore.transaction();

transaction.run(function(err) {
  if (err) {
    // BeginTransaction failed
  }
});
```

Calling anything else from `transaction` will return whatever error the API will.

> Are we worried users may try to do something like: `datastore.transaction().run(....)` and then use datastore instead of transaction inside that run method?

That's a fair concern. We could return the `transaction` as the second argument to the `run` callback:

```js
transaction.run(function(err, tx) { /* transaction === tx */ });
datastore.transaction().run(function(err, transaction) {});
```

jgeewax:
Just to be sure -- there are cases where someone might want to start a transaction, and then from *outside* the transaction execute a query of sorts, right?

If there is **no good reason** to *ever* do:

```js
transaction.run( (err) => {
  datastore.run(somethingElse);
  transaction.commit();
});
```

then we should throw an error if people try to do something like that.... ?

stephenplusplus:
I think I'm a bit confused by your example. `run` is only a method on a Transaction object, so
`datastore.run(somethingElse);` would throw.

jgeewax:
Sorry - I mean it as an example of "running another query from inside that block". Is there a time when that would make sense? That'd I'd `get` an entity while inside a transaction block, but not want to use the transaction ?

stephenplusplus:
Gotcha. In terms of a use case, I'm not sure.

I don't think I'd block those operations, though. I'm not sure how, actually... choosing where a function can be executed or not:

```js
transaction.run(function (err) {
  datastore.get(key, function(err, entity) {})
  //        ^-- need to detect this was executed inside of `transaction.run`
})
```

If this is what we want, maybe @callmehiphop and I can think of something clever. However, I think the user would expect if they have something that works in one place, it would work wherever they choose to put it.

stephenplusplus:
@callmehiphop PTAL

PR #1400, opened 2016-06-30
Docs: Remove extra dash

kevinresol:
This causes the doc generator to think that the param name is a dash


stephenplusplus:
Keep em coming!

PR #1401, opened 2016-06-30
Docs: WritableStream -> WriteStream

kevinresol:
I am not pretty sure about this one though.
Because it writes `WriteStream` in [another place](https://github.com/GoogleCloudPlatform/gcloud-node/blob/464ff6c21e48cfde62a3ed1928e1bebe9e99a0db/lib/bigquery/table.js#L385)

Actually, in `fs` it is called [`fs.WriteStream`](https://nodejs.org/api/fs.html#fs_class_fs_writestream) but the underlying class/interface is [`stream.Writable`](https://nodejs.org/api/stream.html#stream_class_stream_writable). I am not sure what it is referring to...

### DIFF
--- a/lib/bigquery/table.js
+++ b/lib/bigquery/table.js
@@ -356,7 +356,7 @@ Table.prototype.copy = function(destination, metadata, callback) {
  *
  * @resource [Tabledata: list API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/v2/tabledata/list}
  *
- * @return {ReadStream}
+ * @return {ReadableStream}
  *
  * @example
  * var through2 = require('through2');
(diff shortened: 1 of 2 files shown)


stephenplusplus:
I think we should favor `WritableStream`, which is just to say we are returning a stream that is writable. Do you mind basically reversing this PR and changing it in the other file that you found `WriteStream`?

kevinresol:
No problem. Do you want to change ReadStream to ReadableStream as well then?

kevinresol:
I went ahead changing readstream to readable stream. Also added the missing return type for File.create(read/write)stream

stephenplusplus:
Perfect! Thanks again :)

PR #1405, opened 2016-07-01
Incorrect missing project ID error URL

bamapookie:
The URL in the error message does not actually take you to the authentication documentation.  Fixed in the fork.

### DIFF
--- a/test/common/util.js
+++ b/test/common/util.js
@@ -19,6 +19,7 @@
 var assert = require('assert');
 var duplexify;
 var extend = require('extend');
+var format = require('string-format-obj');
 var googleAuth = require('google-auto-auth');
 var is = require('is');
 var mockery = require('mockery-next');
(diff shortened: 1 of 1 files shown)


bamapookie:
I signed it.

stephenplusplus:
Thank you for finding this. This is kinda silly, but we have an 80 char rule, so if you don't mind, can you make these changes?

```diff
var extend = require('extend');
+ var format = require('string-format-obj');
var googleAuth = require('google-auto-auth');

...

var errorMessage = format([
  'Sorry, we cannot connect to Google Cloud Services without a project ID.',
  'You may specify one with an environment variable named "GCLOUD_PROJECT".',
  'See {baseUrl}/{path} for a detailed guide on creating an authenticated',
  'connection.'
].join(' '), {
  baseUrl: 'https://googlecloudplatform.github.io/gcloud-node/#',
  path: '/docs/guides/authentication'
});

var missingProjectIdError = new Error(errorMessage);
```

bamapookie:
Thanks, @stephenplusplus.  I was wondering how I was going to make that URL fit. :)

stephenplusplus:
Awesome, thanks :) And sorry again, but I forgot that we have [this test](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/test/common/util.js#L112) this is going to fail now. That will need the same update. I'm pretty confident that should be it!

stephenplusplus:
Perfect, thank you very much! 💃

PR #1410, opened 2016-07-05
enabling usage with access token only

### LINKED ISSUE #1346: Allow specifying access_token to use
There are situations where you can end up with just the access_token.  (eg. non-refereshable token for an endusers via oauth webflow).

There doesn't seem to be any way to place that raw token and initialize gcloud node. (gcloud-java recently added it in:
https://github.com/GoogleCloudPlatform/gcloud-java/issues/1029 

here are some references for the older googleapis node client:
https://github.com/google/google-api-nodejs-client#making-authenticated-requests

and a citation about overriding the config
https://github.com/GoogleCloudPlatform/gcloud-node/issues/678#issuecomment-131238194

juliusse:
Fixes #1346 
This pull request enables gloud-node users to only specify an access_token for authentication.
As @stephenplusplus suggested, the solution just not uses the auto-auth client, but creates a simple one that just sets the access_token.

### DIFF
--- a/lib/common/token-based-auth-client.js
+++ b/lib/common/token-based-auth-client.js
@@ -18,9 +18,11 @@
  * @module common/token-based-auth-client
  */
 
+'use strict';
+
 /*
  * creates a simpler auth client that works only with an accessToken
- * 
+ *
  * @param {string=} accessToken - The accessToken to use
  */
 function tokenBasedAuthClient(accessToken) {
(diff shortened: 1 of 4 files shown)


stephenplusplus:
Thanks for this :) There are a couple edge cases, though:

  1. Calls made with gRPC APIs won't use the token (Datastore, Logging, Pub/Sub for now)
  1. There are other libraries we use-- [gce-images](http://gitnpm.com/gce-images) & [gcs-resumable-upload](http://gitnpm.com/gcs-resumable-upload)-- that would need to be updated to accept just an access token (and we would have to provide the access token the user give us to those libraries)

@murgatroid99 is there a way to provide gRPC with just an access token?

stephenplusplus:
I'm going to close the PR (thank you, though!) and let the discussion carry on in #1346. As mentioned above, it's rather tricky to implement.

sedouard:
Now that gRPC can easily set metadata (and therefore the access_token) is it in the plan to add access_token to the RESTful interfaces?

PR #1413, opened 2016-07-08
datastore: support long entity IDs

### LINKED ISSUE #1412: Unable to get entity by key id (int64) from Datastore
For example, datastore already contains an entity with id = 100000000000001234 (18 digits). The integers in JavaScript (numbers without a period or exponent notation) are considered accurate up to 15 digits. 100000000000001234 will be 100000000000001000. Maybe the [Long type](https://www.npmjs.com/package/long) support (in the cloud.datastore.key method) is a way how to resolve this issue.

stephenplusplus:
Fixes #1412 
#### To Dos
- [x] Docs
- [x] Tests
  - [x] System
  - [x] Unit
#### Example

``` js
var id = gcloud.datastore.int('100000000000001234');
var key = datastore.key(['Kind', id]);

// everything from this point should work normally.
```

### DIFF
--- a/packages/datastore/src/entity.js
+++ b/packages/datastore/src/entity.js
@@ -45,7 +45,7 @@ var InvalidKeyError = createErrorClass('InvalidKey', function(opts) {
 entity.KEY_SYMBOL = Symbol('KEY');
 
 /**
- * Build a Datastore Double object.
+ * Build a Datastore Double object. For long doubles, a string can be provided.
  *
  * @constructor
  * @param {number} value - The double value.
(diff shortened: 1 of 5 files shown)


stephenplusplus:
@callmehiphop wdyt?

callmehiphop:
LGTM

callmehiphop:
Kind of side note/after thought. But datastore is cool with accepting strings? Might be worth mentioning that Bigtable will bring in a new dependency for supporting Int64 values. Not sure if datastore will accept those as an ID, but the library we're using will convert a string into an Int64.

stephenplusplus:
Good thinking, it might make more sense to use that here, then. I'll wait for the Bigtable merge and play with it.

PR #1420, opened 2016-07-12
updated cluster links and option handling

callmehiphop:
This fixes 2 things:
1. The old documentation for creating clusters went missing, so updates were made to several links
2. Certain options were not being preserved when not using default application credentials.

### DIFF
--- a/test/bigtable/index.js
+++ b/test/bigtable/index.js
@@ -110,6 +110,24 @@ describe('Bigtable', function() {
       fakeUtil.normalizeArguments = normalizeArguments;
     });
 
+    it('should leave the original options unaltered', function() {
+      var fakeOptions = {
+        a: 'a',
+        b: 'b',
+        c: 'c',
+        zone: 'zone'
+      };
+
+      var bigtable = new Bigtable(fakeOptions);
+      var options = bigtable.calledWith_[1];
+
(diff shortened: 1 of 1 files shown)


callmehiphop:
@lesv we noticed that one can optionally provide an API key instead of a keyfile or default application creds. Is there a standard way of providing said key?

stephenplusplus:
Throwing a test on here would be good.

test/bigtable/index.js
```diff
@@ -110,6 +110,24 @@ describe('Bigtable', function() {
       fakeUtil.normalizeArguments = normalizeArguments;
     });
 
+    it('should leave the original options unaltered', function() {
```
callmehiphop: @stephenplusplus PTAL!

stephenplusplus: Can we just see if the same object the user provides (or at least the same properties they defined) make it to `GrpcService.call(this, config, options)`. Sorry if this test is doing that already and I didn't see the connection.

callmehiphop: It does that, I'm doing a `for in` loop on the original options and asserting that they're in the options that get passed to `GrpcService`.

stephenplusplus: Woo hoo!

PR #1427, opened 2016-07-13
logging: allow for encoding options

callmehiphop:
Looks like we're getting some [logging failures on Travis](https://travis-ci.org/GoogleCloudPlatform/gcloud-node/jobs/144255867#L7627).

@stephenplusplus Not sure if I'm way off on this one, but essentially I've added support for "encoding" options when writing a log. LMKWYT and if it looks ok, I'll update the documentation and add some unit tests!

### DIFF
--- a/lib/logging/entry.js
+++ b/lib/logging/entry.js
@@ -133,7 +133,7 @@ Entry.fromApiResponse_ = function(entry) {
  *
  * @private
  */
-Entry.prototype.toJSON = function() {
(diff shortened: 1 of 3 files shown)


stephenplusplus:
I think the problem here was at one point during the PR, I renamed `serialize` to `stringify`, but didn't update it properly. That was a fail I would like to correct in the form of a new PR.

PR #1430, opened 2016-07-13
Create a new issue template

stephenplusplus:
cc: @callmehiphop @jgeewax

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -9,7 +9,7 @@ Please run down the following list and make sure you've tried the usual "quick f
   
 If you are still having issues, please be sure to include as much information as possible:
 
-#### Environment Info
+#### Environment details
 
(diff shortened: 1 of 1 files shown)


callmehiphop:
I really like the tone of the template, but I'm not so sure about the general format. I think I find the use of checkboxes a bit puzzling, especially where we're asking for technical details.

I was kind of expecting something closer to

```md
### OS

### Node version

### gcloud version

### Steps to reproduce
```

Although this isn't nearly as pretty and doesn't convey any sort of tone.

jgeewax:
I kind of like the checkboxes... It means they can check off that they did those things... However it's not a checkbox until they create the issue , so that's... iffy. 

I think the `### OS` , etc is probably a bit too sparse. Any chance we can do something in the middle?

stephenplusplus:
PTAL

callmehiphop:
Looks great to me!

Issue #1411, opened 2016-07-07
Modularization Release Process


stephenplusplus:
Here's the current plan regarding how to publish updates to the core `gcloud` module or the future `@gcloud/{service}` sub-modules. Thoughts very welcome!

- - - 

An npm run script will be created to release a module:

```sh
$ npm run release [SERVICE_NAME] --major|minor|patch
```

Here's what happens if you wanted to publish the _core gcloud module_ with a **major** update.

```sh
$ npm run release --major

# runs:
rm -rf node_modules && npm install
npm run lint
npm run test
npm run system-test
npm version major
rm -rf node_modules && npm publish
git push origin master --follow-tags
```

Here's what happens if you wanted to release a **major** update to the _BigQuery API_.

```sh
$ npm run release bigquery --major

# runs:
cd lib/bigquery
rm -rf node_modules && npm install
npm run lint
npm run test
npm run system-test
npm version major
rm -rf node_modules && npm publish
git push origin master --follow-tags

# Because it was a MAJOR increment (breaking changes):
npm uninstall --save @gcloud/bigquery
npm install --save @gcloud/bigquery # to depend on the new release
git commit -am 'Update @gcloud/bigquery'
npm run release [minor|major] # depending on if gcloud-node is < or > 1.0
```

Here's what happens if you wanted to release a **patch** update to the _BigQuery API_.

```sh
$ npm run release bigquery --patch

# runs:
cd lib/bigquery
rm -rf node_modules && npm install
npm run lint
npm run test
npm run system-test
npm version patch
rm -rf node_modules && npm publish
git push origin master --follow-tags
```

### How about the docs?
The core package will no longer have its own JSON files. Instead, each service will have a directory in gh-pages that the core package doc pages will pull from:

```
$ git checkout gh-pages
$ ls
  docs
  |_ json
    |_ { existing version directories for pre-modularization JSON files }
    |_ bigquery
      |_ v0.1.0
        |_ dataset.json
        |_ index.json
        |_ job.json
        |_ table.json
        |_ types.json
```

#### How does the docs site load the correct JSON files from the modules?

Still thinking! Possible tools that can help:

- JSON.parse(http://gitnpm.com/gcloud/{$scope.activeVersion}/json/dependencies)
- The `semver` client-side library

callmehiphop:
Overall I think it looks great, I only have a couple things to add in regards to the docs.

[...]

PR #1431, opened 2016-07-13
Allow individually publishable API client libraries.

jmdobry:
Lots of discussion floating around on _why_ a PR like this is necessary, like GoogleCloudPlatform/gcloud-common#138.
### Summary of changes

**tldr;**

Move the code for each API client into its own publishable package folder within this repository (rename `lib/` to `packages/`), change a lot of relative import paths to use named package imports, and switch to a `@google-cloud/*` naming scheme.

**Long version**

What is this PR? Right now `gcloud-node` combines all API client library code into a single installable NPM package. This means that to update client code for an individual API, the entire `gcloud-node` package must be version bumped and re-published.

This PR makes it possible for each sub-folder within the `packages/` folder to be published to NPM as its own package. Currently, `lib/index.js` imports all of the `lib/` sub-folders using relative paths, pulling that code directly into the published `gcloud-node` package. Furthermore, individual `lib/` sub-folders sometimes import code from each other using relative paths. In order for `packages/vision/` to be cleanly publishable as an individual NPM package, any code it imports from outside of itself (e.g. `require('../storage/file.js')` or `require('../lib/common/util.js'`), needs to be changed to named package imports (e.g. `require('@google-cloud/storage').File` or `require('@google-cloud/common').util`. This decouples the versioning and publishing of sub-packages from one another.

With this change, some individual packages will be free to move forward toward 1.0.0 while others might need to stay back at alpha or beta versions. The actual version number of the `gcloud-node` becomes less meaningful, as now the "big" package will just depend on a bunch of individually versioned and published packages.

**Before**

```
npm install gcloud
```

``` js
var vision = gcloud.vision();
```

**After**

```
npm install google-cloud
```

``` js
var gcloud = require('google-cloud');
var vision = gcloud.vision();
```

or

```
npm install @google-cloud/vision
```

``` js
var vision = require('@google-cloud/vision')();
```

or

```
npm install @google-cloud/pubsub
```

``` js
var pubsub = require('@google-cloud/pubsub')();
```

etc.
### List of changes
- [x] - `package.json`
  - [x] - Change `"name"` from `"gcloud"` to `"google-cloud"`
  - [x] - Remove existing `"dependencies"` and switch to depending on `@google-cloud/storage`, `@google-cloud/pubsub`, etc.
- [x] - `lib/index.js`
  - [x] - Update `require` calls from using relative paths to using package names, e.g. `require('./storage/index.js')` => `require('google-cloud-storage')`
- [x] - Sub-packages
  - [x] - Add individual `package.json` files for each `lib/<folder>`
  - [x] - Each sub-package has its own dependencies, unused dependencies removed from google-cloud-common
  - [x] - Each sub-package will use a unique User-Agent header, for better usage tracking.
  - [x] - Set sub-package `package.json` "version" field to 0.1.0
  - [x] - Update sub-package use of relative import paths (`../lib/common`) to use `@google-cloud/common`, e.g. `require('../lib/common/utils.js')` => `require('@google-cloud/common').util`
  - [x] - `@google-cloud/common` turned into a module, rather than dependents importing file paths.
- [x] - Tests
  - [x] - When mocking dependencies, update paths to reflect changes in `lib/`
  - [x] - **Note: There seems to be a bug in `mockery` that doesn't like the symbolic links that are now necessary for local development, so I switched to `proxyquire`, which works fine with symbolic links.**
- [x] - `scripts/`
  - [x] - Added `link.sh`, which sets up the symbolic links necessary for local development
  - [x] - Added `unlink.sh`, the reverse of `link.sh`
  - [x] - Added a simple `publish.sh` script, referenced in each sub-package `package.json` file, which can publish a sub-package to NPM with the appropriate legal files included.
- [x] - Docs
  - [x] - Update README.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,11 +20,13 @@
     "through2": "^2.0.0"
   },
   "scripts": {
+    "link-modules": "node ./scripts/link",
+    "install-module-dependencies": "node ./scripts/install",
+    "postinstall": "npm run link-modules && npm run install-module-dependencies",
     "docs": "node ./scripts/docs.js",
     "lint": "jshint packages/ system-test/ test/ && jscs packages/ system-test/ test/",
     "test": "npm run docs && mocha test/docs.js packages/*/test/*.js",
     "system-test": "mocha packages/*/system-test/*.js --no-timeouts --bail",
-    "postinstall": "node ./scripts/install",
     "coveralls": "istanbul cover _mocha --report lcovonly -- --no-timeouts --bail packages/*/test/*.js -R spec && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage"
   },
   "license": "Apache-2.0",
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -547,7 +547,9 @@ util.extendGlobalConfig = extendGlobalConfig;
 function normalizeArguments(globalContext, localConfig, options) {
   options = options || {};
 
-  var config = util.extendGlobalConfig(globalContext.config_, localConfig);
+  var globalConfig = globalContext && globalContext.config_;
+
+  var config = util.extendGlobalConfig(globalConfig, localConfig);
 
   if (options.projectIdRequired !== false && !config.projectId) {
     throw util.missingProjectIdError;
--- a/packages/google-cloud/system-test/google-cloud.js
+++ b/packages/google-cloud/system-test/google-cloud.js
@@ -18,7 +18,7 @@
 
 var assert = require('assert');
 var env = require('../../../system-test/env.js');
-var gcloud = require('../lib');
+var gcloud = require('../');
 var path = require('path');
 
 // Test used to confirm we can perform a successful API operation.
--- a/packages/logging/system-test/logging.js
+++ b/packages/logging/system-test/logging.js
@@ -25,10 +25,10 @@ var prop = require('propprop');
 var uuid = require('node-uuid');
 
 var env = require('../../../system-test/env.js');
-var BigQuery = require('@google-cloud/bigquery/index.js');
+var BigQuery = require('@google-cloud/bigquery');
 var Logging = require('../');
-var PubSub = require('@google-cloud/pubsub/index.js');
-var Storage = require('@google-cloud/storage/index.js');
+var PubSub = require('@google-cloud/pubsub');
+var Storage = require('@google-cloud/storage');
 
 describe('Logging', function() {
   var TESTS_PREFIX = 'gcloud-logging-test';
--- a/packages/translate/system-test/translate.js
+++ b/packages/translate/system-test/translate.js
@@ -21,7 +21,7 @@ var extend = require('extend');
 var prop = require('propprop');
 
 var env = require('../../../system-test/env.js');
-var Translate = require('../lib/translate/index.js').Translate;
+var Translate = require('../').Translate;
 
 var API_KEY = process.env.GCLOUD_TESTS_API_KEY;
 


stephenplusplus:
Thanks for this 🎉 

We should make sure we don't need to use `new`:

```js
var vision = require('google-cloud-vision');
var visionClient = vision();

// or more likely, some opts will be required:
var vision = require('google-cloud-vision');
var visionClient = vision({ projectId: '...', keyFilename: '...' });

// and what I would personally do to make it a one-liner:
var vision = require('google-cloud-vision')({ projectId: '...', keyFilename: '...' });
```

jmdobry:
Ah yes, we don't need `new`, I'll update my examples.

stephenplusplus:
Can you write up a little "Here's how this all works" bit?

Also, are there any similarities between the process outlined [here](https://github.com/GoogleCloudPlatform/gcloud-node/issues/1411) and how this PR works?

jmdobry:
> Can you write up a little "Here's how this all works" bit?

Will do.

> Also, are there any similarities between the process outlined [here](https://github.com/GoogleCloudPlatform/gcloud-node/issues/1411) and how this PR works?

In my PR I didn't really address the publish process as discussed in #1411. My PR is more about organizational changes required in the code itself to facilitate individually publishable API packages. I'll explain better in my summary that I will add to my PR description.

jmdobry:
I updated the PR description.

jmdobry:
Open questions are:

* Versioning
* Release process
* Documentation

stephenplusplus:
A couple of thoughts:

I think we are going to go with scoped packages (https://docs.npmjs.com/misc/scope), so `google-cloud` for the umbrella package, and `@google-cloud/storage` for a submodule.

--

> ```js
var vision = require('google-cloud-vision').Vision();
```

We should be able to lose the `Vision` property so that a user doesn't have to "ask" for vision twice:

```js
var vision = require('google-cloud-vision');
```

Is anything complicating that?

--

> - Versioning

I think the umbrella package should just be the next minor bump from where we are when this PR lands, as opposed to something like 1.0.0-beta.

For the module packages, I would start out at 1.0.0 on each.

--

> - Release process

I'd like Travis to be involved, but pushing to `npm` with Travis has consistently failed for us over the last year or so. This forces a manual release process from the terminal, so the ideas in #1411 should be free to implement after we land the separation of services from this PR.

The manual release process should look like this:

#### Release the umbrella package
```sh
# Start from freshly installed dependencies:
rm -rf node_modules && npm install

# Check for faulty code:
npm run lint && npm run test && npm run system-test

# Create a git tag and increment package.json
npm version major

# Publish to npm from a clean directory:
# (`rm -rf node_modules` is to work around an npm publish bug)
rm -rf node_modules && npm publish

# Push to GitHub:
git push origin master --follow-tags
```

#### Release a module package
```sh
# Start from a clean sub-module directory
cd lib/bigquery && rm -rf node_modules && npm install

# Check for faulty code:
npm run lint && npm run test && npm run system-test

# Create a git tag and increment package.json:
npm version major

# Publish to npm from a clean directory:
# (`rm -rf node_modules` is to work around an npm publish bug)
rm -rf node_modules && npm publish

# Push to GitHub:
git push origin master --follow-tags

# Because it was a major increment with breaking changes,
# we have to manually update the umbrella package's dependency
# on it:
cd ../../ # back to the umbrella package directory
npm uninstall --save @google-cloud/bigquery
npm install --save @google-cloud/bigquery
git commit -am 'Update @gcloud/bigquery'

# Release the umbrella package as:
#   - a minor bump -- if the umbrella package is pre-1.0
#   - a major bump -- if the umbrella package is post-1.0
# [ steps outlined in previous "Release the umbrella package" section ]
```

Once we decide on a versioning pattern (what the umbrella package and each module package starts at), a more specific guide can be whipped up.

--

> - Documentation

@callmehiphop !

jmdobry:
@stephenplusplus 

> I think we are going to go with scoped packages

I'll fix the package names.

>     var vision = require('google-cloud-vision').Vision();`
>
> We should be able to lose the Vision property so that a user doesn't have to "ask" for vision twice:
>
>     var vision = require('google-cloud-vision');
>
> Is anything complicating that?

Well, some submodules need access to constructors from other modules, with the most common example being the `File` constructor which is imported from the Storage submodule into other submodules.

If the overall service constructor *is* the exported module (i.e. `module.exports = Storage;`), then the only way to make the `File` constructor exportable is to attach it to the `Storage` constructor as a static property (i.e. `Storage.File = File;`), which I think is gross. This has always been a limitation of CommonJS, so if you have additional exports but you want the convenience of a default export you've had to attach additional exports to the default export. 

ES2015 module syntax allows a `default` export *and* named exports within a single module, so I switched the submodules to use named exports with the service constructor also being the `default` export as a forward-looking change (these submodules don't exist yet, so here we have a chance to set precedent), i.e.:

```js
// export {
//   Storage as default,
//   Storage,
//   File 
// }
exports.default = Storage;
exports.Storage = Storage;
exports.File = File;
```

So you could do:

```js
import { File, Storage } from '@google-cloud/storage';

const storage = Storage({
  // config
});
```

or

```js
import Storage from '@google-cloud/storage';
import { File } from '@google-cloud/storage';

const storage = Storage({
  // config
});
```

or in CommonJS

```js
const CloudStorage = require('@google-cloud/storage');
const { File, Storage } = CloudStorage;

const storage = Storage({
  // config
});
```

In general I prefer exporting constructors/factory functions, and letting the user instantiate, rather than the act of importing of a module having side effects (magically instantiated client with default options).

Thoughts?

stephenplusplus:
> Well, some submodules need access to constructors from other modules, with the most common example being the File constructor which is imported from the Storage submodule into other submodules.

I see that as our problem that we shouldn't make the user pay for. While gross, I'd prefer exposing the sub-type constructors on the parent type (`Storage.File`), or any other way, as long as the thing exported is the way to use the module directly.

> the act of importing of a module having side effects (magically instantiated client with default options).

It wouldn't. The way to instantiate the module would be to invoke the function returned by the require:

```js
var vision = require('@google-cloud/vision');
var visionClient = vision();

// (aka)
var vision = require('@google-cloud/vision')();
```

jmdobry:
Okay, I'll switch away from named exports.

jmdobry:
Another question: Do we want to move the unit and system tests for a particular submodule into that submodule's folder?

.jscsrc
```diff
@@ -21,6 +21,19 @@
   },
   "excludeFiles": [
     "system-test/data/*",
-    "test/testdata/*"
+    "test/testdata/*",
+    "lib/bigquery/node_modules/",
```
stephenplusplus: A glob can probably help here and the `.jshintignore`, maybe `"*node_modules/*`

lib/bigquery/dataset.js
```diff
@@ -28,13 +28,13 @@ var nodeutil = require('util');
  * @type {module:common/service-object}
  * @private
  */
-var ServiceObject = require('../common/service-object.js');
+var ServiceObject = require('@google-cloud/common/service-object.js');
```
stephenplusplus: Do you think `common` should be more like a module and define access to these files in code, i.e.:

```js
var ServiceObject = require('@google-cloud/common').ServiceObject;
```

lib/bigtable/.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules
```
stephenplusplus: We can probably just catch all of these in the root .gitignore so each directory doesn't need a .gitignore.

I'd also like to figure out only having `AUTHORS`, `CONTRIBUTORS`, and `COPYING` in the root directory of the repo, then including them at the time of `npm publish`, but this might have to be solved later when we get to the release process tooling.

lib/common/util.js
```diff
@@ -38,11 +38,11 @@ var streamEvents = require('stream-events');
...
 var uniq = require('array-uniq');
 
-/** @const {object} gcloud-node's package.json file. */
-var PKG = require('../../package.json');
+/** @const {object} @google-cloud/common's package.json file. */
+var PKG = require('./package.json');
 
 /** @const {string} User agent. */
-var USER_AGENT = 'gcloud-node/' + PKG.version;
+var USER_AGENT = 'google-cloud-node/' + PKG.version;
```
stephenplusplus: Is it useful to include the version of google-cloud/common in the user agent? Maybe we should expose a setter so each module can set a more useful header, i.e. `setUserAgent('@google-cloud/storage', STORAGE_PKG.version)`

lib/bigtable/.gitignore
stephenplusplus: Aaaand I just caught that as part of the publish script. Sorry for not reading the full PR before commenting :) We can probably still save on the `.gitignore` file if we use globbing in the root dir's .gitignore.

lib/datastore/query.js
```diff
@@ -282,7 +282,7 @@ Query.prototype.offset = function(n) {
  * @param {string} options.consistency - Specify either `strong` or `eventual`.
  *     If not specified, default values are chosen by Datastore for the
  *     operation. Learn more about strong and eventual consistency
- *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-google-cloud-datastore).
+ *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-@google-cloud/datastore).
```
stephenplusplus: Find/replace fail? 😄

lib/datastore/request.js
```diff
@@ -244,7 +244,7 @@ DatastoreRequest.prototype.delete = function(keys, callback) {
  * @param {string} options.consistency - Specify either `strong` or `eventual`.
  *     If not specified, default values are chosen by Datastore for the
  *     operation. Learn more about strong and eventual consistency
- *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-google-cloud-datastore).
+ *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-@google-cloud/datastore).
```
stephenplusplus: Oops.

lib/datastore/request.js
```diff
@@ -424,7 +424,7 @@ DatastoreRequest.prototype.insert = function(entities, callback) {
...
```
stephenplusplus: Woops.

stephenplusplus:
We originally used proxyquire, but switched away for some reason. As long as it's working now, I'm glad to ditch `mockery`, so thanks for that :)

test/docs.js
```diff
@@ -34,6 +34,7 @@ function runCodeInSandbox(code, sandbox) {
       timeout: 1000
     });
   } catch(err) {
+    console.log(err.stack);
```
stephenplusplus: Left in by mistake?

Gormartsen:
Subscribed.
I just want to add a feedback here on this topic.

Right now to use gcloud.compute I need to install all other APIs as well. 

It would be awesome if I can install only gcloud-compute with gce-images only.

lib/bigquery/dataset.js
jmdobry: SGTM

lib/bigtable/.gitignore
jmdobry: Done

lib/common/util.js
jmdobry: Done. Each submodule will have a unique user agent, for better usage tracking.

lib/datastore/query.js
jmdobry: Fixed.

lib/datastore/request.js
jmdobry: Fixed.

lib/datastore/request.js
jmdobry: Fixed.

test/docs.js
jmdobry: Fixed.

.jscsrc
jmdobry: Done.

lib/bigquery/dataset.js
jmdobry: Done.

stephenplusplus:
Thanks for the updates! 🍻 

@callmehiphop what are your thoughts on the un-checked TODOs from the opening post? My thoughts: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1431#issuecomment-233048141

Gormartsen:
> @stephenplusplus I'd like Travis to be involved, but pushing to npm with Travis has consistently failed for us over the last year or so.

I am working on CI system right now, where release flow could be separated task connected or not with github context.

If you up to try, I can invest my time to setup a release process example on fork and we can move from there.

Btw: release process could be done from user defined system via SSH.

Let me know what you think.

jmdobry:
@stephenplusplus Do you think I should further modify the directory structure to be something like what Babel has?

```
lib/
  bigquery/
    index.js
    package.json
  bigtable/
    index.js
    package.json
  index.js
system-test/
test/
package.json
```

=>

```
packages/
  google-cloud/
    src/
      index.js
    system-test/
    test/
    package.json
  bigquery/
    src/
      index.js
    system-test/
    test/
    package.json
  bigtable/
    src/
      index.js
    system-test/
    test/
    package.json
```

stephenplusplus:
@jmdobry that looks nice to me. I like keeping the tests close to the source. If someone is contributing a Bigtable bug fix, they wouldn't have to run 1,000 unrelated tests just to see if their change works. And working in an isolated context from the rest of the code sounds better to me. If you're cool with it, I'm totally on board.

@Gormartsen that sounds pretty interesting. I'm down for taking a look at an example. Worth mentioning the project owners (cc: @jgeewax) would have to be involved in the decision to bring in a new tool.

stephenplusplus:
@jmdobry on the versioning, JJ advises to stay in `0.x` land. So, `google-cloud` should start at 0.1.0, as well as all of the sub-packages.

jmdobry:
Okay.

jmdobry:
Re-organized into package folders.

.github/CONTRIBUTING.md
```diff
@@ -7,6 +7,15 @@
 
 ## Testing
 
+### Setup
```
stephenplusplus: Can `npm test` do this automatically?

stephenplusplus:
Is the linking script going to block out any contributors on Windows machines?

jmdobry:
Good question. Right now linking is only necessary in the PR because the packages have not yet been published to NPM. Once they're all published, the linking script can potentially go away.

.github/CONTRIBUTING.md
jmdobry: Yeah, though it only really needs to be done once. If you're iterating with `npm test` it would  slow things down. The linking part can go away once the packages have been published to NPM, at which point this script would just be running `npm install` in a bunch of folders.

jmdobry:
The script that's executed in `pretest` will need to be made cross-platform to support windows collaborators.

stephenplusplus:
Not sure this is a good solution, but found this module:

> Symlinks technically work on Windows, but they require special rights. For users with those rights, symlinks are used, but when not available, a combination of junctions and copying is used to mimic the behavior somewhat performantly.

> \- https://github.com/broccolijs/node-symlink-or-copy

(Source: https://npms.io/search?term=symlink)

stephenplusplus:
I got some feedback that starting the new `google-cloud` package where we left off with `gcloud` would cause the least amount of confusion. I'm cool with that, which would mean `google-cloud` will start at 0.38.0 -- wdyt about that idea?

Gormartsen:
>  wdyt about that idea?

I would add to TODO a "Howto switch from gcloud to google-cloud" documentation.
It should cover API changes.

stephenplusplus:
*Temporary addition to the head of <kbd>readme.md</kbd>*:


## ⚠️ Where did `gcloud-node` go?

`gcloud-node` will live on under a new name, `google-cloud`. Your code will behave the same, simply change your dependency:

```sh
$ npm uninstall --save gcloud
$ npm install --save google-cloud
```

--

The README should also be updated with a quick header under each API that shows how to install independently:


## Google BigQuery

- [API Documentation][gcloud-bigquery-docs]
- [Official Documentation][cloud-bigquery-docs]

#### Preview
```js
// Authenticating on a per-API-basis. You don't need to do this if you auth on a
// global basis (see Authentication section above).
var bigquery = gcloud.bigquery({
  projectId: 'my-project',
});

// Or install only the BigQuery npm module:
//   $ npm install --save @google-cloud/bigquery
var bigquery = require('@google-cloud/bigquery)({/*...*/});

// Access an existing dataset and table.
var schoolsDataset = bigquery.dataset('schools');
var schoolsTable = schoolsDataset.table('schoolsData');

// Import data into a table.
schoolsTable.import('/local/file.json', function(err, job) {});

// Get results from a query job.
var job = bigquery.job('job-id');

// Use a callback.
job.getQueryResults(function(err, rows) {});

// Or get the same results as a readable stream.
job.getQueryResults().on('data', function(row) {});
```


^ better explanations welcome.

callmehiphop:
@stephenplusplus @jmdobry Sorry, I'm a bit late to the party

> My thoughts are the docs can work as they are, we just need to update the overview for each API to show that it can be installed individually, and also show how to instantiate it.

We are going to want a better solution to browse individual sub-module docs. If a user has an older version of a submodule installed it would be pretty frustrating to find the docs for that specific version since we only display/allow switching for the main modules versions (which won't necessarily align if I understand correctly).

IMO we are probably going to want the following:

1. A dropdown on the site allowing us to switch between the main module and all sub-modules. (https://github.com/GoogleCloudPlatform/gcloud-common/issues/52#issuecomment-225016782)
2. When viewing an api on the main module docs (e.g. vision) we'll want an important notice telling the user they may wish to browser the sub-module docs with a link.

To accomplish this we're going to have to capture JSON files for each version of each sub module. I believe that means our <kbd>docs/json</kbd> directory will now resemble the following

```sh
json
 ├─── google-cloud
 |     ├─── master
 |     |     └─── toc.json
 |     └─── v0.1.0
 |           └─── toc.json
 └─── vision
       ├─── master
       |     └─── toc.json
       └─── v0.2.0
             └─── toc.json
```

Each version folder would contain a table of contents (<kbd>toc.json</kbd>) as well as all the JSON files specific to that service + version.

stephenplusplus:
Can you do a PR for https://github.com/GoogleCloudPlatform/gcloud-common/issues/52#issuecomment-225016782 -- is it possible to do it in a way that won't affect the gcloud repos that aren't modularized yet?

> When viewing an api on the main module docs (e.g. vision) we'll want an important notice telling the user they may wish to browser the sub-module docs with a link.

Good call. Maybe in the `@classdesc` for the constructor + the "Getting Started" section.

> To accomplish this we're going to have to capture JSON files for each version of each sub module.

Is this something @jmdobry should add to this PR? Not sure if that's a big undertaking or not.

callmehiphop:
> Can you do a PR for GoogleCloudPlatform/gcloud-common#52 (comment)

It's hard for me to visualize how this is supposed to work 100% - unless you were hoping that PR would dictate how we organize/generate all the things, which would put this PR on hold.

> is it possible to do it in a way that won't affect the gcloud repos that aren't modularized yet?

I think it is possible, I believe we just have to add another config to the <kbd>manifest.json</kbd>. Although, the implementation itself probably wouldn't be as simple as it sounds.

> Good call. Maybe in the @classdesc for the constructor + the "Getting Started" section.

Not really a fan of this, if you add it to the class description it would show up in both the main package docs and the sub-module docs. I think this might be something we want the docs app itself to do and situationally show the links to and from the main package.

> Is this something @jmdobry should add to this PR? Not sure if that's a big undertaking or not.

I would say it is a decent sized undertaking - especially since very few people know how the docs work 😦 . I suppose we could skate by without adding it to this PR, but we would need to get it out the door **very** quickly - as soon as we update a single service, the inability to view sub-module docs is going to seriously impact our users.

stephenplusplus:
Maybe there'll just be a hand-off point of this PR where we can take it over from @jmdobry to implement the docs part. I think all that leaves is updating the versions. Is there anything else you wanted/needed to do, @jmdobry?

> which would put this PR on hold.

Let's do whatever we can do to get this out ASAP. This will quickly get hard to maintain as we move forward in other areas of the code.

> It's hard for me to visualize how this is supposed to work 100% - unless you were hoping that PR would dictate how we organize/generate all the things, which would put this PR on hold.

Maybe throw some questions out so we can all get clarity -- if you think https://github.com/GoogleCloudPlatform/gcloud-common/issues/52 is the right place for that discussion, we can move over there. Your UI changes suggested there look perfect to me.

>  I think this might be something we want the docs app itself to do and situationally show the links to and from the main package.

Sure, whatever you think. The `@classdesc` might not be great, though I think the `Getting Started` would be a logical place to put a note about the ways to use it. We can figure this out as we go, though.

callmehiphop:
> Let's do whatever we can do to get this out ASAP. This will quickly get hard to maintain as we move forward in other areas of the code.

I guess in this case, maybe we should move forward without the doc updates and just try and redirect users to the proper documentation the best we can until the site updates roll out.

jmdobry:
K, let me do a little more testing and fix conflicts.

stephenplusplus:
Thanks 👍 We'll freeze merges for a little bit.

jmdobry:
All fixed.

I updated README.md, specifically [here](https://github.com/jmdobry/gcloud-node/tree/google-cloud#where-did-gcloud-node-go) and [here](https://github.com/jmdobry/gcloud-node/tree/google-cloud#preview). WDYT?

stephenplusplus:
For the first one, I forgot we might want to show the `require` as well:

## Where did gcloud-node go?

gcloud-node lives on under a new name, `google-cloud`. Your code will behave the same, simply change your dependency:

```sh
$ npm uninstall --save gcloud
$ npm install --save google-cloud
```
```diff
- var gcloud = require('gcloud');
+ var gcloud = require('google-cloud');
```

--

For the second example, maybe we should call it out a bit more:

## Google BigQuery

- [API Documentation][gcloud-bigquery-docs]
- [Official Documentation][cloud-bigquery-docs]

##### *Using the all-in-one module*
```sh
$ npm install --save google-cloud
```
```js
var gcloud = require('google-cloud');
var bigquery = gcloud.bigquery;
```

##### *Using the BigQuery API module*
```sh
$ npm install --save @google-cloud/bigquery
```
```js
var bigquery = require('@google-cloud/bigquery');
```

#### Preview
```js
// Authenticating on a per-API-basis. You don't need to do this if you auth on a
// global basis (see Authentication section above).
var bigqueryClient = bigquery({
  projectId: 'my-project',
});

// Access an existing dataset and table.
var schoolsDataset = bigqueryClient.dataset('schools');
var schoolsTable = schoolsDataset.table('schoolsData');

// Import data into a table.
schoolsTable.import('/local/file.json', function(err, job) {});

// Get results from a query job.
var job = bigquery.job('job-id');

// Use a callback.
job.getQueryResults(function(err, rows) {});

// Or get the same results as a readable stream.
job.getQueryResults().on('data', function(row) {});
```

jmdobry:
Done.

jmdobry:
Once this is merged, anyone developing on `gcloud-node` will want to do the following:

1. Cleanup modules that are no longer needed at the root of the repo:

    ```bash
    npm prune
    ```

1. Setup symbolic links for inter-dependent packages (necessary until the inter-dependent packages have been published to NPM, then you will no longer need to do this)

    ```bash
    ./scripts/link.sh
    ```

1. Install NPM dependencies within each sub-package:

    ```bash
    ./scripts/install.sh
    ```

    Or you can manually run `npm install` inside each sub-package folder.

* `npm test` and `npm run system-test` can be run at the root of the repo and in each sub-package folder.
* `npm run lint` can be run at the root of the repo.

sethmurphy18:
Wouldn't a good alternative also be to just delete the gcloud-node folder manually from node_modules, and then reinstall manually?

jmdobry:
I found a really neat way to make the shell scripts cross-platform: https://github.com/shelljs/shelljs

callmehiphop:
@stephenplusplus is this just waiting on the updated docs stuff?

stephenplusplus:
That's a better question for @jmdobry :)

callmehiphop:
I think if the docs are the only thing holding this PR back, we should just merge this in and get any pending PRs resolved. We can time the release with the docs being finished. 

I have a feeling our JSON docs script is going to need quite a bit of work to accommodate the modular docs and I'd hate to delay this more than we need to.

stephenplusplus:
Sure, sounds good to me. Whenever @jmdobry gives the go-ahead, we can merge. The only thing that might be left is the integration of shelljs so Windows users can contribute.

packages/bigquery/system-test/bigquery.js
```diff
@@ -18,16 +18,15 @@
 
 var assert = require('assert');
 var async = require('async');
-var Dataset = require('../lib/bigquery/dataset');
-var Table = require('../lib/bigquery/table');
-var env = require('./env');
+var Dataset = require('../src/dataset.js');
+var Table = require('../src/table.js');
+var env = require('../../system-test/env.js');
```
stephenplusplus: I believe these all need to be `../../../system-test/env.js');

jmdobry:
I'll finalize this today so it can get merged.

jmdobry:
Done, rebased against `master`, conflicts resolved. The scripts I added are now cross-platform via ShellJS (the other scripts could probably be upgraded too).

The scripts I added are:

- `scripts/install` - Installs NPM dependencies in each sub-package. Works in up to 5 directories at a time so as to be fast.
  - Is automatically invoked via `postinstall` when `npm install` is run at the root of the repository
- `scripts/uninstall` - Reverse of `scripts/install`
- `scripts/link` - Creates local symbolic links (via `npm link`) between the various sub-packages in the correct order based on inter-package dependencies. Useful for testing your local changes to multiple sub-packages that are inter-dependent. Also, these links are required for the tests to pass while the sub-packages have not yet been published to NPM.
  - Once the sub-packages have been published to NPM, these lines `.travis.yml` can be removed:

      ```yaml
      - npm install shelljs
      - "./scripts/link"
      ```
- `scripts/unlink` - Reverse of `scripts/link`

Sorry if my rebasing made life harder for other PRs based on this PR.

stephenplusplus:
> Works in up to 5 directories at a time so as to be fast.

Woo!

> Is automatically invoked via postinstall when npm install is run at the root of the repository

I think this means that it will also run when users `npm install google-cloud`.

> Sorry if my rebasing made life harder for other PRs based on this PR.

No problem, I knew I was in for it :)

stephenplusplus:
> I think this means that it will also run when users npm install google-cloud.

^ Ignore me, I got my `package.json`s confused.

stephenplusplus:
Looks good to me. Made a PR with a couple tweaks that this needed to run locally: https://github.com/jmdobry/gcloud-node/pull/2

stephenplusplus:
Just running some tests locally, then merging in. Yay!

stephenplusplus:
Merged in via https://github.com/GoogleCloudPlatform/gcloud-node/commit/1ccd9180058a58f372e040fdfb16030d72853089

Thank you for all of the help! 👍 🎉 The docs are still in progress, but should be completed shortly. We'll start firing off the modules then. MODULES!!!

PR #1436, opened 2016-07-15
grpc: changed error handling to not modify original error

callmehiphop:
Relates to #1433

I believe the grpc streaming APIs are re-using status objects internally, so if we modify the original object to a code other than 0, it will re-emit it that same object as an error. This PR extends the status/error objects, leaving the originals untouched.

### DIFF
--- a/lib/common/grpc-service.js
+++ b/lib/common/grpc-service.js
@@ -244,7 +244,7 @@ GrpcService.prototype.request = function(protoOpts, reqOpts, callback) {
 
       service[protoOpts.method](reqOpts, grpcOpts, function(err, resp) {
         if (err) {
-          respError = GrpcService.getError_(err);
+          respError = GrpcService.decorateError_(err);
 
           if (respError) {
             onResponse(null, respError);
@@ -326,7 +326,7 @@ GrpcService.prototype.requestStream = function(protoOpts, reqOpts) {
     request: function() {
       return service[protoOpts.method](reqOpts, grpcOpts)
         .on('status', function(status) {
-          var grcpStatus = GrpcService.extendGrpcResponse_({}, status);
+          var grcpStatus = GrpcService.decorateStatus_(status);
 
           this.emit('response', grcpStatus || status);
         });
@@ -335,7 +335,7 @@ GrpcService.prototype.requestStream = function(protoOpts, reqOpts) {
 
   return retryRequest(null, retryOpts)
     .on('error', function(err) {
-      var grpcError = GrpcService.getError_(err);
+      var grpcError = GrpcService.decorateError_(err);
 
       stream.destroy(grpcError || err);
     })
@@ -456,7 +456,7 @@ GrpcService.createDeadline_ = function(timeout) {
  * @param {object} response - The grpc response.
  * @return {object|null}
  */
-GrpcService.extendGrpcResponse_ = function(obj, response) {
+GrpcService.decorateGrpcResponse_ = function(obj, response) {
   if (response && GRPC_ERROR_CODE_TO_HTTP[response.code]) {
     var defaultResponseDetails = GRPC_ERROR_CODE_TO_HTTP[response.code];
 
@@ -478,9 +478,21 @@ GrpcService.extendGrpcResponse_ = function(obj, response) {
  * @param {error|object} err - The grpc error.
  * @return {error|null}
  */
-GrpcService.getError_ = function(err) {
+GrpcService.decorateError_ = function(err) {
   var errorObj = is.error(err) ? new Error() : {};
-  return GrpcService.extendGrpcResponse_(errorObj, err);
+  return GrpcService.decorateGrpcResponse_(errorObj, err);
+};
+
+/**
+ * Checks for grpc status code and extends the status object with additional
+ * information
+ *
+ * @private
+ * @param {object} status - The grpc status.
+ * @return {object|null}
+ */
+GrpcService.decorateStatus_ = function(status) {
+  return GrpcService.decorateGrpcResponse_({}, status);
 };
 
 /**
--- a/test/common/grpc-service.js
+++ b/test/common/grpc-service.js
@@ -1122,14 +1122,14 @@ describe('GrpcService', function() {
     });
   });
 
-  describe('extendGrpcResponse_', function() {
+  describe('decorateGrpcResponse_', function() {
     it('should retrieve the HTTP code from the gRPC error map', function() {
       var errorMap = GrpcService.GRPC_ERROR_CODE_TO_HTTP;
       var codes = Object.keys(errorMap);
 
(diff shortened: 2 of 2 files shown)


lib/common/grpc-service.js
```diff
@@ -457,11 +458,12 @@ GrpcService.createDeadline_ = function(timeout) {
 GrpcService.getError_ = function(err) {
   if (err && GRPC_ERROR_CODE_TO_HTTP[err.code]) {
     var defaultErrorDetails = GRPC_ERROR_CODE_TO_HTTP[err.code];
-    err.code = defaultErrorDetails.code;
-    err.message = err.message || defaultErrorDetails.message;
-    return err;
-  }
 
+    return extend(true, new Error(), err, {
```
callmehiphop: @stephenplusplus do you think we should just use a pojo here instead of a `new Error`?

stephenplusplus: I think we just want to be sure:

- We return an `Error` object
- The `.stack` property is useful

This probably assigns a new `stack` to this location in the code, which I suppose is correct?

lib/common/grpc-service.js
```diff
@@ -457,11 +458,13 @@ GrpcService.createDeadline_ = function(timeout) {
 GrpcService.getError_ = function(err) {
   if (err && GRPC_ERROR_CODE_TO_HTTP[err.code]) {
     var defaultErrorDetails = GRPC_ERROR_CODE_TO_HTTP[err.code];
-    err.code = defaultErrorDetails.code;
-    err.message = err.message || defaultErrorDetails.message;
-    return err;
-  }
+    var errorObj = is.error(err) ? new Error() : {};
```
stephenplusplus: Maybe we should rename `getError_` if we're not returning an error, since this gets an error or a status object?

callmehiphop: Sure, WDYT about `getStatus_`?

stephenplusplus: The thing that confuses this for me is a line line:

```js
respError = GrpcService.getStatus_(err);
// ^-- assign an "error" from a method that gets me a "status" from an "error"
```

Given this naming pattern, I think this might be more clear:

```js
var respStatus = GrpcService.getStatus_(err);

if (is.error(respStatus)) {
  respError = respStatus;
}
```

However, can you think of another way to handle this duality? Maybe two methods; `decorateGrpcStatus_` and `decorateGrpcError_` or some kind of utility like `isGrpcError_`?

callmehiphop: Took a crack at this.. PTAL!

lib/common/grpc-service.js
```diff
@@ -325,7 +326,7 @@ GrpcService.prototype.requestStream = function(protoOpts, reqOpts) {
     request: function() {
       return service[protoOpts.method](reqOpts, grpcOpts)
         .on('status', function(status) {
-          var grcpStatus = GrpcService.getError_(status);
+          var grcpStatus = GrpcService.extendGrpcResponse_({}, status);
```
stephenplusplus: How about `GrpcService.decorateStatus_(status);`

lib/common/grpc-service.js
```diff
@@ -446,26 +447,43 @@ GrpcService.createDeadline_ = function(timeout) {
...
+ * @param {object} response - The grpc response.
+ * @return {object|null}
  */
-GrpcService.getError_ = function(err) {
-  if (err && GRPC_ERROR_CODE_TO_HTTP[err.code]) {
-    var defaultErrorDetails = GRPC_ERROR_CODE_TO_HTTP[err.code];
-    err.code = defaultErrorDetails.code;
-    err.message = err.message || defaultErrorDetails.message;
-    return err;
+GrpcService.extendGrpcResponse_ = function(obj, response) {
```
stephenplusplus: I think this method is a little confusing on its own. It might be best to treat this like a utility method used by `decorateStatus_` and `decorateError_`.

I'm choosing `decorate` as opposed to `get` because `var status = getStatus(status)` isn't really what's happening. We already have the status, we're just elaborating on it. And `extend` is okay, except it could be confused with the extend module in that, "am I supposed to provide an object to extend with?".

callmehiphop: I guess I'm having a difficult time understanding why we would create two methods that do the same exact thing. The only difference is that with an error it's possible that it could be an Error instance, which we would want to preserve for the stack trace.

stephenplusplus: Our confusion speaks to the confusing nature of what the code is doing, so I just want to make things as clear as possible.

If all we are really doing is decorating the response from gRPC, we should just *always* do it by calling `decorateGrpcResponse_(grpcResponse)`. If we want to know if it's an error, so we can treat it as an error, we should check if it's an error:

```js
var decoratedGrpcResponse = GrpcService.decorateGrpcResponse_(resp);

if (is.error(decoratedGrpcResponse)) {
  callback(decoratedGrpcResponse);
}
```

When we want to emit the response:

```js
.on('status', function(grpcResponse) {
  var decoratedGrpcResponse = GrpcService.decorateGrpcResponse_(resp);
  this.emit('response', decoratedGrpcResponse);
});
```

callmehiphop: PTAL, I think this is a bit closer to what you're asking for.

stephenplusplus: Hey, looks good to me. That's what I was trying to suggest 2 comments ago. Any reason you chose that instead of my more recent suggestion? You initially didn't seem to be in favor of the two methods that do the same thing.

callmehiphop: I'm not a huge fan tbh, but I just want to get it out the door. Adding the second method felt like the least amount of refactoring.

stephenplusplus: That's very honest of you. We can refactor in the future if we find room for improvement. In the future, please don't just try to humor me and get things out the door... we should do things the right way, no matter how long the discussion or time it takes.

callmehiphop: I wasn't necessarily trying to appease you, but I was felt like we were bike shedding. None of the solutions felt very different to me - they all seemed to be fairly intuitive and solved our problem, hence why I compromised.

stephenplusplus: I think you understood that piece of code differently as its author, but considering my position as a fellow maintainer, it wasn't obvious to me how that function was supposed to work. That's all I was going for... making contributions easier and not leaving bits of code in that aren't as obvious as they can be.

callmehiphop: I think regardless, we both agree that the current solution works well and is intuitive for maintainers. I don't think this conversation is adding anything helpful so perhaps we should take it offline.

stephenplusplus: I will respect your wishes, but am glad we speak in OSS :)

PR #1438, opened 2016-07-18
system tests (logging): increase consistency delay

stephenplusplus:
The recent push to master has introduced a new failed build (https://travis-ci.org/GoogleCloudPlatform/gcloud-node/jobs/145544257#L7638). We currently have two tests that will write log entries, then verify they were written properly by calling `getEntries`. We have a "write consistency delay" of 15 seconds that has worked in the past, but after testing with up to 2 minutes now, the call to `getEntries` still pulls old data.

This PR removes the delay, and simply checks that the "write entries" call doesn't fail.

@filipjs -- is there any reliable way to write a log entry, then pull it down from `getEntries({ orderBy: 'timestamp desc', pageSize: 1 })`?

### DIFF
--- a/system-test/logging.js
+++ b/system-test/logging.js
@@ -32,7 +32,7 @@ var Storage = require('../lib/storage/index.js');
 
 describe('Logging', function() {
   var TESTS_PREFIX = 'gcloud-logging-test';
-  var WRITE_CONSISTENCY_DELAY_MS = 15000;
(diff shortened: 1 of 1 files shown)


filipjs:
We hit a small bump in our processing pipeline earlier today, probably that's why there were failures.
1 minute should be enough for that kind of test.

stephenplusplus:
Ah, okay. Thanks for the info. I've updated the tests to allow 60 seconds before validating the entries.

PR #1440, opened 2016-07-18
Introduce Natural Language API

stephenplusplus:
Upstream API docs:
- https://cloud.google.com/natural-language/docs/getting-started

Docs preview:
- http://stephenplusplus.github.io/gcloud-node/#/docs/master/language
- http://stephenplusplus.github.io/gcloud-node/#/docs/master/language/document
#### To Dos
- [x] Use modularized format from #1431
- [x] Enable API for Travis account
- [x] Add API to toc.json
- [x] Create new language folder with .gitkeep
- [x] Use gRPC
- [x] Add user agent
- [x] Docs
  - [x] JSDocs
  - [x] readme
- [x] System tests
  - [x] HTML input
    - [x] Inline annotation & detection
    - [x] GCS file annotation & detection
  - [x] Text input
    - [x] Inline annotation & detection
    - [x] GCS file annotation & detection
  - [x] Unknown input
    - [x] Inline annotation & detection
    - [x] GCS file annotation & detection
- [x] Unit tests
  - [x] lib/index.js
  - [x] lib/language/document.js
  - [x] lib/language/index.js

### DIFF
--- a/packages/language/test/document.js
+++ b/packages/language/test/document.js
@@ -287,7 +287,7 @@ describe('Document', function() {
         Document.formatTokens_ = util.noop;
       });
 
-      it('should return language if no features are requested', function(done) {
+      it('should always return the language', function(done) {
         var apiResponse = apiResponses.default;
         document.request = createRequestWithResponse(apiResponse);
 


stephenplusplus:
@callmehiphop let me know what you think of the usability. Ready to start on tests after.

callmehiphop:
Could you throw up a gist or something of more code examples? Or is the README example all there is to it?

stephenplusplus:
README + system tests is all for now. It's not a very big API, just 3 methods. I did fail to mention there's a `verbose` mode for each method call.

The two primary ways of calling are

```js
// inline
language.detectEntities('a sentence', function(err, entities) {});

// with a document
var document = language.detectEntities('a sentence');
document.detectEntities(function(err, entities) {});
```

The verbose mode just makes less simplifications. These tests show what the general responses look like in simple/verbose mode for each annotation type (annotation, entities, sentiment): https://github.com/stephenplusplus/gcloud-node/blob/spp--language/system-test/language.js#L314

LMKIYHAQ

callmehiphop:
I see, is there a reason we only provide convenience methods for entities and sentiment? Where as your readme example shows tokens, language and sentences as well.

stephenplusplus:
There are only 3 API endpoints, see https://cloud.google.com/natural-language/reference/rest/v1beta1/documents:

- analyzeEntities (`detectEntities`)
- analyzeSentiment (`detectSentiment`)
- annotateText (`annotate`)

stephenplusplus:
We can add things like `detectSentences()` and `detectTokens()`, it would just be plucking out only part of the `annotate()` response. It might end up with users thinking they have to run both functions to get both types of annotations. WDYT?

callmehiphop:
> We can add things like detectSentences() and detectTokens(), it would just be plucking out only part of the annotate() response.

We took this exact approach with the vision api, did we not? Granted the vision `annotate()` response has more data in it, convenience methods sorta seem like one of the big reasons why gcloud-node exists, no?

> It might end up with users thinking they have to run both functions to get both types of annotations.

Can't this be prevented with documentation? Stating that it's a convenience method for `annotate()`?

stephenplusplus:
> We took this exact approach with the vision api, did we not?

No, the Vision API has one API method that returns a response that matches the type of request you send. So we have `detectFaces`, which makes a request that asks for facial detection, so the response only includes facial detections.

> convenience methods sorta seem like one of the big reasons why gcloud-node exists, no?

What we're talking about here is making a request that includes a lot of info, but only returning a part of it. I can see both sides here, however this is new territory for us. I feel conflicted because we would essentially be throwing away information. Docs can help, but a user simply seeing methods called `detectTokens` and `detectSentences` will make them think these each have to be called to get those types of detections.

callmehiphop:
> No, the Vision API has one API method that returns a response that matches the type of request you send. So we have detectFaces, which makes a request that asks for facial detection, so the response only includes facial detections.

Ah, my mistake. Fair enough then, feel free to ignore!

README.md
```diff
@@ -701,6 +702,71 @@ vision.detectFaces('./image.jpg', function(err, faces) {
 ```
 
 
+## Google Cloud Natural Language (Alpha)
+
+> **This is an Alpha release of Google Cloud Resource Manager.** This feature is not covered by any SLA or deprecation policy and may be subject to backward-incompatible changes.
```
dhermes: This says "Resource Manager" not "Natural Language"

lib/language/index.js
```diff
@@ -0,0 +1,154 @@
...
+ * [syntax analysis](https://cloud.google.com/natural-language/docs/reference/rest/v1beta1/documents/annotateText).
+ *
+ * @constructor
+ * @alias module:language
+ *
+ * @classdesc
+ * The object returned from `gcloud.language` gives you complete control.
+ *
+ * To learn more about Google Cloud Natural Language, see
+ * [What is Google Cloud DNS?](https://cloud.google.com/natural-language)
```
dhermes: Note this links to Cloud DNS

stephenplusplus:
Docs preview: http://stephenplusplus.github.io/gcloud-node/#/docs/master/language

stephenplusplus:
All tested and doc'd @callmehiphop. PTAL when you can.

stephenplusplus:
@jmuk -- I rebased this PR against @jmdobry's [modularization PR](https://github.com/GoogleCloudPlatform/gcloud-node/pulls/1431). That is still in development, but it will hopefully merge soon. You can use the [new structure](https://github.com/stephenplusplus/gcloud-node/tree/spp--language/packages/language) to get started with #1464. If the changes you make can be limited to just creating files, as opposed to amending the files in `packages/language`, that should make it pretty safe in regards to avoiding conflicts.

You will need to run these commands when setting up the repo:

```sh
$ ./scripts/link.sh
$ cd packages/language
$ npm install

# good to go-- helpful development commands:
$ npm run lint
$ npm run test
$ npm run system-test
```

@callmehiphop sorry for the huge PR now... all of the language changes for review are here: https://github.com/stephenplusplus/gcloud-node/tree/spp--language/packages/language

stephenplusplus:
@jmuk @callmehiphop -- I rebased this off of master now.

README.md
```diff
@@ -768,6 +769,92 @@ loggingClient.getEntries(function(err, entries) {
...
+
+#### Preview
+
+```js
+var gcloud = require('google-cloud');
+
+// Authorizing on a per-API-basis. You don't need to do this if you auth on a
+// global basis (see Authorization section above).
+
+var languageClient = language({
```
callmehiphop: Should this be `gcloud.language()`?

stephenplusplus: Oh woops, I have to remove the `var gcloud = require('google-cloud');` line. The `language` variable is created in two different ways above the API example.

packages/common/src/service-object.js
```diff
@@ -308,7 +308,7 @@ ServiceObject.prototype.request = function(reqOpts, callback) {
 
   var uriComponents = [
     this.baseUrl,
-    this.id,
+    this.id || '',
```
callmehiphop: Should we mark this as optional in the `ServiceObject` constructor JSDoc?

README.md
stephenplusplus: Fixed in a new commit

packages/common/src/service-object.js
stephenplusplus: Done

packages/language/src/document.js
```diff
@@ -0,0 +1,741 @@
...
+var extend = require('extend');
+var File = require('@google-cloud/storage').File;
+var format = require('string-format-obj');
+var is = require('is');
+var prop = require('propprop');
+
+/*! Developer Documentation
+ *
+ * @param {module:language} language - The parent Language object.
+ * @param {object=}
```
callmehiphop: `config`

packages/language/src/document.js
```diff
@@ -0,0 +1,741 @@
...
+
+    if (this.reqOpts.document.type === 'TEXT') {
+      this.reqOpts.document.type = 'PLAIN_TEXT';
+    }
+  } else {
+    // Default to plain text.
+    this.reqOpts.document.type = 'PLAIN_TEXT';
+  }
+
+  if (content instanceof File) {
```
stephenplusplus: That's a great idea. Can you open a new issue 'bout it?

packages/language/src/document.js
stephenplusplus: Done.

README.md
```diff
@@ -768,6 +769,90 @@ loggingClient.getEntries(function(err, entries) {
...
+// Analyze the sentiment of the document.
+document.detectSentiment(function(err, positiveSentiment) {
+  // positiveSentiment = true
+});
+
+// Parse the syntax of the document.
+document.annotate(function(err, annotations) {
+  // annotations = {
+  //   language: 'en',
+  //   sentiment: true,
```
callmehiphop: According to the JSDoc for the `Document` object, `sentiment` should be a number, but here it is represented as a boolean.

packages/language/src/document.js
```diff
@@ -0,0 +1,741 @@
...
+ * @param {?error} callback.err - An error occurred while making this request.
+ * @param {object} callback.annotation - The formatted API response.
+ * @param {string} callback.annotation.language - The language detected from the
+ *     text.
+ * @param {number} callback.annotation.sentiment - A value in the range of
+ *     `-100` to `100`. Large numbers represent more postiive sentiments.
+ * @param {object} callback.annotation.entities - The recognized entities from
+ *     the text, grouped by the type of entity.
+ * @param {string[]} callback.annotation.entities.art - Art entities detected
+ *     from the text. This is only present if detections of this type were
```
callmehiphop: Do you think it would be better if we made a note about keys missing within the `callback.annotation.entities` description instead of repeating the text?

stephenplusplus: I originally had that, and decided against it. I imagined our users understanding more clearly the object they get back if they could see a more defined list of features.

README.md
stephenplusplus: Woops, this is wrong as well as L#824

packages/language/src/document.js
```diff
@@ -0,0 +1,741 @@
...
+ *   //       partOfSpeech: {
+ *   //         tag: 'NOUN'
+ *   //       },
+ *   //       dependencyEdge: {
+ *   //         headTokenIndex: 1,
+ *   //         label: 'NSUBJ'
+ *   //       },
+ *   //       lemma: 'Google'
+ *   //     },
+ *   //     ...
```
callmehiphop: We could probably remove this

packages/language/src/document.js
```diff
@@ -0,0 +1,741 @@
...
+ *   // }
+ * });
+ *
+ * //-
+ * // To request only certain annotation types, provide an options object.
+ * //-
+ * var options = {
+ *   entities: true,
+ *   sentiment: true
+ * };
```
callmehiphop: I'm not really sure how I feel about this, it seems a little confusing TBH. I'm assuming it is setup this way because `syntax` is the only one that doesn't have a standalone method?

stephenplusplus: Can you elaborate what is confusing?

callmehiphop: The option toggling functionality. I don't think we have a similar mechanism anywhere else in our codebase. In this example it seems like it would be simpler to just say `syntax: false` and more predictable since all are enabled by default.

I could sort of seeing this being convenient if I only wanted to grab the `syntax` of a document, but that could also be solved by creating a convenience method - at which point we would have methods for grabbing all the available items from an annotation.

packages/language/src/document.js
```diff
@@ -0,0 +1,741 @@
...
+ *   //             text: {
+ *   //               content: 'Google',
+ *   //               beginOffset: -1
+ *   //             }
+ *   //           }
+ *   //         ]
+ *   //       }
+ *   //     ],
+ *   //     places: [
+ *   //       {...}
```
stephenplusplus: Woops, fixed.

README.md
stephenplusplus: Fixed 'em

packages/language/src/document.js
stephenplusplus: I left that in so it doesn't make the user think this is the complete response.

packages/language/src/document.js
```diff
@@ -0,0 +1,758 @@
...
+ *
+ *   // annotation = {
+ *   //   language: 'en',
+ *   //   sentiment: 100,
+ *   //   entities: {
+ *   //     organizations: [
+ *   //       'Google'
+ *   //     ],
+ *   //     places: [
+ *   //       'American'
```
callmehiphop: Is this right? `American` doesn't quite sound like a place

packages/language/src/document.js
```diff
@@ -0,0 +1,758 @@
...
+ * Detect the sentiment of the text in this document.
+ *
+ * @resource [documents.analyzeSentiment API Documentation]{@link https://cloud.google.com/natural-language/reference/rest/v1beta1/documents/analyzeSentiment}
+ *
+ * @param {object=} options - Configuration object.
+ * @param {boolean} options.verbose - Enable verbose mode for more detailed
+ *     results. Default: `false`
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An error occurred while making this request.
+ * @param {object} callback.sentiment - A value in the range of `-100` to `100`.
```
callmehiphop: Should this be `number`?

packages/language/src/document.js
```diff
@@ -0,0 +1,758 @@
...
+  var groupedEntities = entities.reduce(function(acc, entity) {
+    entity = extend(true, {}, entity);
+
+    var groupName = GROUP_NAME_TO_TYPE[entity.type];
+
+    entity.salience *= 100;
+
+    acc[groupName] = arrify(acc[groupName]);
+    acc[groupName].push(entity);
+    acc[groupName] = acc[groupName].sort(Document.sortByProperty_('salience'));
```
callmehiphop: I think `Array#sort` is mutable.

packages/language/src/document.js
stephenplusplus: I think this is just dealing with how `annotate` works-- you can ask for 1, 2, or 3 annotations, so the options just map to the endpoint behavior.

> I don't think we have a similar mechanism anywhere else in our codebase.

Vision has similar behavior with `detect()` where you specify what features you want:
```js
vision.detect('image.jpg', { features: ['face', 'label'] }, function() {}) 
```

I considered an array of strings here, too, but it felt like too much boilerplate:
```js
doc.annotate('image.jpg', { annotations: ['syntax'] }, function() {})
```

Let me know if you think that's better, though.

> In this example it seems like it would be simpler to just say syntax: false and more predictable since all are enabled by default.

I generally think whitelists are better than blacklists, i.e., it's better for a user to be specific about what they want rather than finding out everything that exists, and choosing what they don't want.

I can see how this whole thing is weird, though, since if a user just looks at *our* API, it's like there is only one API method upstream, `annotate` and we just create some convenience methods around it. But that's not the case, since there are `detectEntities` and `detectSentiment` API endpoints upstream. The API must distinguish for a reason, so I'm not comfortable saying `detectSentiment()` is the same as `annotate({ sentiment: true })`. I can see the upstream API merging these two or making the relationship more clear in the future, but for now, we kind of have to follow their schematic, since we don't know what goes on behind the scenes after a request is made.

I can see our API getting an overhaul/simplification in the future when the API makes some moves. For now, though, I think this is the safest path forward.

packages/language/src/document.js
stephenplusplus: Blame the upstream API cause that's what it gives us :)

packages/language/src/document.js
stephenplusplus: 💯

packages/language/src/document.js
```diff
@@ -0,0 +1,758 @@
...
+ *     the API.
+ * @param {boolean} verbose - Enable verbose mode for more detailed results.
+ * @return {object[]|string[]} - The formatted sentences or sentence descriptor
+ *     objects in verbose mode.
+ */
+Document.formatSentences_ = function(sentences, verbose) {
+  sentences = sentences.map(prop('text'));
+
+  if (!verbose) {
+    sentences = sentences.map(prop('content'));
```
callmehiphop: This a style nitpick, so feel free to ignore. But couldn't we just `return` here instead of re-assigning the variable?

packages/language/src/document.js
```diff
@@ -0,0 +1,758 @@
...
+ *     measurements in verbose mode.
+ */
+Document.formatSentiment_ = function(sentiment, verbose) {
+  sentiment = {
+    polarity: sentiment.polarity *= 100,
+    magnitude: sentiment.magnitude *= 100
+  };
+
+  if (!verbose) {
+    sentiment = sentiment.polarity;
```
callmehiphop: RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1440/files#r73750432

packages/language/src/document.js
```diff
@@ -0,0 +1,758 @@
...
+ *
+ * @param {object[]} tokens - A group of syntax detections returned from the
+ *     API.
+ * @param {boolean} verbose - Enable verbose mode for more detailed results.
+ * @return {number|object} - A slimmed down, simplified object or the original
+ *     object in verbose mode.
+ */
+Document.formatTokens_ = function(tokens, verbose) {
+  if (!verbose) {
+    tokens = tokens.map(function(token) {
```
callmehiphop: RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1440/files#r73750432

packages/language/src/document.js
stephenplusplus: I think either is fine, but since I did it the other way, let's keep the other way!

packages/language/src/document.js
stephenplusplus: D'oh.

stephenplusplus: Fixed.

stephenplusplus:
@callmehiphop anything left for me to do here?

packages/language/test/document.js
```diff
@@ -0,0 +1,824 @@
...
+      }
+
+      beforeEach(function() {
+        Document.formatSentiment_ = util.noop;
+        Document.formatEntities_ = util.noop;
+        Document.formatSentences_ = util.noop;
+        Document.formatTokens_ = util.noop;
+      });
+
+      it('should return the language and syntax by default', function(done) {
```
callmehiphop: Why aren't `entities` also returned by default?

callmehiphop: Ah, nevermind I see where you were going with this now. Could we separate these into different tests? I know it would be adding some extra verbosity, but I think if the tests were clearly labeled it would be easier to follow.

packages/language/src/index.js
```diff
@@ -0,0 +1,432 @@
...
+ * @param {function} callback - See {module:language/document#detectEntities}.
+ *
+ * @example
+ * //-
+ * // See {module:language/document#detectEntities} for a detailed breakdown of
+ * // the arguments your callback will be executed with.
+ * //-
+ * function callback(err, entities, apiResponse) {}
+ *
+ * language.detectEntities('Axel Foley is from Detroit', callback);
```
callmehiphop: ❤️

packages/language/src/index.js
```diff
@@ -0,0 +1,432 @@
...
+ * language.detectSentiment(file, callback);
+ *
+ * //-
+ * // Specify HTML content.
+ * //-
+ * var options = {
+ *   type: 'html'
+ * };
+ *
+ * language.detectSentiment('&lt;h1>Document Title&lt;/h1>', options, callback);
```
callmehiphop: Do we also need to encode `&gt;`?

callmehiphop:
@stephenplusplus I had a couple small comments, but nothing really important. I think overall it looks good, my only real complaint is that the system tests were a little tricky to work through.

packages/language/src/index.js
callmehiphop: Nevermind, just noticed your docs render correctly like this.. kind of a weird quirk!

stephenplusplus:
Yeah, the system tests have a unique style for sure. If it gets to be impossible to deal with when adding new features/debugging, we can look into changing that.

PTAL!

packages/language/test/document.js
stephenplusplus: Done!

packages/language/test/document.js
```diff
@@ -287,7 +287,19 @@ describe('Document', function() {
...
+
+        document.annotate(function(err, annotation, apiResponse_) {
+          assert.ifError(err);
+          assert.strictEqual(annotation.language, apiResponse.language);
+          assert.deepEqual(apiResponse_, apiResponse);
+          done();
+        });
+      });
+
+      it('should return the syntax by default', function(done) {
```
callmehiphop: Could we re-word this to something like.. `should return the syntax when no features are requested`?

stephenplusplus: Definitely. 👍

stephenplusplus: PTAL

packages/language/test/document.js
```diff
@@ -287,7 +287,7 @@ describe('Document', function() {
         Document.formatTokens_ = util.noop;
       });
 
-      it('should return the language by default', function(done) {
+      it('should return language if no features are requested', function(done) {
```
callmehiphop: Don't hate me. Isn't language always returned? I was only suggesting we change syntax because it is conditionally returned, while language looked to always be returned.

stephenplusplus: I could never hate you. Fixed!

PR #1444, opened 2016-07-21
grpc: listen for metadata event instead of status

### LINKED ISSUE #1443: Bigtable table.getRows() silently failing on queries > 4000 cells
I'm fairly new to Bigtable, and my apologies if this is an issue on my end. I have a table where all rows have 84 cells; when I submit a getRows request with a limit >= 48 (total > 4000 cells) I get no error and no response. With a lower limit, everything works perfectly. Looking at my networking, it seems like there is data coming in, but otherwise I don't have much else to go off of.

This is happening both when I use the streaming and the callback APIs. Passing in a filter to reduce the number of returned columns doesn't change this behavior. 

Many thanks for getting the Bigtable support out the door!

#### Environment details
`OSX @10.11.4`
`node @6.2.0`
`npm @3.8.9`
`gcloud-node @0.37.0`

#### Steps to reproduce
```
const table = bigtable.table('data');
const rows = [];
const rowCount = 48; //Works <= 47, silently fails >= 48. Each row has 84 cells.

table.getRows({
	limit: rowCount,
})
	.on('error', (err) => {
		console.error(err);
	})
	.on('data', (row) => {
		console.log('got row!');
		rows.push(row);
	})
	.on('end', () => {
		console.log(`Complete! Retrieved ${rows.length}.`);
		//Do something with rows
	});

// Also fails when using callback API
```

callmehiphop:
Closes #1443 

This fixes an issue where a grpc readable stream would hang indefinitely because we weren't draining it. Instead of listening for a status event, which occurs after results have been processed, we're now going to listen instead for a metadata event.

### DIFF
--- a/lib/common/grpc-service.js
+++ b/lib/common/grpc-service.js
@@ -326,6 +326,12 @@ GrpcService.prototype.requestStream = function(protoOpts, reqOpts) {
     request: function() {
       return service[protoOpts.method](reqOpts, grpcOpts)
         .on('metadata', function() {
+          // retry-request requires a server response before it starts emitting
+          // data. The closest mechanism grpc provides is a metadata event, but
+          // this does not provide any kind of response status. So we're faking
+          // it here with code `0` which translates to HTTP 200.
+          //
+          // https://github.com/GoogleCloudPlatform/gcloud-node/pull/1444#discussion_r71812636
           var grcpStatus = GrpcService.decorateStatus_({ code: 0 });
 
           this.emit('response', grcpStatus);


callmehiphop:
@imathews would you mind testing this branch out to see if it resolves your issues?

test/common/grpc-service.js
```diff
@@ -1074,14 +1074,37 @@ describe('GrpcService', function() {
...
+        ProtoService.prototype.method = function() {
+          return fakeStream;
+        };
+
+        retryRequestOverride = function(reqOpts, options) {
+          return options.request();
+        };
+
+        fakeStream
+          .on('response', function(resp) {
```
stephenplusplus: Listen for error as well.

lib/common/grpc-service.js
```diff
@@ -325,7 +325,8 @@ GrpcService.prototype.requestStream = function(protoOpts, reqOpts) {
 
     request: function() {
       return service[protoOpts.method](reqOpts, grpcOpts)
-        .on('status', function(status) {
+        .on('metadata', function() {
+          var status = this.received_status || { code: 0 };
```
stephenplusplus: @murgatroid99 -- is this a reliable way to get the status? Maybe it should be emitted with the metadata event?

callmehiphop: This is not a reliable way to get the status, AFAIK two requests are made - one sets the status the other emits the metadata. Per https://github.com/GoogleCloudPlatform/gcloud-node/issues/1443#issuecomment-234402654 the `metadata` event seems to be the way to go regardless. I can remove the check for `received_status` if you prefer.

stephenplusplus: I think we should get the status from grpc if it's available, we just need to know the right way to do that.

murgatroid99: You should listen for the `status` event from the grpc call to get the grpc status. But that event is emitted when the call completes, so you shouldn't be expecting it at the beginning of a call.

stephenplusplus: Okay, so I guess this should just always fake a status code 0 (200)?

callmehiphop: Done, PTAL!

lib/common/grpc-service.js
```diff
@@ -325,10 +325,10 @@ GrpcService.prototype.requestStream = function(protoOpts, reqOpts) {
 
     request: function() {
       return service[protoOpts.method](reqOpts, grpcOpts)
-        .on('status', function(status) {
-          var grcpStatus = GrpcService.decorateStatus_(status);
+        .on('metadata', function() {
+          var grcpStatus = GrpcService.decorateStatus_({ code: 0 });
```
stephenplusplus: Can you just add a comment explaining why we're doing this / link to this PR?

stephenplusplus:
Thanks :D

imathews:
This is now working beautifully. Many many thanks @callmehiphop for the quick resolution!

PR #1447, opened 2016-07-24
implement VM.prototype.setMachineType

arthuredelstein:
See https://cloud.google.com/compute/docs/reference/v1/instances/setMachineType


arthuredelstein:
(I have signed the Google CLA.)

PR #1448, opened 2016-07-24
implement VM.prototype.setMachineType

arthuredelstein:
See https://cloud.google.com/compute/docs/reference/v1/instances/setMachineType

### DIFF
--- a/lib/compute/vm.js
+++ b/lib/compute/vm.js
@@ -436,9 +436,7 @@ VM.prototype.reset = function(callback) {
  * @param {object} callback.apiResponse - The full API response.
  *
  * @example
- * vm.setMachineType(
- *   'n1-standard-1'
- * }, function(err, operation, apiResponse) {
+ * vm.setMachineType('n1-standard-1', function(err, operation, apiResponse) {
  *   // `operation` is an Operation object that can be used to check the status
  *   // of the request.
  * });


lib/compute/vm.js
```diff
@@ -423,6 +423,42 @@ VM.prototype.reset = function(callback) {
 };
 
 /**
+ * Set the machine type for this instance.
+ *
+ * @resource [Instances: setMachineType API Documentation]{@link https://cloud.google.com/compute/docs/reference/v1/instances/setMachineType}
+ *
+ * @param {string} zone - the compute engine zone
```
stephenplusplus: We don't need the zone, since a VM instance has to be accessed through a Zone. It should be available as this.zone.name.

lib/compute/vm.js
```diff
@@ -423,6 +423,42 @@ VM.prototype.reset = function(callback) {
...
+ * @param {string} machineType - new machine type.
```
stephenplusplus: Can you add a link to the available machine types?

arthuredelstein:
Looks like I need to write unit tests. I will work on that (hopefully) later this week.

stephenplusplus:
Thanks @arthuredelstein!

stephenplusplus:
@arthuredelstein we just merged a PR that modularized the repo, which will allow us to release modules for each service, instead of having them all in one `gcloud` package. That brought some significant structure changes, but updating your PR shouldn't be too bad. The file you edited, `lib/compute/vm.js`, is now `packages/compute/src/vm.js`. Just move your code into that file, and you'll find the test file you need here: `packages/compute/test/vm.js`.

stephenplusplus:
I didn't want to make you do all that, so check it out: #1469. Let me know if it works as you intended!

PR #1452, opened 2016-07-26
bigtable: throw when required information is missing

stephenplusplus:
In Bigtable, we should throw whenever data that is required is not provided. I tried to find those places, but @callmehiphop, could you take a look and let me know if I missed any?
#### To Dos
- [x] Throw
  - [x] Bigtable
  - [x] Bigtable#createTable
  - [x] row#createRules
  - [x] table#createFamily
  - [x] table#family
  - [x] table#row
- [x] Unit tests

### DIFF
--- a/packages/bigtable/src/index.js
+++ b/packages/bigtable/src/index.js
@@ -56,6 +56,9 @@ var PKG = require('../package.json');
  *
  * @resource [Creating a Cloud Bigtable Cluster]{@link https://cloud.google.com/bigtable/docs/creating-compute-instance}
  *
+ * @throws {error} If a cluster is not provided.
+ * @throws {error} If a zone is not provided.
+ *
  * @param {object=} options - [Configuration object](#/docs).
  * @param {string} options.cluster - The cluster name that hosts your tables.
  * @param {string|module:compute/zone} options.zone - The zone in which your
@@ -278,6 +281,14 @@ function Bigtable(options) {
     return new Bigtable(options);
   }
 
+  if (!options.cluster) {
+    throw new Error('A cluster must be provided to interact with Bigtable.');
+  }
+
+  if (!options.zone) {
+    throw new Error('A zone must be provided to interact with Bigtable.');
+  }
+
   options = extend({}, options, {
     zone: options.zone.name || options.zone
   });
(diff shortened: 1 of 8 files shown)


callmehiphop:
Can you clarify what kind of data you want to throw for? For example, in `Row#increment` we don't throw if a user were to provide a bad type for the `column` value - your changes make me think you only want to throw an error where grpc/upstream errors can't give back a detailed message of what is wrong.

stephenplusplus:
> Can you clarify what kind of data you want to throw for?

Somewhere between "let the API handle it all" and "validate all user input". In places like "create____", we know that it's impossible to create a "____" without a name. So, we should blow up and let the user know.

We especially should throw in all cases where our code assumes we have the input and tries to manipulate it, e.g., "create____" takes a string, but it needs to be uppercased, so our code throws on the line we tried to use their input, instead of throwing that their input was wrong and how to fix it.

With `increment`, it's a judgement call. What do you think?

> your changes make me think you only want to throw an error where grpc/upstream errors can't give back a detailed message of what is wrong.

Not exactly, but where possible, the API should do the error reporting for us. I do want to catch completely obvious things, basically falling under the label of "you used *OUR* API wrong", so not things like "does the name match some regex that is enforced by the API".

callmehiphop:
Thanks for that. I think some places you'd definitely want to look at are

* `Row` constructor - `name` param. 
* `Family` constructor - `name` param. 

Possible maybes are

* `Row#createRules` - `rules` param.
* `Row#filter` - `filter` param.
* `Row#deleteCells` - `columns` param.
* `Row#increment` - `column` param.
* `Row#save` - `key` param.
* `Family#setMetadata` - `metadata` param.
* `Table#insert` - `entries` param.
* `Table#mutate` - `entries` param.

packages/bigtable/src/index.js
```diff
@@ -409,6 +422,10 @@ Bigtable.prototype.createTable = function(name, options, callback) {
     options = {};
   }
 
+  if (!name) {
```
callmehiphop: I believe this is currently untested.

stephenplusplus: Woops. Added!

callmehiphop:
LGTM, waiting for CI.

packages/bigtable/test/index.js
```diff
@@ -106,12 +106,27 @@ describe('Bigtable', function() {
       fakeUtil.normalizeArguments = normalizeArguments;
     });
 
+    it('should throw if a cluster is not provided', function() {
+      assert.throws(function() {
+        new Bigtable();
```
callmehiphop: Not sure why, but coveralls is claiming this is untested. Could we try changing this to:

```js
new Bigtable({
  zone: ZONE
});
```

callmehiphop: https://coveralls.io/builds/7328412/source?filename=packages%2Fbigtable%2Fsrc%2Findex.js#L285

stephenplusplus: Okidoke, fixed.

stephenplusplus:
I snuck in a change to add a `covers` task as well here.

PR #1456, opened 2016-07-27
bigtable: column family support when creating tables

callmehiphop:
Relates to https://github.com/GoogleCloudPlatform/gcloud-node/issues/1451

This adds support for specifying column families when creating a table.

### DIFF
--- a/test/bigtable/index.js
+++ b/test/bigtable/index.js
@@ -222,42 +222,74 @@ describe('Bigtable', function() {
       bigtable.createTable(TABLE_ID, options, assert.ifError);
     });
 
-    it('should set the column families', function(done) {
-      var options = {
-        families: [
-          'a',
-          {
-            name: 'b',
-            rule: 'd'
-          },
-          {
-            name: 'c',
-            rule: {}
-          }
-        ]
-      };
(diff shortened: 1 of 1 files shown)


test/bigtable/index.js
```diff
@@ -219,6 +222,44 @@ describe('Bigtable', function() {
       bigtable.createTable(TABLE_ID, options, assert.ifError);
     });
 
+    it('should set the column families', function(done) {
```
stephenplusplus: Can you create these tests:

- describe('options.families', ...
  - it should accept a string family name
  - it should accept a string garbage collection rule
  - it should accept an object garbage collection rule

The reason being, if someone makes a change to your test that currently combines these, they might accidentally remove one of the checks. Having individual tests that can fail will reduce that chance.

callmehiphop: Done! PTAL

stephenplusplus:
Cool, LGTM.

stephenplusplus:
Rebased and merged in: https://github.com/GoogleCloudPlatform/gcloud-node/commit/1a42246accae61d0ce114c9f65610311ab229dae

PR #1459, opened 2016-07-31
bigtable: set true/false filters to null when omitted

### LINKED ISSUE #1458: Bigtable - Condition filter doesn't return rows unless "fail" condition is set
As I understand from the API docs, the fail attribute of a condition filter should be optional, effectively being the same as providing `[{ all: false }]`. However, if I don't provide the fail attribute, I get zero rows back; the condition filter behaves as I'd expect with fail explicitly set to `[{ all: false }]`

#### Environment details
`OSX @10.11.4`
`node @6.2.0`
`npm @3.8.9`
`gcloud-node @0.37.0`

#### Steps to reproduce
````
filter = [{
	condition: {
		test: [
			{
				"column": "SEX",
				"value": "F"
			}
		],
		pass: [{
			all: true
		}],

		// Returns zero rows unless these lines are uncommented

		// fail: [
		// 	{
		// 		all: false
		// 	}
		// ]
	}
}]

table.getRows({
		filter,
		limit: 100,
}).on('data', (row) => {
    // Do things
});
````

callmehiphop:
Closes #1458

### DIFF
--- a/lib/bigtable/filter.js
+++ b/lib/bigtable/filter.js
@@ -765,6 +765,10 @@ Filter.prototype.time = function(time) {
  * execution of the filters will be the order in which they were specified.
  */
 Filter.prototype.toProto = function() {
+  if (!this.filters_.length) {
+    return null;
+  }
+
   if (this.filters_.length === 1) {
     return this.filters_[0];
   }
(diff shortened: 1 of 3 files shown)


stephenplusplus:
Rebased and merged in via https://github.com/GoogleCloudPlatform/gcloud-node/commit/aa4bf3913c0d420fdba94cfbdfdb76f970480aa3

PR #1468, opened 2016-08-05
tests: fix assert.throws() usage

stephenplusplus:
So, I was using mocha wrong in regards to testing thrown errors.

Wrong:

``` js
assert.throw(function() {
  new Bucket();
}, 'You need a name for this bucket.'); // <-- not the real error message
```

Right:

``` js
assert.throw(function() {
  new Bucket();
}, /You need a name for this bucket\./); // <-- not the real error message
```


stephenplusplus:
@callmehiphop how's this one look?

callmehiphop:
LGTM!

PR #1469, opened 2016-08-05
compute: vm: add resize()

stephenplusplus:
This takes over from #1448, adding a `resize` method on the VM object.
#### To Dos
- [x] Start the VM again after setting the new machine type
- [x] Rename method
- [x] Explain the stopping and starting in the docs

``` js
vm.resize('n1-standard-1', function(err, apiResponse) {
  if (!err) {
    // VM resized successfully!
  }
});
```

### DIFF
--- a/packages/compute/src/vm.js
+++ b/packages/compute/src/vm.js
@@ -424,16 +424,45 @@ VM.prototype.reset = function(callback) {
  * @param {string} machineType - Full or partial machine type. See a list of
  *     predefined machine types
  *     [here](https://cloud.google.com/compute/docs/machine-types#predefined_machine_types).
+ * @param {object=} options - Configuration object.
+ * @param {boolean} options.start - Start the VM after successfully updating the
+ *     machine type. Default: `false`.
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request.
  * @param {object} callback.apiResponse - The full API response.
  *
  * @example
- * vm.resize('n1-standard-1', function(err, apiResponse) {});
+ * vm.resize('n1-standard-1', function(err, apiResponse) {
+ *   if (!err) {
+ *     // The VM is running and its machine type was changed successfully.
+ *   }
+ * });
+ *
+ * //-
+ * // By default, calling `resize` will start your server after updating its
+ * // machine type. If you want to leave it stopped, set `options.start` to
+ * // `false`.
+ * //-
+ * var options = {
+ *   start: false
+ * };
+ *
+ * vm.resize('ns-standard-1', options, function(err, apiResponse) {
+ *   if (!err) {
+ *     // The VM is stopped and its machine type was changed successfully.
+ *   }
+ * });
  */
-VM.prototype.resize = function(machineType, callback) {
+VM.prototype.resize = function(machineType, options, callback) {
   var self = this;
   var compute = this.zone.parent;
+
+  if (is.fn(options)) {
+    callback = options;
+    options = {};
+  }
+
+  options = options || {};
 
   var isPartialMachineType = machineType.indexOf('/') === -1;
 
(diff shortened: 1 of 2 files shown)


packages/compute/src/vm.js
```diff
@@ -414,6 +414,61 @@ VM.prototype.reset = function(callback) {
...
+  if (isPartialMachineType) {
+    machineType = format('zones/{zoneName}/machineTypes/{machineType}', {
+      zoneName: this.zone.name,
+      machineType: machineType
+    });
+  }
+
+  callback = callback || common.util.noop;
+
+  this.stop(function(err, operation, apiResponse) {
```
callmehiphop: So a couple questions in regards to this..

1. If the VM is not running, will this throw an error?
2. If it is running and we stop it, should we re-start it also?

stephenplusplus: The API request will fail if it's not stopped, and we should probably restart it, yes!

stephenplusplus:
@callmehiphop could you take a look at the updates and see if this is the best case for the user? It should definitely be the most "automatic", I'm just not sure if it could create surprises. I've played with an `options` object to control the behavior, but it felt too verbose and confusing. LMK if you think this new behavior is okay or if you have any ideas on how to improve.

packages/compute/src/vm.js
```diff
@@ -579,4 +637,24 @@ VM.prototype.request = function(reqOpts, callback) {
   });
 };
 
+VM.prototype.execAfterOperation_ = function(callback) {
```
callmehiphop: Out of curiosity, why do we want this on the prototype chain?

stephenplusplus: It doesn't have to be. And it should really go on `Compute`, as we might want this from many methods. And *really*, it should be exposed to the user, but not like that ^, like this:

```js
vm.stop({ operation: false }, function(err, results) {
  // operation completed
})
```

... but that's a problem for a different day.

callmehiphop:
I'm a bit hesitant making all these guesses, only because to the user it might feel a bit magical. What if instead of just automatically doing all this, we added an option like `restart`? That way it was a bit less magical and if for some reason the user wanted to programmatically restart the VM on their own, they could.

stephenplusplus:
That is an option, but makes the API have different response args to the user's callback. If we just do the API call, which assumes the VM is not running, we would return an operation. If we do a restart, what do we return? I thought it would be weird to return the `vm.start()` operation.

callmehiphop:
Perhaps `setMachineType` shouldn't restart the VM then? We can have it simply set the machine type and nothing else. Then add a convenience method that sits on top of it and will restart the VM?

stephenplusplus:
What would that look like? Having two methods for doing the same thing isn't very "us"-- in this case, I think we just have to pick one or the other. A machine type can only be changed if the machine isn't running, so that means if we just expose the "raw" way (just make the API call), the user will always have to call `vm.stop()` first. And since we're usually "smart" for the user, that's why I keep ending up at doing it for them.

Maybe we should wait for the next meeting and bring it up then.

stephenplusplus:
Feedback was to change to a name that implies magic is happening, i.e. `vm.resize()`. Whereas `setMachineType` doesn't imply "we're restarting your VM".

stephenplusplus:
@callmehiphop PTAL! I renamed the method to `resize`, put a "private" `execAfterOperation_` method on the top level Compute instance*, wrote some tests, and added docs.

\*the method is not static because to use it from "VM", we introduce a cross dependency from `index.js` and `vm.js`. This didn't make the `require` call throw, but did result in an `undefined` `Compute.execAfterOperation_` method. It was easier to put it on the Compute instance.

packages/compute/src/vm.js
```diff
@@ -407,6 +407,74 @@ VM.prototype.reset = function(callback) {
...
+ * Set the machine type for this instance, **stopping and restarting the VM as
+ * necessary**.
+ *
+ * For a list of the standard, high-memory, and high-CPU machines you may choose
+ * from, see
+ * [Predefined machine types]{@link https://cloud.google.com/compute/docs/machine-types#predefined_machine_types}.
+ *
+ * In order to change the machine type, the VM must not be running. This method
+ * will automatically stop the VM if it is running before changing the machine
+ * type. After it is sucessfully changed, the VM will be started.
```
callmehiphop: WDYT about adding a `start: false` option? By default we restart the vm, however they can optionally leave it off if they specify?

stephenplusplus: Good call. Added!

Issue #1471, opened 2016-08-07
Releasing the modules


stephenplusplus:
Our repo has been split up into ready-to-publish modules. This issue will track what needs to be done before we can get the next release out.

- [x] Make a module-friendly docs site: [PR](https://github.com/GoogleCloudPlatform/gcloud-common/pull/157)
- [x] Modify our JSDocs -> JSON build process to suit the changes from [the docs site PR](https://github.com/GoogleCloudPlatform/gcloud-common/pull/157)
  - [x] Combine `@classdesc` docs with `@description` docs
  - [x] Write package specific *Getting Started* docs in `@description` tag
  - [x] Automate <kbd>toc.json</kbd> creation
  - [x] Move configuration object documentation to a *Getting Started* markdown file
  - [x] Update all links to configuration objects
- [x] Publish all `@google-cloud/service` modules to npm at `0.1.0`
- [x] Publish the umbrella package, `google-cloud`, at `0.38.0`
- [x] Revert README change

stephenplusplus:
@callmehiphop please update the post if I got any of those steps wrong / left anything out.

nvartolomei:
`Publish the umbrella package, google-cloud`

next time, please, update readme at the same time you publish new npm package or change the active branch on github so people do not read broken readme/doc or add notices like kubernetes repo does.

stephenplusplus:
Yep, we should have left a message for users directing them to the accurate docs here: https://googlecloudplatform.github.io/gcloud-node/#/docs/v0.37.0/gcloud. I'll do that now, sorry for the confusion.

stephenplusplus:
We're published!

PR #1475, opened 2016-08-09
docs: generate uniform service overviews

stephenplusplus:
RE: #1471

@callmehiphop I wanted to get your thoughts before I move forward. The changes in this PR:
- Removes custom constructor descriptions. They mostly were filler, slight variants of something like "This is how you interact with a {serviceName} object".
- Removes constructor `@example` blocks. This was mostly used to show how to create a client-- that is now covered in the generated description.
- Creates a template file that should be used by all constructors for their description. This file should be parsed with `_.template()`, using a common templating syntax. This is done because each description should be different, depending on the context the user is viewing the docs in: `umbrellaMode` or not.
- Exports a function from `scripts/docs.buildOverview.js` that takes a class name, like `bigquery`, and returns the templated description. This returned description should be pre-pended to any existing description a constructor might have.

Constructor `@example` blocks are currently used to initialize variables that are used throughout the rest of the class's examples. Removing these blocks created a bit of a problem, so as a solution, in `test/docs.js`, it will prepend example code with a manual instantiation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "docs": "node ./scripts/docs/packages.js",
     "bundle": "node ./scripts/docs/bundle.js",
     "lint": "jshint scripts/ packages/ system-test/ test/ && jscs packages/ system-test/ test/",
-    "test": "npm run docs && mocha test/docs.js packages/*/test/*.js",
+    "test": "npm run docs && npm run bundle && mocha test/docs.js packages/*/test/*.js",
     "system-test": "mocha packages/*/system-test/*.js --no-timeouts --bail",
     "cover": "istanbul cover _mocha --report lcovonly -- --no-timeouts --bail packages/*/test/*.js -R spec",
     "coveralls": "npm run cover && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage"
--- a/packages/bigquery/src/table.js
+++ b/packages/bigquery/src/table.js
@@ -48,13 +48,6 @@ var FORMATS = {
  *
  * @param {module:bigquery/dataset} dataset - Dataset instance.
  * @param {string} id - The ID of the table.
- *
- * @example
- * var bigquery = gcloud.bigquery({ projectId: 'grape-spaceship-123' });
- * var Dataset = require('gcloud/lib/bigquery/dataset');
- * var dataset = new Dataset(bigquery, 'dataset-id');
- * var Table = require('gcloud/lib/bigquery/table');
- * var table = new Table(dataset, 'table-id');
  */
 /**
  * Table objects are returned by methods such as
@@ -65,13 +58,6 @@ var FORMATS = {
  * @constructor
  *
  * @example
- * var gcloud = require('google-cloud')({
- *   keyFilename: '/path/to/keyfile.json',
- *   projectId: 'grape-spaceship-123'
- * });
- *
- * var bigquery = gcloud.bigquery();
- *
  * var dataset = bigquery.dataset('my-dataset');
  *
  * var table = dataset.table('my-table');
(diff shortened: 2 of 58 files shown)


scripts/overview.template.html
```diff
@@ -0,0 +1,44 @@
...
+var <%= instanceName || className %> = gcloud.<%= className %>({
+  projectId: 'grape-spaceship-123',
+  keyFilename: '/path/to/keyfile.json'
+});
+<% } %>
+</div>
+
+The full set of options which can be passed to `<%= pkgJson.name %>` are
+outlined in our
+[Authentication guide](#/docs/<%= className %>/<%= pkgJson.version %>/guides/authentication).
```
stephenplusplus: Note to self that this link will need to change if in umbrella mode.

callmehiphop:
Where would the overview content be visible to the user? In the collapsable menu like today? Or before the constructor description?

stephenplusplus:
There is only one body of text that describes a service and its constructor. So the Getting Started and Overview sections are gone. All content would appear as part of the constructor's description.

stephenplusplus:
This has been absorbed by #1479.

stephenplusplus:
@callmehiphop

stephenplusplus:
It looks like the overview changes from this PR weren't carried over into #1479. The docs site still show 3 different sections for each API;

- Getting Started
- Overview
- Constructor description

There is only meant to be one now, under the label of the constructor, i.e.:

screen shot 2016-08-12 at 9 30 11 am
screen shot 2016-08-12 at 9 30 20 am

stephenplusplus:
Updated this PR, PTAL.

I also stuck Datastore constants on the prototype... it was never very convenient accessing them:

```js
var datastore = require('@google-cloud/datastore')({ ... });
datastore.int // undefined (until now)
```

Are there other constants I need to do this for?

Until we get this change in, I don't think we should officially "release" (publishing the release notes), since the docs right now are pretty confusing, e.g. [the first example on the BigQuery module's page](https://googlecloudplatform.github.io/gcloud-node/#/docs/bigquery/0.1.1/bigquery) tells you to `require('google-cloud')`.

package.json
```diff
@@ -29,7 +29,7 @@
     "docs": "node ./scripts/docs/packages.js",
     "bundle": "node ./scripts/docs/bundle.js",
     "lint": "jshint scripts/ packages/ system-test/ test/ && jscs packages/ system-test/ test/",
-    "test": "npm run docs && mocha test/docs.js packages/*/test/*.js",
+    "test": "npm run docs && npm run bundle && mocha test/docs.js packages/*/test/*.js",
```
callmehiphop: This might not work as you intend it to. The tricky thing about bundling the docs together is that what is in the `master` branch of `google-cloud` could be a slightly older version of any one of the dependencies  (e.g. `@google-cloud/bigquery`).

That being said when I wrote the bundle task I had to consider using one of the following approaches:

1. Using a submodule to checkout a tag for each dependency and re-running the docs against said tag.
2. Copying pre-existing JSON and modifying keys to contain code specific to their package.

I ended up going with option 2. That being said, when you bundle, it assumes there's a ghpages submodule that it can find the appropriate version of any `@google-cloud` packages and copies them.

All that being said, I don't think this task will do what you're hoping it does :<

stephenplusplus: I'll have to learn more about this later. I just assumed it was necessary to run before 'npm test'. If not and its the only thing holding the PR back, can you just change it on my fork/branch?

packages/bigquery/src/table.js
```diff
@@ -473,7 +459,7 @@ Table.prototype.createWriteStream = function(metadata) {
  * @throws {Error} If destination format isn't recongized.
  *
  * @example
- * var gcs = gcloud.storage({
+ * var gcs = require('@google-cloud/storage')({
```
callmehiphop: Assuming that the overview gets moved into the constructor description, why do we need this? Shouldn't the tokenized overview contain this already?

stephenplusplus: This is a BigQuery API that is using storage. Let me know if I misunderstood your question.

callmehiphop: Ah, my mistake.

callmehiphop:
@stephenplusplus LGTM, I think we should wait for an update to the docs site before merging this though so that our overview isn't collapsed.

PR #1476, opened 2016-08-10
Introduce Gapic-generated files for language API.

### LINKED ISSUE #1463: Integration with Gapic-generated code
I'm in charge of Gapic codegen and GAX for NodeJS
- gax: https://github.com/googleapis/gax-nodejs
- code generator: https://github.com/googleapis/toolkit

It was an independent effort, but recently we reached to an agreement that bringing the generated code into a part of gcloud itself.

It's still unclear to me where the generated code will be located and how these files are accessible from users, so this is an issue to discuss about these things.

### LINKED ISSUE #1464: Cloud Language API
Bring cloud language API (gRPC and Gapic-generated code). This would be an initial effort for https://github.com/GoogleCloudPlatform/gcloud-node/issues/1463.

The proto files are https://github.com/googleapis/googleapis/tree/master/google/cloud/language/v1beta1

jmuk:
Fixes #1463
Fixes #1464 
- language_service_api.js and language_service_client_config.json:
  auto-generated
- v1beta1/index.js: manually written
- index.js and package.json: manually modified

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "lint": "jshint scripts/ packages/ system-test/ test/ && jscs packages/ system-test/ test/",
     "test": "npm run docs && npm run bundle && mocha test/docs.js packages/*/test/*.js",
     "system-test": "mocha packages/*/system-test/*.js --no-timeouts --bail",
-    "cover": "istanbul cover _mocha --report lcovonly -x 'packages/*/src/v[0-9]*/*.js' -- --no-timeouts --bail packages/*/test/*.js -R spec",
+    "cover": "istanbul cover _mocha --report lcovonly -x 'packages/*/src/v*/*.js' -- --no-timeouts --bail packages/*/test/*.js -R spec",
     "coveralls": "npm run cover && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage"
   },
   "license": "Apache-2.0",


jmdobry:
So, to use it:

```js
var Language = require('@google-cloud/language');

// hand-written client, highest level of abstraction
var language = Language();

// auto-gen client, middle level of abstraction, still has retry, batching, etc.
var api = Language.v1beta1.LanguageServiceApi();

// grpc stub, lowest level of abstraction, just a grpc stub
var grpc = Language.grpc();
```

jmuk:
That's right -- but two notes:
- on the second part, `api` has a property called `stub` for a Promise to the gRPC client stub. So if the user wants to use a gRPC stub, they can always use like `api.stub.then(function(stub) { stub.annotateText(...); }`
- I actually added `grpc()` mostly for enums.  Many requests contain enum properties, and we should allow the access to those constants. For example:

```
var languageGrpc = Language.v1beta1.grpc();
api.annotateText({type: languageGrpc.Document.Type.PLAIN_TEXT, content: "Hello, world!"}, ...);
```

I am not sure what is the best design to reveal those gRPC constants to the users -- please let me know if you have some better ideas.

jmdobry:
> I am not sure what is the best design to reveal those gRPC constants to the users

Wouldn't the hand-written part want to expose them to the user for easiest access? We'd also want JSDoc comments on them so they get into the library's documentation.

stephenplusplus:
> on the second part, `api` has a property called `stub` for a Promise to the gRPC client stub. So if the user wants to use a gRPC stub, they can always use like `api.stub.then(function(stub) { stub.annotateText(...); }`

Just curious why this is a promise. Does anything async need to happen before a stub can be used? In other words, why not just:

```js
var api = Language.v1beta1.LanguageServiceApi();
api.stub.annotateText();
```

I think I like accessing the "stub" through a different method better than the `stub` property, like in @jmdobry's example:

```js
// grpc, lowest level of abstraction, just a grpc stub, have to do auth manually
var LanguageService = Language.v1beta1.grpc().LanguageService;
var stub = new LanguageService('language.googleapis.com', credentials); // credentials acquired manually
stub.annotateText(...);
stub.analyzeEntities(...);
stub.analyzeSentiment(...);
```

How I would personally use that:

```js
var gcloudLanguage = require('@google-cloud/language');
var grpc = gcloudLanguage.v1beta1.grpc();

var language = new grpc.LanguageService('language.googleapis.com', credentials);
language.annotateText({
  // something that needs an enum
  type: grpc.Document.Type.PLAIN_TEXT
}, ...);
language.analyzeEntities(...);
language.analyzeSentiment(...);
```

--

In the case where there are multiple services that a user can instantiate, isn't it common or always the case that they can share an auth client? If that's true, maybe we should make the `v1beta1` property a function instead, that can persist a single auth client / other config details that can be re-used in the services beneath it:

```diff
-var api = language.v1beta1.LanguageServiceApi();
-var language = api.LanguageServiceApi({ projectId: 'grape-spaceship-123', keyFilename: '...' });
-var otherService = api.OtherServiceApi({ projectId: 'grape-spaceship-123', keyFilename: '...' });

+var api = language.v1beta1({ projectId: 'grape-spaceship-123', keyFilename: '...' });
+var language = api.LanguageServiceApi();
+var otherService = api.OtherServiceApi();
// ^-- both share the auth client

language.annotateText(...);
otherService.method(...);
```

--

Anywhere there is an upper-camelcase identifier, `new` is expected to be used. Code linters will throw something like "`new` expected on constructor function" if they see:

```js
var language = api.LanguageServiceApi();
```

My preference would be to not require `new` for the auto-gen client, and just rename the methods to lower camelcase (`api.languageServiceApi()`).

--

Would you explain more about what kind of documentation is expected for these new APIs? Specifically, will the auto gen client have a separate page that lists all of the methods and arguments, with examples? And will the "stub" have a separate page with the same? Since we're going to have 3 ways of doing the same thing, I'm concerned about the added complexity on our docs site. Let me know if there's an existing vision for how these docs will intertwine.

jmuk:
Pushed a new patch, but that fixes minor syntactic things.

> > on the second part, api has a property called stub for a Promise to the gRPC client stub. So if the user wants to use a gRPC stub, they can always use like api.stub.then(function(stub) { stub.annotateText(...); }
>
> Just curious why this is a promise. Does anything async need to happen before a stub can be used? 

That's because gRPC stub needs to wait for the authentication. `languageServiceApi` handles the authentication and gRPC channel setup, and then creates the stub. Promise packages these asynchronous behavior.

Also `languageServiceApi` offers methods, so normally users don't have to access `stub` directly unless they want to access the bare gRPC client behind the API instance.

jmuk:
> In the case where there are multiple services that a user can instantiate, isn't it common or always the case that they can share an auth client?

Right now Google-GAX has its own auth factory as https://github.com/googleapis/gax-nodejs/blob/master/lib/grpc.js#L36, so multiple service instances share the same client instance if the service instances are under the same package.

By adopting google-auto-auth library, your design would make more sense -- but thinking further, the auth client can't be shared among multiple APIs (like Language API and Speech API), right?
```javascript
var language = require('@google-cloud/language').v1beta1({keyFile: '...'});
var speech = require('@google-cloud/speech').v1({keyFile: '...'});
var langApi = language.languageServiceApi();
var speechApi = speech.speechApi();
langApi.annotateText(...);
speechApi.nonStreamingRecognize(...);
// langApi and speechApi don't share the auth client...
```

So rather than doing this, it would probably better to accept an 'auth' parameters, wouldn't it?
```javascript
var Auth = require('google-auto-auth');
var auth = new Auth({keyFile: '...'});
var language = require('@google-cloud/language').v1beta1;
var speech = require('@google-cloud/speech').v1;
var langApi = language.languageServiceApi({auth: auth});
var speechApi = speech.speechApi({auth: auth});
...
```

jmuk:
> Would you explain more about what kind of documentation is expected for these new APIs? Specifically, will the auto gen client have a separate page that lists all of the methods and arguments, with examples? And will the "stub" have a separate page with the same? 

We want to have a separate page for the auto-gen clients, with the list of methods, arguments, and examples.

I don't care about stubs. We don't require showing their documents, it does not affect the usability of auto-gen clients.

jmdobry:
> So rather than doing this, it would probably better to accept an 'auth' parameters, wouldn't it?

That was my initial thought on the subject.

> I don't care about stubs. We don't require showing their documents, it does not affect the usability of auto-gen clients.

Yeah, the stubs are just a proto file turned into JavaScript by Protobuf.js, and a user can look at the proto file or the API's gRPC reference docs to understand the resulting structure of the stub.

stephenplusplus:
```js
var language = require('@google-cloud/language').v1beta1({keyFile: '...'});
var speech = require('@google-cloud/speech').v1({keyFile: '...'});
var langApi = language.languageServiceApi();
var speechApi = speech.speechApi();
langApi.annotateText(...);
speechApi.nonStreamingRecognize(...);
```

I would rather not require the user to install & learn an external dependency (google-auto-auth) in order to be able to use the autogen layer. Also, google-auto-auth returns an auth client, already bound to the required scopes. The user would have to know the scopes the API requires to create the correct auth client. We should just force the cost on the user of duplicate config objects containing their keyfile, projectId, and other required settings from the API.

--

> That's because gRPC stub needs to wait for the authentication. languageServiceApi handles the authentication and gRPC channel setup, and then creates the stub. Promise packages these asynchronous behavior.

I think our libraries somewhat clash here, since we don't support promises (yet), and I think when we do, we would put anything that returns a promise behind a function call, e.g. `getStub().then(...)` rather than `stub.then(...)`. I know `stub` will technically exist on the instantiated object, but can we only document and advise using the `grpc()` method?

I'm on board with `stub` if we can get it to work like this:

```js
var api = Language.v1beta1.LanguageServiceApi();
api.stub.annotateText();
```

--

> multiple service instances share the same client instance if the service instances are under the same package.

```js
var service1 = Language.v1beta1.ServiceOneApi({ auth details });
var service2 = Language.v1beta1.ServiceTwoApi(); // how does this know about `{auth details}`
```

The user would have to provide the auth details twice, resulting in two auth clients being created, right? If it moves a level up, I believe those issues would go away:

```js
var languageServices = Language.v1beta1({ auth details });
var service1 = languageServices.ServiceOneApi();
```

jmuk:
I didn't intend to advertise the existence of `stub`.  It's not expected to be used by normal users. If it attracts your attentions, I now think we should just hide it. That would be simpler.

jmuk:
I wanted to allow injecting gRPC module itself too, to share the same C-extension. This is quite limited usage, but this would be helpful to some expert users in some case.

I am in the process of modifying `v1beta1` as a function (taking auth params), and thinking this gRPC injection as a part of this option. This means:
```javascript
var language = require('@google-cloud/language');
var service = language.v1beta1({keyFile: ...});
var api = service.LanguageServiceApi();
api.annotateText({type: service.grpc.Document.Type.PLAIN_TEXT, ... });
```

If the user has some special need to customize gRPC module:
```javascript
var grpc = ...;
var service = language.v1beta1({keyFile: ..., grpc: grpc});
...
```

jmuk:
Forgotten to mention: I didn't think of the difference of scopes, and indeed that would affect the design. Thank you for pointing it out.

stephenplusplus:
Removing `stub`, making `v1beta1` as a function that takes auth params & a gRPC argument sounds good to me 👍

jmuk:
Uploaded a new patchset. It relies on a patch for google-gax at https://github.com/jmuk/gax-nodejs/commit/65052ae1e147a213de6e75821ef8a3399dc1c0c2

jmdobry:
`grpc` is no longer exported? How would the user get access to the grpc stub?

jmuk:
I think normally users won't access to grpc stubs, but in case they want, `grpc` will be exported like
```javascript
var language = require('@google-cloud/language');
var v1beta1 = language.v1beta1({keyFile: ...});
var stub = new v1beta1.grpc.LanguageService(...);
```

jmuk:
Another idea for `grpc` came up through our offline-discussion, simply grpc objects and Gapic-generated code are in the same namespace, i.e.
```javascript
var language = require('@google-cloud/language');
var v1beta1 = language.v1beta1({keyFile: ...});
var api = v1beta1.LanguageServiceApi();
api.annotateText({type: v1beta1.Document.Type.PLAIN_TEXT, ...});
// not v1beta1.grpc.Document.Type.PLAIN_TEXT
```

jmdobry:
> `var stub = new v1beta1.grpc.LanguageService(...);`

Got it.

jmdobry:
So are we pretty satisfied with how this works? Can your manual edits be automated at all? The manual stuff seems pretty minimal, seems like it could be automated.

packages/language/src/v1beta1/language_service_api.js
```diff
@@ -0,0 +1,249 @@
...
+   *   The customized config to build the call settings. See
+   *   {@link gax.constructSettings} for the format.
+   * @param {number} opts.timeout
+   *   The default timeout, in seconds, for calls made through this client.
+   * @param {number} opts.appName
+   *   The codename of the calling service.
+   * @param {String} opts.appVersion
+   *   The version of the calling service.
+   */
+  function LanguageServiceApi(opts) {
```
stephenplusplus: Can you move `LanguageServiceApi` from outside of the `build()` function? Having it this way causes it to be re-created each time `build()` is called.

packages/language/src/v1beta1/index.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
```
stephenplusplus: I believe this file needs the copyright header comment.

packages/language/src/v1beta1/index.js
```diff
@@ -0,0 +1,11 @@
...
+
+var languageServiceApi = require('./language_service_api');
+
+function v1beta1(options) {
```
stephenplusplus: You should be able to remove the `v1beta1` function and just export it later:

```js
module.exports = languageServiceApi;
```

The constants will be picked up as well, which I believe makes this file unnecessary. Maybe `langage_service_api.js` should just be `index.js`?

packages/language/src/v1beta1/language_service_api.js
jmuk: Done

packages/language/src/v1beta1/index.js
jmuk: Done

packages/language/src/v1beta1/index.js
jmuk: I modified the handling of the options a bit, and I believe this file is necessary.

In case that an API consists of multiple services (like pubsub -- it has Publisher and Subscriber), I think it's more natural to have this index.js, which creates the auth client (through GAX library) and passes to the service client files.

Please take another look.

packages/language/src/v1beta1/index.js
```diff
@@ -0,0 +1,35 @@
...
+  options = options || {};
+  if (!('scopes' in options)) {
+    options.scopes = v1beta1.ALL_SCOPES;
+  }
+  var gaxGrpc = gax.grpc(options);
+  var result = {};
+  extend(result, languageServiceApi(gaxGrpc));
+  return result;
```
jmdobry: What is the purpose of `result`? Why not `return languageServiceApi(gaxGrpc);`?

jmuk: Removed.  It was added to make sure the pattern of multiple services in an API (in that case, services are merged into `result`, but not necessary in this case).

jmdobry:
> v1beta1/index.js: manually written
> index.js and package.json: manually modified

* `v1beta1/index.js` - does this really need to be written manually?
* `index.js` - Check for string `Language.v1beta1 = require('./v1beta1');`, if not there append to file. It's okay for it to be after `module.exports = Language`.
* `package.json` - NPM can update this for you, just have the generator run `npm install --save arguejs@0.2.3 google-gax@0.6.0` in the `language` directory.

jmuk:
> v1beta1/index.js - does this really need to be written manually?

This can be generated automatically, but I manually created to keep this discussion.
I'll set up a generator after we settle down the basic structure.

> index.js - Check for string Language.v1beta1 = require('./v1beta1');, if not there append to file. It's okay for it to be after module.exports = Language.

I may miss your point, but modified index.js so that the edit is simply an addition of a line to index.js, and then this edit can also be automatically generated.

> package.json - NPM can update this for you, just have the generator run npm install --save arguejs@0.2.3 google-gax@0.6.0 in the language directory.

Thanks. Will do in the next API (probably adding that in the code generation pipeline).

packages/language/src/v1beta1/index.js
stephenplusplus: Would you mind doing a quick mock of what index.js would look like for PubSub? I think that would help me understand.

jmdobry:
> I may miss your point, but modified index.js so that the edit is simply an addition of a line to index.js, and then this edit can also be automatically generated.

Your edit should work fine.

packages/language/src/v1beta1/index.js
jmuk: That should be something like:
```javascript
var publisherApi = require('./publisher_api');
var subscriberApi = require('./subscriber_api');

function v1(options) {
  options = options || {};
  if (!('scopes' in options)) {
    // might be good to use lodash.union?
    options.scopes = [];
    options.scopes.push.apply(options.scopes, publisherApi.ALL_SCOPES);
    options.scopes.push.apply(options.scopes, subscriberApi.ALL_SCOPES);
  }
  var gaxGrpc = gax.grpc(options);
  var result = {};
  extend(result, publisherApi(gaxGrpc));
  extend(result, subscriberApi(gaxGrpc));
  return result;
}
v1.SERVICE_ADDRESS = publisherApi.SERVICE_ADDRESS;
module.exports = v1;
```

notes:
- server_address is common among files, simply picking up one is fine
- scopes might be different, needs to be merged
- the second `extend` might replace some of `publisherApi()` result, but essentially they point to the same structure, so this would be fine

packages/language/src/v1beta1/index.js
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+'use strict';
+
+var languageServiceApi = require('./language_service_api');
+var gax = require('google-gax');
+
+function v1beta1(options) {
+  options = options || {};
+  if (!('scopes' in options)) {
```
stephenplusplus: In what scenario would the user provide their own scopes?

jmuk: I don't have a specific use case for this. I simply think it's good to allow customizing it.

stephenplusplus: I agree with customization as well, but in the case of scopes, won't the API refuse to work if the user changes them? I don't think we want to support that.

jmuk: It would work as long as it contains the scopes for the operations. For example, when 'https://www.googleapis.com/auth/logging.write' is specified but 'https://www.googleapis.com/auth/logging.read' isn't, the user code can't read any logs but can write logs.

But anyways this sounds a really rare use case, we might not have to care about that (and probably it might be better to be handled at the dev console screen, not here).

stephenplusplus: Oh, okay, I didn't consider that. I'm good with whatever you think.

packages/language/src/index.js
```diff
@@ -434,3 +434,4 @@ Language.prototype.text = function(content, options) {
 };
 
 module.exports = Language;
+module.exports.v1beta1 = require('./v1beta1');
```
stephenplusplus: (RE: #1492)

We should probably put some documentation here to show the user what they're getting. Should we wait to see how #1492 resolves, or put up something basic for now to make this merge-able?

jmuk: As the initial step, I think we can skip comments here -- so it exists but nobody would notice. That's not our final goal but won't cause any problems.
Then let me look into the doc scripts for #1492, which may end up with more edits here, but I think that would be acceptable.

packages/language/src/v1beta1/index.js
```diff
@@ -0,0 +1,32 @@
...
+    options.scopes = v1beta1.ALL_SCOPES;
+  }
+  var gaxGrpc = gax.grpc(options);
```
stephenplusplus: All of the preceding lines can be replaced with:

```js
options = extend({
  scopes: v1beta1.ALL_SCOPES
}, options);
```

stephenplusplus:
Is this the correct intended usage so far?

```js
var languageGrpc = require('@google-cloud/language').v1beta1({
  keyFilename: '/Users/stephen/dev/keyfile.json'
});

var language = languageGrpc.languageServiceApi();

language.analyzeEntities(..., function(err, apiResponse) {});
```

packages/language/src/v1beta1/index.js
jmuk: Done

jmuk:
Yes, that's the intended usage.

stephenplusplus:
@callmehiphop do you think the API looks okay (basically: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1476#issuecomment-240513318) Any other thoughts on the generated code or anything else?

callmehiphop:
A little late to the party, so please forgive me. I kinda think that the version being a method name is a little confusing - only because it could imply that it's loading a different version of our library when in reality they'll have a lot of differences.

What about something like..

```js
var languageGrpc = require('@google-cloud/language').grpc({
  version: 'v1beta1',
  keyFilename: '/Users/stephen/dev/keyfile.json'
});

var language = languageGrpc.languageServiceApi();

language.analyzeEntities(..., function(err, apiResponse) {});
```

Although my concern can probably be remedied easily with clear documentation. So aside from that LGTM!

stephenplusplus:
That's a good point, I like `grpc` as the method name accepting a version well. Or, just move what we have in this PR currently behind a `grpc` object:

```js
var languageGrpc = require('@google-cloud/language').grpc.v1beta1({
  keyFilename: '/Users/stephen/dev/keyfile.json'
});

var language = languageGrpc.languageServiceApi();

language.analyzeEntities(..., function(err, apiResponse) {});
```

What do you think, @jmdobry @jmuk?

jmuk:
I want to invite @omaray here.

I like the design itself but I'm a bit hesitant about the word `grpc`, because it contains both our generated code and gRPC code, I personally like to name it something else.

An idea is `gax` which we've used in some other places:

```javascript
var languageGax = require('@google-cloud/language').gax({
  version: 'v1beta1', 
  ...
```

jmdobry:
What about

```js
var languageApi = require('@google-cloud/language').api({
  version: 'v1beta1', 
  ...
});

var language = languageApi.languageServiceApi();
language.analyzeEntities(..., function(err, apiResponse) {});
```
or


```js
var languageClient = require('@google-cloud/language').client({
  version: 'v1beta1', 
  ...
});

var language = languageClient.languageServiceApi();
language.analyzeEntities(..., function(err, apiResponse) {});
```

stephenplusplus:
@jmdobry I think we'd run into the same issue as @callmehiphop mentioned; confusion between handwritten vs autogen. The proper names are the differentiators. I'm cool with `gax`, but it feels like there might be a better term out there that can succinctly describe *what* it is.

jmuk:
Well, then `autogen` might be an idea. Not cool nor concise, but descriptive and probably acceptable/good enough. Anyways those files are automatically generated.

```javascript
var languageAutogen = require('@google-cloud/language').autogen({
  version: 'v1beta1',
  ...
});
var languageApi = languageAutogen.languageServiceApi();
languageApi.analyzeEntities(...)
```

omaray:
So let's not use "autogen" or "grpc".

To put things in perspective, here's how we do it for Python:

``
from google.cloud import pubsub # handwritten
from google.cloud.pubsub.v1 import publisher_api # autogen
``

Because keep in mind that for APIs like PubSub, the autogen layer has two services: the Publisher and the Subscriber. But here we see that we focus on a specific version and then on the specific service. We do call it "api" but that's fine.

Can we do something similar? I know it's not super easy :)

stephenplusplus:
@omaray here's our version of that, as the PR stands now (@jmuk correct me if this is wrong):

```js
var handwritten = require('@google-cloud/pubsub')({
  keyFilename: '/Users/stephen/dev/keyfile.json'
});

handwritten.getTopics(...);

var autogen = require('@google-cloud/pubsub').v1({
  keyFilename: '/Users/stephen/dev/keyfile.json'
});

var publisher = autogen.publisherServiceApi();

publisher.listTopics(...);
```

How does that look?

jmuk:
@stephenplusplus, right, your sample explains how this PR works currently. And I believe that @omaray will be okay with this.

What @omaray is saying is about @callmehiphop 's suggestion of introducing `grpc` (or something) between `require('@google-cloud/pubsub')` and `v1`, or passing the version as a parameter. He doesn't want to have any words there unless it's really required.

jmuk:
Are there any other additional comments on https://github.com/GoogleCloudPlatform/gcloud-node/pull/1476#issuecomment-240901161 or https://github.com/GoogleCloudPlatform/gcloud-node/pull/1476#issuecomment-240889200?  Are there anything I'm missing?

stephenplusplus:
I think we're set. We just need to find a way to keep our unit test coverage at 100%.

We will need a unit test that checks `module.exports.v1beta1` is being exported properly from <kbd>packages/language/src/index.js</kbd> by creating a test in <kbd>packages/language/test/index.js</kbd>.

And since we aren't worried about writing tests for the generated files, we'll have to find a way to have our linter skip those files. The easiest way to do that will probably be from using a naming convention and adding it to <kbd>https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/.jshintignore</kbd>

So maybe we want to add:

```
packages/*/src/v* // catch all version directories
```

Or to make it a little more robust, maybe we would place generated files in their own directory to allow:

```
packages/*/src/generated // or something similar
```

jmdobry:
Wouldn't we want the generated code to pass JSHint? At the least it could catch undefined variables, etc., especially since we won't have actual tests for the generated code.

stephenplusplus:
Hmm, maybe I'm getting my dev tools mixed up. What's causing the coverage drop?

package.json
```diff
@@ -31,7 +31,7 @@
     "lint": "jshint scripts/ packages/ system-test/ test/ && jscs packages/ system-test/ test/",
     "test": "npm run docs && npm run bundle && mocha test/docs.js packages/*/test/*.js",
     "system-test": "mocha packages/*/system-test/*.js --no-timeouts --bail",
-    "cover": "istanbul cover _mocha --report lcovonly -- --no-timeouts --bail packages/*/test/*.js -R spec",
+    "cover": "istanbul cover _mocha --report lcovonly -x 'packages/*/src/v[0-9]*/*.js' -- --no-timeouts --bail packages/*/test/*.js -R spec",
```
stephenplusplus: I think this would prevent beta versions from being caught, e.g. `v1beta1`. Just `v*` is probably sufficient.

jmuk: Done

jmuk:
Note that JSHint (and JSCS) doesn't exclude the auto-generated files. I also found their output helpful for me to fix the code generator.

stephenplusplus:
It's all looking good to me. Travis failed because of our repo rename.

Anything left to do? // @jmdobry @jmuk @callmehiphop

jmdobry:
LGTM

jmuk:
LGTM too

stephenplusplus:
Thanks! Now we just need to figure out the docs: #1492

PR #1479, opened 2016-08-11
docs: individual package support

callmehiphop:
Closes #1475 
Relates to #1471 
Depends on https://github.com/GoogleCloudPlatform/gcloud-common/pull/157

[Demo](https://callmehiphop.github.io/gcloud-common-modular/#/docs/google-cloud/v0.1.0/google-cloud)

### DIFF
--- a/scripts/publish.js
+++ b/scripts/publish.js
@@ -57,9 +57,7 @@ setTimeout(function() {
     cwd: cwd
   });
 
-  exec('echo "Now push to master: git push origin master --follow-tags"', {
-    cwd: cwd
-  });
+  console.log('echo "Now push to master: git push origin master --follow-tags"');
 
   // Remove the duplicated files
   rm([


callmehiphop:
@stephenplusplus I'm kinda puzzled on the deployment strategy for this.. seems like there's a lot of duplicated logic between 

* <kbd>scripts/release.sh</kbd>
* <kbd>scripts/build.sh</kbd>
* <kbd>scripts/publish.js</kbd>

Should travis be where we update the individual modules JSON in ghpages? Or in the publish script? I'm a little lost about which scripts fire when now.

scripts/docs/bundle.js
```diff
@@ -0,0 +1,223 @@
...
+var globby = require('globby');
+var semver = require('semver');
+var mkdirp = require('mkdirp');
+var fs = require('fs');
+var async = require('async');
+var flatten = require('lodash.flatten');
+var parser = require('./parser');
+var config = require('./config');
+
+// since we version the JSON in ghpages and we don't create tags for
```
stephenplusplus: My PR had the code in `publish.js` to create tags. If you wanted the changes to `publish` to come after this PR, that's cool. But once we do the tagging, does that change anything here?

stephenplusplus:
I think anything touching `gh-pages` should be handled by Travis.

`publish.js` should just be a convenience wrapper around several commands that we would have to run manually, which will

- create a git tag
- run npm version
- run npm docs

Because the publish script in my PR will create a custom git tag (e.g. `bigquery-0.1.0`), the release.sh script will run because it's a tagged commit. It will need to be modified to parse the git tag (tag.split('-')[0] = module name, tag.split('-')[1] = version number).

stephenplusplus:
fixed type above (`npm publish` not `npm docs`)

scripts/docs/bundle.js
callmehiphop: Maybe, for some reason I didn't think we were going to tag individual packages.

stephenplusplus: I think it's because `npm version v0.1.0` will make a tag called `0.1.0`, which is going to clash with bigquery's 0.1.0, and storage's 0.1.0, etc. But what we will do instead is create a custom tag and tell `npm version` not to auto-create the `0.1.0` tag.

stephenplusplus: For ref, that code's here: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1475/files#diff-e6889d205eee7cf0d8b917c5576051c8R41

scripts/release.sh
```diff
@@ -24,12 +24,20 @@ npm run system-test
 git config --global user.name "travis-ci"
 git config --global user.email "travis@travis-ci.org"
 
+if [[ ${TRAVIS_TAG} == *"-"* ]]; then
```
callmehiphop: @stephenplusplus so based on what we said about tagging, WDYT about these changes?

stephenplusplus:
LGTM!

PR #1480, opened 2016-08-11
Publishing

stephenplusplus:
This renames our `publish` script as `publish-module`. Usage:

`npm run publish-module 0.1.2` from a module directory (issue opened to add docs to explain this: #1481)

Also removed the linking script since all of our deps are on npm now.


packages/bigquery/package.json
```diff
@@ -31,7 +31,7 @@
   ],
   "main": "./src/index.js",
   "files": [
-    "./src/*",
+    "src",
```
stephenplusplus: This is required, otherwise packages are published without any files from `src`... which is why the first usable version of each module starts at `0.1.1`.

PR #1488, opened 2016-08-15
allow Buffer as an input image for vision

calibr:
Hi! I think it would be useful if vision api could allow images as Buffers.
I've seen issue https://github.com/GoogleCloudPlatform/gcloud-node/issues/1449, but I think Buffer is easier to detect than base64 string, I just use `Buffer.isBuffer(image)` and perhaps buffer is more natural way for passing binary data in node. I will be glad to make changes if something wrong with this patch

### DIFF
--- a/packages/vision/src/index.js
+++ b/packages/vision/src/index.js
@@ -1282,10 +1282,8 @@ Vision.findImages_ = function(images, callback) {
 
   function findImage(image, callback) {
     if (Buffer.isBuffer(image)) {
-      process.nextTick(function() {
-        callback(null, {
-          content: image.toString('base64')
-        });
+      callback(null, {
+        content: image.toString('base64')
       });
       return;
     }


calibr:
I signed it!

callmehiphop:
@calibr looks like there are a couple small linting issues, would you mind making those small changes?

You can lint locally via `npm run lint`

calibr:
@callmehiphop of course, should I submit a fix in this merge request?

callmehiphop:
@calibr That would be perfect, we'll just squash the commits when we merge :)

packages/vision/src/index.js
```diff
@@ -1281,6 +1281,14 @@ Vision.findImages_ = function(images, callback) {
   images = arrify(images);
 
   function findImage(image, callback) {
+    if (Buffer.isBuffer(image)) {
+      process.nextTick(function() {
```
callmehiphop: @calibr I think we could probably safely drop the `process.nextTick` part off, WDYT?

calibr: yes I think we can, I will do it

callmehiphop: Awesome, thanks!

calibr:
@callmehiphop  is everything OK now with the patch?

callmehiphop:
@calibr yep, merged! Thanks!

calibr:
@callmehiphop thanks! when the update is going to be on the npm?

callmehiphop:
I think closer to the end of the week

stephenplusplus:
@callmehiphop can you add docs for this?

callmehiphop:
Sure!

PR #1489, opened 2016-08-15
Update home.html

murilolobato:
Fix package name on the faq


callmehiphop:
@murilolobato nice catch! Could you make create a new PR that makes this change in the master branch? (<kbd>docs/home.html</kbd>). Our CI job will copy it to `gh-pages`, so this change would get overwritten.

murilolobato:
Ok, I'll do that!

PR #1500, opened 2016-08-16
docs: updated vision detect params/examples

callmehiphop:
Relates to https://github.com/GoogleCloudPlatform/gcloud-node/pull/1488#issuecomment-240179248

### DIFF
--- a/packages/vision/src/index.js
+++ b/packages/vision/src/index.js
@@ -133,6 +133,7 @@ Vision.prototype.annotate = function(requests, callback) {
   });
 };
 
+// jscs:disable maximumLineLength
 /**
  * Detect properties from an image (or images) of one or more types.
  *
@@ -213,14 +214,15 @@ Vision.prototype.annotate = function(requests, callback) {
  * //-
  * // Run feature detection over a Buffer.
  * //-
- * var fs = require('fs');
+ * var level = require('level');
+ * var db = level('./users-database');
  *
- * fs.readFile('image.jpg', function(err, contents) {
+ * db.get('user-image', { encoding: 'binary' }, function(err, image) {
  *   if (err) {
  *     // Error handling omitted.
  *   }
  *
- *   vision.detect(contents, types, function(err, detection, apiResponse) {});
+ *   vision.detect(image, function(err, detection, apiResponse) {});
  * });
  *
  * //-
@@ -533,6 +535,7 @@ Vision.prototype.detect = function(images, options, callback) {
     });
   });
 };
+// jscs:enable maximumLineLength
 
 /**
  * Run face detection against an image.
--- a/.jscsrc
+++ b/.jscsrc
@@ -17,7 +17,7 @@
   "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties",
   "maximumLineLength": {
     "value": 80,
-    "allowComments": true
+    "allowUrlComments": true
   },
   "excludeFiles": [
     "system-test/data/*",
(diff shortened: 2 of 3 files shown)


packages/vision/src/index.js
```diff
@@ -157,9 +157,10 @@ Vision.prototype.annotate = function(requests, callback) {
  * For the response in the original format, review the `apiResponse` argument
  * your callback receives.
  *
- * @param {string|string[]|module:storage/file|module:storage/file[]} images -
- *     The source image(s) to run the detection on. It can be either a local
- *     image path, a remote image URL, or a @google-cloud/storage File object.
+ * @param {string|string[]|buffer|buffer[]|module:storage/file|module:storage/file[]} images - The
```
callmehiphop: I know this looks a bit weird, but dox doesn't parse it quite right if all the types, the name and at least part of the description are on a single line.

stephenplusplus: I figured, dox is cray.

packages/vision/src/index.js
```diff
@@ -211,6 +212,19 @@ Vision.prototype.annotate = function(requests, callback) {
  * vision.detect(img, types, function(err, detection, apiResponse) {});
  *
  * //-
+ * // Run feature detection over a Buffer.
+ * //-
+ * var fs = require('fs');
+ *
+ * fs.readFile('image.jpg', function(err, contents) {
```
stephenplusplus: I know this is exactly what we support, but examples sometimes appear as "the right way™ to do things". Is there another use case we could write up? @calibr -- how will you be using this feature?

calibr: I'm using this feature with a Buffer received from s3, like below:

```js
S3.getObject({
  Bucket: "my-s3-bucket",
  Key: "s3-file-key"
}, function(err, objectData) {
  vision.detectText(objectData.Body, function(err, text) {
  	
  });
});
```

callmehiphop:
@stephenplusplus WDYT?

stephenplusplus:
I think we just need a different example (https://github.com/GoogleCloudPlatform/gcloud-node/pull/1500#discussion_r75020420). Sorry that I'm struggling to come up with something specific to suggest... the only thing is that it shouldn't confuse people into thinking they have to use fs.readFile() to have an image on disk run through feature detection.

calibr:
I think that there are many sorts of buffers sources, like DBs, remote services, incoming http requests, etc., IMHO doc shouldn't describe all these sources and maybe example should be transformed to a common form of getting buffer like:

```js
getBufferFromSomewhere(function(err, buffer) {
  vision.detect(buffer, types, function(err, detection, apiResponse) {});
});
```

stephenplusplus:
```js
var db = require('level')('./users-database');

db.get('user-image', { encoding: 'binary' }, function(err, img) {
  if (err) {
    // Error handling omitted.
  }

  vision.detect(img, function(err, detection, apiResponse) {});
});
```

callmehiphop:
@stephenplusplus WDYT about using @calibr's example? IIRC a lot of our customers use AWS services as well, so I think it has the added benefit of being a common use case.

stephenplusplus:
Let's see if @jgeewax or @omaray has a preference of which example best demonstrates passing a Buffer with our library off to Vision's feature detection.

packages/vision/src/index.js
```diff
@@ -519,6 +535,7 @@ Vision.prototype.detect = function(images, options, callback) {
     });
   });
 };
+// jscs:enable maximumLineLength
```
callmehiphop: I had to put this below the function itself (which I know is undesirable) because dox was flipping out.

callmehiphop:
@stephenplusplus @calibr updated the example, PTAL

PR #1501, opened 2016-08-17
Fix typos in pub/sub docs

RobertHerhold:
`pulling` with regards to pub/sub should be `polling`

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,13 +421,13 @@ var options = {
 };
 
 topic.subscribe('subscription-name', options, function(err, subscription) {
-  // Register listeners to start pulling for messages.
+  // Register listeners to start polling for messages.
   function onError(err) {}
   function onMessage(message) {}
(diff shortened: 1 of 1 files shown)


RobertHerhold:
@googlebot Signed CLA

callmehiphop:
This actually isn't a typo, albeit I can definitely see where you're coming from on this. When we check for messages on a Subscription we use the [Subscription#pull](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/packages/pubsub/src/subscription.js#L508) method, so when we say *pulling* we mean polling the pull method (via [Subscription#startPulling_](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/packages/pubsub/src/subscription.js#L685)).

RobertHerhold:
Makes sense. Thanks for correcting me on this.

PR #1507, opened 2016-08-18
Create .npmignore

knownasilya:
To cut down on the size of the module on npm

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
 /.github
 /docs
 /test
+/packages/**/test


stephenplusplus:
Thank you for this, but the `package.json` in the root of the repo is a private package: https://github.com/GoogleCloudPlatform/gcloud-node/blob/d606b2aecf8033a96cb512d509fa98ee33a3dc89/package.json#L3

The rest of our packages that are published to npm use the `files` property in their package.json's to limit what is published, e.g. `google-cloud`: https://github.com/GoogleCloudPlatform/gcloud-node/blob/d606b2aecf8033a96cb512d509fa98ee33a3dc89/packages/google-cloud/package.json#L37

Did you install one of the modules from npm that included tests and docs?

knownasilya:
Ah, good!

PR #1524, opened 2016-08-23
translate: Add support for target option to translate.getLanguages.

jmdobry:
Did I do the JSDoc comments right?

This adds support for:

``` js
translate.getLanguages({
  target: 'en'
}, function (err, languages) {
  console.log(languages[0]); // { language: 'ar', name: 'Arabic' }
});
```

``` js
translate.getLanguages({
  target: 'ru'
}, function (err, languages) {
  console.log(languages[0]); // { language: 'ar', name: 'арабский' }
});
```

### DIFF
--- a/packages/translate/src/index.js
+++ b/packages/translate/src/index.js
@@ -166,8 +166,8 @@ Translate.prototype.detect = function(input, callback) {
  *
  * @resource [Discovering Supported Languages]{@link https://cloud.google.com/translate/v2/discovering-supported-languages-with-rest}
  *
- * @param {string=} target - Include the languages names in the response
- *     translated to a target language.
+ * @param {string=} target - Get the language names in a language other than
+ *     English:
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request.
  * @param {string[]|object[]} callback.languages - The supported ISO 639-1
@@ -176,23 +176,6 @@ Translate.prototype.detect = function(input, callback) {
  *
  * @example
  * translate.getLanguages(function(err, languages) {
- *   if (!err) {
- *     // languages = [
- *     //   'af',
- *     //   'ar',
- *     //   'az',
- *     //   ...
- *     // ]
- *   }
- * });
- *
- * //-
- * // Include the languages names in the response translated to a target
- * // language:
- * //-
- * translate.getLanguages({
- *   target: 'en'
- * }, function(err, languages, apiResponse) {
  *   if (!err) {
  *     // languages = [
  *     //   {
@@ -211,11 +194,34 @@ Translate.prototype.detect = function(input, callback) {
  *     // ]
  *   }
  * });
+ *
+ * //-
+ * // Get the language names in a language other than English:
+ * //-
+ * translate.getLanguages('es', function(err, languages) {
+ *   if (!err) {
+ *     // languages = [
+ *     //   {
+ *     //     code: 'af',
+ *     //     name: 'afrikáans'
+ *     //   },
+ *     //   {
+ *     //     code: 'ar',
+ *     //     name: 'árabe'
+ *     //   },
+ *     //   {
+ *     //     code: 'az',
+ *     //     name: 'azerí'
+ *     //   }
+ *     //   ...
+ *     // ]
+ *   }
+ * });
  */
 Translate.prototype.getLanguages = function(target, callback) {
   if (is.fn(target)) {
     callback = target;
-    target = undefined;
+    target = 'en';
   }
 
   var reqOpts = {
@@ -234,18 +240,12 @@ Translate.prototype.getLanguages = function(target, callback) {
       return;
     }
 
-    var languages = resp.data.languages;
-
-    if (!reqOpts.qs.target) {
-      languages = languages.map(prop('language'));
-    } else {
-      languages = languages.map(function(language) {
-        return {
-          code: language.language,
-          name: language.name
-        };
-      });
-    }
+    var languages = resp.data.languages.map(function(language) {
+      return {
+        code: language.language,
+        name: language.name
+      };
+    });
 
     callback(null, languages, resp);
   });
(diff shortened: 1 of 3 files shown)


packages/translate/src/index.js
```diff
@@ -166,10 +166,12 @@ Translate.prototype.detect = function(input, callback) {
  *
  * @resource [Discover Supported Languages]{@link https://cloud.google.com/translate/v2/using_rest#supported-languages}
```
stephenplusplus: Can you add a new resource?:

```
@resource [Discovering Supported Languages]{@link https://cloud.google.com/translate/v2/discovering-supported-languages-with-rest}
```

packages/translate/src/index.js
```diff
@@ -166,10 +166,12 @@ Translate.prototype.detect = function(input, callback) {
...
  *
+ * @param {options=} options - Configuration options.
+ * @param {options} options.target - Configuration options.
```
stephenplusplus: Let's just accept a string that will be used as the target language:

```diff
* @param {string=} target - The language you intend to translate text to.
```

^ Feel free to re-word... just trying to be overly-helpful to the reader.

packages/translate/src/index.js
```diff
@@ -166,10 +166,12 @@ Translate.prototype.detect = function(input, callback) {
...
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request.
- * @param {string[]} callback.languages - The supported ISO 639-1 language
- *     codes.
+ * @param {string[]|object[]} callback.languages - The supported ISO 639-1
```
stephenplusplus: Let's just always return an array of strings. If the user wants the display name property, they can get that from `apiResponse`.

packages/translate/src/index.js
```diff
@@ -183,17 +185,60 @@ Translate.prototype.detect = function(input, callback) {
  *     // ]
  *   }
  * });
+ *
+ * //-
+ * // Get the names of the languages in a target language:
```
stephenplusplus: See the languages to which you can translate text of a specific language.

^ If that reads weird, feel free to translate that to better English.

packages/translate/src/index.js
jmdobry: Done.

packages/translate/src/index.js
jmdobry: Done.

packages/translate/src/index.js
jmdobry: Done.

packages/translate/src/index.js
jmdobry: Done.

packages/translate/src/index.js
```diff
@@ -164,12 +164,14 @@ Translate.prototype.detect = function(input, callback) {
 /**
  * Get an array of all supported languages.
  *
- * @resource [Discover Supported Languages]{@link https://cloud.google.com/translate/v2/using_rest#supported-languages}
+ * @resource [Discovering Supported Languages]{@link https://cloud.google.com/translate/v2/discovering-supported-languages-with-rest}
  *
+ * @param {string=} target - Include the languages names in the response
```
stephenplusplus: Oops, I misunderstood what the `target` option was for. If the purpose is to get more information from the API, I feel bad not giving it directly to the user as you originally were.

So, this is what I would expect:

```js
translate.getLanguages('en', function(err, languages) {
  languages = [
    {
      code: 'af', // changed key from "language" to "code"
      name: 'Afrikaans'
    },
    // ...
  ]
})
```

I think `code` is easier for this case:

```js
translate.getLanguages('en', function(err, languages) {
  languages.forEach(function(language) {
    var code = language.language; // `language.language` :(
  })
})
```

Sorry for the frustration I've surely caused :\

jmdobry: Fixed.

stephenplusplus: Do you think we should just default to English and always return objects?

packages/translate/src/index.js
```diff
@@ -183,17 +185,75 @@ Translate.prototype.detect = function(input, callback) {
...
+ *
+ * //-
+ * // Include the languages names in the response translated to a target
+ * // language:
+ * //-
+ * translate.getLanguages({
+ *   target: 'en'
+ * }, function(err, languages, apiResponse) {
+ *   if (!err) {
+ *     // languages = [
```
stephenplusplus: This needs to be updated, please.

packages/translate/src/index.js
jmdobry: Sounds fine to me, but it's a breaking change.

packages/translate/src/index.js
jmdobry: Fixed.

packages/translate/src/index.js
stephenplusplus: I think this one is okay. I'll post a warning on Twitter just in case anyone is listening :)

packages/translate/src/index.js
stephenplusplus: *(not capped at 80chars):*

```js
 * @param {object[]} callback.languages - The languages supported by the Translate API.
 * @param {string} callback.languages[].code - The ISO 639-1 language code.
 * @param {string} callback.languages[].name - The language name. This can be translated into your preferred language with the `target` option described above.
```

@callmehiphop did I do this JSDocs right?

packages/translate/src/index.js
jmdobry: Done.

stephenplusplus:
Made some small style changes to the tests and docs, merged in https://github.com/GoogleCloudPlatform/google-cloud-node/commit/581c3b76582027b82ada5bd4f507ec609d94bd6d.

stephenplusplus:
Thank you for this!

jmdobry:
Cool! When can I expect 0.2.0 to be published? Is there a set release schedule?

stephenplusplus:
A batch of releases is coming today 👍

stephenplusplus:
Published: https://www.npmjs.com/package/@google-cloud/translate

PR #1439, opened 2016-07-18
bigtable: v2 support

callmehiphop:
Fixes #1526
##### TODOS:
- [x] New APIs
  - [x] Instance
  - [x] Cluster
  - [x] grpc Operations
- [x] Old API Updates
  - [x] ReadRows
- [x] Refactor request options for existing APIs (where applicable)
- [x] Tests
  - [x] Linted
  - [x] Unit
    - [x] Update outdated tests
    - [x] Add tests for new APIs
  - [x] System
- [x] Documentation
  - [x] Add docs for new APIs
  - [x] Update old docs to reflect param changes, etc.
# 

[Docs Demo](http://callmehiphop.github.io/gcloud-node/#/docs/bigtable/master/bigtable)

PR #1525, opened 2016-08-24
common: create grpc operation object

callmehiphop:
Moving the `GrpcOperation` section of the [Bigtable PR](https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1439) to it's own PR.

### DIFF
--- a/packages/common/test/grpc-operation.js
+++ b/packages/common/test/grpc-operation.js
@@ -286,7 +286,7 @@ describe('GrpcOperation', function() {
     });
 
     describe('operation pending', function() {
-      var apiResponse = { status: 'PENDING' };
+      var apiResponse = { done: false };
       var setTimeoutCached = global.setTimeout;
 
       beforeEach(function() {


.jscsrc
```diff
@@ -17,7 +17,7 @@
   "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties",
   "maximumLineLength": {
     "value": 80,
-    "allowUrlComments": true
+    "allowComments": true
```
callmehiphop: See https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1525/files#diff-531b9ca2a9a7327c46d2ff2ffd657de6R51

stephenplusplus: We're gonna have to figure something out... how about an inline jscs ignore?

packages/common/src/grpc-operation.js
```diff
@@ -0,0 +1,212 @@
...
+    return;
+  }
+
+  this.getMetadata(function(err, resp) {
+    if (err || resp.error) {
+      self.emit('error', err || GrpcService.decorateStatus_(resp.result));
+      return;
+    }
+
+    if (!resp.done && self.status !== GrpcOperation.STATUS.running) {
```
stephenplusplus: Looking over https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto, there isn't a `status` property:

```
message Operation {
  string name = 1;
  google.protobuf.Any metadata = 2;
  bool done = 3;
  oneof result {
    google.rpc.Status error = 4;
    google.protobuf.Any response = 5;
  }
}
```

stephenplusplus: Here is what I ended up with after I took out the `status` stuff: https://github.com/stephenplusplus/gcloud-node/blob/4660c1182f903416c19e338884f5a23a217a7bda/packages/common/src/grpc-operation.js

.jscsrc
callmehiphop: Sure, PTAL

packages/common/src/grpc-operation.js
callmehiphop: I just stole your refactor and removed some dead tests, PTAL

callmehiphop:
@stephenplusplus I think this should be good to go - PTAL!

packages/common/src/grpc-operation.js
```diff
@@ -0,0 +1,192 @@
...
+
+  this.listenForEvents_();
+}
+
+modelo.inherits(GrpcOperation, GrpcServiceObject, events.EventEmitter);
+
+/**
+ * Cancel the operation.
+ *
+ * @param {function=} callback - The callback function.
```
stephenplusplus: If this is optional, I think our code needs a `callback || common.util.noop`

stephenplusplus: (and a test)

packages/common/test/grpc-operation.js
```diff
@@ -0,0 +1,337 @@
...
+          assert.strictEqual(err, formattedError);
+          done();
+        });
+
+        grpcOperation.startPolling_();
+      });
+    });
+
+    describe('operation pending', function() {
+      var apiResponse = { status: 'PENDING' };
```
stephenplusplus: Since there is no `status`, we should just use `done: false` to indicate a running operation.

packages/common/src/grpc-operation.js
callmehiphop: Done!

packages/common/test/grpc-operation.js
callmehiphop: Done!

PR #1529, opened 2016-08-25
common: support checking custom types

### LINKED ISSUE #1467: Decoupling packages
RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1440/files#r73738579

There are a number of instances within our code where we (sort of) couple packages together. For example, in the NL package we offer optional support for annotating GCS objects. I think it would be great if we could remove the actual storage module dependency while still supporting this feature.

Possible solutions could include:

* Using [`optionalDependencies`](https://docs.npmjs.com/files/package.json#optionaldependencies)
* Check for inheritance of `ServiceObject`
* Check the constructor name

stephenplusplus:
Fixes #1467 

In use:

``` js
if (common.util.isCustomType(image, 'storage/file')) {
  // ...
}
```

I used the constructor `@alias` strings as the canonical name for a class.

### DIFF
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -619,7 +619,7 @@ function isCustomType(unknown, module) {
     if (getConstructorName(walkingModule) === parentModuleName) {
       return true;
     }
-  } while (walkingModule.parent && (walkingModule = walkingModule.parent));
+  } while ((walkingModule = walkingModule.parent));
 
   return false;
 }


packages/common/test/util.js
```diff
@@ -1370,4 +1370,30 @@ describe('common/util', function() {
...
+      this.parent = new PubSub();
+    }
+
+    function Subscription() {
+      this.parent = new MiddleLayer();
+    }
+
+    var subscription = new Subscription();
+
+    it('should match a custom type by constructor names', function() {
```
callmehiphop: Not sure if it will ever come up, but what do you think about adding a test to check the type of a `new PubSub()` - testing against Service and not a ServiceObject.

stephenplusplus: Oh jeez, good catch.

stephenplusplus: Done! (Also, it wasn't working!)

callmehiphop:
LGTM!

packages/common/src/util.js
```diff
@@ -599,3 +599,29 @@ function createLimiter(makeRequestFn, options) {
...
+
+  var parentModuleName = moduleNameParts[0] && moduleNameParts[0].toLowerCase();
+  var subModuleName = moduleNameParts[1] && moduleNameParts[1].toLowerCase();
+
+  if (subModuleName && getConstructorName(unknown) !== subModuleName) {
+    return false;
+  }
+
+  var walkingModule = unknown;
+  do {
```
callmehiphop: WHAT IS THIS SORCERY

stephenplusplus: Hahaha

PR #1531, opened 2016-08-25
bigquery/language/logging/prediction/vision: decouple packages

stephenplusplus:
After #1529 ships, this does the de-coupling of packages using the new method, `common.util.isCustomType()`

### DIFF
--- a/packages/logging/package.json
+++ b/packages/logging/package.json
@@ -50,6 +50,7 @@
     "logging"
   ],
   "dependencies": {
+    "@google-cloud/common": "^0.3.0",
     "arrify": "^1.0.0",
     "extend": "^3.0.0",
     "google-proto-files": "^0.2.1",
@@ -57,11 +58,10 @@
     "string-format-obj": "^1.0.0"
   },
   "devDependencies": {
-    "async": "^1.4.2",
     "@google-cloud/bigquery": "*",
-    "@google-cloud/common": "*",
     "@google-cloud/pubsub": "*",
     "@google-cloud/storage": "*",
+    "async": "^1.4.2",
     "methmeth": "^1.0.0",
     "mocha": "^3.0.1",
     "node-uuid": "^1.4.3",
--- a/packages/bigquery/package.json
+++ b/packages/bigquery/package.json
@@ -50,18 +50,18 @@
     "bigquery"
   ],
   "dependencies": {
+    "@google-cloud/common": "^0.3.0",
     "arrify": "^1.0.0",
     "duplexify": "^3.2.0",
     "extend": "^3.0.0",
-    "@google-cloud/common": "^0.3.0",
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
     "docs": "node ./scripts/docs/packages.js",
     "bundle": "node ./scripts/docs/bundle.js",
     "lint": "jshint scripts/ packages/ system-test/ test/ && jscs packages/ system-test/ test/",
-    "test": "npm run docs && npm run bundle && mocha test/docs.js packages/*/test/*.js",
+    "test": "npm run docs && npm run bundle && mocha --timeout 5000 test/docs.js packages/*/test/*.js",
```
callmehiphop: What happened to `--no-timeouts`?

stephenplusplus: We have that for system tests, but not unit.

packages/logging/package.json
```diff
@@ -52,16 +52,16 @@
...
-    "@google-cloud/common": "^0.1.0",
-    "@google-cloud/pubsub": "^0.1.0",
-    "@google-cloud/storage": "^0.1.0",
     "google-proto-files": "^0.2.1",
     "is": "^3.0.1",
     "string-format-obj": "^1.0.0"
   },
   "devDependencies": {
     "async": "^1.4.2",
+    "@google-cloud/bigquery": "*",
```
callmehiphop: Why do we need these on dev dependencies?

callmehiphop: system tests?

stephenplusplus: Yep.

packages/vision/package.json
```diff
@@ -53,8 +53,7 @@
     "arrify": "^1.0.0",
     "async": "^1.4.2",
     "extend": "^3.0.0",
-    "@google-cloud/common": "^0.1.0",
-    "@google-cloud/storage": "^0.1.0",
```
callmehiphop: I'm assuming there is a system test for this? Don't we still need it as a dev dependency?

packages/logging/package.json
```diff
@@ -52,16 +52,16 @@
...
+    "@google-cloud/common": "*",
```
callmehiphop: Also I believe that @google-cloud/common needs to stay a regular `dependency`.

stephenplusplus: Woops.

packages/vision/package.json
stephenplusplus: There wasn't a system test that uses a file actually. Should I write one?

callmehiphop: Assuming there is code that uses storage, I think that would be a great idea!

stephenplusplus: Done!

stephenplusplus:
PTAL.

Issue #1542, opened 2016-08-30
Emit header for gRPC messages


pcostell:
gRPC connections currently don't have enough information in the request for us to do efficient traffic management/routing. 

Will you add a header that allows for us to optimize for this on the server? The header format is:

```
name: google-cloud-resource-prefix
value: projects/my-project-id
```

In general, the prefix would be the same as the JSON-over-http URL up to any custom method. For example, a Cloud Datastore commit message over JSON would hit the url `projects/my-project-id:commit`.

Here is an example usage in the cloud-bigtable-client: https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/915/files#diff-1078787e69bc1d7c5bb8507f0a87a143R83

dhermes:
@pcostell This should be true in Python as well?

```python
headers = [
    ('authorization', 'Bearer ' + access_token),
    ('user-agent', self._user_agent),
    ('google-cloud-resource-prefix', 'projects/my-project-id'),
]
```

[1]: https://github.com/GoogleCloudPlatform/gcloud-python/blob/6ce152cc097a113874a7713cceb99ce2eda13818/gcloud/_helpers.py#L572-L575

pcostell:
Yes -- note that we don't need it for proto/JSON-over-HTTP, only for gRPC calls.

PR #1546, opened 2016-08-30
allow setting gRPC metadata

stephenplusplus:
RE: #1542 
#### To Dos
- [x] Unit tests

@murgatroid99 -- could you take a look and let me know if this is the correct way to set metadata? Thanks!

### DIFF
--- a/packages/common/test/grpc-service.js
+++ b/packages/common/test/grpc-service.js
@@ -243,7 +243,7 @@ describe('GrpcService', function() {
       assert.strictEqual(calledWith[1], OPTIONS);
     });
 
-    it('should default grpcMetadata to null', function() {
+    it('should default grpcMetadata to empty metadata', function() {
       var fakeGrpcMetadata = {};
 
       GrpcMetadataOverride = function() {


packages/common/src/grpc-service.js
```diff
@@ -635,18 +642,27 @@ GrpcService.structToObj_ = function(struct) {
...
       return;
     }
 
-    var credentials = grpc.credentials.combineChannelCredentials(
+    var callCredentialObjects = [
       grpc.credentials.createSsl(),
       grpc.credentials.createFromGoogleCredential(authClient)
+    ];
+
+    callCredentialObjects = callCredentialObjects.concat(self.callCredentials);
```
stephenplusplus: This is where the metadata CallCredential object [created here](https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1546/files#diff-78d3ba232fe7028b5068f61d6d20a988R737) is combined with the `createSsl` and `createFromGoogleCredential` objects created above.

packages/common/src/grpc-service.js
```diff
@@ -717,5 +733,19 @@ GrpcService.prototype.getService_ = function(protoOpts) {
   return service;
 };
 
+GrpcService.prototype.setGrpcMetadata_ = function(metadata) {
+  var cc = grpc.credentials.createFromMetadataGenerator(function(_, cb) {
+    var grpcMetadata = new grpc.Metadata();
+
+    for (var prop in metadata) {
+      grpcMetadata.add(prop, metadata[prop]);
+    }
```
murgatroid99: I would recommend creating and filling the metadata object outside of the metadata generator function. Some validation is done there, and it would be better to get the validation errors immediately, rather than waiting until a method is called.

stephenplusplus: Thanks for the quick review! Could you show what that looks like? I tried to model this after an example snippet you gave here: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1346#issuecomment-240912476

murgatroid99:
This is an unexpected use of call credentials, but it should work.

Another option would be to add the metadata directly to the per call metadata when making a call, by doing something like this wherever you make grpc calls:

```js
for (prop in metadata) {
  perCallMetadata.add(prop, metadata[prop]);
}
service.callMethod(argument, perCallMetadata, callback);
```

packages/common/src/grpc-service.js
```diff
@@ -227,8 +229,17 @@ GrpcService.prototype.request = function(protoOpts, reqOpts, callback) {
   delete reqOpts.autoPaginateVal;
 
   var service = this.getService_(protoOpts);
-  var grpcOpts = {};
 
+  var metadata = null;
```
murgatroid99: Yes, this is what I meant. Either this or the previous iteration should work fine, but what you have now is closer to how I originally expected the API to be used.

stephenplusplus: Cool, thanks for the guidance!

packages/common/src/grpc-service.js
```diff
@@ -158,6 +159,18 @@ function GrpcService(config, options) {
     this.grpcCredentials = grpc.credentials.createInsecure();
   }
 
+  this.grpcMetadata = null;
```
callmehiphop: Have you noticed any system test failures in the event that grpcMetadata is `null`? I pulled your branch down and linked common to all other packages and was met with an immediate Bigtable failure.

Doing this fixed the issue for me locally.

```js
this.grpcMetadata = new grpc.Metadata();

if (config.grpcMetadata) {
  for (var prop in config.grpcMetadata) {
    if (config.grpcMetadata.hasOwnProperty(prop)) {
      this.grpcMetadata.add(prop, config.grpcMetadata[prop]);
    }
  }
}
```

stephenplusplus: Oh sheesh, good catch. I'll make the fix shortly.

stephenplusplus: Done!

packages/common/test/grpc-service.js
```diff
@@ -241,11 +241,17 @@ describe('GrpcService', function() {
     });
 
     it('should default grpcMetadata to null', function() {
```
callmehiphop: *should default grpcMetadata to empty metadata**

stephenplusplus: Fixed!

PR #1548, opened 2016-08-31
Add Gapic-generated files to logging API.

jmuk:
The same structure as #1476.

### DIFF
--- a/packages/logging/src/v2/index.js
+++ b/packages/logging/src/v2/index.js
@@ -20,7 +20,6 @@ var loggingServiceV2Api = require('./logging_service_v2_api');
 var metricsServiceV2Api = require('./metrics_service_v2_api');
 var extend = require('extend');
 var gax = require('google-gax');
-var lodash = require('lodash');
 
 function v2(options) {
   options = extend({
@@ -35,8 +34,5 @@ function v2(options) {
 }
 
 v2.SERVICE_ADDRESS = loggingServiceV2Api.SERVICE_ADDRESS;
-v2.ALL_SCOPES = lodash.union(
-    configServiceV2Api.ALL_SCOPES,
-    loggingServiceV2Api.ALL_SCOPES,
-    metricsServiceV2Api.ALL_SCOPES);
+v2.ALL_SCOPES = loggingServiceV2Api.ALL_SCOPES;
 module.exports = v2;
--- a/packages/logging/package.json
+++ b/packages/logging/package.json
@@ -57,7 +57,6 @@
     "google-gax": "^0.6.0",
     "google-proto-files": "^0.2.1",
     "is": "^3.0.1",
-    "lodash": "^4.15.0",
     "string-format-obj": "^1.0.0"
   },
   "devDependencies": {


stephenplusplus:
Cool, thanks! Some linting issues:

```sh
$ npm run lint
> google-cloud-node@ lint /home/travis/build/GoogleCloudPlatform/google-cloud-node

packages/logging/src/v2/config_service_v2_api.js: line 448, col 2, Unnecessary semicolon.
packages/logging/src/v2/logging_service_v2_api.js: line 451, col 2, Unnecessary semicolon.
packages/logging/src/v2/metrics_service_v2_api.js: line 447, col 2, Unnecessary semicolon.
```

packages/logging/src/v2/index.js
```diff
@@ -0,0 +1,42 @@
...
+  var gaxGrpc = gax.grpc(options);
+  var result = {};
+  extend(result, configServiceV2Api(gaxGrpc));
+  extend(result, loggingServiceV2Api(gaxGrpc));
+  extend(result, metricsServiceV2Api(gaxGrpc));
+  return result;
+}
+
+v2.SERVICE_ADDRESS = loggingServiceV2Api.SERVICE_ADDRESS;
+v2.ALL_SCOPES = lodash.union(
```
stephenplusplus: We can save ourselves the `lodash` dependency if we just do:

```js
v2.ALL_SCOPES = [].concat(
  configServiceV2Api.ALL_SCOPES,
  loggingServiceV2Api.ALL_SCOPES,
  metricsServiceV2Api.ALL_SCOPES
);
```

jmuk: Many parts of those `ALL_SCOPES` can overlap each other and `[].concat` will leave duplication. Is it okay?

The logging API has the exactly same set of scopes among those services, so simply using `loggingServiceV2Api.ALL_SCOPES` might be okay.

stephenplusplus: Sorry, I missed that. In cases like that, using `union` makes sense, but we can trim down the dependency tree by just installing `lodash.union`: http://npmjs.com/lodash.union.

If all of these `ALL_SCOPES` values are the same, we might as well skip it in this case if possible.

jmuk: Modified as `v2.ALL_SCOPES = loggingServiceV2Api.ALL_SCOPES;`

packages/logging/src/v2/logging_service_v2_api.js
```diff
@@ -0,0 +1,454 @@
...
+  if ('filter' in args.otherArgs) {
+    req.filter = args.otherArgs.filter;
+  }
+  if ('orderBy' in args.otherArgs) {
+    req.order_by = args.otherArgs.orderBy;
+  }
+  if ('pageSize' in args.otherArgs) {
+    req.page_size = args.otherArgs.pageSize;
+  }
+  return this._listLogEntries(req, args.options, args.callback);
```
stephenplusplus: @jmuk The handwritten library's version of this currently looks like this:

```js
logging.getEntries(function(err, entries) {});
```

Or in stream mode:

```js
logging.getEntries()
  .on('error', function(err) {})
  .on('data', function(entry) {})
  .on('end', function() {});
```

Behind the scenes, if the API response from "listLogEntries()" returns a `nextPageToken`, we automatically run the query again, without the user even realizing that extra call was made. If the user wants to paginate themselves, they do this:

```js
logging.getEntries({ autoPaginate: false }, function(err, firstResultSet, nextQuery) {
  if (nextQuery) {
    logging.getEntries(nextQuery, function(err, secondResultSet, nextQuery) {
      ...
    });
  }
});
```

Does this method (`LoggingServiceV2Api.prototype.listLogEntries()`) do automatic-pagination as well?

If yes, this method auto-paginates: how should the handwritten methods that allow manual cursoring continue to work?

If no, this doesn't auto-paginate: how can I get a hold of the `nextPageToken` value to continue cursoring for the user?

Thanks!

jmuk: Yes, it returns a stream and it automatically does the automatic pagination (it's called page-streaming, in our side). It automatically fetches the next response by using `nextPageToken` and emits an instance of log entry instead of listLogEntriesResponse upon 'data' event.

In order to iterate over pages, it actually require to specify a value in the call options.
```javascript
var v2 = require('@google-cloud/logging').v2();
var api = v2.loggingServiceV2Api();

api.listLogEntries([projectId], new gax.CallOptions({page_token: gax.FIRST_PAGE})).on('data', function(resp) {
   // resp is a listLogEntriesResponse.
}).on('end', function() { ...});
```
`page_token` here will be supplied from the last result, or use `gax.FIRST_PAGE` object when starting from a page.

This design borrows from what Python is doing, so might not look like the way JS programmers will do. In that case please tell me how this can be improved.

jmuk: Filed https://github.com/googleapis/gax-nodejs/issues/35, please add comments there.

PR #1552, opened 2016-08-31
datastore: deep clone user objects

### LINKED ISSUE #1551: `datastore.save` transforming JS Array into arrayValue
When saving data to the datastore, it seems that when a Javascript Array is contained inside an object that is a class instance (as opposed of a pure JS object),  the JS Array is being transformed into an 'arrayValue'.

Interestingly, the same does not happen with a pure JS Object. The code below demonstrates this happening and also shows how arrays inside JS Objects remain unchanged.

Sample JS:
```js
'use strict';
const gcloud = require('gcloud');
const config = require('../config/config');

const ds = gcloud.datastore({
  projectId: config.get('GCLOUD_PROJECT')
});

class MyClass {

}

const myInstance = new MyClass();
myInstance.myArray = ['A', 'B', 'C'];

const myObject = {
  myArray: ['A', 'B', 'C', 'D']
}

const data =
  [
    {
      name: 'key1',
      value: 'value1'
    },
    {
      name: 'key2',
      value: ['A', 'B', 'C', 'D']
    },
    {
      name: 'myObject',
      value: myObject
    },
    {
      name: 'myInstance',
      value: myInstance
    }
  ];

const key = ds.key('test');

const entity = {
  key: key,
  data: data
}

console.log('myInstance before save: ', myInstance);
console.log('myObject before save: ', myObject);
ds.save(entity, err => {
  if (err) {
    console.log(err);
  }
  console.log('myInstance after save: ', myInstance, '// arrayValue instead of a regular JS Array');
  console.log('myObject after save: ', myObject, '// Arrays in Objects keep unchanged!');  
});
```

Output:
```
myInstance before save:  MyClass { myArray: [ 'A', 'B', 'C' ] }
myObject before save:  { myArray: [ 'A', 'B', 'C', 'D' ] }
myInstance after save:  MyClass { myArray: { arrayValue: { values: [Object] } } } // arrayValue instead of a regular JS Array
myObject after save:  { myArray: [ 'A', 'B', 'C', 'D' ] } // Arrays in Objects keep unchanged!
```

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS: OS X El Capitain (10.11.6 )
  - Node.js version: v4.2.6
  - npm version: 2.14.12
  - google-cloud-node version: ~0.36.0

#### Steps to reproduce

  1. run the script above

stephenplusplus:
Fixes #1551

### DIFF
--- a/packages/datastore/test/entity.js
+++ b/packages/datastore/test/entity.js
@@ -442,9 +442,10 @@ describe('entity', function() {
 
       var originalValue = extend(true, {}, value);
 
-      entity.encodeValue(value);
+      var encodedValue = entity.encodeValue(value);
 
       assert.deepEqual(value, originalValue);
+      assert.notStrictEqual(value, encodedValue);
     });
 
     it('should encode an empty object', function() {


packages/datastore/test/entity.js
```diff
@@ -430,6 +431,22 @@ describe('entity', function() {
...
+            obj: true
+          }
+        }
+      };
+
+      var originalValue = extend(true, {}, value);
+
+      entity.encodeValue(value);
+
+      assert.deepEqual(value, originalValue);
```
callmehiphop: Can we also do a `notStrictEqual` here?

stephenplusplus: Done-o.

Issue #1493, opened 2016-08-15
OS testing


callmehiphop:
One area that we need to improve upon before we go beta is automating tests on Windows and OSX.

It appears [Travis offers osx support](https://docs.travis-ci.com/user/multi-os/), but it is considered beta still.

- [x] Windows: Appveyor
- [x] OS X: Travis

stephenplusplus:
Link dropping: https://www.appveyor.com/

stephenplusplus:
@omaray @jgeewax We will need Appveyor enabled for our repo: https://ci.appveyor.com/project/googlecloudplatform/google-cloud-node -- then you'll need to give access to the "team" here: https://ci.appveyor.com/team

dhermes:
Ditto on enabling https://ci.appveyor.com/project/googlecloudplatform/gcloud-python

omaray:
No this is great!

@jonparrott and @elibixby would it be possible to give access to Node and Python.

Thanks!

mziccard:
Can this also be enabled for `gcloud-java`?  https://ci.appveyor.com/project/googlecloudplatform/gcloud-java

elibixby:
the @GoogleCloudPlatform/gcloud team has admin access on all these repos, and it doesn't look like AppVeyor requires Org level access. Are you unable to enable this yourself?

dhermes:
@elibixby When I go to https://ci.appveyor.com/projects/new and click "Add" for the "google-cloud-python" project under the "GoogleCloudPlatform" it produces it under `/dhermes`, e.g.
https://ci.appveyor.com/project/dhermes/google-cloud-python-bqg0j

elibixby:
It does the same for me. I don't see any way to produce a project under the org banner as those links request. Does anybody know if that's possible and if so how?

dhermes:
http://help.appveyor.com/discussions/questions/1154-appveyor-account-for-github-organizations seems to have an answer (I am midway through reading)

dhermes:
NVM this seems more up to date: https://www.appveyor.com/docs/team-setup/#setting-up-appveyor-account-for-github-organization

elibixby:
Alright, I have created GoogleCloudPlatform as an AppVeyor account. And made @GoogleCloudPlatform/gcloud an admin. Try logging in with github again after signing out and hopefully you'll see a drop down (if I understand everything correctly).

dhermes:
[...]

PR #1553, opened 2016-08-31
tests: add OS X to travis

stephenplusplus:
RE: #1493 

Let's see how this goes!


stephenplusplus:
Things are looking up for this PR. It looks like Travis is running both the "PR" tests and "Push" tests. That's not normal, is it @callmehiphop?

callmehiphop:
https://github.com/travis-ci/travis-ci/issues/3241#issuecomment-77120970 ..?

stephenplusplus:
Between "Build pushes" and "Build pull requests" Travis options, I chose to only Build pushes. If a "push" build runs from a PR, then we apparently don't need to run the "pull request" build. Working theory...

PR #1556, opened 2016-09-01
bigtable: return Buffer instead of base64 when decode is true

### LINKED ISSUE #1526: Binary data in BigTable column
Hi, I am trying to store an encode protobuf in BigTable and I'm not able to insert binary data (represented as a Buffer) into a BigTable column. I have tried the following ways: 

```js
this.table.insert([
  key: `foo`,
  data: {
    event: new Buffer(),
  },
]);

this.table.insert([
  key: `foo`,
  data: {
    event: {
     data: new Buffer(),
    }
  },
]);
```

It seems to get stringified somewhere in the client library, which is making the data invalid.

callmehiphop:
Closes #1526

### DIFF
--- a/packages/bigtable/src/mutation.js
+++ b/packages/bigtable/src/mutation.js
@@ -60,14 +60,18 @@ var methods = Mutation.methods = {
  * Parses "bytes" returned from proto service.
  *
  * @param {string} bytes - Base64 encoded string.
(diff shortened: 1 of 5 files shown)


callmehiphop:
@arbesfeld is this closer to the behavior you are expecting?

PR #1559, opened 2016-09-02
docs: remove `nextQuery` docs & annotate arrays correctly

stephenplusplus:
Was looking through the docs, and noticed we were inconsistent with documenting the `callback.nextQuery` param. I think it makes sense to remove it and show how that flow works in the @&shy;example section. I also found various places where we weren't specifying a user would get an array of objects back, but only an object, i.e. `{module:compute/vm} callback.vms` instead of `{module:compute/vm[]} callback.vms`.

### DIFF
--- a/packages/bigquery/src/dataset.js
+++ b/packages/bigquery/src/dataset.js
@@ -272,8 +272,6 @@ Dataset.prototype.delete = function(options, callback) {
  * @param {?error} callback.err - An error returned while making this request
  * @param {module:bigquery/table[]} callback.tables - The list of tables from
  *     your Dataset.
- * @param {?object} callback.nextQuery - If present, query with this object to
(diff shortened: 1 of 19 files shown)


callmehiphop:
So why are we removing `nextQuery` everywhere? I can't help but think only documenting it in examples would be confusing to the user.

stephenplusplus:
Well, it started because we were inconsistent. It was in some APIs and not in others. Frankly, I removed it because the whole thing is pretty confusing. We support the `nextQuery` flow because there *might* be *some* users who find it important to manually paginate through results. But if we show documentation suggesting it's always there, I think that introduces confusion to the majority user who isn't thinking about pagination... they're just trying to get all the results in one method call, and no such `nextQuery` object will exist.

PR #1567, opened 2016-09-06
bigtable: support local emulator

### LINKED ISSUE #1506: BigTable - local instance
Is there a way to specify the host/port of the BigTable instance? We'd like to use https://github.com/spotify/docker-bigtable for local testing.

stephenplusplus:
Fixes #1506 
#### To Dos
- [x] Overview docs showing how to connect with the emulator

### DIFF
--- a/packages/bigtable/src/index.js
+++ b/packages/bigtable/src/index.js
@@ -53,6 +53,24 @@ var Instance = require('./instance.js');
  *
  * @example
  * //-
+ * // <h3>The Bigtable Emulator</h3>
+ * //
+ * // Make sure you have the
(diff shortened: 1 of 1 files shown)


callmehiphop:
@stephenplusplus I've added a commit, I think the only thing that was missing was making sure that all the service (admin/instance/etc.) were also updated. Obviously the instance API doesn't work with the mentioned emulator, but parts of the table admin api do.

@arbesfeld any chance you'd want to give this a test drive?

garye:
@callmehiphop What's the status of this? Is it something we can merge or does it need more testing?

callmehiphop:
@garye should be good to go!

PR #1568, opened 2016-09-06
all modules: ensure all User-Agents are set

stephenplusplus:
#### To Dos
- [x] Merge #1587 
- [x] Release @google-cloud/common@0.5.0

Each module will now pass its own `package.json` to ensure the correct UA string is set.

### DIFF
--- a/packages/pubsub/src/iam.js
+++ b/packages/pubsub/src/iam.js
@@ -73,7 +73,8 @@ function IAM(pubsub, id) {
     scopes: [
       'https://www.googleapis.com/auth/pubsub',
       'https://www.googleapis.com/auth/cloud-platform'
-    ]
+    ],
+    packageJson: require('../package.json')
   };
 
   this.id = id;
--- a/packages/pubsub/test/iam.js
+++ b/packages/pubsub/test/iam.js
@@ -65,6 +65,7 @@ describe('IAM', function() {
         'https://www.googleapis.com/auth/pubsub',
         'https://www.googleapis.com/auth/cloud-platform'
       ]);
+      assert.deepEqual(config.packageJson, require('../package.json'));
 
       assert.strictEqual(options, PUBSUB.options);
     });


packages/common/src/grpc-service.js
```diff
@@ -161,6 +161,11 @@ function GrpcService(config, options) {
 
   this.grpcMetadata = new grpc.Metadata();
 
+  this.grpcMetadata.add('User-Agent', [
```
stephenplusplus: @murgatroid99 -- is this the way to set a UA with gRPC?

packages/common/src/grpc-service.js
```diff
@@ -161,6 +161,11 @@ function GrpcService(config, options) {
...
+    config.packageJson.name.replace('@google-cloud', 'gcloud-node'),
```
callmehiphop: This would make the header appear as so `gcloud-node/bigtable/v0.3.0`, correct? Do we want the service name in there?

stephenplusplus: Not sure, that's still being discussed in https://github.com/GoogleCloudPlatform/gcloud-common/issues/179. But without the service, the version would easily conflict with other service versions... although, I suppose if the Bigtable API is getting a request, it could assume that the version number refers to the Bigtable module.

callmehiphop: Doesn't it pull in the local <kbd>package.json</kbd> already? Meaning even if we're using the umbrella module, it's sending the individual package version with the request..?

stephenplusplus: Yes, that is correct.

stephenplusplus: In case I wasn't being clear, I originally thought the service name was important to include with the version. Mapping `gcloud-node/version` to `google-cloud/bigtable@version` might not be obvious, since gcloud-node/google-cloud-node might share that same version.

To use the UA string correctly, I think we're supposed to describe a "product" with its version. The product is the module, so describing it as `gcloud-node` isn't accurate. But this stuff is being talked about in https://github.com/GoogleCloudPlatform/gcloud-common/issues/179, so I'm just waiting to see which way they want to go.

packages/common/src/grpc-service.js
murgatroid99: We expect the user agent to be set with the channel argument "grpc.primary_user_agent". So, you would construct a client like this:

```js
client = new Client(address, credentials, {'grpc.primary_user_agent': user_agent_string});
```

Note that this will add your user agent string to the existing grpc user agent string.

packages/common/test/grpc-service.js
```diff
@@ -43,6 +44,7 @@ function fakeRetryRequest() {
   return (retryRequestOverride || retryRequest).apply(null, arguments);
 }
 
+
```
callmehiphop: extra newline!

stephenplusplus: Not anymore!

callmehiphop:
@stephenplusplus Not trying to be a pain in the butt, but could we separate this into 2 different PRs? 

1. All the `@google-cloud/common` updates.
2. All the modules upgrading to the latest common + `packageJson` fields

The way all these changes are lumped together makes testing this locally  .. challenging.

callmehiphop:
Looks like we're missing `packageJson` within `PubSub/IAM`

stephenplusplus:
Added.

PR #1574, opened 2016-09-06
translate: support format option to specify html or text

### LINKED ISSUE #1573: Translate api should support 'format' request parameter
I am using the Translate module the Google cloud package.
Google translate API allows the user to specify the format for translation as a query parameter (text, or html, defaults to html) https://cloud.google.com/translate/v2/translate-reference

Defaulting to HTML format translations means the translation results contain HTML entities, which is not desired in most cases when using the API, because results are not displayed directly on a web page. This package should allow sending the format as a parameter, probably as part of the options parameter to the translate function.

#### Environment details

  - OS: Any
  - Node.js version: 6.5.1
  - npm version: 3.8.9
  - google-cloud-node version: 0.39.0

#### Steps to reproduce
 

    
    var googleCloud = require('google-cloud');
    var gtranslate = googleCloud.translate;

    var translator = gtranslate({
      key: API_KEY,
    });

    translator.translate('How are you today?', {source: 'en', target: 'fr', format: 'text'}, (err, result) => {
        console.log(result);
    });

Expected results: Comment allez-vous aujourd'hui?
Actual results: Comment allez-vous aujourd\&#39;hui?

Thanks!

WarTech9:
Fixes #1573

Adds the 'format' parameter to the translate request if specified by caller.

### DIFF
--- a/packages/translate/src/index.js
+++ b/packages/translate/src/index.js
@@ -272,6 +272,10 @@ Translate.prototype.getLanguages = function(target, callback) {
  *     is written in.
  * @param {string} options.to - The ISO 639-1 language code to translate the
  *     input to.
+ * @param {string} options.format - This optional parameter allows you to indicate the text
+ *     to be translated is either plain-text or HTML.
+ *     A value of 'html' indicates HTML and a value of 'text' indicates plain-text.
+ *     Defaults to 'html' if not specified.
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request.
  * @param {object|object[]} callback.translations - If a single string input was
@@ -295,7 +299,8 @@ Translate.prototype.getLanguages = function(target, callback) {
  * //-
  * var options = {
  *   from: 'en',
- *   to: 'es'
+ *   to: 'es',
+ *   format: 'text'
  * };
  *
  * translate.translate('Hello', options, function(err, translation) {


stephenplusplus:
Do you think something like [is-html](https://github.com/sindresorhus/is-html) would be reliable?

innovative1:
You mean using is-html to test the input? That could work, but what if the input is not html but the caller wants html output, or the other way around? Having an option parameter override covers that case.

Btw, I signed the CLA.

stephenplusplus:
> but what if the input is not html but the caller wants html output

Are there reasons why someone would want to do that? I assume the API wouldn't do its job as well if the format didn't match the input.

innovative1:
>  I assume the API wouldn't do its job as well if the format didn't match the input.

That is the current behaviour: The API auto detects the source language but doesn't auto detect the format type, it defaults to html if the format parameter is missing, regardless of the actual input format.
https://cloud.google.com/translate/v2/translate-reference#format

So the caller can choose the output format. That is also how the Google Node.js client works as well, by passing a format parameter. Which is why I thought the a parameter should be used here too, as I was coming from that client. 
https://github.com/google/google-api-nodejs-client/blob/master/apis/translate/v2.js#L141

I suppose other devs coming from the node client would expect it to work the same way.

stephenplusplus:
I think this is a case where we can try to out-smart the API. If we can detect that the user's input is HTML or text, then they shouldn't need to override it. However, I'm not certain [`is-html`](https://github.com/sindresorhus/is-html) is going to always be correct for the various snippets of input the user will be giving us. So, I think you were right that we should do both. If the user doesn't provide `options.format`, can you set it for them?

```diff
var query = {
  q: arrify(input),
+ format: options.format || isHtml(query.q[0]) ? 'html' : 'text'
};
```

stephenplusplus:
Continued in #1647.

PR #1576, opened 2016-09-07
Introduce Gapic-generated files for cloud vision API.

### LINKED ISSUE #1463: Integration with Gapic-generated code

### LINKED ISSUE #1575: Introduce Gapic-generated files for cloud vision API.
Bring cloud vision API (gRPC and Gapic-generated code). This would be an initial effort for #1463.

The proto files are https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1/image_annotator.proto

swcloud:
Fixes #1463
Fixes #1575

auto-generated:
image_annotator_api.js
image_annotator_client_config.json

v1/index.js: manually written
index.js and package.json: manually modified

### DIFF
--- a/packages/vision/package.json
+++ b/packages/vision/package.json
@@ -56,6 +56,7 @@
     "async": "^1.4.2",
     "extend": "^3.0.0",
     "google-gax": "^0.6.0",
+    "google-proto-files": "^0.7.0",
     "is": "^3.0.1",
     "propprop": "^0.3.0",
     "request": "^2.70.0",


packages/vision/src/v1/image_annotator_api.js
```diff
@@ -0,0 +1,188 @@
...
+  return this._batchAnnotateImages(req, args.options, args.callback);
+};
+
+function ImageAnnotatorApiBuilder(gaxGrpc) {
+  if (!(this instanceof ImageAnnotatorApiBuilder)) {
+    return new ImageAnnotatorApiBuilder(gaxGrpc);
+  }
+
+  var grpcClient = gaxGrpc.load([{
+    root: require('google-proto-files')('..'),
```
stephenplusplus: We need to add this as an npm dependency for this module since we currently use the JSON API for the handwritten layer.

swcloud: "google-proto-files": "^0.7.0" is part of common/package.json. I think vision/package.json has a dependency on 'common', right?

stephenplusplus: Yes, but a module should define its own dependencies instead of borrowing from another. That can lead to unexpected disappearances of modules if the middle layer changes its dependencies.

swcloud:
@stephenplusplus the CI build fails on nodejs 0.12:

/home/travis/build/GoogleCloudPlatform/google-cloud-node/packages/google-cloud/node_modules/@google-cloud/bigtable/node_modules/dot-prop/index.js:
const isObj = require('is-obj');
^^^^^
SyntaxError: Use of const in strict mode.

This does not seem to have anything to do with this PR. Sounds like another build flakiness issue. Can you rebuild? Is there anyway we can improve this since each time the build takes long time and we don't want to waste too much time on this.

stephenplusplus:
> Is there anyway we can improve this since each time the build takes long time and we don't want to waste too much time on this.

We simply have 3 OSes to run tests for 3 different versions of Node. Running the system tests takes a big portion of the time, the other is the installation of all of our dependencies. I have improved the performance somewhat by forcing Travis to use npm 3, which flattens the dependency tree *before* downloading. I think there might be room to improve if we only run the system tests for the module that the modified files affected, e.g., this PR wouldn't run Bigtable tests. I'll open a new issue to talk that through.

Regarding the error itself, it was from a mistake I made yesterday, giving @google-cloud/bigtable a dependency on a module that only worked on Node > v4. I released a patch earlier today and just restarted the build. If that goes well, I'll merge this right after.

Thanks!

PR #1589, opened 2016-09-12
common: use gzip content encoding when making http requests

KidkArolis:
This should speed up the transfer of larger bigquery responses.

Should this be done only in bigquery instead of globally?

### DIFF
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -29,6 +29,7 @@ var googleAuth = require('google-auto-auth');
 var is = require('is');
 var request = require('request').defaults({
   timeout: 60000,
+  gzip: true,
   pool: {
     maxSockets: Infinity
   }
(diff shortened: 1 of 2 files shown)


stephenplusplus:
Thanks for this @KidkArolis! I'm going to test this locally soon against our system tests. I think putting it global is the right move 👍 Would you running `npm test` locally and fixing the broken test? It should be just a small fix.

KidkArolis:
Updated. Detailed tests! I couldn't get them all running locally, but the `packages/common` tests pass now.

dhermes:
@stephenplusplus Non-sequitur: Not sure if you guys prefer the @coveralls update on every PR but the check also does the same thing. You can disable on coveralls.io:

screen_shot_003

stephenplusplus:
The comments do get kind of annoying, but I like that it's loud when there's a drop so that the contributor will notice more easily. It would be nice to disable the comment unless it drops, but I don't think that's possible. Not sure what's up with the "Changes Unknown" though :\

stephenplusplus:
Thanks for this! #perfmatters

PR #1592, opened 2016-09-14
add system test running capabilities

stephenplusplus:
We now have a secure variable `GCLOUD_TESTS_CREDENTIALS` in AppVeyor. It is not exposed to PR builds.


.appveyor.yml
```diff
@@ -1,5 +1,6 @@
 environment:
   matrix:
+  fast_finish: true
```
stephenplusplus: If one version fails, the build is canceled.

.appveyor.yml
```diff
@@ -13,5 +14,10 @@ test_script:
   - node --version
   - npm --version
   - npm test
+  - ps: if (!$env:APPVEYOR_PULL_REQUEST_NUMBER) { npm run system-test }
```
stephenplusplus: Only run system tests on merges to master.

PR #1596, opened 2016-09-15
tests: ci: allow parallel npm installs

stephenplusplus:


stephenplusplus:
@callmehiphop this has sped up our builds quite a bit!

image

image

Previously, each environment would require around ~15-17 minutes just to run `npm install`s, *before* running any tests. Now we've got it down to ~8-17 for the tests to complete.

.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: node_js
 node_js:
-  - 6
-  - 4
   - 0.12
+  - 4
+  - 6
```
stephenplusplus: Re-ordered these. It seems it's easiest to break 0.12, so I put it up front for the quickest response.

.appveyor.yml
```diff
@@ -1,13 +1,31 @@
...
+  - packages\dns\node_modules -> packages\dns\package.json
+  - packages\google-cloud\node_modules -> packages\google-cloud\package.json
+  - packages\language\node_modules -> packages\language\package.json
+  - packages\logging\node_modules -> packages\logging\package.json
+  - packages\prediction\node_modules -> packages\prediction\package.json
+  - packages\pubsub\node_modules -> packages\pubsub\package.json
+  - packages\resource\node_modules -> packages\resource\package.json
+  - packages\storage\node_modules -> packages\storage\package.json
+  - packages\translate\node_modules -> packages\translate\package.json
+  - packages\vision\node_modules -> packages\vision\package.json
```
callmehiphop: Caching these doesn't cause a version mismatch anywhere?

stephenplusplus: What do you mean?

stephenplusplus: The pattern here is: "cache these files -> until this file changes"

stephenplusplus:
Even though it took 100 commits, this should be an improvement over what we have now. It might take more tweaking, but that's just another 100-commit PR away.

.appveyor.yml
callmehiphop: Sorry, I meant a version mismatch between node versions. Like if the first install uses Node 0.12, would we be re-installing the same dependencies for 6?

This is purely anecdotal, but by default I run 0.12 on my local machine. Sometimes I'll use nvm to switch to 6 for pet projects and if I forget to do so I get a version mismatch error.

stephenplusplus: Oh, gotcha. With Travis, it separates caches by the Node version and/or (?) OS:

> There is one cache per branch and language version/ compiler version/ JDK version/ Gemfile location/ etc.
> \- https://docs.travis-ci.com/user/caching/#Fetching-and-storing-caches

With AppVeyor, it's not as clear from their docs. The AppVeyor builds with caching from this PR have been succeeding, so I'd say we're safe.

PR #1597, opened 2016-09-15
core: ci scripts overhaul

### LINKED ISSUE #1579: Only run tests for modules affected by a PR
*Cont'd from https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1576#issuecomment-245700733*

We now have 3 OSes to run tests for 3 different versions of Node. Running the system tests takes a big portion of the time, the other is the installation of all of our dependencies. I have improved the performance somewhat by forcing Travis to use npm 3, which flattens the dependency tree before downloading.

I think there might be room to improve if we only run the system tests for the module that the modified files affected, e.g., a merge that only changed Storage files wouldn't run Bigtable tests.

@callmehiphop any ideas on tooling that might exist for this? I know we've indirectly used things like this for Grunt/Gulp pipelines.

callmehiphop:
Closes #1579 
Don't merge until after #1606

This PR aims to do 2 major things
1. Only run tests for modules that have actually been updated.
2. Converts scripts from Bash to JS

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -43,8 +43,8 @@ require('shelljs/global');
 
 var helpers = require('./helpers');
 
-var Module = helpers.Module;
 var ci = helpers.ci;
+var Module = helpers.Module;
 
 // Get a list of the modules that have code changes.
 var modules = Module.getUpdated();
@@ -60,9 +60,12 @@ if (ci.isFirstPass()) {
   Module.runCoveralls();
 }
 
+Module.buildDocs();
+
 modules.forEach(function(mod) {
   mod.install();
   mod.runUnitTests();
+  mod.runSnippetTests();
 });
 
 // If this is a push to master, let's run system tests
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "array-uniq": "^1.0.3",
     "async": "^1.4.2",
     "coveralls": "^2.11.2",
+    "create-error-class": "^3.0.2",
     "dox": "0.8.1",
     "extend": "^3.0.0",
     "glob": "^5.0.9",
--- a/scripts/build-docs.js
+++ b/scripts/build-docs.js
@@ -63,7 +63,7 @@ var ghpages = git.submodule('gh-pages', 'ghpages');
 var GHPAGES_JSON = path.join(ghpages.cwd, 'json');
 
 // generate latest json docs
-run('npm run docs');
+Module.buildDocs();
 
 // copy docs to gh-pages
 cp('-R', 'docs/json/*', GHPAGES_JSON);
--- a/scripts/helpers.js
+++ b/scripts/helpers.js
@@ -93,6 +93,15 @@ Module.getUpdated = function() {
 };
 
 /**
+ * Builds docs for all modules
+ *
+ * @static
+ */
+Module.buildDocs = function() {
+  run('npm run docs', { cwd: ROOT_DIR });
+};
+
+/**
  * Returns a list containing ALL the modules.
  *
  * @static
@@ -162,6 +171,13 @@ Module.prototype.link = function(mod) {
  */
 Module.prototype.runUnitTests = function() {
   run('npm run test', { cwd: this.directory });
+};
+
+/**
+ * Runs snippet tests for this module.
+ */
+Module.prototype.runSnippetTests = function() {
+  run(['TEST_MODULE=' + this.name, 'mocha test/docs.js'], { cwd: ROOT_DIR });
 };
 
 /**
--- a/test/docs.js
+++ b/test/docs.js
@@ -16,204 +16,201 @@
 
 'use strict';
 
+var createErrorClass = require('create-error-class');
 var format = require('string-format-obj');
 var fs = require('fs');
-var gcloud = require('../packages/google-cloud');
 var glob = require('glob');
 var mitm = require('mitm');
-var packageOverviewConfigs = require('../scripts/docs/config.js').OVERVIEW;
+var overviews = require('../scripts/docs/config').OVERVIEW;
+var path = require('path');
 var prop = require('propprop');
 var vm = require('vm');
 
+var rootFile = 'index.json';
 var util = require('../packages/common').util;
 
-function runCodeInSandbox(code, sandbox) {
-  vm.createContext(sandbox);
-  try {
-    vm.runInNewContext(code, sandbox, {
-      filename: 'assert-code.vm',
-      timeout: 5000
-    });
-  } catch(err) {
-    // rethrow the error with code for context and resolving issues faster.
-    var lineCol = err.stack.match('assert-code\.vm:(.+)');
-    lineCol.line = lineCol[1];
+var DocsError = createErrorClass('DocsError', function(err, code) {
+  var lineCol = err.stack.match('assert-code\.vm:(.+)');
+  lineCol.line = lineCol[1];
 
-    var lines = code.split('\n')
-      .filter(function(line, index) {
-        if (index < lineCol.line) {
-          return line;
-        }
-      })
-      .join('\n');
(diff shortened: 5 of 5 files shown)


stephenplusplus:
Could you go comment-crazy in the files, in terms of headings to each major file / function on what it's purpose is?

scripts/build.js
```diff
@@ -0,0 +1,103 @@
...
+
+var TRAVIS_BRANCH = process.env.TRAVIS_BRANCH;
+var IS_PULL_REQUEST = process.env.TRAVIS_PULL_REQUEST === 'true';
+var TEST_ONLY = process.env.TEST_ONLY === 'true';
+
+var modules = Module.getUpdated();
+
+if (!modules.length) {
+  echo('No code changes found, exiting early.');
+  exit();
```
stephenplusplus: Does this need a `return;`?

scripts/build.js
```diff
@@ -0,0 +1,103 @@
...
+coveralls.handleInput(coverage, function(err) {
+  if (err) {
+    echo('Coveralls error!');
+    echo(err.message || err);
+  }
+});
+
+if (TRAVIS_BRANCH !== 'master' || IS_PULL_REQUEST) {
+  echo('Skipping system tests.');
+  exit();
```
stephenplusplus: `return;`?

scripts/release.js
```diff
@@ -0,0 +1,130 @@
...
+  var parts = TRAVIS_TAG.split('-');
+
+  MODULE_VERSION = parts.pop();
+  MODULE_NAME = parts.pop();
+}
+
+var mod = new Module(MODULE_NAME);
+
+mod.install();
+mod.runSystemTests();
```
stephenplusplus: What is a release -- a merge to master?

scripts/build.js
callmehiphop: Per [shelljs](https://github.com/shelljs/shelljs) I don't think so. I'm guessing it calls `process.exit()` underneath.

scripts/release.js
stephenplusplus: > does not test to see if the release integrates with other modules

A change to common shouldn't run other modules' tests because common changes have to come first before the other modules update their dependency on common. We have to trust the integrity of our `common` test suite to know that we aren't breaking anything.

> google-cloud does not run tests for any other modules (PubSub, BigQuery, etc.)

If our dev->release process goes right, it shouldn't have to. Only modules that pass unit & system tests should be released, so google-cloud would never release a non-working module.

Am I not thinking of something where this wouldn't work?

stephenplusplus:
I think we might need an overhaul on what a build is and what a release is. Since Travis doesn't publish to npm for us anymore, what distinguishes the two?

stephenplusplus:
Topics that would be awesome if they were covered in the "Developer's Documentation" code comments:

- What a build is, what a release is, what triggers each, etc.
- An overview of the chain of events that take place
- How CI runs code coverage (is it only running and reporting it once now?)

I have some more code-specific confusions, but I'll wait to see something like ^^ before I confuse us both by asking.

scripts/release.js
callmehiphop: > Am I not thinking of something where these thoughts aren't good enough?

Nope! I'm probably just overthinking things :)

stephenplusplus: That's okay, with my under-thinking we should land somewhere in the middle :)

callmehiphop:
@stephenplusplus I went ahead and added a bunch of docs/comments to <kbd>build.js</kbd> - PTAL and let me know if you're cool with the format, then I'll do the others.

scripts/build.js
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * - First we check to see if any modules have had code updates. A code update
+ *   is either a .js file update or a dependency update. In the event that a
+ *   module did not receive a code update (could be a markdown update), we'll
+ *   exit early.
+ *
+ * - For each module that has updates, we'll install their dependencies and run
+ *   their unit tests.
+ *
+ * - Assuming all the unit tests pass, we'll then attempt to get code coverage
```
stephenplusplus: Ooh, good point... how do we get code coverage for the whole library?

callmehiphop: I think we'd have to install all the dependencies for all the modules and run unit tests against those. I can add in some code to install everything for merges to master (I'm assuming that's the only time we'll care about it)

scripts/build.js
```diff
@@ -0,0 +1,161 @@
...
+ * - At this point, if this "build" was triggered by a PR or is a merge to a
+ *   branch other than master, we'll exit early here. All steps after this point
+ *   assume a merge to the master branch.
+ *
+ * - Next we'll run system tests for any and all modules that have recieved code
+ *   updates.
+ *
+ * - Assuming all system tests pass, we'll create symlinks for each module
+ *   via `npm link`. We'll then search for any modules that have the updated
+ *   modules listed as dependencies or devDependencies. For each dependent
```
stephenplusplus: Aside from `common`, we only use `devDependencies` when depending on other modules, right? Is this on purpose, so that changes to common will run all of the other module's system tests? In both cases (devDeps and deps), this could be a problem. I don't think we can require that changes to a module won't break other modules using the old version, since the change might intend to be a breaking change. Sorry if I led us down this path without fully thinking it through :(

scripts/build.js
```diff
@@ -0,0 +1,161 @@
...
+ *   module we find, we'll run it's system tests using the symlinks in place
+ *   of the latest stable dependency version.
+ *
+ * - We'll then check to see if this build requires a documentation update.
+ *   This is determined by the `TEST_ONLY` environment variable. This is set
+ *   in the .travis.yml file. If this build does not require a documentation
```
stephenplusplus: This will be run by AppVeyor as well, will this all still work?

stephenplusplus:
> I went ahead and added a bunch of docs/comments to build.js - PTAL and let me know if you're cool with the format, then I'll do the others.

They're beautiful 😢

scripts/build.js
stephenplusplus: Just a crazy idea... should we have a code coverage report for each module?

scripts/build.js
callmehiphop: Ohh right, I'll change it to `GENERATE_DOCS`, then if we don't bother setting it in AppVeyor it will come back false and we'll just never generate docs there.

scripts/build.js
stephenplusplus: I'm thinking coveralls won't let us do that.

> I can add in some code to install everything for merges to master (I'm assuming that's the only time we'll care about it)

We'll need our coverage reports to be generated for each commit to a PR, so that we (maintainers and contributors) can see if we're busting stuff.

scripts/build.js
stephenplusplus: Cool cool.

scripts/build.js
callmehiphop: That's true, I can add an exception for `common` in that case. I did it on purpose, but I see where you're coming from. I guess we'll just have to be more diligent about testing common locally.

stephenplusplus: Yeah, and I think that's implied, since if we're messing with common, that means we need it for the benefit of another module. IOW, we're not just playing around in there.

scripts/build.js
callmehiphop: That should work as is (I think?) But it only runs the report against the files that have changed. Not sure if that's going to mess with the % or not though..

stephenplusplus: Yeah, it'll be like when I leave it.only and our coverage drops by 98.82% :(

stephenplusplus:
Is there anything left to do on this before I review?

callmehiphop:
@stephenplusplus I think it's good to go!

PR #1600, opened 2016-09-19
storage: add warning for resumable uploads

### LINKED ISSUE #1598: Improve storage createWriteStream options docs for resumable
If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS: Linux, OS X
  - Node.js version: 6.5.x
  - npm version: 3.10.x
  - google-cloud-node version: 0.38.3

#### Steps to reproduce

  1. See https://github.com/GoogleCloudPlatform/google-cloud-node/issues/397#issuecomment-101916003

Can we please do better at documenting [these default performance problems and workarounds when uploading many medium-sized or smaller files](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/397#issuecomment-101916003) in the API docs? 

My team just ran into these and had to dig into a closed issue to find the solution. It seems like there should be a big warning about resumable being defaulted to true and how it can affect performance in the docs!

Instead the docs say this:
```
Resumable uploads are automatically enabled and must be shut off explicitly by setting options.resumable to false.
```

It gives no reason why you would shut it off. No mention of performance impact at all.

callmehiphop:
Closes #1598 

Preview:
screen shot 2016-09-19 at 11 10 42 am

### DIFF
--- a/packages/storage/src/file.js
+++ b/packages/storage/src/file.js
@@ -730,10 +730,10 @@ File.prototype.createResumableUpload = function(options, callback) {
  * by setting `options.resumable` to `false`.
  *
  * <p class="notice">
- *   Resumable uploads have been known to cause performance issues. For each
- *   file being uploaded we store the first 16kb of data, this can cause
- *   considerable performance degredation when uploading a large number of
- *   files.
+ *   There is some overhead when using a resumable upload that can cause
+ *   noticeable performance degradation while uploading a series of small files.
+ *   When uploading files less than 10MB, it is recommended that the resumable
+ *   feature is disabled.
  * </p>
  *
  * For faster crc32c computation, you must manually install


packages/storage/src/file.js
```diff
@@ -729,6 +729,13 @@ File.prototype.createResumableUpload = function(options, callback) {
  * Resumable uploads are automatically enabled and must be shut off explicitly
  * by setting `options.resumable` to `false`.
  *
+ * <p class="notice">
+ *   Resumable uploads have been known to cause performance issues. For each
+ *   file being uploaded we store the first 16kb of data, this can cause
+ *   considerable performance degredation when uploading a large number of
+ *   files.
+ * </p>
```
callmehiphop: Why do we want to note that it only happens to small files? Both of the issues reported mention medium sized files - and based on what I read I thought this issue pertained the volume of images and not the image sizes themselves.

stephenplusplus: Small and medium are loose terms which is why I threw out 10 MB. I don't think resumable uploads are even possible upstream if the file size is less than that, so wasting time caching information about the upload is where the performance hit comes from.

callmehiphop: PTAL

Splaktar:
This will help quite a few people! Thank you!

callmehiphop:
@Splaktar Thanks for opening an issue about this!

PR #1606, opened 2016-09-21
speech: doc fixes

callmehiphop:
There were a couple errors in the README examples and I was seeing a documentation test error on Windows machine, looks like the `tmp` path on Windows was coming out funky.

### DIFF
--- a/test/docs.js
+++ b/test/docs.js
@@ -54,28 +54,22 @@ function runCodeInSandbox(code, sandbox) {
 }
 
 describe('documentation', function() {
-  var tempFile = 'temp-test-file.tmp';
   var MITM;
 
-  before(function(done) {
+  before(function() {
     // Set a global to indicate to any interested function inside of gcloud that
     // this is a sandbox environment.
     global.GCLOUD_SANDBOX_ENV = true;
 
     // Turn off the network so that API calls aren't actually made.
     MITM = mitm();
-
-    // Create a temporary file for any examples that try to write to disk.
-    fs.writeFile(tempFile, 'test', done);
   });
 
-  after(function(done) {
+  after(function() {
     delete global.GCLOUD_SANDBOX_ENV;
 
     // Re-enable the network.
     MITM.disable();
-
-    fs.unlink(tempFile, done);
   });
 
   var ignore = [
@@ -150,6 +144,26 @@ describe('documentation', function() {
         };
       }
 
+      function FakeFs() {
+        return {
+          writeFile: function(name, contents, callback) {
+            callback(null);
+          },
+          readFile: function(filename, callback) {
+            callback(null, new Buffer(''));
+          },
+          readFileSync: function() {
+            return new Buffer('');
+          },
+          createWriteStream: function() {
+            return require('through2')();
+          },
+          createReadStream: function() {
+            return require('through2')();
+          }
+        };
+      }
+
       fileDocBlocks.methods.forEach(function(method) {
         var code = method.examples.map(prop('code')).join('\n');
         var lowercaseId = method.id.toLowerCase();
@@ -188,10 +202,7 @@ describe('documentation', function() {
           )
           .replace('require(\'express\')', FakeExpress.toString())
           .replace('require(\'level\')', FakeLevel.toString())
-
-          // in:  fs.anyMethod('--any-file-path--
-          // out: fs.anymethod('--a-temporary-file--
-          .replace(/(fs\.[^(]+\(['"])([^'"]*)/g, '$1' + tempFile);
+          .replace('require(\'fs\')', '(' + FakeFs.toString() + '())');
 
         var displayName = filename
           .replace('docs/json/master/', '')
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "docs": "node ./scripts/docs/packages.js",
     "bundle": "node ./scripts/docs/bundle.js",
     "lint": "jshint scripts/ packages/ system-test/ test/ && jscs packages/ system-test/ test/",
-    "test": "npm run docs && npm run bundle && mocha --timeout 5000 --bail test/docs.js packages/*/test/*.js",
+    "test": "npm run docs && mocha --timeout 5000 --bail test/docs.js packages/*/test/*.js",
     "system-test": "mocha packages/*/system-test/*.js --no-timeouts --bail",
     "cover": "istanbul cover _mocha --report lcovonly -x 'packages/*/src/v*/*.js' -- --no-timeouts --bail packages/*/test/*.js -R spec",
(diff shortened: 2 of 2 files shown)


callmehiphop:
@jmdobry could you PTAL? @stephenplusplus is out of town at the moment.

test/docs.js
```diff
@@ -172,9 +176,6 @@ describe('documentation', function() {
           code = moduleInstantationCode + code;
         }
 
-        // Create a temporary file for any examples that try to write to disk.
```
stephenplusplus: Can you re-add this comment back above the new line?

stephenplusplus:
What was the Windows error? Can you see if a module exists to create a temp file that is Windows compatible? Maybe even 'tmp' that we already use some places.

callmehiphop:
We were already using `tmp` and the path that it was giving back was bad, Windows was throwing an error because it couldn't read the file. It could be because we were calling it synchronously? All the system-tests seem to be using the async version.

stephenplusplus:
Can you paste the error?

callmehiphop:
https://ci.appveyor.com/project/GoogleCloudPlatform/google-cloud-node/build/1.0.111/job/twamtlqs0rh2m91n#L10508

jmdobry:
The README changes look fine to me.

test/docs.js
```diff
@@ -189,7 +191,7 @@ describe('documentation', function() {
 
           // in:  fs.anyMethod('--any-file-path--
           // out: fs.anymethod('--a-temporary-file--
-          .replace(/(fs\.[^(]+\(['"])([^'"]*)/g, '$1' + temporaryFile.name);
```
stephenplusplus: I know this isn't fun, but can you `console.log(temporaryFile.name)` here so that when AppVeyor runs it, we can see it?

It seems like something with the `replace` line could be messed up, since the error is:

> ENOENT, no such file or directory 'C:\UsersappveyorAppDataLocalTemp	mp-2676Xj4IRwkvTWvY.tmp'

callmehiphop: `C:\Users\\appveyor\\AppData\\Local\\Temp\\mp-2676Xj4IRwkvTWvY.tmp`

stephenplusplus: So presumably, the line that's failing looks like:

```js
speech.recognize({
  contents: fs.readFileSync('C:\Users\\appveyor\\AppData\\Local\\Temp\\mp-2676Xj4IRwkvTWvY.tmp')
}, ...
```

I think we might need to use `C:\\`, which can be done by calling `path.normalize(temporaryFile.name)`

callmehiphop: I tried this locally and I still get the same error.

stephenplusplus: You get the error on Windows?

callmehiphop: Correct

stephenplusplus: Okay, did you try manually putting in two \\ after C:?

stephenplusplus: Feel free to try anything else you can think of. This module definitely supports Windows, so it's just a matter of figuring out what might be going wrong. The reason I want to use a system temp file is because we inevitably will end up leaving a temp file on disk. If we absolutely can't get this module to work, I'm okay with checking in a "test-file.temp" that's meant just for this use.

callmehiphop: - I tried your previous suggestion and it didn't work
- The `after` hook has an `fs.unlink` call, so I think the file should get cleaned up.
- We could even go so far as to add it to the <kbd>.gitignore</kbd> if it concerns you.

jmdobry: Instead of rewriting the code, what about mocking `fs` and intercepting the filename?

stephenplusplus: While there is the manual unlink in the after hook, we know that's not always reliable from experience-- check how many buckets/clusters/instances/etc etc are in your GCP console that start with "GCLOUD-TEST" 😉. In general, the thought was since we have a module that's for this exact purpose, we shouldn't have to be writing it out ourselves.

> I tried your previous suggestion and it didn't work

I'm not sure if that means normalize, manual C:\ -> C:\\ replacement, or both. You also had an idea about using the async way of creating a temp file. But if you're done playing with the temp module, you can just merge the docs changes and I'll give it a go later this week.

stephenplusplus: > Instead of rewriting the code, what about mocking fs and intercepting the filename?

I like that idea as well, but the downside is we have to recreate all of the `fs` methods -- we'd have to figure out what is our using of `fs` vs. some other internal API that might actually need it to work.

jmdobry: I think you could programmatically mock _all_ of the `fs` methods in a loop. Just check if the first argument passed to the method is a string, and if so replace it with the temp filename, then call original `fs` method. It appears most of the `fs` methods take a `path` as their first argument.

stephenplusplus: Yeah, I think the possible problem could be if some of the code needs to actually use fs. Can't hurt to try though!

callmehiphop: So I found a different workaround using `tmp` which is pretty much the following.

```js
var tmpFile = tmp.fileSync();
var name = process.platform === 'win32' ? tmpFile.name.split('\\').join('\\\\') : tmpFile.name;
```

On a scale of 1-10 how much do you hate it? lul

jmdobry:
On another topic, when do we want to publish 0.1.0? Wait for the first gax-nodejs PR?

callmehiphop:
@stephenplusplus @jmdobry I changed the code to mock `fs` and doc tests seem to be passing.

stephenplusplus:
Awesome, glad that mocking worked out. The AppVeyor build log shows a failure:

```
1) gcloud should export static speech:
     AssertionError: undefined === [Function: FakeApi]
      at Context.<anonymous> (C:\projects\google-cloud-node\packages\google-cloud\test\index.js:117:12)
      at callFn (C:\projects\google-cloud-node\node_modules\mocha\lib\runnable.js:326:21)
      at Test.Runnable.run (C:\projects\google-cloud-node\node_modules\mocha\lib\runnable.js:319:7)
      at Runner.runTest (C:\projects\google-cloud-node\node_modules\mocha\lib\runner.js:422:10)
      at C:\projects\google-cloud-node\node_modules\mocha\lib\runner.js:528:12
      at next (C:\projects\google-cloud-node\node_modules\mocha\lib\runner.js:342:14)
      at C:\projects\google-cloud-node\node_modules\mocha\lib\runner.js:352:7
      at next (C:\projects\google-cloud-node\node_modules\mocha\lib\runner.js:284:14)
      at Immediate._onImmediate (C:\projects\google-cloud-node\node_modules\mocha\lib\runner.js:320:5)
```

callmehiphop:
That's because unit tests for google-cloud#speech were created before
speech was released. First we need to fix the docs here, then publish it.
Finally install it as a dependency on google-cloud.

stephenplusplus:
Ah, okay. Merge time.

PR #1608, opened 2016-09-23
umbrella: add speech

callmehiphop:
Depends on speech 0.1.0 release.

### DIFF
--- a/packages/google-cloud/package.json
+++ b/packages/google-cloud/package.json
@@ -88,6 +88,9 @@
     "google cloud resource manager",
     "cloud resource manager",
     "resource manager",
+    "google cloud speech",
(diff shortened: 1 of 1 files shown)


callmehiphop:
@stephenplusplus I think this should be good to go!

stephenplusplus:
Can you add speech-related keywords to the package.json?

packages/google-cloud/package.json
```diff
@@ -88,6 +88,8 @@
     "google cloud resource manager",
     "cloud resource manager",
     "resource manager",
+    "google speech",
```
stephenplusplus: google cloud speech,
cloud speech,
speech

PR #1614, opened 2016-09-23
Introduce auto-generated Gapic client for speech API

jmuk:


jmuk:
cc: @jmdobry

jmdobry:
LGTM. Once this is merged, I feel we should evaluate whether to cut a release before changing the hand-written stuff to use the autogen stuff under the hood (my vote is to update the hand-written stuff). It probably wouldn't be much work, and it'd be nice to catch any possible breaking changes sooner rather than after we've already released 0.1.0.

packages/speech/src/index.js
```diff
@@ -137,6 +137,7 @@ Speech.prototype.endpointerTypes = {
  */
 Speech.detectEncoding_ = function(filename) {
   if (!is.string(filename)) {
+    // istanbul ignore next
```
stephenplusplus: Oops, did this drop our coverage? I'd rather just add a test for this. I can do this in a separate PR.

jmuk: I just added a test case.

packages/speech/test/index.js
```diff
@@ -158,6 +158,10 @@ describe('Speech', function() {
         Speech.detectEncoding_('blah.mp3');
       }, /Encoding could not be determined for file: blah\.mp3/);
     });
+
+    it('should return nothing if the argument is not a string', function() {
+      assert.equal(Speech.detectEncoding_({}), undefined);
```
stephenplusplus: Can you switch this to `assert.strictEqual`?

bjwatson:
@jmuk Looks like we need an update with the latest protos: https://github.com/googleapis/googleapis/pull/153

stephenplusplus:
I subscribed to the PR, but if I miss it, please ping me and I'll make a new release to `google-proto-files` with the update after it merges.

packages/speech/test/index.js
jmuk: done

stephenplusplus:
Thanks for adding the test and syncing with master. Do I need to cut a new `google-proto-files` release with the changes from https://github.com/googleapis/googleapis/commit/57bf5353625669e2bd42adc85356ce3740f8f456?

jmuk:
Seems it's not necessary to create a new version of `google-proto-files`, because the change itself is mostly for comments and no changes on the protocol structures (it adds a new `AsyncRecognizeMetadata` message, but that is not used at all, so I'm still not sure why it's added).

stephenplusplus:
Okay, we can wait on that then. I'll merge this once CI says it's happy.

bjwatson:
@jmuk @stephenplusplus Looks like @landrito has pushed the proto updates to Github: https://github.com/googleapis/googleapis/commit/57bf5353625669e2bd42adc85356ce3740f8f456

PR #1616, opened 2016-09-23
testing: updating check for travis pr

callmehiphop:


scripts/helpers.js
```diff
@@ -361,8 +361,12 @@ var BRANCH = process.env.TRAVIS_BRANCH || process.env.APPVEYOR_REPO_BRANCH;
  *
  * @alias ci.IS_PR
  */
-var IS_PR = process.env.TRAVIS_PULL_REQUEST === 'true' ||
-  !!process.env.APPVEYOR_PULL_REQUEST_NUMBER;
+var IS_PR = (function() {
+  if (process.env.TRAVIS_PULL_REQUEST) {
+    return process.env.TRAVIS_PULL_REQUEST !== 'false';
+  }
+  return !!process.env.APPVEYOR_PULL_REQUEST_NUMBER;
```
stephenplusplus: Instead of the IIFE, how about something like:

```js
var PR_NUMBER = process.env.APPVEYOR_PULL_REQUEST_NUMBER ||
  process.env.TRAVIS_PULL_REQUEST;

var IS_PR = parseInt(PR_NUMBER, 10) !== NaN;
```

callmehiphop: I like it, PTAL!

PR #1626, opened 2016-09-26
maintenance: add auto-updating of npm packages

stephenplusplus:
This script will update our sub-modules to the latest compatible dependencies when run:

``` sh
$ npm run update-deps
```

Also included in the PR are the modified package.json's after running.

### DIFF
--- a/scripts/update-deps.js
+++ b/scripts/update-deps.js
@@ -91,7 +91,6 @@ function updatePackage(pkg, callback) {
 }
 
 function shouldUpdateDep(dependency, callback) {
-  console.log(BLACKLIST, dependency.name)
   if (BLACKLIST.indexOf(dependency.name) > -1) {
     setImmediate(function() {
       callback(null, false);


packages/dns/package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "@google-cloud/common": "^0.5.0",
     "arrify": "^1.0.0",
-    "dns-zonefile": "0.1.18",
+    "dns-zonefile": "^0.1.18",
```
stephenplusplus: @callmehiphop We intentionally wanted 0.1.18 and not any updates, didn't we? I can exclude this if so.

callmehiphop: I think that's right, but I'd have to search for the reasoning behind that.

stephenplusplus: Added it to a blacklist anyway.

scripts/update-deps.js
```diff
@@ -0,0 +1,145 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+var async = require('async');
+var david = require('david');
```
callmehiphop: this is my new favorite package

stephenplusplus: 💃

callmehiphop:
So what's the long term plan for this? Run it once in a while to see if we have updates..? I think it would be great it we could tie this into CI somewhere and have it just run weekly or so..?

stephenplusplus:
That's interesting. My plan was to simply run it manually before publishing.

PR #1631, opened 2016-09-26
Make Speech client use autogen layer

jmdobry:
My initial attempt at using the autogen layer in the handwritten layer. I have a concern:

It would appear that `gax-nodejs` does not translate camel case option names into the snake case names defined the proto. Where `sampleRate` worked before, thy tests for `recognize` and `createRecognitionStream` started failing until I switched it to `sample_rate`.

Other than that it seems to work fine.

I found some other `gax-nodejs` usability issues with pagination that don't apply here.

### DIFF
--- a/packages/speech/src/index.js
+++ b/packages/speech/src/index.js
@@ -88,6 +88,7 @@ function Speech(options) {
   };
 
   common.GrpcService.call(this, config, options);
+
   this.api = {
     Speech: v1beta1(options).speechApi(options)
   };
--- a/packages/speech/test/index.js
+++ b/packages/speech/test/index.js
@@ -582,6 +582,8 @@ describe('Speech', function() {
       Speech.findFile_ = function(files, callback) {
         callback(null, FOUND_FILE);
       };
+
+      speech.api.Speech.syncRecognize = util.noop;
     });
 
     it('should find the files', function(done) {
@@ -594,16 +596,13 @@ describe('Speech', function() {
     });
 
     it('should make the correct request', function(done) {
-      speech.request = function(protoOpts, reqOpts) {
-        assert.deepEqual(protoOpts, {
-          service: 'Speech',
-          method: 'syncRecognize'
+      speech.api.Speech.syncRecognize = function(config, file) {
+        var expectedConfig = extend({}, CONFIG, {
+          encoding: DETECTED_ENCODING
         });
+        assert.deepEqual(config, expectedConfig);
 
-        assert.deepEqual(reqOpts, {
-          config: extend({}, CONFIG, { encoding: DETECTED_ENCODING }),
-          audio: FOUND_FILE
-        });
+        assert.strictEqual(file, FOUND_FILE);
 
         done();
       };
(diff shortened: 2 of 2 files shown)


stephenplusplus:
// @jmuk

packages/speech/src/index.js
```diff
@@ -87,6 +88,7 @@ function Speech(options) {
   };
 
   common.GrpcService.call(this, config, options);
+  this.api = new v1beta1(options).speechApi();
```
stephenplusplus: Is this the line you wished was simpler? Ideally, what would it look like?

jmdobry: Yeah, though looking back through https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1476 it makes more sense why it works the way it works. Also, looks like the `new` is unnecessary.

packages/speech/src/index.js
```diff
@@ -87,6 +88,9 @@ function Speech(options) {
   };
 
   common.GrpcService.call(this, config, options);
+  this.api = {
```
jmdobry: I figure other packages can follow this same convention, (nest autogen services under `api`) as some packages will actually need to instantiate multiple autogen services, i.e. Pub/Sub.

packages/speech/src/index.js
```diff
@@ -87,6 +88,9 @@ function Speech(options) {
...
+    speechApi: v1beta1(options).speechApi()
```
jmdobry: Maybe the property name could just be `speech` instead of `speechApi`.

packages/speech/src/index.js
stephenplusplus: I'm cool with that. Small tweak, maybe `this.api.speech` since `this.api.speechApi` is somewhat redundant.

packages/speech/src/index.js
jmdobry: Renamed to `Speech`, to match the spelling and capitalization of the actual gRPC service name. With this convention, `this.api` will actually look quite similar to `this.protos`.

stephenplusplus: Awesome 👍

stephenplusplus:
Just adding the `don't merge` while we wait to see if we can stick to 🐫 💼 .

stephenplusplus:
> I found some other gax-nodejs usability issues with pagination that don't apply here.

Anything to call out here, or are you going to file something somewhere else?

jmdobry:
Regarding camel case, it looks like `gax-nodejs` might need to be updated to output code that passes `{ convertFieldsToCamelCase: true }` here: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/speech/src/v1beta1/speech_api.js#L220

Regarding pagination, will file elsewhere.

stephenplusplus:
There's a new release of gax out (0.7.0) which camelCases and upgrades to gRPC@1.0. I believe we're now just waiting on the API files to be generated.

jmdobry:
I'll wait for that PR to be merged then.

stephenplusplus:
The PR with the re-generated files has been merged.

packages/speech/src/index.js
```diff
@@ -651,12 +650,7 @@ Speech.prototype.recognize = function(file, config, callback) {
       return;
     }
 
-    var reqOpts = {
-      audio: foundFile,
-      config: config
-    };
-
-    self.request(protoOpts, reqOpts, function(err, apiResponse) {
+    self.api.Speech.syncRecognize(config, foundFile, function(err, apiResponse) {
```
stephenplusplus: To get around this line being >80 chars, let's just rename `apiResponse` to `resp`.

jmdobry:
Updated

packages/speech/src/index.js
jmdobry: Done

PR #1639, opened 2016-09-28
speech: fix docs to use correct event

evancohen:
After a little bit of head scratching I'm pretty sure that the docs are incorrect. The `requestStream` emits a `'response'` event: [`index.js#L484`](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/speech/src/index.js#L484)

Right now they the docs claim it is `'data'`. The event name should either be changed or the docs should be updated to reflect the actual name to prevent further confusion.

Note: this may impact #1631


evancohen:
Nope. Please disregard.

PR #1647, opened 2016-09-29
translate: allow setting format of text

stephenplusplus:
_Continuing #1574_

See #1574 for the discussion of this feature.

@innovative1 thanks for the idea and the original implementation! Can you take a look at this one?

### DIFF
--- a/packages/translate/src/index.js
+++ b/packages/translate/src/index.js
@@ -270,6 +270,8 @@ Translate.prototype.getLanguages = function(target, callback) {
  *     `en` for English). If an object, you may also specify the source
  *     language.
  * @param {string} options.format - Set the text's format as `html` or `text`.
+ *     If not provided, we will try to auto-detect if the text given is HTML. If
+ *     not, we set the format as `text`.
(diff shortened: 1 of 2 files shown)


packages/translate/src/index.js
```diff
@@ -268,6 +269,7 @@ Translate.prototype.getLanguages = function(target, callback) {
  *     target ISO 639-1 language code to translate the source input to. (e.g.
  *     `en` for English). If an object, you may also specify the source
  *     language.
+ * @param {string} options.format - Set the text's format as `html` or `text`.
```
callmehiphop: We should probably mark this as optional and mention that the default is text.

callmehiphop:
@stephenplusplus can we add a system test for this?

stephenplusplus:
PTAL!

packages/translate/src/index.js
stephenplusplus: `options` itself is marked as optional. I elaborated on what happens if a user doesn't provide a format.

Issue #1607, opened 2016-09-22
datastore: decoding dates does not load nanos


joel31415:
The nanos field is set when encoding a date, but not loaded when decoding.

I would expect Dates to be encoded and decoded with millisecond precision.

It seems intentional looking at the unit tests, but I hope it's a bug!

PR #1648, opened 2016-09-29
bigquery: convert nested objects to their native types

### LINKED ISSUE #1593: standard bigquery v2, support nested schema in query result
Please support nested schema field names in complex queries like `SELECT ARRAY[STRUCT(1 AS a, 2 AS b, 3 AS c)] AS d`

```js
const BigQuery = require('@google-cloud/bigquery');
> BigQuery().query({ query: 'SELECT 1 AS cnt', useLegacySql: false }, function (err, data) { if (err) return console.error(err); console.dir(data, { depth: null, colors: true }); });
> [ { cnt: 1 } ]
```

This is right, the field name `cnt` is correctly used

```js
> BigQuery().query({ query: 'SELECT ARRAY[STRUCT(1 AS a, 2 AS b, 3 AS c)] AS d', useLegacySql: false }, function (err, data) { if (err) return console.error(err); console.dir(data, { depth: null, colors: true }); });
> [ { d: [ { v: { f: [ { v: '1' }, { v: '2' }, { v: '3' } ] } } ] } ] 
```

- only the outer level field name `d` is in use, not others a, b, c,
- all get default `f` and `v`, and all their values type, the 1,2,3 here should be integers

I have even more complex queries than this, all are like this only the outer one level schema field names / value types are got populated

A raw query over HTTP, the http tool is from [httpie](https://pypi.python.org/pypi/httpie), just an advanced version of curl; after I get access token with same httpie tool from token allocation, I can do queries like this on command line, here the server response included correct schema with nested fields, so technically it should be possible to populate all inner level field names and value types correct

```console
➸ time http --verbose --pretty=colors POST https://www.googleapis.com/bigquery/v2/projects/XXXXXXX/queries 'Authorization: Bearer XXXXXXXXXXXX' useLegacySql=false useQueryCache=false query='SELECT ARRAY[STRUCT(1 AS a, 2 AS b, 3 AS c)] AS d'
POST /bigquery/v2/projects/XXXXXXXXXXXXXX/queries HTTP/1.1
Content-Length: 113
Accept-Encoding: gzip, deflate
Host: www.googleapis.com
Accept: application/json, */*
User-Agent: HTTPie/0.9.6
Connection: keep-alive
Content-Type: application/json
Authorization: Bearer XXXXXXXXXXXXXXXXXXXXXX

{"useLegacySql": "false", "useQueryCache": "false", "query": "SELECT ARRAY[STRUCT(1 AS a, 2 AS b, 3 AS c)] AS d"}

HTTP/1.1 200 OK
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Pragma: no-cache
Expires: Mon, 01 Jan 1990 00:00:00 GMT
Date: Wed, 14 Sep 2016 17:37:03 GMT
ETag: "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
Vary: Origin
Vary: X-Origin
Content-Type: application/json; charset=UTF-8
Content-Encoding: gzip
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
Server: GSE
Alt-Svc: quic=":443"; ma=2592000; v="36,35,34,33,32"
Transfer-Encoding: chunked

{
 "kind": "bigquery#queryResponse",
 "schema": {
  "fields": [
   {
    "name": "d",
    "type": "RECORD",
    "mode": "REPEATED",
    "fields": [
     {
      "name": "a",
      "type": "INTEGER",
      "mode": "NULLABLE"
     },
     {
      "name": "b",
      "type": "INTEGER",
      "mode": "NULLABLE"
     },
     {
      "name": "c",
      "type": "INTEGER",
      "mode": "NULLABLE"
     }
    ]
   }
  ]
 },
 "jobReference": {
  "projectId": "XXXXXXXXXXXXX",
  "jobId": "job_XXXXXXXXXXXXXXXXXXXXXXXXXXX"
 },
 "totalRows": "1",
 "rows": [
  {
   "f": [
    {
     "v": [
      {
       "v": {
        "f": [
         {
          "v": "1"
         },
         {
          "v": "2"
         },
         {
          "v": "3"
         }
        ]
       }
      }
     ]
    }
   ]
  }
 ],
 "totalBytesProcessed": "0",
 "jobComplete": true,
 "cacheHit": false
}

real	0m3.040s
user	0m0.364s
sys	0m0.020s
```

#### Environment details

  - OS: Linux
  - Node.js version: v6

I have search other issues here #1564 #1135 about nested schema, but for different purpose, I believe this ticket is different.

stephenplusplus:
Fixes #1593

### DIFF
--- a/packages/bigquery/src/table.js
+++ b/packages/bigquery/src/table.js
@@ -1031,7 +1031,7 @@ Table.prototype.query = function(query, callback) {
  * @param {string} metadata.name - A descriptive name for the table.
  * @param {string|object} metadata.schema - A comma-separated list of name:type
  *     pairs. Valid types are "string", "integer", "float", "boolean", "bytes",
- *     "recodr", and "timestamp". If the type is omitted, it is assumed to be
+ *     "record", and "timestamp". If the type is omitted, it is assumed to be
  *     "string". Example: "name:string, age:integer". Schemas can also be
  *     specified as a JSON array of fields, which allows for nested and repeated
  *     fields. See a [Table resource](http://goo.gl/sl8Dmg) for more detailed


c0b:
looks good :+1:

c0b:
wow, it still somewhat not working because mode 'REPEATED' could be on every possible type! not only on 'STRUCT';

try this query:

```console
$ bq query --use_legacy_sql=false 'SELECT GENERATE_ARRAY(10, 0, -3) AS example_array'

[
  {
    "example_array": [
      "10", 
      "7", 
      "4", 
      "1"
    ]
  }
]

```

```js
{ err: null,
  rows: [ { example_array: [ { v: '10' }, { v: '7' }, { v: '4' }, { v: '1' } ] } ],
  nextQuery: null,
  apiResponse: 
   { kind: 'bigquery#queryResponse',
     schema: { fields: [ { name: 'example_array', type: 'INTEGER', mode: 'REPEATED' } ] },
     ...
     totalRows: '1',
     rows: [ { f: [ { v: [ { v: '10' }, { v: '7' }, { v: '4' }, { v: '1' } ] } ] } ],
     totalBytesProcessed: '0',
     jobComplete: true,
     cacheHit: true } }

```

stephenplusplus:
Phew, thanks for catching that. I reorganized a bit, and I think I've covered that case now. Let me know what you think.

packages/bigquery/src/table.js
```diff
@@ -188,42 +188,64 @@ Table.createSchemaFromString_ = function(str) {
...
   }
 
+  function convert(schemaField, value) {
+    switch (schemaField.type) {
+      case 'BOOLEAN': {
+        value = value === 'true';
+        break;
+      }
+      case 'FLOAT': {
+        if (!is.nil(value)) {
```
c0b: since now `value === null` has been checked above; I assume all `is.nil` check can be dropped?

packages/bigquery/src/table.js
```diff
@@ -188,42 +188,64 @@ Table.createSchemaFromString_ = function(str) {
...
+          value = parseFloat(value);
+        }
+        break;
+      }
+      case 'INTEGER': {
+        if (!is.nil(value)) {
```
c0b: and here

packages/bigquery/src/table.js
```diff
@@ -188,42 +188,64 @@ Table.createSchemaFromString_ = function(str) {
...
-      var fieldObject = {};
       fieldObject[schemaField.name] = value;
       return fieldObject;
     });
   }
 
+  function convert(schemaField, value) {
+    switch (schemaField.type) {
+      case 'BOOLEAN': {
+        value = value === 'true';
```
c0b: a NULLABLE BOOLEAN have 3 states: true or false or null; above null check helped eliminate the behavior of converting null to false before;

c0b:
from some more testing I just realized bigquery has more data types than JavaScript, like TIME, DATE, DATETIME (which is different than TIMESTAMP), only DATE is convertable to a JavaScript Date but you may not want to convert?

```console
➸ node -p 'new Date("2016-10-01")'
2016-10-01T00:00:00.000Z
```

```console
node ./bigquery/queries.js sync 'SELECT DATETIME "2016-10-01 03:45:23.34583"'
{ err: null,
  rows: [ { f0_: '2016-10-01T03:45:23.345830' } ],
  nextQuery: null,
  apiResponse: 
   { kind: 'bigquery#queryResponse',
     schema: { fields: [ { name: 'f0_', type: 'DATETIME', mode: 'NULLABLE' } ] },
     jobReference: { ... },
     totalRows: '1',
     rows: [ { f: [ { v: '2016-10-01T03:45:23.345830' } ] } ],
     totalBytesProcessed: '0',
     jobComplete: true,
     cacheHit: true } }
Received 1 row(s)!
[ { f0_: '2016-10-01T03:45:23.345830' } ]
```

losing precision is another problem, all TIME DATETIME TIMESTAMP in BigQuery can be up to micro seconds but a JavaScript Date is only up to milliseconds,

```console
node ./bigquery/queries.js sync 'SELECT TIMESTAMP "2016-10-01 03:45:23.34583 Asia/Singapore"'
{ err: null,
  rows: [ { f0_: 2016-09-30T19:45:23.345Z } ],
  nextQuery: null,
  apiResponse: 
   { kind: 'bigquery#queryResponse',
     schema: { fields: [ { name: 'f0_', type: 'TIMESTAMP', mode: 'NULLABLE' } ] },
     jobReference: { ... },
     totalRows: '1',
     rows: [ { f: [ { v: '1.47526472334583E9' } ] } ],
     totalBytesProcessed: '0',
     jobComplete: true,
     cacheHit: false } }
Received 1 row(s)!
[ { f0_: 2016-09-30T19:45:23.345Z } ]
```

and losing precision in converting a INT64 to JavaScript Number, due to no true Integer in JavaScript; a JavaScript Number is only FLOAT64

```console
$ node ./bigquery/queries.js sync 'SELECT ARRAY<INT64>[0x7fff1234deadbeef, -0x8000000000000000] AS example_array'
{ err: null,
  rows: [ { example_array: [ 9223110580161593000, -9223372036854776000 ] } ],
  nextQuery: null,
  apiResponse: 
   { kind: 'bigquery#queryResponse',
     schema: { fields: [ { name: 'example_array', type: 'INTEGER', mode: 'REPEATED' } ] },
     jobReference: { ... },
     totalRows: '1',
     rows: [ { f: [ { v: [ { v: '9223110580161593071' }, { v: '-9223372036854775808' } ] } ] } ],
     totalBytesProcessed: '0',
     jobComplete: true,
     cacheHit: false } }
Received 1 row(s)!
[ { example_array: [ 9223110580161593000, -9223372036854776000 ] } ]
```

the queries.js is a modified https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/bigquery/queries.js to print raw apiResponse

I'm not sure if you guys discussed this before, this can be different issue which I don't have a good solution; #1607 #1611 were about Datastore.

packages/bigquery/src/table.js
stephenplusplus: I believe that caught null for a top-level property, but in the case it is an array, we still map over the values after the null check.

packages/bigquery/src/table.js
stephenplusplus: This actually helps convert the string `'true'` into a proper boolean.

stephenplusplus:
@c0b thanks for bringing that up! I will look into that.

packages/bigquery/src/table.js
c0b: you may instead do once nil check at beginning of convert function; while the docs saying `query result ARRAYs can never have NULL elements`

https://cloud.google.com/bigquery/sql-reference/data-types#array-type

in practice, if I try to have null in array, it ends up an error:

```js
node ./bigquery/queries.js sync 'SELECT ARRAY[1,3,2,NULL]'
{ err: 
   { ApiError: Array 'f0_' cannot have a null element

     errors: 
      [ { domain: 'global',
          reason: 'invalidQuery',
          message: 'Array \'f0_\' cannot have a null element',
          locationType: 'other',
          location: 'query' } ],
     response: undefined,
     message: 'Array \'f0_\' cannot have a null element' },
  rows: null,
  nextQuery: null,
  apiResponse: 
   { error: 
      { errors: 
         [ { domain: 'global',
             reason: 'invalidQuery',
             message: 'Array \'f0_\' cannot have a null element',
             locationType: 'other',
             location: 'query' } ],
        code: 400,
        message: 'Array \'f0_\' cannot have a null element' } } }

```

stephenplusplus: Gotcha, makes sense.

c0b:
I don't have a good solution for all above losing precision datatype conversions; but `BYTES` is wonderfully convertable to a nodejs Buffer type that I think you may add a clause into the convert function

```js
node ./bigquery/queries.js sync '
SELECT b"abc©"'
{ err: null,
  rows: [ { f0_: 'YWJjwqk=' } ],
  nextQuery: null,
  apiResponse: 
   { kind: 'bigquery#queryResponse',
     schema: { fields: [ { name: 'f0_', type: 'BYTES', mode: 'NULLABLE' } ] },
     jobReference: { ... },
     totalRows: '1',
     rows: [ { f: [ { v: 'YWJjwqk=' } ] } ],
     totalBytesProcessed: '0',
     jobComplete: true,
     cacheHit: false } }
Received 1 row(s)!
[ { f0_: 'YWJjwqk=' } ]
```

a `Buffer.from` call would suffice https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_string_encoding

```js
➸ node -p 'Buffer.from("YWJjwqk=", "base64")'
<Buffer 61 62 63 c2 a9>
```

when you do so; please also make sure stream writing also works, make sure writing a Buffer directly to a BYTES field works

here are my local changes with this pr:

```diff
 packages/bigquery/src/table.js | 24 +++++++++++-------------
 1 file changed, 11 insertions(+), 13 deletions(-)

diff --git a/packages/bigquery/src/table.js b/packages/bigquery/src/table.js
index 761f862..1b53cf1 100644
--- a/packages/bigquery/src/table.js
+++ b/packages/bigquery/src/table.js
@@ -191,60 +191,58 @@ Table.mergeSchemaWithRows_ = function(schema, rows) {
   return arrify(rows).map(mergeSchema).map(flattenRows);
 
   function mergeSchema(row) {
     return row.f.map(function(field, index) {
       var schemaField = schema.fields[index];
       var value = field.v;
 
-      var fieldObject = {};
-
-      if (value === null) {
-        fieldObject[schemaField.name] = null;
-        return fieldObject;
-      }
-
       if (schemaField.mode === 'REPEATED') {
         value = value.map(function(val) {
           return convert(schemaField, val.v);
         });
       } else {
         value = convert(schemaField, value);
       }
 
+      var fieldObject = {};
       fieldObject[schemaField.name] = value;
       return fieldObject;
     });
   }
 
   function convert(schemaField, value) {
+    if (is.nil(value)) {
+      return null;
+    }
+
     switch (schemaField.type) {
       case 'BOOLEAN': {
         value = value === 'true';
         break;
       }
       case 'FLOAT': {
-        if (!is.nil(value)) {
-          value = parseFloat(value);
-        }
+        value = parseFloat(value);
         break;
       }
       case 'INTEGER': {
-        if (!is.nil(value)) {
-          value = parseInt(value, 10);
-        }
+        value = parseInt(value, 10);
         break;
       }
       case 'RECORD': {
         value = Table.mergeSchemaWithRows_(schemaField, value).pop();
         break;
       }
       case 'TIMESTAMP': {
         value = new Date(value * 1000);
         break;
       }
+      case 'BYTES': {
+        value = Buffer.from(value, 'base64');
+        break;
+      }
     }
 
     return value;
   }
 
   function flattenRows(rows) {
     return rows.reduce(function(acc, row) {

```

stephenplusplus:
Alright, support for BYTES has been added, as well as pre-API request encoding, so that Buffer & Date objects are converted to their JSON API-safe values (`buffer.toString('base64')` & `date.toJSON()`).

Regarding `TIME`, `DATE`, `DATETIME`, and `TIMESTAMP`, we haven't discussed this before, so a new issue would be great.

Thank you again for continuing to catch these things and make our BigQuery support much better 👍

qw3r:
+1

packages/bigquery/src/table.js
```diff
@@ -977,11 +1030,12 @@ Table.prototype.query = function(query, callback) {
  *     table.
  * @param {string} metadata.name - A descriptive name for the table.
  * @param {string|object} metadata.schema - A comma-separated list of name:type
- *     pairs. Valid types are "string", "integer", "float", "boolean", and
- *     "timestamp". If the type is omitted, it is assumed to be "string".
- *     Example: "name:string, age:integer". Schemas can also be specified as a
- *     JSON array of fields, which allows for nested and repeated fields. See
- *     a [Table resource](http://goo.gl/sl8Dmg) for more detailed information.
+ *     pairs. Valid types are "string", "integer", "float", "boolean", "bytes",
+ *     "recodr", and "timestamp". If the type is omitted, it is assumed to be
```
callmehiphop: `record`*

stephenplusplus: Woops, fixed!

c0b:
:+1:

PR #1656, opened 2016-09-30
core: read projectId from authClient

### LINKED ISSUE #1653: Auto-detect a project ID
### Problem

After running `gcloud beta auth application-default login` I should be able to use google-cloud-node to access Google Cloud Platform services, but I can't. I _must_ set the `GCLOUD_PROJECT` env var or explicitly pass a project ID to a client for google-cloud-node to work.
### Steps to reproduce
1. Install the Google Cloud SDK, git, Node.js and NPM
2. Run `gcloud init`, you'll be prompted to set a default project
3. Run `gcloud beta auth application-default login` and follow the OAuth 2.0 flow to login
4. Clone https://github.com/GoogleCloudPlatform/nodejs-docs-samples.git
5. Run `cd nodejs-docs-samples/storage`
6. Run `npm install`
7. Run `node buckets.js list`

Observe the error:

```
Error: Sorry, we cannot connect to Google Cloud Services without a project ID. You may
specify one with an environment variable named "GCLOUD_PROJECT". See
https://googlecloudplatform.github.io/google-cloud-node/#//docs/guides/authentication
for a detailed guide on creating an authenticated connection.
```
1. Modify `buckets.js` to explicitly pass in a project ID
2. Run `node buckets.js list`
3. It works
4. Undo your changes to `buckets.js`
5. Set the `GCLOUD_PROJECT` env var to your project ID
6. Run `node buckets.js list`
7. It works
### Expected
1. Open terminal
2. Run `cd nodejs-docs-samples/storage`
3. Run `gcloud beta auth application-default login`
4. Run `node buckets.js list` (No env vars are set, code does not pass project ID to client)
5. It works

stephenplusplus:
Fixes #1653
#### To Dos
- [x] https://github.com/google/google-auth-library-nodejs/issues/98
  - [x] followed by: google-auto-auth update
- [x] https://github.com/google/google-auth-library-nodejs/pull/101
  - [x] followed by: google-auth-library update
- [x] https://github.com/stephenplusplus/google-auto-auth/issues/7
  - [x] follow by: google-auto-auth update
- [x] Recurse gRPC reqOpts
- [x] Docs
- [x] Tests

This PR assumes `google-auth-library` will auto-detect a projectId when available and make it available as a property on the authClient instance: `authClient.projectId`.

This PR makes it so we won't require a projectId to be specified at the time of instantiation of a class:

``` js
// Previously:
var bigquery = require('@google-cloud/bigquery')({ no projectId provided })
// throws error
```

Instead, it will internally default the `projectId` to `{{projectId}}`. This token is then looked for when a request is made, and replaced with the value we get from the google-auth-library authClient.

In the case a projectId was not found by google-auth-library, the user's callback will be executed with the same error that we previously threw.

This change is complex due to the nature of async vs. sync expectations. Previously, we were able to do everything synchronously at the time of instantiation. Requiring the project ID made making API requests simple, because it's frequently required as part of the URI that we send requests to. In the case of gRPC, it's commonly used in the request option/arguments to a service.


packages/common/src/util.js
```diff
@@ -631,3 +638,27 @@ function getUserAgentFromPackageJson(packageJson) {
...
+
+    var activeConfig = fs.readFileSync(path.join(configDir, 'active_config'));
+    // "default"
+
+    var configFile = path.join(configDir, 'configurations', 'config_' + activeConfig);
+    // ~/.config/gcloud/configurations/config_default
+
+    var iniFile = fs.readFileSync(configFile, { encoding: 'utf8' });
+
+    return parseIni(iniFile).core.project;
```
stephenplusplus: Not sure if this is always under `core`. My config_default file looks like:

```ini
[core]
project = nth-circlet-705
account = sawchuk@gmail.com
```

stephenplusplus:
I think we're going to go another route: https://github.com/google/google-auth-library-nodejs/issues/97

stephenplusplus:
@callmehiphop I thought I tagged you in this for a review, but I can't find it. Could you take a dive into this one? The architecture here is difficult (see first post), so I'm very open to better solutions :)

packages/common/src/service.js
```diff
@@ -51,14 +51,13 @@ function Service(config, options) {
...
   this.makeAuthenticatedRequest = util.makeAuthenticatedRequestFactory(reqCfg);
-
   this.authClient = this.makeAuthenticatedRequest.authClient;
   this.baseUrl = config.baseUrl;
   this.getCredentials = this.makeAuthenticatedRequest.getCredentials;
   this.globalInterceptors = arrify(options.interceptors_);
   this.interceptors = [];
   this.packageJson = config.packageJson;
-  this.projectId = options.projectId;
+  this.projectId = options.projectId || '{{projectId}}';
```
callmehiphop: I could be missing parts of the picture still, but why can't we just do something like

```js
this.projectId = options.projectId || this.authClient.projectId;
```

And then handle the error as we did previously?

stephenplusplus: Getting the projectId is async.

callmehiphop: Got it, so then my next question would be if `grpc` services offer `interceptors` and if so, could we not just create a globalInterceptor at the `Service` level that would replace the `{{projectId}}`?

stephenplusplus: GrpcService doesn't currently support interceptors. That should work, though.

Task list, if we go this route:

- [ ] Support request interceptors for GrpcService
- [ ] Make all request interceptors compatible with async modifications

callmehiphop: That sounds like maybe more effort than we should give this? My only real critique is that it looks like there might be some unnecessary duplication and I think it would be nice if we could consolidate this logic somehow.

stephenplusplus:
@callmehiphop please take another look before I get into the tests.

I've updated the docs, which is a good place to start for the review, since it will describe the expectations of this PR. However, we still need a release of `google-auth-library` which supports automatic project ID detection in DAC environments. Our code won't have to change to compensate.

Currently, the only way to see that this PR works is by providing a JSON key file as the env var `GCLOUD_TESTS_KEY` and unsetting the `process.env.GCLOUD_TESTS_PROJECT_ID` env var.

packages/common/src/grpc-service.js
```diff
@@ -351,6 +358,13 @@ GrpcService.prototype.requestStream = function(protoOpts, reqOpts) {
     shouldRetryFn: GrpcService.shouldRetryRequest_,
 
     request: function() {
+      try {
+        reqOpts = util.decorateRequest(reqOpts, { projectId: self.projectId });
+      } catch (e) {
+        stream.destroy(e);
```
callmehiphop: Should we be destroying within a `setImmediate` and returning the stream here like in https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1656/files#diff-78d3ba232fe7028b5068f61d6d20a988R439?

callmehiphop: Yep!

stephenplusplus:
@callmehiphop Tests have been added-- PTAL!

callmehiphop:
Can we update the google-cloud system tests to conditionally run if and when the environment variables are set?

stephenplusplus:
Good point. I ended up removing the `system-test` directory altogether. The service's system tests themselves will fail as they should when the environment lacks an appropriate level of authentication.

PR #1661, opened 2016-10-02
streamline auth - normalize `project_id` to `projectId`

nw:
This patch helps to streamline authentication with google-cloud modules. 

Example:

``` js
const bq = require('@google-cloud/bigquery')(require('./gcloud_key.json'))
```

`gcloud_key.json` is a non modified exported key from IAM service accounts.

This patch normalizes the name of `project_id` found in the keyfile to `projectId` to work with google-auto-auth library.

### DIFF
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -541,7 +541,7 @@ function normalizeArguments(globalContext, localConfig, options) {
 
   var config = util.extendGlobalConfig(globalConfig, localConfig);
 
-  if(!config.projectId && config.project_id) {
+  if (!config.projectId && config.project_id) {
     config.projectId = config.project_id;
   }
 
--- a/packages/common/test/util.js
+++ b/packages/common/test/util.js
@@ -1215,21 +1215,21 @@ describe('common/util', function() {
       config = util.normalizeArguments(fakeContext, local);
       assert.strictEqual(config, fakeContext.config_);
     });
-    
+
     describe('project_idNormalized', function() {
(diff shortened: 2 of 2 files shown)


stephenplusplus:
Thank you for this! It's very interestingly-timed, because we're working on getting the projectId in a similar way with one of our dependencies: https://github.com/google/google-auth-library-nodejs/pull/98. I think we should wait for that, since it will work for all of the various ways we accept credentials-- keyfile, environment variable, and a credentials object-- and we won't have to special-case this scenario.

```js
const bq = require('@google-cloud/bigquery')(require('./gcloud_key.json'))
```

This confuses me a bit because when a user provides a parsed key file, we accept it under the `credentials` key like this:

```js
const bq = require('@google-cloud/bigquery')({
  credentials: require('./keyfile.json')
})
```

Was your example a typo, or does it work without using the `credentials` key as well?

nw:
@stephenplusplus Yeah I saw the discussion about `google-auth` just after putting this PR.

The example is correct. The json key files have all the credentials needed with the exception of normalizing the projectId.

Maybe this example is clearer

```js
const creds = require('./keyfile.json')
const bq = require('@google-cloud/bigquery')(creds)
```

stephenplusplus:
When I try that, I get:

```
Error: Could not load the default credentials. Browse to https://developers.google.com/accounts/docs/application-default-credentials for more information.
    at /Users/stephen/dev/play/gcloud-1475518276/node_modules/google-auth-library/lib/auth/googleauth.js:145:21
```

PR #1663, opened 2016-10-03
storage: support rewrite operation

### LINKED ISSUE #1373: Add support for File rewrite operations.
This appears to be new! [Upstream docs](https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite). This operation also support [encryption key headers](https://cloud.google.com/storage/docs/encryption#customer-supplied).

stephenplusplus:
Fixes #1373

Compare https://cloud.google.com/storage/docs/json_api/v1/objects/copy and https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite. It sounds like `rewrite` is basically `copy`, except with some added capabilities. I kept our method named `copy`, but we only use `rewrite` in this PR. LMKWYT.

### DIFF
--- a/packages/storage/test/file.js
+++ b/packages/storage/test/file.js
@@ -369,7 +369,7 @@ describe('File', function() {
       it('should make the subsequent correct API request', function(done) {
         var newFile = new File(BUCKET, 'new-file');
 
-        file.request = function(reqOpts, callback) {
+        file.request = function(reqOpts) {
           assert.strictEqual(reqOpts.qs.rewriteToken, apiResponse.rewriteToken);
           done();
         };


stephenplusplus:
@capstan is it safe to always send our users' "copy" requests to the rewrite API endpoint? It seems like it accepts more types of requests, but will ultimately do what the copy API endpoint does. Let me know if that's wrong and we need to support both endpoints individually. Thanks!

// @omaray

Capstan:
Yes, it is always safe to send Copy requests to the Rewrite API endpoint. It will always handle cross-location, cross-storage class requests, whereas Copy may time out.

stephenplusplus:
Cool, thanks!

packages/storage/src/file.js
```diff
@@ -269,6 +269,12 @@ function File(bucket, name, options) {
 
 util.inherits(File, common.ServiceObject);
 
+/*! Developer Documentation
+ *
+ * @param {object=} options - Configuration object.
```
callmehiphop: Just to clarify, we're not publicly documented options because the user will never manually make the subsequent requests for large files?

stephenplusplus: Yep yep, we're hiding that and just doing it for 'em.

PR #1666, opened 2016-10-03
logging: change resource argument to metadata

### LINKED ISSUE #1348: logging: Public API for setting timestamp, severity, etc. on Entry
I can't find a public API for setting the [metadata fields](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/logging/entry.js#L124-L133) like timestamp and severity on a log entry.

It works to do `var entry = log.entry(...); entry.severity = "INFO";`, which is similar to what's done internally in [`Log.assignSeverityToEntries_`](https://github.com/GoogleCloudPlatform/gcloud-node/blob/master/lib/logging/log.js#L113), but I think there should be an options object argument in the Entry constructor.

stephenplusplus:
Fixes #1348
#### Before

``` js
var resource = {
  type: 'gce_instance',
  labels: {
    zone: 'global',
    instance_id: '3'
  }
}

var entry = log.entry(resource, data)
entry.otherMetadataProperty = '...' // <- not documented this way, because it wasn't intended to be done this way
```
#### After

``` js
var metadata = {
  otherMetadataProperty: '...', // <- logical place for this
  resource: {
    type: 'gce_instance',
    labels: {
      zone: 'global',
      instance_id: '3'
    }
  }
}

var entry = log.entry(metadata, data)
```

### DIFF
--- a/packages/logging/system-test/logging.js
+++ b/packages/logging/system-test/logging.js
@@ -20,6 +20,7 @@ var assert = require('assert');
 var async = require('async');
 var BigQuery = require('@google-cloud/bigquery');
 var exec = require('methmeth');
+var extend = require('extend');
 var format = require('string-format-obj');
 var is = require('is');
 var prop = require('propprop');
@@ -368,6 +369,35 @@ describe('Logging', function() {
               matchUser: logEntry.data.matchUser.toString(),
               matchUserError: logEntry.data.matchUserError.toString()
             });
+
+            done();
+          });
+        }, WRITE_CONSISTENCY_DELAY_MS);
+      });
+    });
+
+    it('should write a log with metadata', function(done) {
+      var metadata = extend({}, options, {
+        severity: 'DEBUG'
+      });
+
+      var data = {
+        embeddedData: true
+      };
+
+      var logEntry = log.entry(metadata, data);
+
+      log.write(logEntry, function(err) {
+        assert.ifError(err);
+
+        setTimeout(function() {
+          log.getEntries({ pageSize: 1 }, function(err, entries) {
+            assert.ifError(err);
+
+            var entry = entries[0];
+
+            assert.strictEqual(entry.metadata.severity, metadata.severity);
+            assert.deepEqual(entry.data, data);
 
             done();
           });


stephenplusplus:
@zbjornson mind taking a quick look? Thanks!

packages/logging/src/entry.js
```diff
@@ -32,7 +32,7 @@ var is = require('is');
  * @alias module:logging/entry
  * @constructor
  *
- * @param {object=|string=} resource - See a
+ * @param {object=} metadata - See a
  *     [Monitored Resource](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/MonitoredResource).
```
zbjornson: Is this the correct URL still? Might be this instead/in addition:
https://cloud.google.com/logging/docs/api/ref/rest/v1beta3/projects.logs.entries/write#LogEntryMetadata

packages/logging/src/log.js
```diff
@@ -206,22 +208,24 @@ Log.prototype.emergency = function(entry, options, callback) {
  *
  * @resource [LogEntry JSON representation]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/LogEntry}
  *
- * @param {object=|string=} resource - See a
+ * @param {object=} metadata - See a
  *     [Monitored Resource](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/MonitoredResource).
```
zbjornson: Ditto

zbjornson:
Don't have enough time to try it out at the moment, but the API and quick glance at the code look good. Thanks! Let me know if you want me to test before merging, but it looks straightforward enough.

packages/logging/src/entry.js
zbjornson: Edit: that was the old version.
https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/LogEntry for v2beta1

stephenplusplus: Thank you!

callmehiphop:
Can we add some system tests for this?

stephenplusplus:
Yes, good idea. Added!

zbjornson:
Thanks for the fix.

Do you know what version this will come out in -- 0.42.0 or 0.41.3? It's a breaking change and I need to [update a dependency accordingly](https://github.com/mlazarov/bunyan-stackdriver/issues/6).

stephenplusplus:
RE: the comment on https://github.com/mlazarov/bunyan-stackdriver/pull/6, we do follow semver. It'll be 0.42.0, although we're backed up for a while (3-7 days) trying to get promises in. It'll also release independently in `@google-cloud/logging@1.0.0-beta`.

zbjornson:
Thanks.

> we do follow semver

Er, right -- I guess because it's pre-1.0.0 breaking changes are allowed. :)

stephenplusplus:
Pre-1.0, a minor bump signals a breaking change. `npm install --save google-cloud` saves the value `^0.41.0` in package.json, which is `>=0.41.0 && < 0.42.0`. Compared to if we were post 1.0, `^1.41.0` would mean `>= 1.41.0 && < 2.0.0`.

PR #1665, opened 2016-10-03
core: created streaming methods

callmehiphop:
This PR moves all streaming APIs to their own methods.

e.g.

`bigquery.getJobs()` -> `bigquery.getJobsStream()`
#### TODOS
- [x] rename methods
- [x] docs
- [x] linting
- [x] unit tests
- [x] system tests

---
### Changes

All callback style methods are left intact, however as mentioned earlier, all streaming methods have been moved to their own methods. Here is a map of where streaming functionality has moved to.
#### BigQuery
- `BigQuery#query` -> `BigQuery#createQueryStream`
- `BigQuery#getDatasets` -> `BigQuery#getDatasetsStream`
- `BigQuery#getJobs` -> `BigQuery#getJobsStream`
- `Dataset#query` -> `Dataset#createQueryStream`
- `Dataset#getTables` -> `Dataset#getTablesStream`
- `Job#getQueryResults` -> `Job#getQueryResultsStream`
- `Table#query` -> `Table#createQueryStream`
- `Table#getRows` -> `Table#createReadStream`
#### Bigtable
- `Bigtable#getInstances` -> `Bigtable#getInstancesStream`
- `Instance#getClusters` -> `Instance#getClustersStream`
- `Instance#getTables` -> `Instance#getTablesStream`
- `Table#getRows` -> `Table#createReadStream`
- `Table#sampleRowKeys` -> `Table#sampleRowKeysStream`
#### Compute
- `Compute#getAddresses` -> `Compute#getAddressesStream`
- `Compute#getAutoscalers` -> `Compute#getAutoscalersStream`
- `Compute#getDisks` -> `Compute#getDisksStream`
- `Compute#getInstanceGroups` -> `Compute#getInstanceGroupsStream`
- `Compute#getFirewalls` -> `Compute#getFirewallsStream`
- `Compute#getHealthChecks` -> `Compute#getHealthChecksStream`
- `Compute#getMachineTypes` -> `Compute#getMachineTypesStream`
- `Compute#getNetworks` -> `Compute#getNetworksStream`
- `Compute#getOperations` -> `Compute#getOperationsStream`
- `Compute#getRegions` -> `Compute#getRegionsStream`
- `Compute#getRules` -> `Compute#getRulesStream`
- `Compute#getServices` -> `Compute#getServicesStream`
- `Compute#getSnapshots` -> `Compute#getSnapshotsStream`
- `Compute#getSubnetworks` -> `Compute#getSubnetworksStream`
[...]

Issue #1306, opened 2016-05-11
Return Jobs/Operations directly from the method that starts them


stephenplusplus:
RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1294#discussion_r62872247

## Blocked
Let's wait for [our support of promises](https://github.com/GoogleCloudPlatform/gcloud-node/pull/1142) to engineer this.

--

Before:
```js
vm.delete(function(err, operation, apiResponse) {
  operation
    .on('error', function(err) {})
    .on('running', function(operationMetadata) {})
    .on('complete', function(operationMetadata) {});
});
```

After:
```js
vm.delete()
  .on('error', function(err) {})
  .on('running', function(apiResponse) {}) // is this what you had in mind, or is it the operationMetadata?
  .on('complete', function(operationMetadata) {});
```

@callmehiphop What should we do in these situations (regarding the second argument, `vm`):
```js
compute.createVM('vm-name', function(err, vm, operation, apiResponse) {});
```

Emit `vm` with `running`?
```js
compute.createVM('vm-name')
  .on('error', function(err) {})
  .on('running', function(vm, apiResponse) {})
  .on('complete', function(operationMetadata) {});
```

... or is that getting weird?

callmehiphop:
I think if the operation/job was responsible for creating something (like a VM) that the created VM would be emitted via `complete`. 

I forgot about the previously existing `running` event, perhaps `created` (or similar) would be better suited.

```js
compute.createVM('vm-name')
  .on('error', function(err) {})
  .on('created', function(apiResponse) {})
  .on('running', function(operationMetadata) {})
  .on('complete', function(vm, operationMetadata) {});
```

stephenplusplus:
`created` and `running` feel a little too similar, maybe we should just not return the original `apiResponse`. If an error occurred with the `createVM` API call, then the apiResponse is already included there. The initial state of the operation shouldn't be vital... but it theoretically should be available from `operation.metadata` at some point:

```js
var operation = compute.createVM('vm-name')
  .on('error', function(err) {}) // the "POST /vm-name" call failed OR the operation failed
  .on('running', function() {}) // arg removed. `operation.metadata` can be checked for the status
  .on('complete', function(vm) {}); // took out `operationMetadata`. they can check `operation.metadata` for the status
```

callmehiphop:
[...]

PR #1670, opened 2016-10-05
core: adding promise support

callmehiphop:
~~Waiting on #1665~~

When it comes time to ship this we'll have to ship the code out in phases. To test this locally you'll need to run

``` sh
$ npm run link-common
```

To run all the system tests, additionally links may need to be made where modules have `devDependencies` on modules like Storage/PubSub/etc.

### DIFF
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -656,7 +656,9 @@ function promisify(originalMethod) {
       return originalMethod.apply(context, args);
     }
 
-    return new (PromiseOverride || Promise)(function(resolve, reject) {
+    var PromiseCtor = PromiseOverride || Promise;
+
+    return new PromiseCtor(function(resolve, reject) {
       args.push(function() {
         var callbackArgs = slice.call(arguments);
         var err = callbackArgs.shift();
--- a/packages/common/test/service.js
+++ b/packages/common/test/service.js
@@ -150,7 +150,7 @@ describe('Service', function() {
       assert.strictEqual(service.projectIdRequired, true);
     });
 
-    it('should call util.setPromiseOverride when promise is set', function(done) {
+    it('should call setPromiseOverride when promise is set', function(done) {
       var FakePromise = function() {};
 
       util.setPromiseOverride = function(override) {
@@ -158,7 +158,7 @@ describe('Service', function() {
         done();
       };
 
-      var service = new Service({}, { promise: FakePromise });
+      new Service({}, { promise: FakePromise });
     });
   });
 
--- a/packages/google-cloud/src/index.js
+++ b/packages/google-cloud/src/index.js
@@ -406,6 +406,8 @@ var apis = {
  *     true)
  * @param {number} options.maxRetries - Maximum number of automatic retries
  *     attempted before returning the error. (default: 3)
+ * @param {function} options.promise - Custom promise module to use instead of
+ *     native Promises.
  *
  * @example
  * var gcloud = require('google-cloud')({


stephenplusplus:
@callmehiphop should we close #1142?

callmehiphop:
@stephenplusplus I think this is the cleanest route to promisifiy our library. PTAL a let me know what you think about this basic implementation.

Assuming there are no red flags with the implementation, my next question would be in regards to system tests - do we want to write a system test for every method that can leverage promises? I think this would almost double the number of system tests.

callmehiphop:
> should we close #1142?

Actually this PR is requesting to merge into that branch :)

stephenplusplus:
Ah, fancy.

I think this implementation looks good. For system tests, I would say we don't have to system test, but we do need to smoke test / docs test. I think we're already planning on showing "@&shy;examples" that use both callbacks and promises, right?

callmehiphop:
> I think we're already planning on showing "@­examples" that use both callbacks and promises, right?

Sure do, I plan to handle them in the same fashion as we used to document streams.

stephenplusplus:
That should be good enough for me then to show that our promises work.

packages/bigquery/src/index.js
```diff
@@ -175,6 +184,14 @@ BigQuery.prototype.dataset = function(id) {
  * }, callback);
  *
  * //-
+ * // If the callback is omitted, we'll return a Promise.
+ * //-
+ * bigquery.getDatasets()
```
callmehiphop: This is an example of how the docs will look when a single parameter is given back to the callback.

callmehiphop:
@stephenplusplus I've updated a few of the examples, WDYT?

How do you want me to organize all the things? First commit is `util.promisify()` then each commit after is a single API? Or should I put the common stuff in a separate PR?

stephenplusplus:
> How do you want me to organize all the things? First commit is util.promisify() then each commit after is a single API? Or should I put the common stuff in a separate PR?

Whatever is easiest for your workflow.

callmehiphop:
> I wonder if we can always return a single argument and make apiResponse accessible in a different way

There will still be scenarios where we have to return multiple arguments (`autoPaginate: false`). However we could either make the api response the context of the callback or we could make it a property of whatever we pass in? I think this is how we planned on handling errors.

```js
bigquery.getDatasets()
  .then(function(datasets) {
    console.log(datasets.apiResponse);
  })
  .catch(function(err) {
    console.log(err.apiResponse);
  });
```

WDYT?

stephenplusplus:
`err.apiResponse` is okay. I think we already have `response` available on ApiError objects, so we should stick with that for custom errors.

I don't want to collide any extra properties like `apiResponse` with the returned argument, as that might conflict with other data / unintentionally be logged/printed/looped over by the user.

Keeping in mind this is describing the less common use case of manual pagination & needing the apiResponse, I think this would be the best separation.

```js
bigquery
  .getDatasets({ autoPaginate: false })
  .then(function(datasets) {
    var apiResponse = this.request.response
    var nextQuery = this.request.nextQuery
    
    if (nextQuery) {
      return bigquery.getDatasets(nextQuery)
    }
  })
```

So we're fighting between good UX for the common case or good UX for the uncommon. If multiple args combined in one object is completely fine and expected in an average promise workflow, then just ignore me on all of this. That does bother me, because "getDatasets()" sounds like it's going to return "datasets", not an object that has datasets. But the other bother is that we'll have some methods that return single and others that return multiple.

callmehiphop:
> If multiple args combined in one object is completely fine and expected in an average promise workflow, then just ignore me on all of this.

I think it is pretty common to take that approach, however we had [previously decided](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/551#issuecomment-189446956) that using Arrays + destructuring would be the way to go. Then in the event that a user is using Bluebird in some fashion, they can take advantage of the `spread()` method.

```js
bigquery.getDatasets()
  .then(function([datasets, apiResponse]) {});

bigquery.getDatasets()
  .spread(function(datasets, apiResponse) {});
```

The only downside of this of course is that `0.12.x` users still have to access items via index, however I'm not sure if object notation is really much better.

stephenplusplus:
Okay, that makes sense. I forgot about destructuring making the lives of devs on modern stacks easier. It looks like we didn't get votes on arrays vs objects in that linked post. Can you show what the spread workflow looks like for someone using Bluebird? Also, why would they be using Bluebird in the first place-- does it do something we aren't doing with our promise support that is required to conform to a promise spec-- or is it just for features like spread that aren't supported by native promises?

callmehiphop:
>  Can you show what the spread workflow looks like for someone using Bluebird

All spread does is transform an Array returned from an earlier call it separate parameters, very similarly to the spread operator. So assuming a user makes Bluebird the global Promise object (I don't think that's very uncommon).

```js
global.Promise = require('bluebird');

bigquery.getDatasets({ autoPaginate: false })
  .spread(function(datasets, nextQuery, apiResponse) {});
```

however, even if they don't re-assign the Promise object, it's still plausible they'll use some of their convenience methods.

```js
Bluebird.delay(1000)
  .then(function() {
    return bigquery.getDatasets({ autoPaginate: false });
  })
  .spread(function(datasets, nextQuery, apiResponse) {
    // ...
  });
```

> Also, why would they be using Bluebird in the first place-- does it do something we aren't doing with our promise support that is required to conform to a promise spec-- or is it just for features like spread that aren't supported by native promises?

I would guess the latter. It's also possible that they aren't using Bluebird with our library specifically, but perhaps with another as it offers `promisify()` methods of it's own.

stephenplusplus:
Ah, okay. So would you lean towards using the array syntax instead of the boxed in object so that Bluebird users are happy?

callmehiphop:
> So would you lean towards using the array syntax instead of the boxed in object so that Bluebird users are happy?

I think that's fair to say. If you think we're better off with the object syntax, I think I'll be pretty easily swayed. It's still destructurable and perhaps slightly better for users who can't take advantage of that feature yet.

However I think this is going to be a bit trickier to implement because we'll need to map each returned item to a specific key.

stephenplusplus:
> we'll need to map each returned item to a specific key.

Can you elaborate on that? I'm not sure I follow.

callmehiphop:
> Can you elaborate on that? I'm not sure I follow.

Certainly! I think what I'm trying to say is that using an object format will require more code changes. Our methods would probably have to resemble something like this..

```js
BigQuery.prototype.getDatasets = function(query, callback) {
  var that = this;

  if (is.fn(query)) {
    callback = query;
    query = {};
  }

  query = query || {};

  this.request({
    uri: '/datasets',
    qs: query
  }, function(err, resp) {
    var response = {
      apiResponse: resp
    };

    if (err) {
      callback(err, response);
      return;
    }

    response.nextQuery = null;

    if (resp.nextPageToken) {
      response.nextQuery = extend({}, query, {
        pageToken: resp.nextPageToken
      });
    }

    response.datasets = (resp.datasets || []).map(function(dataset) {
      var ds = that.dataset(dataset.datasetReference.datasetId);
      ds.metadata = dataset;
      return ds;
    });

    callback(null, response);
  });
};
```

Because we have no way of mapping objects to an appropriate key (`datasets`, `nextQuery`) otherwise.

stephenplusplus:
Oh, right. Thanks :) Arrays it is! Seems to benefit the most amount of people.

callmehiphop:
I've been looking through a lot of promisify libraries, seeing if any of them suited our needs and a common theme among them is [specifying whether or not you want multiple arguments](https://github.com/sindresorhus/pify/blob/master/index.js#L16).

I think we can all agree that for older Node users who do not use some kind of `Promise#spread` convenience code, returning an Array is annoying, which makes me a little trepidatious to do so.

I think in majority of cases, the user is generally only going to care about the first argument. On rare occasion they may wish to control pagination themselves, but otherwise I can't really think of a scenario when they'll want multiple args. That being said, I think we should offer a similar parameter within our methods that would enhance the user experience.

```js
bigquery.createDataset('my-dataset')
  .then(function(dataset) {});

bigquery.createDataset('my-dataset', { multiArgs: true })
  .then(function([dataset, apiResponse]) {});
```

In most cases we assign the `apiResponse` as the `metadata` field on the objects being returned, so I can't really think of a single scenario where the user would actually want to do this.

In the cases of paginated lists, we could treat `autoPaginate: false` in the same fashion.

```js
bigquery.getDatasets()
  .then(function(datasets) {});

bigquery.getDatasets({ autoPaginate: false })
  .then(function([datasets, nextQuery, apiResponse]) {});
```

The only place this doesn't feel quite right is for APIs that return operations. However, I think those APIs may be changing sometime in future to return operations directly from a method as opposed to from within a callback. (https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1306)

Thoughts?

/cc @jgeewax @omaray @jmdobry

stephenplusplus:
> In most cases we assign the apiResponse as the metadata field on the objects being returned, so I can't really think of a single scenario where the user would actually want to do this.

There are API calls like "query", "detectText", etc. where `apiResponse` is very useful to the power users because we "shut out" or simplify certain bits of the raw API response. Some users need that other stuff, which is why we have `apiResponse` at all. Historically, we implemented it after receiving "Hey, where's the ___ property?" then a month later "Okay, what about the ___ property?" It was easiest just to return the whole thing, untouched.

I'm 100% with `multiArgs` as an option and `autoPaginate: false === multiArgs: true`. I think it would be *amazing* if we figured out how to handle operations at the same time we're landing promise support. I think what we really want is to eliminate operations altogether and just execute the `then` callbacks when the operation is done.

jmdobry:
We're going to make the event emitter a property of the promise right? In that case we need to return the promise regardless of whether the user provides a callback, because a user who passes callbacks might still want to access the event emitter, e.g. to cancel the request.

stephenplusplus:
That's definitely relevant, because we've really just been thinking of the non-gapic APIs. I'm not sure what @callmehiphop thinks, but if we have to do something goofy with the UX just to return the event emitter, I'm still okay with not offering the EE through the handwritten layer.

jmdobry:
I don't think we'd have to do anything goofy. If I recall the plan is for @jmuk to change `gax-nodejs` to return the promise with the event emitter attached.

sindresorhus:
I don't think a `multiArgs` option here is a good user experience. You should be consistent with what you return. If `apiResponse` can sometimes be useful, then it should always be returned. If you don't like returning an array of arguments, you could always return an object, although I would strongly urge you to optimize for destructuring, many consumers are already on Node.js 6 and can take advantage of it. In reality, most consumers will be using Bluebird anyways, so they'll have `Promise#spread` even though they don't have destructuring.

Another approach could be to have a dedicated API to get the `apiResponse`. You could either embed the `apiResponse` in the `dataset` with a Symbol:

```js
bigquery.getApiResponse = x => x[Symbol.ApiResponse];

bigquery.getApiResponse(dataset);
//=> the API response
```

Or use a Weak Map (probably best solution):

```js
// the key is the `dataset` itself
bigquery.getApiResponse = x => bigquery._someApiResponseWeakMap.get(dataset);

bigquery.getApiResponse(dataset);
//=> the API response
```

jmdobry:
> If `apiResponse` can sometimes be useful, then it should always be returned.

Agreed 👍 

My vote is for:

```js
// callback, this will continue to work for current users
bigquery.getDatasets({ autoPaginate: false }, (datasets, nextQuery, apiResponse) => { 
  // ...
});

// destructuring, we don't have promises yet, so it's okay
// for us to say "this is how it will work when we do"
bigquery.getDatasets({ autoPaginate: false })
  .then(([datasets, nextQuery, apiResponse]) => {
    // ...
  });

// spread, thanks Bluebird! (unless we want to just polyfill Promise.prototype.spread)
bigquery.getDatasets({ autoPaginate: false })
  .spread((datasets, nextQuery, apiResponse) => {
    // ...
  });
```

My vote is also for always returning the promise so the user can access the event emitter, regardless of whether the user passed a callback or not. In `gax-nodejs`, the promise is always  returned. If the user provides a callback, then it is called _in addition_ to the promise being resolved/rejected. Once @jmuk updates `gax-nodejs` to return the promise with the event emitter attached, handwritten portions of this library will easily be able to just forward that promise to the user.

stephenplusplus:
Sounds good to me. It sounds always returning an array is the conventional approach.

#### options.promise
Should we accept a "promise" option, where a user provides their Promise library for us to use internally? Or do we just use `global.Promise`, and the user just wraps our library with Bluebird?

> My vote is also for always returning the promise so the user can access the event emitter, regardless of whether the user passed a callback or not.

This only applies to the APIs with a generated library, but will introduce inconsistencies with the others, e.g. BigQuery, Storage.

@callmehiphop how would this work behind the scenes, i.e., how many callbacks get executed / is there a "winner"?:

```js
pubsub.getTopics(function(err, topics) {})
  .then(function(topics) {})
  .catch(function(err) {});
```

As I understand it, these are the features the EE provides:

- Listening to the `response` event
  - This should be the same thing we return as `apiResponse`

- Canceling a request
  - We don't have this currently and I don't expect the majority of our users needing to cancel their requests. It sounds like promises themselves will be cancelable someday (only going on what I've heard). At that time, I imagine it will be possible to wait for the cancel event from the promise and send it up to the EE's `cancel` function.

I know this is getting old, but I'm still not convinced that we need to provide the EE through the handwritten layer *at all*, even in promise mode. It's a useful feature to a subset of the users, and it will be available if they use the generated library directly, which we provide. It's causing a bit too many downsides for what I think it's worth:

- We have to return a promise when a user is providing a callback (expected: `undefined`)
- We will have inconsistencies between generated libraries and HTTP APIs
- The EE has to be "caught" from the returned promise before the promise is chained. This isn't a pattern I've seen or used before. It feels like we're trying to squeeze it in somewhere, which we are. This would require good documentation, which is another liability: a repeated message in every method's docs where we return a promise through a generated library

Regarding our API being different than gax's; I think gax should provide all the features it can, as it is positioned as a generated, low-level dependency as opposed to an opinionated one, such as google-cloud-node. It's up to consumers of gax (in this case, us) to determine where the value is for its users. What we think is important might not be the same for another library. I think it's okay to be different, given this architecture.

#### callback -> promise vs. promise -> callback 

@callmehiphop in the generated libraries, we'll eventually be interacting with promises through gax, as opposed to collecting results with `service.request()` method calls. Will we need a method like "callbackify()" to match "promisify()" for these methods?

jmdobry:
> Should we accept a "promise" option, where a user provides their Promise library for us to use internally?

SGTM

> how would this work behind the scenes, i.e., how many callbacks get executed / is there a "winner"?:

At least in `gax-nodejs`, the promise always gets resolved or rejected. If the user provides a callback, it will be called as `callback(null, response)` _in addition_ to the promise being resolved, or as `callback(err)` _in addition_ to the promise being rejected. The promise gets resolved or rejected either way. If a user decides to pass a callback _and_ consume the promise with `.then` or `.catch`, then the callback function will be called first due to the forced asynchronous nature of the promise.

This is how Bluebird's `.asCallback/.nodeify` methods work (used for creating dual promise-callback APIs, see the [Bluebird author's recommendation](http://stackoverflow.com/a/23512411)): the promise is always returned, and always resolved or rejected. If a callback is provided it will be called _in addition_ to the promise being resolved/rejected. I used this very pattern (using Bluebird.js) nearly 3 years ago in an open-source Node.js library of mine. I don't think we'd going against the grain if we use that pattern here. A consistent return value is less surprising than a conditionally present return value.

Setting aside whether we want to user to be able to access the event emitter from the handwritten layer, I think always returning a consistent return value (the promise) has merit on its own. The promise becomes the expected API, the callback an optional fallback.

stephenplusplus:
> I think always returning a consistent return value (the promise) has merit on its own. The promise becomes the expected API, the callback an optional fallback.

I do like how simple that sounds. I definitely want to do what is conventional, avoiding having any quirks or extra things that need to be learned with our API. If it's normal to return a promise from a method you're invoking with a callback, then we should do that. I'd still welcome feedback from others on if they're familiar with that pattern as well.

jonparrott:
>  A consistent return value is less surprising than a conditionally present return value.

I agree with this, principle of least surprise applies here.

JustinBeckwith:
A big +1 to having a consistent return value here.  Mutating the return value type based on passed parameters would be pretty surprising to me.  I'm *really* excited to see promise support land in this API folks.

callmehiphop:
> Will we need a method like "callbackify()" to match "promisify()" for these methods?

@stephenplusplus Yes, that would be correct. I was looking into making promises the normal way we handle things but it would require a significantly larger time investment since we would need to refactor every single method, which in turn would require us to refactor every single system test, unit test and doc snippet. I'm trepidatious to make such a large volume of changes if we're going to eventually move everything over to gapic anyways. IMO we should just hold out on refactoring all the things and instead tackle one API at a time while making the switch to gapic.

WDYT?

stephenplusplus:
> I was looking into making promises the normal way we handle things but it would require a significantly larger time investment since we would need to refactor every single method, which in turn would require us to refactor every single system test, unit test and doc snippet.

Yeah, it's basically a whole new library at that point. Maybe it'll come down to a managerial decision on where our effort should be spent, but I still want our library to be the easiest to understand for future maintainers / contributors. So I'm +1 to supporting promises the easiest way now, then refactoring later... as long as we actually refactor later.

> if we're going to eventually move everything over to gapic anyways.

- [ ] Google BigQuery
- [x] Google Cloud Bigtable
- [x] Google Cloud Datastore
- [ ] Google Cloud DNS
- [x] Google Cloud Pub/Sub
- [ ] Google Cloud Storage
- [ ] Google Compute Engine
- [ ] Google Prediction API
- [ ] Google Translate API
- [x] Google Cloud Natural Language (Beta)
- [ ] Google Cloud Resource Manager (Beta)
- [x] Google Cloud Speech (Beta)
- [x] Google Cloud Vision (Beta)
- [x] Stackdriver Logging (Beta)

Only half of our APIs are gapic-compatible. Just wanted to call that out, so we still show love to the other half of the codebase ❤️

packages/bigquery/src/dataset.js
```diff
@@ -414,4 +470,15 @@ Dataset.prototype.table = function(id) {
  */
 common.paginator.extend(Dataset, ['getTables']);
 
+/*! Developer Documentation
+ *
+ * All async methods (except for streams) will return a Promise in the event
+ * that a callback is omitted.
+ */
+common.util.promisify(Dataset, {
+  filter: function(methodName) {
```
stephenplusplus: How about just:

```js
ignore: ['table']
```

packages/bigquery/src/index.js
```diff
@@ -90,6 +90,15 @@ util.inherits(BigQuery, common.Service);
  *
  * @example
  * bigquery.createDataset('my-dataset', function(err, dataset, apiResponse) {});
+ *
+ * //-
+ * // If the callback is omitted, we'll return a Promise.
+ * //-
+ * bigquery.createDataset('my-dataset')
```
stephenplusplus: Is there a system for when `.then` goes on a new line? I think we should always put it on a new line for consistency.

packages/bigquery/src/dataset.js
```diff
@@ -414,4 +470,15 @@ Dataset.prototype.table = function(id) {
  */
 common.paginator.extend(Dataset, ['getTables']);
 
+/*! Developer Documentation
+ *
+ * All async methods (except for streams) will return a Promise in the event
+ * that a callback is omitted.
+ */
+common.util.promisify(Dataset, {
```
stephenplusplus: What do you think about having `Promisify` as a base class that `Service` inherits from? Not sure if it's doable without micro-configuration.

packages/bigquery/src/dataset.js
```diff
@@ -70,6 +78,11 @@ function Dataset(bigQuery, id) {
      *
      * @example
      * dataset.exists(function(err, exists) {});
+     *
+     * //-
+     * // If the callback is omitted, we'll return a Promise.
+     * //-
+     * dataset.exists().then(function(exists) {});
```
stephenplusplus: I'm unsure about the inconsistency in sometimes returning an array vs. the unboxed response. I don't know if we should needlessly arrayify single-arg responses, but then again, maybe that's expected / better for consistency? Curious on your thoughts!

// @jmdobry

packages/bigquery/src/dataset.js
callmehiphop: Not really sure how that would work? Currently this essentially just wraps all callback style functions to conditionally use promises, pretty much in the same exact manner as `streamRouter`. I think if this were something we extended, we'd probably have to wrap all methods on the fly everytime a class was instantiated..?

packages/bigquery/src/dataset.js
callmehiphop: Sure, I can dig it!

packages/bigquery/src/dataset.js
stephenplusplus: Ah right, perf hit indeed.

stephenplusplus:
@jgeewax @jmdobry @omaray 

Do you guys think we should show error handling in our example docs, e.g.:

```js
If the callback is omitted, we'll return a Promise.

bigquery.createDataset('my-dataset')
  .then(function(data) {
    var dataset = data[0];
    var apiResponse = data[1];
  })
  .catch(function(err) {
    // err is an API error if one occurred.
  });
```

callmehiphop:
IMO I think it's implied that if an error occurs it will be catchable via `catch(func)` or `then(null, func)`.

packages/vision/test/index.js
```diff
@@ -101,6 +108,10 @@ describe('Vision', function() {
       ]);
       assert.deepEqual(calledWith.packageJson, require('../package.json'));
     });
+
+    it('should promisify all the things', function() {
+      assert(promisified);
+    });
   });
 
   describe('constants', function() {
```
stephenplusplus: constants are fixed variables.

packages/common/src/util.js
```diff
@@ -631,3 +631,66 @@ function getUserAgentFromPackageJson(packageJson) {
 }
 
 util.getUserAgentFromPackageJson = getUserAgentFromPackageJson;
+
+/**
+ * Promisifies certain Class methods. This will not promisify private or
+ * streaming methods. By default it will promisify any method that is
+ * camel cased with the exception of `query`.
```
stephenplusplus: I think this bit on `query` is no longer accurate, unless I'm missing something in the code.

stephenplusplus:
Do settings like `maxApiCalls` and `maxRetries` still work when using promises?

packages/bigquery/src/dataset.js
jmdobry: I feel like we should choose consistency here, that way users always know what to expect and won't be surprised. Otherwise, I feel like one what constantly have to be checking the API ref docs for google-cloud-node to figure out which methods return arrays vs which ones don't.

callmehiphop:
> Do settings like maxApiCalls and maxRetries still work when using promises?

Sure do, I can add a system test or two for good measure/peace of mind.

stephenplusplus:
> Sure do, I can add a system test or two for good measure/peace of mind.

That's good news! I suppose that wouldn't hurt, since we randomly test stream functionality in places.

packages/common/src/util.js
callmehiphop: Right-o. Fixed!

callmehiphop:
@stephenplusplus I just pushed some updates.

* A few system tests (they live in BigQuery)
* Removed `filter` function in favor of `exclude` array.
* Updated logic/examples to always return arrays as opposed to single arguments.

packages/common/src/util.js
```diff
@@ -631,3 +633,90 @@ function getUserAgentFromPackageJson(packageJson) {
...
+
+util.promisifyAll = promisifyAll;
+
+/**
+ * Allows user to override the Promise constructor without the need to touch
+ * globals. Override should be ES6 Promise compliant.
+ *
+ * @param {promise} override
+ */
+function setPromiseOverride(override) {
```
stephenplusplus: I'm wondering if we need to localize this somehow, as a user might need to override a promise in one place and not another. Since `deduping` is default, this file is going to be re-used across all of the APIs.

And in a "ya never know" case, if a library wraps `google-cloud-node` and another library wraps *that* library, whoever sets an override would be affected all the way through.

callmehiphop: Yeah, that's a good point I hadn't really considered. My initial thought is to just localize it at the Service level. [WDYT](https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1689/commits/12b0fec9e76370f17d960cdf4fd154428b2cd14e)?

stephenplusplus: Looks good!

PR #1676, opened 2016-10-06
tests: check for broken links

### LINKED ISSUE #1203: Test docs for dead links
On http://googlecloudplatform.github.io/gcloud-node/#/docs/v0.30.1/datastore/query

Looks like we moved these up a level (to the datastore instance) rather than defining datasets on their own. 

Can we do two things:
1. ~~Fix the dead links?~~
2. Make sure that dead links are caught in the build?

stephenplusplus:
Fixes #1203

This is an early implementation of checking the existence of links we use in our docs.
#### To Dos
- [ ] Check all links, not just from example snippets
- [ ] Use the network (see https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1676#issuecomment-252086079)

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,9 @@
     "semver": "^5.3.0",
     "shelljs": "^0.7.3",
     "string-format-obj": "^1.0.0",
-    "through2": "^2.0.0"
+    "through2": "^2.0.0",
+    "url-exists": "^1.0.3",
+    "url-regex": "^3.2.0"
   },
   "scripts": {
     "postinstall": "node ./scripts/install.js",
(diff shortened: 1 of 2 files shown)


stephenplusplus:
@callmehiphop I'm stuck on this, which is why this PR is a little early. I'm having trouble using the network, even after `MITM.disable()` is called. I was hoping, whenever you had a chance, you could take a look and let me know if I'm doing something obviously wrong, or if you can think of a way around this.

stephenplusplus:
I'll close this, but if we ever want this in the future, this can be a reference.

PR #1682, opened 2016-10-10
datastore: use a Symbol to get the key from an entity

### LINKED ISSUE #508: Access properties directly from the entity
When modifying or accessing entity properties, consistently you have to step into the data property:
```
my_entity.data.my_property
my_entity.data.my_other_property
```

Compare this to the other gcloud datastore APIs:

Python:
```
my_entity['my_property']
```
Ruby:
```
my_entity[:my_property]
````

It would be nice if properties could be accessed on the entity directly.

### PR DESCRIPTION
**Pull request by stephenplusplus**
Fixes #508
#### Before

``` js
var key = datastore.key(['Kind', 'id'])

datastore.get(key, function (err, entity) {
  entity = {
    key: key,
    data: { ... }
  }
});
```
#### After

``` js
var key = datastore.key(['Kind', 'id'])

datastore.get(key, function (err, entity) {
  entity = {...}
})
```
##### So, how do I get the key from an entity?

``` js
var key = datastore.key(['Kind', 'id'])

datastore.get(key, function (err, entity) {
  var key = entity[datastore.KEY]
})
```

`datastore.KEY` is a [Symbol](https://www.keithcirkel.co.uk/metaprogramming-in-es6-symbols/), so it won't collide with any other property that might be stored on the entity.

### DIFF
--- a/packages/datastore/src/entity.js
+++ b/packages/datastore/src/entity.js
@@ -38,6 +38,13 @@ var InvalidKeyError = createErrorClass('InvalidKey', function(opts) {
 });
 
 /**
+ * A symbol to access the Key object from an entity object.
+ *
+ * @type {symbol}
+ */
+entity.KEY_SYMBOL = Symbol('KEY');
+
+/**
  * Build a Datastore Double object.
  *
  * @constructor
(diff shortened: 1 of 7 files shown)


callmehiphop:
@stephenplusplus are there other places in our library that could similarly benefit from Symbols? Obviously that's out of the scope of this PR, but maybe we should take a look/open an issue..?

stephenplusplus:
I think we can just stay aware of it being an option and implement it wherever it fits. No need for an issue.

PR #1688, opened 2016-10-13
Add support for Cloud Functions

jmdobry:
- [x] - GAPIC code
- [x] - Implementation
  - [x] - Functions#cloudfunction
  - [x] - Functions#createFunction
  - [x] - Functions#getFunctions
  - [x] - Functions#getFunctionsStream
  - [x] - Functions#operation
  - [x] - CloudFunction#call
  - [x] - CloudFunction#create
  - [x] - CloudFunction#delete
  - [x] - CloudFunction#exists
  - [x] - CloudFunction#get
  - [x] - CloudFunction#setMetadata
- [x] - unit tests
- [x] - system tests
- [ ] - docs

### DIFF
--- a/packages/functions/README.md
+++ b/packages/functions/README.md
@@ -1,0 +1,79 @@
+# @google-cloud/functions
+> Google Cloud Functions Client Library for Node.js
+
+> **This is a Alpha release of Google Cloud Functions.** This feature is not covered by any SLA or deprecation policy and may be subject to backward-incompatible changes.
+
+*Looking for more Google APIs than just Cloud Functions? You might want to check out [`google-cloud`][google-cloud].*
+
+- [API Documentation][gcloud-functions-docs]
+- [Official Documentation][cloud-functions-docs]
+
+
+```sh
+$ npm install --save @google-cloud/functions
+```
+```js
+var functions = require('@google-cloud/functions')({
+  projectId: 'grape-spaceship-123',
+  keyFilename: '/path/to/keyfile.json'
+});
+
+var functions = functions({
+  projectId: 'grape-spaceship-123',
+  keyFilename: '/path/to/keyfile.json'
+});
+```
+
+## Authentication
+
+It's incredibly easy to get authenticated and start using Google's APIs. You can set your credentials on a global basis as well as on a per-API basis. See each individual API section below to see how you can auth on a per-API-basis. This is useful if you want to use different accounts for different Google Cloud services.
+
+### On Google Compute Engine
+
+If you are running this client on Google Compute Engine, we handle authentication for you with no configuration. You just need to make sure that when you [set up the GCE instance][gce-how-to], you add the correct scopes for the APIs you want to access.
+
+``` js
+// Authenticating on a global basis.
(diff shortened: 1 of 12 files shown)


jmdobry:
@stephenplusplus @callmehiphop 

The Cloud Functions API `ListFunctions` method doesn't support pagination yet. I attempted to use the paginator for the `getFunctions` method, but limiting doesn't appear work. Here's [the skipped test](https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1688/files?diff=unified#diff-7d7afb526358698f6ceb0a51994a4d52R182).

The PR is done (except for docs), barring any requested changes during review.

I experimentally implemented this PR using some ES2015 features (see #1855), and set the `engines` field in `package.json` to `"node": ">=4.3.2"` (it's the oldest 4.x version shown at [node.green](http://node.green/). I imagine the `@google-cloud/functions` package will be "alpha" for a while, then "beta", and by the time it's GA google-cloud-node will have dropped support for Node.js 0.12, so perhaps it can be a pilot for #1855.

ofrobots:
@jmdobry not sure what view you're seeing, I do see more recent versions of Node.js 4.x on node.green (not that there is a difference in the list of supported language features that I am aware of between 4.3.x and 4.7.x)

jmdobry:
I see the more recent 4.x versions too, but I thought it was prudent to stay compatible with older 4.x versions.

jmdobry:
Instead of dropping support for 0.10 and 0.12 straight away in google-auth-library-nodejs, perhaps a better experiment would be to drop support for 0.10 and 0.12 in a brand new package, such as this one.

stephenplusplus:
What's the plan / status of this?

jmdobry:
Good question. cc @omaray 

## Option 1

Continue forward with this PR. There have been many changes to the google-cloud-node common libraries since I submitted this, so it's needs some work to get up-to-date. When I left off the PR was ready for review.

## Option 2

Expose the Gapic code to the user and delete the majority of my handwritten code, keeping only the helper function that zips and uploads a directory to Google Cloud Storage in preparation for creating/deploying a new Cloud Function (because it's a beast, and I'd hate to have to make users do all that themselves).

lukesneeringer:
> Instead of dropping support for 0.10 and 0.12 straight away in google-auth-library-nodejs, perhaps a better experiment would be to drop support for 0.10 and 0.12 in a brand new package, such as this one.
 :+1: 1  

Worth noting: This has been adjudicated already. 0.12 support is gone.

lukesneeringer:
> ### Option 1
>
> Continue forward with this PR. There have been many changes to the google-cloud-node common libraries since I submitted this, so it's needs some work to get up-to-date. When I left off the PR was ready for review.
>
> ### Option 2
>
> Expose the Gapic code to the user and delete the majority of my handwritten code, keeping only the helper function that zips and uploads a directory to Google Cloud Storage in preparation for creating/deploying a new Cloud Function (because it's a beast, and I'd hate to have to make users do all that themselves).

Our team is working on the concept of a "partial GAPIC", which is basically option 2. We should almost certainly do that. I should have a design doc soon and this could be a useful proof case for it.

PR #1697, opened 2016-10-17
storage: add promise support

callmehiphop:
### Breaking changes ahead! ⚠️

> This adds Promise support for the `@google-cloud/storage` module!
#### Summary of changes

The following methods previously returned a `stream` in the event that the `callback` parameter was omitted. This functionality has since moved to a different method entirely.
- `Storage#getBuckets` -> `Storage#getBucketsStream`
- `Bucket#getFiles` -> `Bucket#getFilesStream`

Each callback accepting change has also been updated to return a `Promise` instance in the event that the`callback` was omitted.

### DIFF
--- a/packages/storage/README.md
+++ b/packages/storage/README.md
@@ -53,6 +53,11 @@ localReadStream.pipe(remoteWriteStream);
 bucket.upload('/photos/zoo/zebra.jpg').then(function(data) {
   var file = data[0];
 });
+
+// It's also possible to integrate with third-party Promise libraries.
+var gcs = require('@google-cloud/storage')({
+  promise: require('bluebird')
+});
 ```
 
 


packages/storage/README.md
```diff
@@ -48,6 +48,11 @@ remoteReadStream.pipe(localWriteStream);
 var localReadStream = fs.createReadStream('/photos/zoo/zebra.jpg');
 var remoteWriteStream = bucket.file('zebra.jpg').createWriteStream();
 localReadStream.pipe(remoteWriteStream);
+
+// Promises are also supported by omitting callbacks.
+bucket.upload('/photos/zoo/zebra.jpg').then(function(data) {
+  var file = data[0];
+});
```
stephenplusplus: I'm just wondering... should we be promoting promise use first, as in "this is the way to use our library", with a mention about callbacks? @jgeewax @omaray

callmehiphop: All the JSDoc in our library promotes callbacks first and promises second, which is why I chose to do that here.

stephenplusplus: Is it possible and/or necessary for the user to understand how our API works to change the JSDocs? Since the demand for promises is so strong, I'm wondering if we might be better off changing our position to a promise library that supports callbacks instead of the other way around.

stephenplusplus: Discussion moved to: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1699

packages/storage/README.md
```diff
@@ -53,6 +53,11 @@ localReadStream.pipe(remoteWriteStream);
 bucket.upload('/photos/zoo/zebra.jpg').then(function(data) {
   var file = data[0];
 });
+
+// It's also possible to integrate with third-party Promise libraries.
```
stephenplusplus: I think we should actually leave this out for now, and wait for a more "global" place to document this that we can link to (either the main README or a new page in the Getting Started section of the docs site)

callmehiphop: Part of me thinks it's ok to leave this here since it's what gets displayed on npm for the package overview. I think if anything we can just be much more elaborate on the umbrella docs.

PR #1720, opened 2016-10-18
docs: properly document autoPaginate

### LINKED ISSUE #1684: core: normalize auto-paginated method examples
I'm noticing there are some inconsistencies around our auto-paginated methods. Sometimes we only give examples that imply the user has to [manually paginate requests](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/pubsub/src/topic.js#L232) when using callback mode. I think a sweep of these methods to make sure the examples show both the auto and manual paginated examples would be a great.

stephenplusplus:
Fixes #1684

@callmehiphop -- did you see any others in your journey to promiseland?

### DIFF
--- a/packages/bigquery/src/dataset.js
+++ b/packages/bigquery/src/dataset.js
@@ -360,9 +360,24 @@ Dataset.prototype.delete = function(options, callback) {
  * @param {object} callback.apiResponse - The full API response.
  *
  * @example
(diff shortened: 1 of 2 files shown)


callmehiphop:
hmm, not sure. I'll take another look through.

PR #1721, opened 2016-10-18
Regenerate existing GAPIC auto-generated files.

jmuk:
Updates parameter passing style (now accepting an object) and return style (now returning a Promise)
- [x] Language: #1723
- [ ] Logging: #1735
- [ ] Monitoring: #1733
- [ ] Pub/Sub: #1734
- [ ] Speech: #1737
- [ ] Vision: #1736

### DIFF
--- a/packages/language/package.json
+++ b/packages/language/package.json
@@ -55,7 +55,7 @@
     "@google-cloud/common": "^0.7.0",
     "arrify": "^1.0.1",
     "extend": "^3.0.0",
-    "google-gax": "^0.8.0",
+    "google-gax": "^0.8.1",
     "google-proto-files": "^0.8.3",
     "is": "^3.0.1",
     "propprop": "^0.3.1"
--- a/packages/logging/package.json
+++ b/packages/logging/package.json
@@ -55,7 +55,7 @@
     "@google-cloud/common": "^0.7.0",
     "arrify": "^1.0.0",
     "extend": "^3.0.0",
-    "google-gax": "^0.8.0",
+    "google-gax": "^0.8.1",
     "google-proto-files": "^0.8.0",
     "is": "^3.0.1",
     "is-circular": "^1.0.1",
--- a/packages/monitoring/package.json
+++ b/packages/monitoring/package.json
@@ -52,7 +52,7 @@
     "stackdriver"
   ],
   "dependencies": {
-    "google-gax": "^0.8.0",
+    "google-gax": "^0.8.1",
     "google-proto-files": "^0.8.3",
     "extend": "^3.0.0"
   },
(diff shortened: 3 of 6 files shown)


stephenplusplus:
Thank you! I'm going to mark this PR as <kbd>don't merge</kbd> while I pull pieces from them and update our APIs one at a time.

stephenplusplus:
@jmuk -- I tried to run the Language system tests and got an error:

```
  1) Language HTML inline annotation should work without creating a document:
     Uncaught Error: Exception occurred in retry method that was not classified as transient
      at Object.callback (node_modules/google-gax/lib/api_callable.js:147:19)
      at node_modules/grpc/src/node/src/client.js:420:14
```

stephenplusplus:
Wait, diving deeper gave me some more info:

```js
{ Error: Exception occurred in retry method that was not classified as transient
    at Object.callback (/Users/stephen/dev/gcloud-node/packages/language/node_modules/google-gax/lib/api_callable.js:147:19)
    at /Users/stephen/dev/gcloud-node/packages/language/node_modules/grpc/src/node/src/client.js:420:14
  cause:
   { Error: The document.type is not specified
       at /Users/stephen/dev/gcloud-node/packages/language/node_modules/grpc/src/node/src/client.js:417:17 code: 3, metadata: Metadata { _internal_repr: [Object] } } }
```

This is probably something simple, so assume everything's okay 👍

stephenplusplus:
Yep, a simple fix, my mistake. However, instead of that error, I think it would be more helpful to return the message as the one that is buried currently, ie:

```js
// Before
{ Error: Exception occurred in retry method that was not classified as transient
    at Object.callback (/Users/stephen/dev/gcloud-node/packages/language/node_modules/google-gax/lib/api_callable.js:147:19)
    at /Users/stephen/dev/gcloud-node/packages/language/node_modules/grpc/src/node/src/client.js:420:14
  cause:
   { Error: The document.type is not specified
       at /Users/stephen/dev/gcloud-node/packages/language/node_modules/grpc/src/node/src/client.js:417:17 code: 3, metadata: Metadata { _internal_repr: [Object] } } }

// After
{ Error: The document.type is not specified
    at /Users/stephen/dev/gcloud-node/packages/language/node_modules/grpc/src/node/src/client.js:417:17 code: 3, metadata: Metadata { _internal_repr: [Object] } } }
```

Is that something can happen in the generated layer, or should we wrap it and return the `cause` error to the user from our handwritten layer?

jmuk:
Updated a new patchset to fix travis-ci failures.

Regarding the `error.cause` -- that was embedded in gax-nodejs, and that was simply because I borrowed that from Python, and that design is actually beneficial in Python.
Filed as https://github.com/googleapis/gax-nodejs/issues/56

stephenplusplus:
Ah, okay. Thanks for the background.

jmuk:
Removed `cause` -- it's gone when you update `google-gax` to 0.8.1.

PR #1729, opened 2016-10-19
pubsub: fixed test typos

callmehiphop:


stephenplusplus:
I don't think the method exists on the PubSub class, but let me know if I just missed it.

packages/pubsub/test/topic.js
```diff
@@ -195,28 +195,28 @@ describe('Topic', function() {
     it('should return a stream', function(done) {
```
stephenplusplus: We don't need this test!

PR #1732, opened 2016-10-20
Bug fix - stop promisifying Job#promise and Operation#promise

jmdobry:
This PR represents a test case allowing one to reproduce a bizarre bug I encountered when trying to use `job.promise()`.

Checkout my branch and do the following:
1. `cd packages/bigquery`
2. Set the necessary env vars to run the system tests
3. Do `npm run system-test`

The test will start, and the BigQuery import job will be created, and then the process will be killed inexplicably, with nothing else being printed to the console. I tested this in Node v4.2.0 and v6.7.0.

I believe that somehow, when the job attempts to resolve its promise after emitting the "complete" event, the process crashes. I dunno why.

EDIT: This was caused by the `Job#promise` and `Operation#promise` methods getting promisified.

### DIFF
--- a/packages/bigquery/system-test/bigquery.js
+++ b/packages/bigquery/system-test/bigquery.js
@@ -478,8 +478,7 @@ describe('BigQuery', function() {
           })
           .then(function(results) {
             var metadata = results[0];
-            assert(metadata.status);
-            assert.equal(metadata.status.state, 'DONE');
+            assert.strictEqual(metadata.status.state, 'DONE');
           });
       });
 


stephenplusplus:
If you add a catch block to both promises, I wonder if it would error out.

jmdobry:
I think I might have figured it out. I think the `Job.prototype.promise` method is erroneously getting promisified. Investigating some more...

jmdobry:
That was it! Some kind of crazy promise inception. I'll update my PR to fix it.

callmehiphop:
@jmdobry good catch - an easier fix would probably just be to [ignore methods](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/common/src/util.js#L702) called `promise` by default. This has the added benefit of just releasing a patch to common as opposed to several patch releases.

jmdobry:
Done

callmehiphop:
Would you mind adding a unit test for this? An additional assertion in the [current test](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/common/test/util.js#L1422) should be good enough.

packages/bigquery/system-test/bigquery.js
```diff
@@ -471,6 +471,18 @@ describe('BigQuery', function() {
...
 
+      it('should import data from a file via promises', function() {
+        return table.import(file)
+          .then(function(results) {
+            return results[0].promise();
+          })
+          .then(function(results) {
+            var metadata = results[0];
+            assert(metadata.status);
+            assert.equal(metadata.status.state, 'DONE');
```
stephenplusplus:
```js
assert.strictEqual(metadata.status.state, 'DONE');
```

packages/bigquery/system-test/bigquery.js
```diff
@@ -471,6 +471,18 @@ describe('BigQuery', function() {
...
       });
 
+      it('should import data from a file via promises', function() {
+        return table.import(file)
+          .then(function(results) {
+            return results[0].promise();
+          })
+          .then(function(results) {
+            var metadata = results[0];
+            assert(metadata.status);
```
stephenplusplus: Remove.

packages/bigquery/system-test/bigquery.js
```diff
@@ -471,6 +471,18 @@ describe('BigQuery', function() {
         });
       });
 
+      it('should import data from a file via promises', function() {
+        return table.import(file)
+          .then(function(results) {
+            return results[0].promise();
+          })
```
stephenplusplus: Do we need to handle the errors from both of these promises, or does returning the promise to mocha figure that out?

PR #1740, opened 2016-10-21
monitoring: add to google-cloud

stephenplusplus:


stephenplusplus:
@landrito @jmuk we're getting an error trying to use the generated API:

https://github.com/GoogleCloudPlatform/google-cloud-node/blob/97207ec97bea51bfc9049c6a818a6ee7338bdadf/packages/monitoring/src/v3/group_service_api.js#L524

```js
var groupServiceClient = require('grpc-google-monitoring-v3').client;
```

`grpc-google-monitoring-v3` doesn't exist on npm and isn't a dependency. Is this a typo?

landrito:
I suspect that I generated the file with an incorrect package name which caused this error because the gapic-yaml package name is not [here](https://github.com/googleapis/googleapis/blob/master/google/monitoring/v3/monitoring_gapic.yaml#L14). I'll regenerate. Sec

landrito:
#1741

PR #1747, opened 2016-10-24
docs: Fix bigquery#import format correctly

e-jigsaw:
Hello,

A sample code in Bigquery#import document are incorrect. I tried and faild Job sveral times, so I investigate this issue.

``` js
  if (format) {
    metadata.sourceFormat = FORMATS[format.toLowerCase()];
    delete metadata.format;
  }
```

Above block remove I assigned format. :(

Please fix it.

Thanks

---

This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-node/1747)

### DIFF
--- a/packages/bigquery/src/table.js
+++ b/packages/bigquery/src/table.js
@@ -826,7 +826,7 @@ Table.prototype.getRows = function(options, callback) {
  * //-
  * var metadata = {
  *   encoding: 'ISO-8859-1',
- *   sourceFormat: 'NEWLINE_DELIMITED_JSON'
+ *   format: 'JSON'
  * };
  *
  * table.import('./my-data.csv', metadata, function(err, job, apiResponse) {});


packages/bigquery/src/table.js
```diff
@@ -826,7 +826,7 @@ Table.prototype.getRows = function(options, callback) {
  * //-
  * var metadata = {
  *   encoding: 'ISO-8859-1',
- *   sourceFormat: 'NEWLINE_DELIMITED_JSON'
+ *   format: 'JSON'
```
e-jigsaw: @stephenplusplus Thank you to review.

I wrote

```js
      table.import(files, {
        sourceFormat: 'NEWLINE_DELIMITED_JSON',
        schema: {
          fields: [...]
        },
        writeDisposition: 'WRITE_EMPTY'
      }, done)
```

like this.

https://github.com/e-jigsaw/google-cloud-node/blob/5d1c04f22ce48a0365854da656ea1a95a368f8a8/packages/bigquery/src/table.js#L870

This line assign `NEWLINE_DELIMITED_JSON` to `format` and then

https://github.com/e-jigsaw/google-cloud-node/blob/5d1c04f22ce48a0365854da656ea1a95a368f8a8/packages/bigquery/src/table.js#L872

This line returns `undefined` because `FORMATS` is

https://github.com/e-jigsaw/google-cloud-node/blob/5d1c04f22ce48a0365854da656ea1a95a368f8a8/packages/bigquery/src/table.js#L40-L44

Finally it is configured by `csv`. Then I caught error.

Is this unexpected behavior?

stephenplusplus: Oh, I see now. Thank you, this is indeed a bug! I will send a PR with a fix that will honor both `format` and `sourceFormat`.

stephenplusplus:
I've sent a PR so that `sourceFormat` is honored, which was the intended behavior: #1748.

Thank you for reporting this!

PR #1750, opened 2016-10-25
Change default hash to CRC32c and improve error message if MD5 requested but absent

### LINKED ISSUE #1749: Minor clarification of GCS data validation docs; change default hash algo?
The docs for [createReadStream](https://googlecloudplatform.github.io/google-cloud-node/#/docs/storage/0.4.0/storage/file?method=createReadStream) and [createWriteStream](https://googlecloudplatform.github.io/google-cloud-node/#/docs/storage/0.4.0/storage/file?method=createWriteStream) say

> "By default, data integrity is validated with an MD5 checksum for maximum reliability, falling back to CRC32c when an MD5 hash wasn't returned from the API."

It appears that actually _both_ are calculated, and _both_ are verified if present (although the md5 check overrules the crc32c check), so I think it would be more accurate to say

> "By default, checksums are calculated using both MD5 and CRC32c on the fly; MD5 is used for the final validation for maximum reliability, falling back to CRC32c when an MD5 hash wasn't returned from the API (i.e. in the case of composite objects)."

But, using a cryptographic hash seems like an overkill default. What do you think of changing it to by default calculating CRC32c only, and not calculating MD5 unless the validation method is "MD5"?

(Also, if you specify "md5" then it does not appear to fallback to CRC32c if there was no MD5 returned -- ideally it would fallback instead of doing zero validation, I think?)

zbjornson:
First commit changes validation, fixes #1749.

Second commit improves the error message if MD5 is requested, but not present.

(Noticed that CONTRIBUTING now says to add my name to AUTHORS and CONTRIBUTORS, but looks like not many people do that -- actually supposed to do that?)

### DIFF
--- a/packages/storage/src/file.js
+++ b/packages/storage/src/file.js
@@ -689,7 +689,15 @@ File.prototype.createReadStream = function(options) {
         failed = !validateStream.test('md5', hashes.md5);
       }
 
-      if (failed) {
+      if (md5 && !hashes.md5) {
+        var hashError = new Error([
+          'MD5 verification was specified, but is not available for the',
+          'requested object. MD5 is not available for composite objects.'
+        ].join(' '));
+        hashError.code = 'MD5_NOT_AVAILABLE';
+
+        throughStream.destroy(hashError);
+      } else if (failed) {
         var mismatchError = new Error([
           'The downloaded data did not match the data from the server.',
           'To be sure the content is the same, you should download the',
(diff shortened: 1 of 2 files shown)


stephenplusplus:
Thank you!

@omaray @jgeewax does this PR sound like a good idea? Also, please see the CONTRIBUTORS vs AUTHORS question above.

stephenplusplus:
Thanks for lookin' out for us as always!

PR #1760, opened 2016-10-31
treat partial errors as callback(err)ors

### LINKED ISSUE #1644: expose failed operations as callback(err)
*Thread hijacked by @stephenplusplus*

Some upstream API methods exist, which can return a code 200, however, a portion of the request failed-- partial failure. We've been returning these partial failures to users separate from the `err` property of a `callback(err)`. This is confusing, and we should just do it like this:

```
methodThatCanPartiallyFail('...', function (err, apiResponse) {
  if (err) {
    // err.code = 'INSERT_ERROR'
    // err.errors = [insert errors]
  }
})
```

#### Methods that need to be updated

- bigquery/table#insert
- bigtable/table#mutate
- vision#detect

*Back to you, @c0b...*

am playing around streaming data into tables, with this nodejs-docs-samples code,

https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/bigquery/tables.js#L194-L207

a table created with `name:integer,value:string` as schema but inserted name as string, the program runs ok says 1 row inserted but nothing show up in bigquery console, until I print the `apiResponse` I found the error, but it turns out the callback function's first err is a null, where it shouldn't be; if user has to check `apiResponse` anyway, that makes the first err check not meaningful

``` js
  table.insert(rows, function (err, insertErrors, apiResponse) {
    if (err) {          // why this branch not triggered
      return callback(err);
    }
    ...
```

``` console
$ bq mk new_dataset.new_table name:integer,value:string

$ node ./bigquery/tables.js insert new_dataset newtable '[{ "name": "abc" }]'
Inserted 1 row(s)!
{ datasetId: 'new_dataset',
  tableId: 'newtable',
  rows: [ { name: 'abc' } ],
  callback: [Function],
  err: null,
  insertErrors: 
   [ { errors: 
        [ { message: 'Cannot convert value to integer (bad value).',
            reason: 'invalid' } ],
       row: { name: 'abc' } } ],
  apiResponse: 
   { kind: 'bigquery#tableDataInsertAllResponse',
     insertErrors: 
      [ { index: 0,
          errors: 
           [ { reason: 'invalid',
               location: 'name',
               debugInfo: 'generic::invalid_argument: Cannot convert value to integer (bad value).',
               message: 'Cannot convert value to integer (bad value).' } ] } ] } }
```

stephenplusplus:
Fixes #1644

#### To Dos

  - [x] Split apart common/util into a separate PR: #1764
  - **BigQuery**
    - [x] Docs
    - [x] Unit tests
    - [x] System tests
  - **Bigtable**
    - [x] Docs
    - [x] Unit tests
    - [x] System tests
  - **Vision**
    - [x] Docs
    - [x] Unit tests
    - [x] System tests

### DIFF
--- a/packages/bigquery/src/table.js
+++ b/packages/bigquery/src/table.js
@@ -945,6 +945,7 @@ Table.prototype.import = function(source, metadata, callback) {
  * {module:bigquery/table#import} instead.
  *
  * @resource [Tabledata: insertAll API Documentation]{@link https://cloud.google.com/bigquery/docs/reference/v2/tabledata/insertAll}
+ * @resource [Troubleshooting Errors]{@link https://developers.google.com/bigquery/troubleshooting-errors}
  *
  * @param {object|object[]} rows - The rows to insert into the table.
  * @param {object=} options - Configuration object.
@@ -964,7 +965,9 @@ Table.prototype.import = function(source, metadata, callback) {
  *     for considerations when working with templates tables.
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request.
- * @param {array} callback.insertErrors - A list of errors for insert failures.
+ * @param {object[]} callback.err.errors - If present, these represent partial
+ *     failures. It's possible for part of your request to be completed
+ *     successfully, while the other part was not.
  * @param {object} callback.apiResponse - The full API response.
  *
  * @example
@@ -1011,22 +1014,21 @@ Table.prototype.import = function(source, metadata, callback) {
  * table.insert(row, options, insertHandler);
  *
  * //-
- * // Handling the response.
+ * // Handling the response. See <a href="https://developers.google.com/bigquery/troubleshooting-errors">
+ * // Troubleshooting Errors</a> for best practices on how to handle errors.
  * //-
- * function insertHandler(err, insertErrors, apiResponse) {
- *   // err (object):
- *   //   An API error or partial failure occurred.
+ * function insertHandler(err, apiResponse) {
+ *   if (err) {
+ *     // An API error or partial failure occurred.
  *
- *   if (err.name === 'PartialFailureError') {
- *     // Some rows failed to insert, while others may have succeeded.
+ *     if (err.name === 'PartialFailureError') {
+ *       // Some rows failed to insert, while others may have succeeded.
  *
- *     // err.errors (object[]):
- *     // err.errors[].row (original row object passed to `insert`)
- *     // err.errors[].errors[].reason
- *     // err.errors[].errors[].message
- *
- *     // See https://developers.google.com/bigquery/troubleshooting-errors for
- *     // recommendations on handling errors.
+ *       // err.errors (object[]):
+ *       // err.errors[].row (original row object passed to `insert`)
+ *       // err.errors[].errors[].reason
+ *       // err.errors[].errors[].message
+ *     }
  *   }
  * }
  *
--- a/packages/bigquery/system-test/bigquery.js
+++ b/packages/bigquery/system-test/bigquery.js
@@ -485,7 +485,7 @@ describe('BigQuery', function() {
       it('should convert values to their schema types', function(done) {
         var data = {
           name: 'dave',
-          breed: 'british shorthair',
+          breed: 'husky',
           id: 99,
           dob: new Date(),
           around: true,
@@ -531,7 +531,7 @@ describe('BigQuery', function() {
       it('should return partial errors', function(done) {
         var data = {
           name: 'dave',
-          breed: 'british shorthair',
+          breed: 'british husky',
           id: 99,
           dob: new Date(),
           around: true,
--- a/packages/vision/src/index.js
+++ b/packages/vision/src/index.js
@@ -264,24 +264,28 @@ Vision.prototype.annotate = function(requests, callback) {
  * // a single feature request was not successful.
  * //-
  * vision.detect('malformed-image.jpg', types, function(err, detections) {
- *   if (err && err.name === 'PartialFailureError') {
- *     // err.errors = [
- *     //   {
- *     //     image: 'malformed-image.jpg',
- *     //     errors: [
- *     //       {
- *     //         code: 400,
- *     //         message: 'Bad image data',
- *     //         type: 'faces'
- *     //       },
- *     //       {
- *     //         code: 400,
- *     //         message: 'Bad image data',
- *     //         type: 'labels'
- *     //       }
- *     //     ]
- *     //   }
- *     // ]
+ *   if (err) {
+ *     // An API error or partial failure occurred.
+ *
+ *     if (err.name === 'PartialFailureError') {
+ *       // err.errors = [
+ *       //   {
+ *       //     image: 'malformed-image.jpg',
+ *       //     errors: [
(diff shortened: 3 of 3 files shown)


stephenplusplus:
@callmehiphop PTAL

packages/vision/src/index.js
```diff
@@ -385,16 +386,30 @@ Vision.prototype.detect = function(images, options, callback) {
...
       var detections = images
         .map(groupDetectionsByImage)
         .map(assignTypeToEmptyAnnotations)
         .map(combineErrors)
         .map(flattenAnnotations)
-        .map(decorateAnnotations);
+        .map(decorateAnnotations)
+        .filter(function(annotation) {
+          // Remove annotations that had errors.
+          return annotation;
```
callmehiphop: Should we pushing errors into `partialFailureErrors` somewhere around here?

callmehiphop: nvm! spoke too soon

stephenplusplus: Yeah, that sounds cleaner!

packages/bigquery/src/table.js
```diff
@@ -1015,18 +1015,19 @@ Table.prototype.import = function(source, metadata, callback) {
...
- *   //
- *   // insertErrors[].row (original individual row object passed to `insert`)
- *   // insertErrors[].errors[].reason
- *   // insertErrors[].errors[].message
- *
- *   // See https://developers.google.com/bigquery/troubleshooting-errors for
- *   // recommendations on handling errors.
+ *   //   An API error or partial failure occurred.
+ *
+ *   if (err.name === 'PartialFailureError') {
```
stephenplusplus: Maybe a `code`?

```js
if (err.code === 'PARTIAL_FAILURE') {
  // ...
}
```

callmehiphop: Well, my thought with the property was to avoid a string comparison, so I don't really see that as being any different than the check against the name. If you think that's the way to go, then LGTM.

stephenplusplus: I think the convention of error handling is either the name or the code (short of, e.g. `err instanceof TyperError`), so let's just stick with name.

stephenplusplus:
Ready for a final sign off, @callmehiphop!

packages/bigquery/src/table.js
```diff
@@ -1015,18 +1015,19 @@ Table.prototype.import = function(source, metadata, callback) {
  * //-
  * function insertHandler(err, insertErrors, apiResponse) {
```
callmehiphop: I think you need to kill `insertErrors` here

packages/bigquery/src/table.js
callmehiphop: Should this be `if (err && err.name === 'PartialFailureError') {`? I'm just thinking that if users copy and paste this could would throw an error if `err` is `null`.

packages/bigquery/src/table.js
```diff
@@ -1015,18 +1015,19 @@ Table.prototype.import = function(source, metadata, callback) {
...
- *   // See https://developers.google.com/bigquery/troubleshooting-errors for
- *   // recommendations on handling errors.
+ *   //   An API error or partial failure occurred.
+ *
+ *   if (err.name === 'PartialFailureError') {
+ *     // Some rows failed to insert, while others may have succeeded.
+ *
+ *     // err.errors (object[]):
+ *     // err.errors[].row (original row object passed to `insert`)
+ *     // err.errors[].errors[].reason
```
callmehiphop: Is this right? `.errors[].errors[]`?

packages/bigquery/src/table.js
```diff
@@ -1015,18 +1015,19 @@ Table.prototype.import = function(source, metadata, callback) {
...
+ *     // err.errors[].errors[].message
+ *
+ *     // See https://developers.google.com/bigquery/troubleshooting-errors for
```
callmehiphop: Maybe instead add this link as a `@resource`?

packages/bigquery/system-test/bigquery.js
```diff
@@ -533,6 +528,55 @@ describe('BigQuery', function() {
         });
       });
 
+      it('should return partial errors', function(done) {
+        var data = {
+          name: 'dave',
+          breed: 'british shorthair',
```
callmehiphop: But I wanna be a husky :(((

packages/bigtable/src/table.js
```diff
@@ -769,34 +760,24 @@ Table.prototype.insert = function(entries, callback) {
...
+ *     failures. It's possible for part of your request to be completed
+ *     successfully, while the other part was not.
  *
  * @example
  * //-
  * // Insert entities. See {module:bigtable/table#insert}
  * //-
- * var callback = function(err, mutationErrors) {
+ * var callback = function(err) {
  *   if (err) {
```
callmehiphop: Should we document this consistently across the board? For bigquery we aren't checking if err exists, here we do the err + partial error checks in different blocks and for vision we do it in a single block.

packages/vision/src/index.js
```diff
@@ -262,14 +261,31 @@ Vision.prototype.annotate = function(requests, callback) {
...
- * // a single feature request was not successful. Each returned detection will
- * // have an `errors` array, including any of these errors which may have
- * // occurred.
+ * // a single feature request was not successful.
  * //-
  * vision.detect('malformed-image.jpg', types, function(err, detections) {
- *   if (detections.faces.errors.length > 0) {
- *     // Errors occurred while trying to use this image for a face annotation.
+ *   if (err && err.name === 'PartialFailureError') {
+ *     err.errors = [
```
callmehiphop: Should this be commented? e.g. `// err.errors = [`

stephenplusplus: ✅

packages/bigquery/src/table.js
stephenplusplus: Yeah :(

`err.errors` is the standard format from PartialFailureError. What goes in that array is up to the implementor; so in this case, a row can fail to insert for multiple reasons (according to [this doc](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/insertAll))

packages/bigtable/src/table.js
stephenplusplus: Yes, good catch. Done.

packages/bigquery/system-test/bigquery.js
stephenplusplus: Well, fine then.

packages/bigquery/src/table.js
stephenplusplus: It's a goner.

stephenplusplus:
PTAL!

packages/bigquery/system-test/bigquery.js
```diff
@@ -533,6 +528,55 @@ describe('BigQuery', function() {
         });
       });
 
+      it('should return partial errors', function(done) {
+        var data = {
+          name: 'dave',
+          breed: 'british husky',
```
callmehiphop: A classy husky I suppose

stephenplusplus: You're welcome.

PR #1764, opened 2016-11-02
common: create PartialFailureError

stephenplusplus:
*This PR is a part of #1760*

This exposes a `PartialFailureError` object, which represents an API request that succeeded, but with some type of failure.

### DIFF
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -72,7 +72,7 @@ util.noop = noop;
  *
  * @param {object} errorBody - Error object.
  */
-var ApiError = createErrorClass('ApiError', function(errorBody) {
+util.ApiError = createErrorClass('ApiError', function(errorBody) {
   this.code = errorBody.code;
   this.errors = errorBody.errors;
   this.response = errorBody.response;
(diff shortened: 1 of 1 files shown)


packages/common/src/util.js
```diff
@@ -108,6 +108,30 @@ util.ApiError = function(errorBody) {
...
+  var defaultErrorMessage = 'A failure occurred during this request.';
+  this.message = errorObject.message || defaultErrorMessage;
+});
+
+/**
+ * Wrap the PartialFailureError constructor so context isn't lost.
+ *
+ * @param {object} errorBody - Error object.
+ */
+util.PartialFailureError = function(errorBody) {
```
callmehiphop: Do we need a factory here? The class casing makes me think our intention is not to use this as a factory.

stephenplusplus: Took this out. This is a bit scary, because I originally needed this wrapper when we implemented ApiError (this same pattern exists for that error type). The problem is, I don't remember why, except for my comment in the JSDocs, "...so that context isn't lost."

I haven't been able to produce any error when running our tests, so I took both wrapper functions `util.ApiError` and `util.PartialFailureError`, and just exposed what `create-error-class` returns naturally.

Can you think of any reason we needed the wrapper?

callmehiphop: Sure can't, there are a couple errors floating around BigTable that make use of `createErrorClass` and I've never seen/heard any issues with them. `ApiError` predates our use of `createErrorClass`, is it possible that was an old comment that never got cleaned up?

stephenplusplus: ¯\_(ツ)_/¯ -- this PR's ready to merge if it lgty.

Issue #1766, opened 2016-11-02
Upgrade resource manager to v1


stephenplusplus:
In v1, creating a project now returns an Operation, instead of a Project.

PR #1770, opened 2016-11-04
common: add Operation object

stephenplusplus:
This introduces the `Operation` object into @google-cloud/common. This is a generic class, which will be inherited by:

- common/grpcOperation (in this PR)
- bigquery/job (#1771)
- compute/operation (#1773)
- resource/operation (will exist in a future PR for #1766)

### DIFF
--- a/packages/common/src/grpc-operation.js
+++ b/packages/common/src/grpc-operation.js
@@ -20,20 +20,19 @@
 
 'use strict';
 
-var events = require('events');
 var modelo = require('modelo');
-
-/**
- * @type {module:common/grpcService}
- * @private
- */
-var GrpcService = require('./grpc-service.js');
 
 /**
  * @type {module:common/grpcServiceObject}
  * @private
  */
 var GrpcServiceObject = require('./grpc-service-object.js');
+
+/**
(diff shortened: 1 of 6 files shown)


packages/common/src/util.js
```diff
@@ -99,19 +99,13 @@ util.ApiError = createErrorClass('ApiError', function(errorBody) {
 });
 
 /**
- * Custom error type for partial errors returned from the API.
+ * Wrap the ApiError constructor so context isn't lost.
  *
- * @param {object} b - Error object.
+ * @param {object} errorBody - Error object.
  */
-util.PartialFailureError = createErrorClass('PartialFailureError', function(b) {
```
callmehiphop: But we just added this!

stephenplusplus: Woopsies.

stephenplusplus: Fixedsies!

stephenplusplus:
@callmehiphop after this goes, we can cut a common release, which will clear the way for #1760 (implementing partial failure errors) as well as the follow-up PRs for implementing Operation across the other APIs.

PR #1771, opened 2016-11-07
bigquery: incorporate common/Operation

stephenplusplus:
*Related: #1770*

This converts `bigquery/job` to use `common/operation`.

### DIFF
--- a/packages/bigquery/src/job.js
+++ b/packages/bigquery/src/job.js
@@ -21,7 +21,6 @@
 'use strict';
 
 var common = require('@google-cloud/common');
-var events = require('events');
 var is = require('is');
 var util = require('util');
 
@@ -144,12 +143,11 @@ function Job(bigQuery, id) {
     getMetadata: true
   };
 
-  var self = this;
-
   common.Operation.call(this, {
     parent: bigQuery,
     baseUrl: '/jobs',
-    id: id
+    id: id,
+    methods: methods
   });
 
   this.bigQuery = bigQuery;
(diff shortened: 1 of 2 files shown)


packages/bigquery/src/job.js
```diff
@@ -144,15 +144,14 @@ function Job(bigQuery, id) {
...
 
-  common.ServiceObject.call(this, {
+  var self = this;
+
+  common.Operation.call(this, {
     parent: bigQuery,
     baseUrl: '/jobs',
-    id: id,
-    methods: methods
+    id: id
```
callmehiphop: don't we need to pass in `methods` here?

packages/bigquery/src/job.js
```diff
@@ -313,93 +307,33 @@ Job.prototype.getQueryResultsStream = function(options) {
...
-  var self = this;
-
-  if (!this.hasActiveListeners) {
-    return;
-  }
-
+Job.prototype.poll_ = function(callback) {
   this.getMetadata(function(err, metadata, apiResponse) {
     if (apiResponse.status && apiResponse.status.errors) {
       err = common.util.ApiError(apiResponse.status);
```
callmehiphop: My apologies because I think I missed it in the previous PR, but why isn't this the default behavior of `poll_`? Seems like we should be overwriting it in `GrpcOperation` to handle gRPC related errors instead of handling them by default.

stephenplusplus: It's just about the structure of the response. A Job has a response object that is unique with the `status` property. The default, `common/operation#poll_` handles the most common response object format, using a GrpcService helper method to convert the rpc code to an HTTP status code.

stephenplusplus: If you want, feel free to wait on my other PRs for the other services to see how each Operation-like class uses `common/operation`.

packages/bigquery/src/job.js
stephenplusplus: I ended up passing them in anyway to avoid a lint error.

PR #1773, opened 2016-11-07
compute: use common/Operation

stephenplusplus:
*Related: #1770*

This converts `compute/operation` to use `common/operation`.


callmehiphop:
@stephenplusplus I think the code looks ok, but it looks like there a couple linting issues that need to be addressed.

stephenplusplus:
Woops, sorry about that. Pushed with the fixes.

PR #1775, opened 2016-11-08
Regenerate speech API. Now it contains streamingRecognize method.

jmuk:
This does not change src/index.js yet.

### DIFF
--- a/packages/speech/src/index.js
+++ b/packages/speech/src/index.js
@@ -472,18 +472,13 @@ Speech.formatResults_ = function(resultSets, verboseMode) {
 Speech.prototype.createRecognizeStream = function(config) {
   var self = this;
 
-  var protoOpts = {
-    service: 'Speech',
-    method: 'streamingRecognize'
-  };
-
   var verboseMode = config.verbose === true;
   delete config.verbose;
 
   var recognizeStream = streamEvents(pumpify.obj());
 
   recognizeStream.once('writing', function() {
-    var requestStream = self.requestWritableStream(protoOpts);
+    var requestStream = self.api.Speech.streamingRecognize();
 
     requestStream.on('response', function(response) {
       recognizeStream.emit('response', response);
--- a/packages/speech/test/index.js
+++ b/packages/speech/test/index.js
@@ -448,22 +448,20 @@ describe('Speech', function() {
 
       stream.setPipeline = util.noop;
 
-      speech.requestWritableStream = function() {
-        requestStream = through.obj();
-        return requestStream;
+      speech.api.Speech = {
+        streamingRecognize: function() {
+          requestStream = through.obj();
+          return requestStream;
+        }
       };
     });
 
     it('should make the correct request once writing started', function(done) {
-      speech.requestWritableStream = function(protoOpts) {
-        assert.deepEqual(protoOpts, {
-          service: 'Speech',
-          method: 'streamingRecognize'
-        });
-
-        setImmediate(done);
-
-        return through.obj();
+      speech.api.Speech = {
+        streamingRecognize: function() {
+          setImmediate(done);
+          return through.obj();
+        }
       };
 
       stream.emit('writing');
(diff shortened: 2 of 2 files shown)


stephenplusplus:
@callmehiphop -- having some CI trouble here with AppVeyor:

> npm run docs
> google-cloud-node@ docs C:\projects\google-cloud-node
> node ./scripts/docs
git rev-parse --abbrev-ref HEAD
HEAD
Building docs for bigquery@HEAD
ShellJS: internal error
Error: ENOENT, no such file or directory 'C:\projects\google-cloud-node\docs\json\bigquery\HEAD
'
    at Error (native)
    at Object.fs.mkdirSync (fs.js:747:18)
    at mkdirSyncRecursive (C:\projects\google-cloud-node\node_modules\shelljs\src\mkdir.js:32:6)
    at C:\projects\google-cloud-node\node_modules\shelljs\src\mkdir.js:78:9
    at Array.forEach (native)
    at _mkdir (C:\projects\google-cloud-node\node_modules\shelljs\src\mkdir.js:58:8)
    at C:\projects\google-cloud-node\node_modules\shelljs\src\common.js:338:25
    at Builder.build (C:\projects\google-cloud-node\scripts\docs\builder.js:76:19)
    at build (C:\projects\google-cloud-node\scripts\docs\builder.js:441:11)
    at C:\projects\google-cloud-node\scripts\docs\builder.js:469:5
npm ERR! Windows_NT 6.3.9600

https://ci.appveyor.com/project/GoogleCloudPlatform/google-cloud-node/build/1.0.519/job/9mgbx2cxbrecrcgq#L721

Any ideas?

callmehiphop:
hmm, that's weird. Appears to be failing on a `mkdir -p`. Not sure why that would happen .. I'll try and dig in.

stephenplusplus:
I've tried to integrate this with src/index, and it seems to work, except for the `response` event. Does gax expose a way to get the API's raw response? Currently:

```js
speech.createRecognizeStream({...})
  .on('response', function(response) {
    response === {
      "code": 200,
      "details": "",
      "metadata": {
        "_internal_repr": {
          "content-disposition": [
            "attachment"
          ]
        }
      },
      "message": "OK"
    }
  })
  .on('data', function() {...})
```

stephenplusplus:
@jmuk to clarify, this `response` is from the grpc stream event named `metadata`: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/116436fa789d8b0f7fc5100b19b424e3ec63e6bf/packages/common/src/grpc-service.js#L355

We rename it to `response` as its more common for users familiar with the tradition HTTP request/response workflow, with the `response` event being standard with the Node.js native HTTP module: https://nodejs.org/api/http.html#http_event_response

jmuk:
I see. It seems gax loses `metadata` and `status` event which gRPC supplies. Filed as https://github.com/googleapis/gax-nodejs/issues/67 and I'm on it.

jmuk:
Could it be `status` instead of `metadata` which we want for `response`?

stephenplusplus:
Maybe, what is the difference between the status and response objects?

jmuk:
`metadata` only contains the part of `{metadata: {_internal_repr: { ... }}}` while `status` provides what you wrote -- including `code` or something. And it seems `status` comes at the end of the stream while `metadata` comes earlier.

stephenplusplus:
@callmehiphop wdyt? I think you were more closely connected with our implementation.

callmehiphop:
I think something to note is that status and metadata occur at different times. Where metadata happens once the request starts, status doesn't happen until after all results have been processed.

jmuk:
Ah, okay. So https://github.com/GoogleCloudPlatform/google-cloud-node/blob/116436fa789d8b0f7fc5100b19b424e3ec63e6bf/packages/common/src/grpc-service.js#L355 does take 'metadata' event, and create the response object with "code: 0" (i.e. succeeds). I'll do the same thing.

jmuk:
Now it's added.

stephenplusplus:
I added a commit to use the new streaming method from our API.

@callmehiphop could you take a look and push a commit that gets rid of anything we don't need, re: GrpcService?

callmehiphop:
@stephenplusplus I don't think you can cut out `GrpcService` unless GAX also supports LRO.

stephenplusplus:
Okay, thanks for taking a look.

PR #1781, opened 2016-11-09
translate: api updates

callmehiphop:
The most notable change with this request is that ap keys are no longer required and users can now authenticate in a similar fashion to our other APIs. 

#### TODO

- [x] Docs
- [x] Unit tests
- [x] System tests
- [x] Support model parameter - this needs more discussion as it is a beta feature!

### DIFF
--- a/packages/translate/src/index.js
+++ b/packages/translate/src/index.js
@@ -312,7 +312,7 @@ Translate.prototype.getLanguages = function(target, callback) {
  *     is written in.
  * @param {string} options.model - **Note:** Users must be whitelisted to use
  *     this parameter. Set the model type requested for this translation. Please
- *     refer to the upstread documentation for possible values.
+ *     refer to the upstream documentation for possible values.
  * @param {string} options.to - The ISO 639-1 language code to translate the
  *     input to.
  * @param {function} callback - The callback function.
@@ -376,35 +376,35 @@ Translate.prototype.getLanguages = function(target, callback) {
 Translate.prototype.translate = function(input, options, callback) {
   input = arrify(input);
 
-  var query = {
+  var body = {
     q: input,
     format: options.format || (isHtml(input[0]) ? 'html' : 'text')
   };
 
   if (is.string(options)) {
-    query.target = options;
+    body.target = options;
   } else {
     if (options.from) {
-      query.source = options.from;
+      body.source = options.from;
     }
 
     if (options.to) {
-      query.target = options.to;
+      body.target = options.to;
     }
 
     if (options.model) {
-      query.model = options.model;
+      body.model = options.model;
     }
   }
 
-  if (!query.target) {
+  if (!body.target) {
     throw new Error('A target language is required to perform a translation.');
   }
 
   this.request({
     method: 'POST',
     uri: '',
-    json: query
+    json: body
   }, function(err, resp) {
     if (err) {
       callback(err, null, resp);
@@ -413,7 +413,7 @@ Translate.prototype.translate = function(input, options, callback) {
 
     var translations = resp.data.translations.map(prop('translatedText'));
 
-    if (query.q.length === 1) {
+    if (body.q.length === 1) {
       translations = translations[0];
     }
 
--- a/packages/translate/test/index.js
+++ b/packages/translate/test/index.js
@@ -101,6 +101,7 @@ describe('Translate', function() {
         'https://www.googleapis.com/auth/cloud-platform'
       ]);
       assert.deepEqual(calledWith.packageJson, require('../package.json'));
+      assert.strictEqual(calledWith.projectIdRequired, false);
     });
 
     describe('Using an API Key', function() {


callmehiphop:
@stephenplusplus @jgeewax @omaray The thing I'm missing here is the model parameter, my understanding is that it is in beta and users will have to be whitelisted to use it. Do we want to cut a separate release to support this parameter?

It's just a string, so I think we could probably get away with just adding it here and making note that users need to be whitelisted to use it. WDYT?

callmehiphop:
@stephenplusplus I'm not sure any exists yet :( but it would essentially just be a string that the user supplies.

callmehiphop:
Seeing an e2e failure, not sure if it's an actual error or not.

If we send the word *Hello!* to be translated into Spanish it comes back as *¡Hola!*. (This is expected). However when we send the same text in the form of html `<body>Hello!</body>` we are returned `<body>Hola!</body>` (The *¡* is no longer present).

bjwatson:
@callmehiphop I've pinged @akeran to see if he has any thoughts on that behavior.

akeran:
If you set format=html you will get "\u003cbody\u003e ¡Hola! \u003c/body\u003e"
html is the default value for this API, please check that you don't overwrite it to format=text

stephenplusplus:
> Seeing an e2e failure, not sure if it's an actual error or not.

Which test? `should autodetect HTML` returns what I would expect:

```
<body> ¡Hola! </body>
```

callmehiphop:
@akeran the assertion we made actually included the html tags wrapping the content - however, it would appear the issue I was seeing disappeared at some point in the night!

packages/translate/src/index.js
```diff
@@ -312,6 +312,9 @@ Translate.prototype.getLanguages = function(target, callback) {
  *     is written in.
  * @param {string} options.to - The ISO 639-1 language code to translate the
  *     input to.
+ * @param {string} options.model - **Note:** Users must be whitelisted to use
```
akeran: You can add something like "The `model` type requested for this translation. Valid values are listed in public documentation."  We'll add the link ready.

packages/translate/README.md
```diff
@@ -59,6 +58,71 @@ var translate = require('@google-cloud/translate')({
...
+
+// ...you're good to go!
+```
+
+### With an API key
+
+It's also possible to authenticate with an API key. To create an API key:
+
+1. Visit the [Google Developers Console][dev-console].
+2. 2. Create a new project or click on an existing project.
```
stephenplusplus: `2. 2.`

packages/translate/src/index.js
```diff
@@ -305,6 +312,9 @@ Translate.prototype.getLanguages = function(target, callback) {
...
```
stephenplusplus: Alpha-sort the options.

packages/translate/system-test/translate.js
```diff
@@ -23,16 +23,8 @@ var prop = require('propprop');
...
-
-// Only run the tests if there is an API key to test with.
-(API_KEY ? describe : describe.skip)('translate', function() {
-  if (!API_KEY) {
-    // The test runner still executes this function, even if it is skipped.
-    return;
-  }
-
-  var translate = new Translate(extend({}, env, { key: API_KEY }));
+describe('translate', function() {
```
stephenplusplus: We should make sure to test that authenticating with an API key works. Just a test that uses the key if the env var is set, `this.skip()`s otherwise, and hits an endpoint to check that it doesn't error.

packages/translate/test/index.js
```diff
@@ -38,9 +42,16 @@ var fakeUtil = extend({}, util, {
   }
 });
 
+function FakeService() {
+  this.calledWith_ = arguments;
+  Service.apply(this, arguments);
```
stephenplusplus: We shouldn't need to call through to the real Service.

packages/translate/test/index.js
```diff
@@ -85,27 +97,37 @@ describe('Translate', function() {
...
+      });
+
+      it('should localize the options', function() {
+        var options = { key: '...' };
+        var translate = new Translate(options);
+        assert.strictEqual(translate.options, options);
+      });
+
+      it('should localize the api key', function() {
+        assert.equal(translate.key, KEY_OPTIONS.key);
```
stephenplusplus: `strictEqual`

packages/translate/test/index.js
```diff
@@ -85,27 +97,37 @@ describe('Translate', function() {
       fakeUtil.normalizeArguments = normalizeArguments;
     });
 
-    it('should throw if an API key is not provided', function() {
-      assert.throws(function() {
-        new Translate({});
-      }, /An API key is required to use the Translate API\./);
-    });
+    it('should inherit from Service', function() {
+      assert(translate instanceof Service);
```
stephenplusplus: `FakeService`

packages/translate/system-test/translate.js
callmehiphop: Done!

packages/translate/README.md
callmehiphop: Doh. Fixed!

packages/translate/src/index.js
callmehiphop: Done - originally I had not only because I thought `to` and `from` made more sense being grouped together.

packages/translate/test/index.js
callmehiphop: Done!

packages/translate/test/index.js
callmehiphop: Done!

packages/translate/test/index.js
callmehiphop: Done!

packages/translate/src/index.js
```diff
@@ -70,21 +68,30 @@ function Translate(options) {
...
+  if (options.key) {
+    this.options = options;
+    this.key = options.key;
+  }
+
+  var config = {
+    baseUrl: baseUrl,
+    scopes: ['https://www.googleapis.com/auth/cloud-platform'],
+    packageJson: require('../package.json'),
+    projectIdRequired: false
```
stephenplusplus: Untested.

packages/translate/src/index.js
```diff
@@ -303,6 +310,9 @@ Translate.prototype.getLanguages = function(target, callback) {
  *     not, we set the format as `text`.
  * @param {string} options.from - The ISO 639-1 language code the source input
  *     is written in.
+ * @param {string} options.model - **Note:** Users must be whitelisted to use
+ *     this parameter. Set the model type requested for this translation. Please
+ *     refer to the upstread documentation for possible values.
```
stephenplusplus: upstream*

packages/translate/src/index.js
```diff
@@ -381,16 +391,20 @@ Translate.prototype.translate = function(input, options, callback) {
     if (options.to) {
       query.target = options.to;
     }
+
+    if (options.model) {
+      query.model = options.model;
```
stephenplusplus: Let's rename `query` to `body` now that this is a post request.

packages/translate/test/index.js
```diff
@@ -462,44 +497,83 @@ describe('Translate', function() {
...
-      };
+    describe('OAuth mode', function() {
+      var request;
 
-      var reqOpts = {
-        uri: '/test',
-        a: 'b',
-        c: 'd',
-        qs: {
+      beforeEach(function() {
```
stephenplusplus: I think this should just be a `before` hook.

callmehiphop: Actually there's an issue with the order of execution of `before` hooks, so we start seeing errors when using it.

packages/translate/src/index.js
callmehiphop: Done

packages/translate/src/index.js
callmehiphop: Whoops! Fixed

packages/translate/src/index.js
callmehiphop: Sure, done.

packages/translate/test/index.js
stephenplusplus: I switched it to `before` locally and it worked okay. What error do you get?

callmehiphop:
```sh
  1) Translate request API key mode should make the correct request:
     TypeError: Cannot read property 'call' of undefined
      at Translate.request (src/index.js:437:37)
      at Context.<anonymous> (test/index.js:576:19)
```

It would seem it is trying to use `Service#request` to make the request. My guess is that the `before` hooks are not happening in the expected order and the Translate instance using an API key gets overwrote by an instance using the standard credentials.

stephenplusplus:
Please re-assign if it needs another look.

PR #1783, opened 2016-11-10
Language updates

callmehiphop:
Updates Natural Language to 1.0 proto files!

# ~:warning: BLOCKED~

- [x] https://github.com/googleapis/googleapis/pull/190
- [x] A new [google-proto-files](https://github.com/stephenplusplus/google-proto-files) release

#### Notable Changes

* New rcp `AnalyzeSyntax` in the form of `document#detectSyntax`
* Sentences can now contain sentiments
* `sentiment.polarity` was renamed to `sentiment.score`
* `partOfSpeech` received 11 new attributes
* English, Japanese and Spanish now supported for all features.
* There are additional changes, but generally only affect verbose mode.

#### Possible Conflicts (per API team review)

* Multiplying values by `100` appears to be problematic. While score (polarity) returns a value between -1 and 1, magnitude is unbounded. That being said I'm not sure we're creating a meaningful number for magnitude by multiplying it.

#### TODO

- [x] Unit tests

### DIFF
--- a/packages/language/system-test/language.js
+++ b/packages/language/system-test/language.js
@@ -360,17 +360,22 @@ describe('Language', function() {
           text: 'Hello',
           partOfSpeech: 'Other: foreign words, typos, abbreviations',
           tag: 'X',
-          aspect: false,
-          case: false,
-          form: false,
-          gender: false,
-          mood: false,
-          number: false,
-          person: false,
-          proper: false,
-          reciprocity: false,
-          tense: false,
-          voice: false
+          aspect: undefined,
+          case: undefined,
+          form: undefined,
+          gender: undefined,
+          mood: undefined,
+          number: undefined,
+          person: undefined,
+          proper: undefined,
+          reciprocity: undefined,
+          tense: undefined,
+          voice: undefined,
+          dependencyEdge: {
+            description: 'Root',
+            label: 'ROOT',
+            headTokenIndex: 0
+          }
         });
 
         assert(is.object(apiResponse));
@@ -517,9 +522,9 @@ describe('Language', function() {
       try {
         assert.ifError(err);
 
-        assert(is.object(sentiment.sentiment));
-        assert(is.number(sentiment.sentiment.score));
-        assert(is.number(sentiment.sentiment.magnitude));
+        assert(is.object(sentiment));
+        assert(is.number(sentiment.score));
+        assert(is.number(sentiment.magnitude));
         assert.strictEqual(sentiment.language, 'en');
         assert.strictEqual(sentiment.sentences.length, 2);
 
@@ -533,29 +538,31 @@ describe('Language', function() {
   }
 
   function validateSyntaxSimple(callback) {
-    return function(err, syntax, apiResponse) {
+    return function(err, tokens, apiResponse) {
       try {
         assert.ifError(err);
-        assert.strictEqual(syntax.sentences.length, 2);
-        assert(is.string(syntax.sentences[0]));
-        assert.strictEqual(syntax.tokens.length, 17);
-        assert.deepEqual(syntax.tokens[0], {
-          aspect: false,
-          case: false,
-          form: false,
-          gender: false,
-          mood: false,
-          number: false,
+        assert.strictEqual(tokens.length, 17);
+        assert.deepEqual(tokens[0], {
+          aspect: undefined,
+          case: undefined,
+          form: undefined,
+          gender: undefined,
+          mood: undefined,
+          number: undefined,
           partOfSpeech: 'Other: foreign words, typos, abbreviations',
-          person: false,
-          proper: false,
-          reciprocity: false,
+          person: undefined,
+          proper: undefined,
+          reciprocity: undefined,
           tag: 'X',
-          tense: false,
+          tense: undefined,
           text: 'Hello',
-          voice: false
+          voice: undefined,
+          dependencyEdge: {
+            description: 'Root',
+            headTokenIndex: 0,
+            label: 'ROOT'
+          }
         });
-        assert.strictEqual(syntax.language, 'en');
 
         assert(is.object(apiResponse));
 
--- a/packages/language/package.json
+++ b/packages/language/package.json
@@ -56,7 +56,7 @@
     "arrify": "^1.0.1",
     "extend": "^3.0.0",
     "google-gax": "^0.9.0",
-    "google-proto-files": "^0.8.3",
+    "google-proto-files": "^0.8.5",
     "is": "^3.0.1",
     "propprop": "^0.3.1"
   },


callmehiphop:
@stephenplusplus PTAL!

stephenplusplus:
+1 on not multiplying magnitude.

callmehiphop:
> +1 on not multiplying magnitude.

Awesome, that being said do you think we should be consistent and not multiply score as well?

stephenplusplus:
No, they're not the same value range, so we can't achieve consistency between them. All of our APIs multiple -1.0 through 1.0 ranges.

packages/language/src/document.js
```diff
@@ -567,16 +626,20 @@ Document.prototype.detectEntities = function(options, callback) {
...
  * @param {object=} options - Configuration object. See
- *     [documents.annotateText](https://cloud.google.com/natural-language/reference/rest/v1beta1/documents/analyzeSentiment#request-body).
+ *     [documents.annotateText](https://cloud.google.com/natural-language/reference/rest/v1/documents/analyzeSentiment#request-body).
  * @param {boolean} options.verbose - Enable verbose mode for more detailed
  *     results. Default: `false`
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error occurred while making this request.
- * @param {number} callback.sentiment - A value in the range of `-100` to `100`.
- *     Large numbers represent more positive sentiments.
+ * @param {number} callback.sentiment.sentiment - A value in the range of `-100`
```
stephenplusplus: Let's keep our default return the same; just return the score.

Default:
```js
sentiment = 33
```

Verbose:
```js
sentiment = {
  score: 33,
  magnitude: 4,
  sentences: [
    {
      text: {
        content: '...',
        beginOffset: '...'
      }
    }
  ],
  language: 'en'
}
```

packages/language/src/document.js
```diff
@@ -601,8 +671,21 @@ Document.prototype.detectEntities = function(options, callback) {
...
+ *   //     score: 100,
+ *   //     magnitude: 4
+ *   //   },
+ *   //   sentences: [
+ *   //     {
+ *   //       text: {
+ *   //         content:
+ *   //           'Google is an American multinational technology company' +
+ *   //           'specializing in Internet-related services and products.',
+ *   //         beingOffset: -1
```
stephenplusplus: `begin`

packages/language/system-test/language.js
```diff
@@ -331,7 +359,18 @@ describe('Language', function() {
         assert.deepEqual(annotation.tokens[0], {
           text: 'Hello',
           partOfSpeech: 'Other: foreign words, typos, abbreviations',
-          partOfSpeechTag: 'X'
+          tag: 'X',
+          aspect: 'ASPECT_UNKNOWN',
```
stephenplusplus: Can we normalize the `0` case for these `{{type}}_UNKNOWN` and return `undefined` or `false`? That would help the user:

```js
detectSyntax(..., function(err, syntax) {
  // :(
  var genders = syntax.tokens.filter(function(token) {
    return token.gender !== 'GENDER_UNKNOWN'
  })

  // :)
  var genders = syntax.tokens.filter(function(token) {
    return token.gender
  })
})
```

packages/language/src/document.js
```diff
@@ -196,13 +197,34 @@ Document.PART_OF_SPEECH = {
  *     text.
  * @param {object[]} callback.annotation.tokens - Parts of speech that were
  *     detected from the text.
- * @param {object[]} callback.annotation.tokens.text - The piece of text
+ * @param {string} callback.annotation.tokens[].text - The piece of text
```
callmehiphop: I believe so? I'll double check but I don't see why it wouldn't.

jgeewax:
Are we going to change the version to 1.0? Or is this meaning something else? Not sure if we're ready to say 1.0 here....

packages/language/src/document.js
callmehiphop: For reference, Translate actually does this as well.

http://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.43.0/translate?method=detect

callmehiphop:
@jgeewax sorry about the confusion, we're moving from the upstream v1beta1 api to the upstream v1 api. We will **not** be publishing the library itself as 1.0

packages/language/src/document.js
callmehiphop: Done

packages/language/src/document.js
callmehiphop: Fixed

packages/language/system-test/language.js
callmehiphop: SGTM - done.

packages/language/src/document.js
stephenplusplus: Cool.

packages/language/src/document.js
```diff
@@ -567,16 +626,16 @@ Document.prototype.detectEntities = function(options, callback) {
...
  * @param {object=} options - Configuration object. See
- *     [documents.annotateText](https://cloud.google.com/natural-language/reference/rest/v1beta1/documents/analyzeSentiment#request-body).
+ *     [documents.annotateText](https://cloud.google.com/natural-language/reference/rest/v1/documents/analyzeSentiment#request-body).
  * @param {boolean} options.verbose - Enable verbose mode for more detailed
  *     results. Default: `false`
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error occurred while making this request.
- * @param {number} callback.sentiment - A value in the range of `-100` to `100`.
- *     Large numbers represent more positive sentiments.
+ * @param {number} callback.sentiment - A value in the range of `-100`
```
stephenplusplus: I think more of this might fit on the first line. Let me know if that's wrong!

packages/language/src/document.js
```diff
@@ -601,8 +660,21 @@ Document.prototype.detectEntities = function(options, callback) {
  *   }
  *
  *   // sentiment = {
- *   //   polarity: 100,
- *   //   magnitude: 40
+ *   //   sentiment: {
```
stephenplusplus: Let's bump the properties in this object up a level so we don't have "sentiment.sentiment"

packages/language/src/document.js
```diff
@@ -634,11 +706,160 @@ Document.prototype.detectSentiment = function(options, callback) {
...
+    if (err) {
+      callback(err, null, resp);
+      return;
+    }
+
+    var originalResp = extend(true, {}, resp);
+
+    var syntax = {
+      sentences: Document.formatSentences_(resp.sentences, verbose),
+      tokens: Document.formatTokens_(resp.tokens, verbose),
```
callmehiphop: @stephenplusplus since `detectSentiment` only returns sentences & language via verbose mode, should we follow that convention and only return the tokens by default here?

packages/language/src/document.js
callmehiphop: yep yep!

packages/language/src/document.js
```diff
@@ -634,11 +704,162 @@ Document.prototype.detectSentiment = function(options, callback) {
...
+ * @param {boolean} options.verbose - Enable verbose mode for more detailed
+ *     results. Default: `false`
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An error occurred while making this request.
+ * @param {object} callback.syntax - The recognizd syntax from the text.
+ * @param {string[]} callback.syntax.sentences - Sentences detected from the
+ *     text.
+ * @param {object[]} callback.syntax.tokens - Parts of speech that were
+ *     detected from the text.
+ * @param {string} callback.syntax.language - The language detected from the
```
stephenplusplus: Alpha-sort.

packages/language/src/document.js
```diff
@@ -634,11 +704,162 @@ Document.prototype.detectSentiment = function(options, callback) {
...
+ *
+ * @resource [documents.analyzeSyntax API Documentation]{@link https://cloud.google.com/natural-language/reference/rest/v1/documents/analyzeSyntax}
+ *
+ * @param {object=} options - Configuration object. See
+ *     [documents.annotateText](https://cloud.google.com/natural-language/reference/rest/v1/documents/analyzeSyntax#request-body).
+ * @param {boolean} options.verbose - Enable verbose mode for more detailed
+ *     results. Default: `false`
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An error occurred while making this request.
+ * @param {object} callback.syntax - The recognizd syntax from the text.
```
stephenplusplus: The syntax recognized from the text.

packages/language/src/document.js
```diff
@@ -634,11 +704,162 @@ Document.prototype.detectSentiment = function(options, callback) {
...
   });
 };
 
 /**
+ * Detect syntax from the document.
+ *
+ * @resource [documents.analyzeSyntax API Documentation]{@link https://cloud.google.com/natural-language/reference/rest/v1/documents/analyzeSyntax}
+ *
+ * @param {object=} options - Configuration object. See
+ *     [documents.annotateText](https://cloud.google.com/natural-language/reference/rest/v1/documents/analyzeSyntax#request-body).
```
stephenplusplus: Should this be `[documents.analyzeSyntax]`?

packages/language/src/document.js
```diff
@@ -634,11 +704,162 @@ Document.prototype.detectSentiment = function(options, callback) {
...
+ *
+ *   // syntax = {
+ *   //   sentences: [
+ *   //     'Google is an American multinational technology company ' +
+ *   //     'specializing in Internet-related services and products.'
+ *   //   ],
+ *   //   tokens: [
+ *   //     {
+ *   //       text: 'Google',
+ *   //       partOfSpeech: 'Noun (common and proper)'
```
stephenplusplus: Let's keep the `tag` property here as well in case the user needs our map to help them work with the data.

packages/language/src/document.js
```diff
@@ -634,11 +704,162 @@ Document.prototype.detectSentiment = function(options, callback) {
...
+ *   //         number: 'SINGULAR',
+ *   //         person: 'FIRST',
+ *   //         proper: 'PROPER',
+ *   //         reciprocity: 'RECIPROCAL',
+ *   //         tense: 'PAST',
+ *   //         voice: 'PASSIVE'
+ *   //       },
+ *   //       dependencyEdge: {
+ *   //         headTokenIndex: 1,
+ *   //         label: 'NSUBJ'
```
stephenplusplus: Is there a map that we can use to friendly this up?

packages/language/src/document.js
```diff
@@ -745,12 +964,19 @@ Document.formatSentiment_ = function(sentiment, verbose) {
...
-      };
+    tokens = tokens.map(function(rawToken) {
+      var token = extend({}, rawToken.partOfSpeech, {
+        text: rawToken.text.content,
+        partOfSpeech: Document.PART_OF_SPEECH[rawToken.partOfSpeech.tag]
+      });
+
+      for (var part in token) {
+        if (token.hasOwnProperty(part) && /UNKNOWN/.test(token[part])) {
+          token[part] = false;
```
stephenplusplus: I'm not sure if `false` or `undefined` is better here. `false` implies there was an answer, but `undefined` means "we don't know", which better matches `UNKNOWN`. What do you think?

callmehiphop: I think that makes sense, truthfully I could go either way - all other possible values are strings so it's likely that our users will treat it as a falsey value, in which case `undefined` and `false` are going to behave in the same fashion.

packages/language/src/document.js
stephenplusplus: This would be similar to partsOfSpeech. We can return both-- the code is good for computers, the friendly name is good for humans; an example use would be for logging/printing to a user.

packages/language/src/document.js
stephenplusplus: Let's go undefined, so we don't have even a chance of misinterpretation by the user.

stephenplusplus:
Please re-assign when it's ready for another look.

packages/language/src/document.js
callmehiphop: I've added it in a new field called `description`. PTAL

packages/language/src/document.js
callmehiphop: It was there, just undocumented.

packages/language/src/document.js
callmehiphop: Sure

packages/language/src/document.js
callmehiphop: I think so, probably copy/pasta.

packages/language/src/document.js
stephenplusplus: Looks good, I think this example needs an update to add `description: 'Nominal subject'`.

callmehiphop: Sure does, I was just waiting to hear your feedback before I wrote docs and tests.

callmehiphop: Tests & docs updated.

stephenplusplus:
LGTM. Re-assign when the blockers are cleared.

jmuk:
https://github.com/googleapis/googleapis/pull/190 has been merged, and the new proto is already in the master of https://github.com/googleapis/googleapis.  @stephenplusplus, please update google-proto-files package.

bjwatson:
@callmehiphop @jmdobry https://github.com/googleapis/googleapis/pull/190 has been merged. Can one of you take care of updating google-proto-files?

jmdobry:
https://github.com/stephenplusplus/google-proto-files/pull/11

bjwatson:
Thanks @jmdobry and @stephenplusplus.

jmdobry:
Language v1 proto is available in `google-proto-files v0.8.5`.

bjwatson:
Thanks @jmdobry.

@stephenplusplus have your requested changes been satisfied?

bjwatson:
Dave explained that the unhappiness is just depending on the google-proto-files update.

stephenplusplus:
Yep, just update google-proto-files to 0.9.0 (@callmehiphop simultaneously published to npm).

callmehiphop:
Should be good to go on this.

jmdobry:
@callmehiphop Remember to get your google-proto-files `0.9.0` commit pushed to GitHub

callmehiphop:
@jmdobry I ended up unpublishing it since you beat me to it :)

PR #1789, opened 2016-11-14
bigtable: promisify table#mutate methods

### LINKED ISSUE #1788: Bigtable.insert isn't promisified?
https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.43.0/bigtable/table?method=insert

Looks like we don't have the promisified verison of insert which is a bummer because the code ends up looking like...

```js
bigtable.getInstances().then(function(data) {
  var instance = data[0][0];
  var table = instance.table(...);
  table.insert(entries, function(err, insertErrors) {
    ...
  });
});
```

Mixing the promise version and callbacks makes me feel :(

callmehiphop:
Closes #1788 

`Bigtable#insert` and `Bigtable#mutate` were not being promisified, I believe they got skipped because they have a unique dual functionality. gRPC returns an object stream for these methods in an attempt to process errors as they occur as opposed to when all mutations have been processed.

My solution was to create `mutateStream` and `insertStream` methods that still allow the user to process errors in this fashion and use promises for `mutate` and `insert`.

### DIFF
--- a/packages/bigtable/src/table.js
+++ b/packages/bigtable/src/table.js
@@ -763,7 +763,7 @@ Table.prototype.insert = function(entries, callback) {
  *
  * @example
  * //-
- * // Insert entities. See {module:bigtable/table#insert}
+ * // Insert entities. See {module:bigtable/table#insert}.
  * //-
  * var callback = function(err) {
  *   if (err) {
@@ -792,7 +792,7 @@ Table.prototype.insert = function(entries, callback) {
  * table.mutate(entries, callback);
  *
  * //-
- * // Delete entities. See {module:bigtable/row#deleteCells}
+ * // Delete entities. See {module:bigtable/row#deleteCells}.
  * //-
  * var entries = [
  *   {
@@ -848,7 +848,6 @@ Table.prototype.insert = function(entries, callback) {
  * table.mutate(entries).then(function() {
  *   // All requested mutations have been processed.
  * });
- * //-
  */
 Table.prototype.mutate = function(entries, callback) {
   entries = flatten(arrify(entries));
@@ -864,40 +863,34 @@ Table.prototype.mutate = function(entries, callback) {
     entries: entries.map(Mutation.parse)
   };
 
-  var stream = pumpify.obj([
-    this.requestStream(grpcOpts, reqOpts),
-    through.obj(function(data, enc, next) {
-      var throughStream = this;
+  var mutationErrors = [];
 
-      data.entries.forEach(function(entry) {
-        // mutation was successful, no need to notify the user
+  this.requestStream(grpcOpts, reqOpts)
+    .on('error', callback)
+    .on('data', function(obj) {
+      obj.entries.forEach(function(entry) {
+        // Mutation was successful.
         if (entry.status.code === 0) {
           return;
         }
 
         var status = common.GrpcService.decorateStatus_(entry.status);
-
         status.entry = entries[entry.index];
-        throughStream.push(status);
+
+        mutationErrors.push(status);
       });
-
-      next();
     })
-  ]);
-
-  stream
-    .on('error', callback)
-    .pipe(concat(function(mutationErrors) {
+    .on('end', function() {
       var err = null;
 
-      if (mutationErrors && mutationErrors.length > 0) {
+      if (mutationErrors.length > 0) {
         err = new common.util.PartialFailureError({
           errors: mutationErrors
         });
       }
 
       callback(err);
-    }));
+    });
 };
 
 /**
--- a/packages/bigtable/test/table.js
+++ b/packages/bigtable/test/table.js
@@ -976,6 +976,39 @@ describe('Bigtable/Table', function() {
         });
       });
     });
+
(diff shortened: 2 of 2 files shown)


callmehiphop:
@stephenplusplus Going to hold off on tests/linting/etc. until I get an OK from you on API design.

packages/bigtable/src/table.js
```diff
@@ -735,20 +735,51 @@ Table.prototype.getRows = function(options, callback) {
...
+
+  return this.mutate(entries, callback);
+};
+
+/**
+ * Insert or update rows in your table. It should be noted that gRPC only allows
+ * you to send payloads that are less than or equal to 4MB. If you're inserting
+ * more than that you may need to send smaller individual requests.
+ *
+ * Use this method over {bigtable/table#insert} if you want to handle insert
```
stephenplusplus: Can we insert a similar line in `insert` that suggests this method?

packages/bigtable/src/table.js
```diff
@@ -735,20 +735,51 @@ Table.prototype.getRows = function(options, callback) {
...
+ *         gwashington: 1
+ *       }
+ *     }
+ *   }
+ * ];
+ *
+ * table.insertStream(entries)
  *   .on('error', console.error)
- *   .on('complete', function() {
+ *   .on('finish', function() {
```
stephenplusplus: Can you catch me up on why we were using `complete` and how it can be replaced with `finish` now?

callmehiphop: Originally you had requested that we do not return a stream, but rather an EventEmitter (I can't remember the exact details as to why tbh). So we bound the streams `finish` event to the event emitters `complete` event.

packages/bigtable/src/table.js
callmehiphop: Sure

packages/bigtable/src/table.js
stephenplusplus: Ah, right. This isn't very stream-y, and I don't think we should call this new method a stream, if I'm understanding it right. A stream would imply to me that I'm getting a writable stream and I can pipe data to it, i.e.

```js
fs.createReadStream('./my-entries.json')
  .pipe(table.createWriteStream())
  .on('error', function(err) {})
  .on('finish', function() {})
```

The way `mutateStream` appears to work isn't as a stream, since all of the data is required up front. The reason we have the event emitter is just to communicate errors as they occur, right? I wonder if we can just keep things simple, and only communicate the errors after the request is complete, and return it in the callback... which I believe is how `mutate` works currently if you give it a callback.

Ideally, we could return a writable stream so the code from my example is possible. We can build the `reqOpts.entries` array progressively as we receive data from a source stream, but it would end up building up memory before we can send the request and flushing it. So... that doesn't work.

callmehiphop: I'm find with renaming it something else, but obviously we need to separate this functionality some how. This is probably going to require new since this is pretty unique to our API. Any suggestions?

stephenplusplus: Does that mean you aren't keen to just remove the event emitter concept and simply wait until the call is complete to return errors? Again, I just think we're making things a bit more confusing by splitting methods based on this, but let me know if this method is required or motivated by a real need.

callmehiphop: I think we can remove it, I believe this functionality was implemented to notify users of the transaction states for large mutations, my understanding is that they can become lengthy in duration.

stephenplusplus: I'm okay with removing, and waiting until we hear this is crucial to some users to reinstate.

callmehiphop:
@stephenplusplus made some changes per our convo, PTAL

package.json
```diff
@@ -39,7 +39,7 @@
     "prepare-ghpages": "node ./scripts/docs/prepare.js",
     "remove-ghpages": "node ./scripts/docs/remove.js",
     "lint": "jshint scripts/ packages/ system-test/ test/ && jscs packages/ system-test/ test/",
-    "test": "npm run docs && npm run snippet-test && npm run unit-test",
+    "test": "rm -rf docs/json && npm run docs && npm run snippet-test && npm run unit-test",
```
stephenplusplus: Is this Windows-safe?

stephenplusplus: I guess we'll find out once AppVeyor runs :)

stephenplusplus: We might also need to do this programmatically in case `npm test` is run from the `packages` directory, the local path, "docs/json", might not locate the correct directory.

stephenplusplus:
Looks like our coverage of bigtable/src/table.js dropped... not sure if in this PR or somewhere else. And maybe the coveralls UI is doing something wrong, but it appears to be a drop on a line in our JSDocs: https://coveralls.io/jobs/20152098/source_files/1964040321#L875

... any chance that gives you any idea what might be up?

stephenplusplus:
Resolved the test coverage and simplified how the `mutate` method collects & returns errors, now that we're not concerned with the stream/event emitter pattern.

PR #1802, opened 2016-11-17
Bigtable prefix

### LINKED ISSUE #1790: Bigtable doesn't have a magical prefix range helper?
A common thing with Bigtable is the ability to scan based on a prefix, which is basically a range with start=prefix and end=someAlteration(prefix), but its tricky to do manually. 

Any chance we can add a `prefix` filter that "just works"?

---

Go does this a bit trickily (see https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/bigtable/bigtable.go#L307), the gist seems to be (for those not familiar with go):

- if the string is empty, the end range marker should just be empty string
- otherwise, find the last character in the string that can be incremented (ie, `abcde` can increment position 4, `abc\xff\xff` can increment position 2.
- using that pivot, take `start.substr(0, n-1)` and append the incremented character (`start[n]+1`)

I think -- in JS -- this would look something like....

```js
var getPrefixEndRange = function(start) {
  var maxChar = String.fromCharCode(0xff);
  var position = start.length-1;

  // Walk backwards until we get to a character we can increment.
  while (start[position] == maxChar && position >= 0) position--;

  // If the position is -1, there is no reasonable end range for the prefix.
  if (position == -1) return '';

  var nextChar = String.fromCharCode(start.charCodeAt(position)+1)
  return start.substring(0, position) + nextChar;
}
```

Some test cases...
```js
getPrefixEndRange('start'); // -> 'staru'
getPrefixEndRange('X' + String.fromCharCode(0xff)); // -> 'Y'
getPrefixEndRange('xoo' + String.fromCharCode(0xff)); // -> 'xop'
getPrefixEndRange('com.google.'); // -> 'com.google/'
getPrefixEndRange(String.fromCharCode(0xff)); // -> ''
getPrefixEndRange(''); // -> ''
```

callmehiphop:
Closes #1790

### DIFF
--- a/packages/bigtable/test/table.js
+++ b/packages/bigtable/test/table.js
@@ -202,7 +202,7 @@ describe('Bigtable/Table', function() {
     });
   });
 
-  describe('createEndRange_', function() {
+  describe('createPrefixRange_', function() {
     it('should create a range from the prefix', function() {
       assert.deepEqual(Table.createPrefixRange_('start'), {
         start: 'start',


stephenplusplus:
Code looks good to me. I pinged our friend @sindresorhus to hopefully check out the `createPrefixRange_` function. If I recall correctly, character codes and string lengths in JS have been areas of trickery, so I just want to make sure that's gotcha-proof.

PR #1807, opened 2016-11-19
Update README - GCS Signature

cristiancavalli:
Update the readme to match the current GCS
IIFE invocation signature. Without updated
signature invoking on property bucket will
cause an exception to be thrown.


stephenplusplus:
Thanks!

PR #1808, opened 2016-11-20
Add default resource generator for log entries

cristiancavalli:
#### To Dos

- [x] Add environment detection helpers to google-auto-auth: https://github.com/stephenplusplus/google-auto-auth/pull/8
- [x] Update google-cloud/common with new google-auto-auth release
- Tests
  - [x] System
  - Unit
    - [x] log.js
    - [x] metadata.js

To accommodate invocations like:
  `syslog.entry({textPayload: 'test'})`
Create a basic class which attempts to automatically
establish resource values based on connectivity to
the metadata service and environmental variables.

This addition does not affect the existing API and
does not modify or augment user supplied resources.
The meta data class will attempt to determine if the
resource is a cloud function, app engine application
or compute engine instance. If it can't match on any
of these configurations but can still locate a project
id then it will produce a global resource id.

This addition adds behavioral tests and adds a new
Metadata class-specific set of unit tests which reach
100% module coverage:
https://coveralls.io/github/cristiancavalli/google-cloud-node

cc @ofrobots

### DIFF
--- a/scripts/docs/config.js
+++ b/scripts/docs/config.js
@@ -25,6 +25,7 @@ module.exports = {
     'common',
     'datastore/src/entity.js',
     'datastore/src/request.js',
+    'logging/src/metadata.js',
     'pubsub/src/iam.js',
     'storage/src/acl.js',
     'bigtable/src/mutation.js'


packages/logging/src/log.js
```diff
@@ -136,6 +137,27 @@ Log.formatName_ = function(projectId, name) {
   return path + name;
 };
 
+Log.prototype.assignDefaultMetadata_ = function(entry, callback) {
+  var entryMetadata = entry.metadata;
```
stephenplusplus: I believe this function will currently modify the object the user is still holding. We should leave theirs untouched, but decorate it internally.

packages/logging/src/metadata.js
```diff
@@ -0,0 +1,297 @@
...
+/**
+ * Invoke to attempt to retrieve the project id from the current environment or
+ * from the metadata service.
+ *
+ * @param {Function} callback - the error-back style callback to invoke once
+ *  the project id has been determined.
+ * @returns {module:metadata}
+ * @chainable
+ */
+Metadata.prototype.getProjectId = function(callback) {
```
stephenplusplus: Let's wait until this function is called to try to get the project ID from the metadata server.

packages/logging/src/metadata.js
```diff
@@ -0,0 +1,297 @@
...
+}
+
+/**
+ * Attaches core request lifecycle event listeners on the request stream
+ * provided to the constuctor.
+ *
+ * @private
+ */
+Metadata.prototype.attachRequestStreamListeners_ = function(stream) {
+  var self = this;
```
stephenplusplus: Let's make this request in a traditional callback-style:

```js
common.util.makeRequest({
  url: 'http://metadata.google.internal/computeMetadata/v1/project/project-id',
  headers: {
    'Metadata-Flavor': 'Google'
  }
}, function(err, resp) {})`

File: packages/logging/src/metadata.js [hunk 4]
```diff
@@ -0,0 +1,297 @@
...
+ * @chainable
+ */
+Metadata.prototype.getProjectId = function(callback) {
+  var self = this;
+  if (!is.function(callback)) {
+    throw new TypeError(
+      'Must supply a callback function as an argument to invocation.');
+  } else if (!is.empty(self.processProjectId_()) &&
+    is.string(self.processProjectId_())) {
+    // The process's GCLOUD_PROJECT env variable has been set
```
stephenplusplus: Should we also be aware of a project ID explicitly provided by the user during instantiation of the Logging client?

```js
require('@google-cloud/logging')({
  projectId: 'project-id' // <--
})
```

packages/logging/src/log.js
```diff
@@ -136,6 +137,27 @@ Log.formatName_ = function(projectId, name) {
...
+  // If the resource specifcally was not supplied attempt to determine it
+  // automatically
+  if (!is.object(entryMetadata) || !is.object(entryMetadata.resource)) {
+    this.metadata_.getDefaultResource(function(err, resource) {
+      if (err) {
+        callback(err);
```
stephenplusplus: Should we just continue making the request with the object as it was provided by the user, instead of returning an error because we were unable to decorate it? Presumably, if something was required upstream that we were unable to find, the upstream API would have its own error to communicate this, that we would then provide to the user.

stephenplusplus:
I noticed this PR doesn't decorate requests sent through the `write` method, but only ones the severity shortcuts. Is this intentional?

packages/logging/src/metadata.js
cristiancavalli: Where does this come into play? User provided resource data always overwrites auto generated resources so that function the same as it did before, otherwise this change wouldn't affect it?

packages/logging/src/metadata.js
cristiancavalli: Any reason why? This request is normally only a couple bytes in payload and executes fairly quickly. Heating up this path as fast as possible would seem like a good way of not head-of-line blocking a log flush when the request could have been done in the background during startup.

packages/logging/src/metadata.js
cristiancavalli: Any particular reason why? The stream interface decouples the interface nicely, make instrumentation particularly easy, and allows multiple callers to idiomatically listen on the same event without creating extra state inside the class or another 'in-the-middle' event emitter.

packages/logging/src/log.js
cristiancavalli: This has been updated to call-back with the error

packages/logging/src/log.js
cristiancavalli: This function will be updated, but given the path from Log.entry this function will not modify any object from user-space - it will instead modify the object being generate in this function with properties from the user-space object overwriting via extend.

packages/logging/src/metadata.js
stephenplusplus: I simply think that's more complicated than we need. We only need to make the call once and cache the result, so all of the parsing and caching of the pieces of this type of stream request introduces a lot of complexity in the codebase.

packages/logging/src/metadata.js
stephenplusplus: I'd rather wait until the user asks us to do something for us to do something. The delay will be minimal and last-minute is the way we generally work in Node/this library.

packages/logging/src/metadata.js
stephenplusplus: Just trying to follow along with this function. It's trying to find the project ID from various places, but it doesn't handle if the user already provided one manually during instantiation of the class. Is it possible the user has provided one, and still needs the metadata generation features of `getDefaultResource()`?

cristiancavalli:
refactored the call chain to only affect the `write` method

packages/logging/src/metadata.js
cristiancavalli: Sure, but in that case the default resource would still get written to the entry. In essence this wouldn't affect a user-supplied project id in the resource field, since: if one is not provided it will be automatically provided and if one is provided by the user then that one will overwrite the auto generated one in the resource

packages/logging/src/metadata.js
cristiancavalli: So, sorry. But I think making a callback style interface would cause more complexity. Given the case where we have multiple callers I still have to have a way of queuing callbacks on the resolution of the request to the metadata service and with the the callback-style request I would either have to create another event emitter (like the metadata class itself) to emit events too or create another single event to callback invoker on the class which is separate from the request itself. This way the request stream is the event emitter which unifies management of the event chain and reduces unneeded redundancy. This is also the same basic request functionality that the library itself exhibits.

packages/logging/src/metadata.js
```diff
@@ -0,0 +1,297 @@
...
+ * @param {String} projectId - the project id
+ * @returns {Object}
+ */
+Metadata.prototype.produceGAEDescriptor_ = function(projectId) {
+  return {
+    type: 'gae_app',
+    labels: {
+      project_id: projectId,
+      module_id: process.env.GAE_SERVICE,
+      version_id: process.env.GAE_VERSION
```
ofrobots: I think it would make sense to also support AppEngine ManagedVMs here. The module (service) name is available via `GAE_MODULE_NAME` (and the version in `GAE_MODULE_VERSION`)I think it would make sense to also support AppEngine ManagedVMs here. The module (service) name is available via `GAE_MODULE_NAME` (and the version in `GAE_MODULE_VERSION`). The `isAppEngine_` function would also need to be changed above.

packages/logging/src/metadata.js
stephenplusplus: There's actually a major simplification we can make, which is using the auth client to get the project ID. It detects it from all of the same places you are here.

You could cache the Logging instance in the Metadata constructor, and call `logging.authClient.getProjectId(function(err, projectId) {})`.

```js
function Metadata(logging) {
  this.logging = logging
}

Metadata.prototype.getProjectId = function(callback) {
  if (this.logging.projectId) {
    setImmediate(function() { callback(null, projectId) })
    return
  }

  this.logging.getProjectId(callback)
}
```

> Given the case where we have multiple callers I still have to have a way of queuing callbacks on the resolution

How do we end up in that situation, using a code call like this:

```js
log.write([entries], function(err) {})
```

> so no user-space queuing

I'm still opposed to running phantom requests in the background. It should run at the time the user has a request to send.

stephenplusplus: (Although, please consider using the auth client's `getProjectId()` -- if you'd rather leave this for me to implement, that's fine as well.)

cristiancavalli: @stephenplusplus Awesome, would be happy to use the auth libs implementation but I'm trying to get it to work with a GCE instance I'm shelled into and it keep returning null even after setting the env variable:
screen shot 2016-11-21 at 12 08 10 pm

Not sure if this is intended behavior. I also updated the code to use project id's from lib instantiation if given a project id at instantiation.

packages/logging/src/metadata.js
cristiancavalli: @ofrobots is the distinction that App engine apps have a `GAE_SERVICE` & `GAE_VERSION` while managed vm apps have a `GAE_SERVICE`, `GAE_VERSION` and `GAE_MODULE_NAME`?

packages/logging/src/metadata.js
ofrobots: /cc @jmdobry

packages/logging/src/metadata.js
ofrobots: ManagedVMs are the slightly older variant of AppEngine Flexible – you get this variant when you have `vm: true` in your `app.yaml`. As far as I know `GAE_SERVICE` is not defined on ManagedVMs. From the user's point of view both of these are 'AppEngine' and we should magically discover & configure either. /cc @JustinBeckwith in case I got these details wrong.

packages/logging/src/metadata.js
jmdobry: It's not working for you because `google-auth-library-nodejs` `0.9.10` needs to be released.

cristiancavalli: @ofrobots @stephenplusplus Should I go forward with implementing use of `getDefaultProjectID` expecting the auth lib to be released or @stephenplusplus, would you rather implement this change later?

packages/logging/src/metadata.js
cristiancavalli: @ofrobots this code's behaviour has been update to detect this

packages/logging/src/metadata.js
cristiancavalli: Also @ofrobots @stephenplusplus consider that the success of the request to the metadata service is part of the heuristic used to determine whether the library instance is running in a Google Cloud service instance or elsewhere since the metadata service is only addressable from within a Google Cloud environment. Therefore, to maintain the integrity of this check I would still have to make a request to a metadata service endpoint unless something else is doing it already that we can tap into. Therefore, my thought is (given the current state of affairs) that it might be best to keep the project id metadata request in the current code as it serves two purposes, one of which is currently solved by the use of `getDefaultProjectId` since that function will prefer env-based project id's to immediately requesting against a metadata endpoint.

packages/logging/src/metadata.js
cristiancavalli: [sorry posted the comment to the wrong thread]
Also @ofrobots @stephenplusplus consider that the success of the request to the metadata service is part of the heuristic used to determine whether the library instance is running in a Google Cloud service instance or elsewhere since the metadata service is only addressable from within a Google Cloud environment. Therefore, to maintain the integrity of this check I would still have to make a request to a metadata service endpoint unless something else is doing it already that we can tap into. Therefore, my thought is (given the current state of affairs) that it might be best to keep the project id metadata request in the current code as it serves two purposes, one of which is currently solved by the use of getDefaultProjectId since that function will prefer env-based project id's to immediately requesting against a metadata endpoint.

stephenplusplus: I believe a new release of the auth-library is coming today via @jmuk. It would be nice to use this, but I can see the dilemma. @jmdobry is there anything that the auth lib exposes that we can use to confirm the environment we're in?

jmdobry: > is there anything that the auth lib exposes that we can use to confirm the environment we're in?

`google-auth-library-nodejs` makes requests to the metadata service in at least two places:

- [`GoogleAuth#_checkIsGCE(callback)`](https://github.com/google/google-auth-library-nodejs/blob/master/lib/auth/googleauth.js#L329)
- [`GoogleAuth#_getGCEProjectId(callback)`](https://github.com/google/google-auth-library-nodejs/blob/master/lib/auth/googleauth.js#L248)

---- 

`0.9.10` has been published, so `projectId` should be available from `google-auto-auth` once it's using `0.9.10`

cristiancavalli: @jmdobry Those are private functions, right? And the main interface path prefers a process id value over cloud values:
https://github.com/google/google-auth-library-nodejs/blob/6469eea371800205d3f6b0189289f3cf216df5ab/lib/auth/googleauth.js#L136

As it stands unless I call a private member this issue can't be solved from the public interface.
Also, why are those two functions broken apart? Couldn't you elide the success of a call to the metadata project id endpoint with being able to reach the metadata service in the first place and accomplish two objectives in one request?

jmdobry: You are correct. I think `google-auth-library-nodejs` could use a lot of improvement.

cristiancavalli: @jmdobry sorry about the misplacement of that last comment. Thanks for doing the leg-work to get this exposed, despite that nit this is a great solution and I will formalize my thoughts in a pull-request.

@ofrobots @stephenplusplus are either of you opposed to proceeding with the current api usage? I don't mind owning the job of circling back around and updating the api's later once we have solution closer suited to the immediate needs.

packages/logging/src/log.js
```diff
@@ -136,6 +138,42 @@ Log.formatName_ = function(projectId, name) {
...
+    }
+  }
+  return entry;
+};
+
+Log.prototype.getDefaultMetadata_ = function(entry, callback) {
+  var self = this;
+  this.metadata_.getDefaultResource(function(err, resource) {
+    if (err) {
+      callback(err);
```
ofrobots: Need a return here. Else we fall into a second call of the same callback.

cristiancavalli: Thanks!

packages/logging/src/metadata.js
cristiancavalli: hey @stephenplusplus thanks a ton for doing the legwork to get this integrated - your new changes look great. If I come up with any thoughts I'll post but as of now your changes lgtm

stephenplusplus: Cool, thanks for checking it out. I pushed the changes here so it's easier for others to review as well.

packages/logging/src/metadata.js
```diff
@@ -0,0 +1,185 @@
...
+Metadata.prototype.getDefaultResource = function(callback) {
+  var self = this;
+
+  this.getProjectId(function(err, projectId) {
+    if (err) {
+      callback(err, null);
+      return;
+    }
+
+    self.logging.authClient.determineEnvironment(function(err, env) {
```
ofrobots: `s/determineEnvironment/getEnvironment/` based on stephenplusplus/google-auto-auth#8.

stephenplusplus:
@google-cloud-common@0.8.1 released so work is free to continue on this. Feel free to test it out in the various environments to catch any issues.

stephenplusplus:
@cristiancavalli can you try this out on the various environments to see that it's working as intended? I've made a small zip you can run (edit the `env` object in `system-test/logging.js`, `npm install && npm run system-test` after installing), https://storage.googleapis.com/stephen-has-a-new-bucket/pr-1808-test.zip

cristiancavalli:
@stephenplusplus Sorry for the latency -- just finished testing on GCE and all passes test

stephenplusplus:
This is ready for a final review. @callmehiphop PTAL.

packages/logging/src/log.js
```diff
@@ -591,14 +609,22 @@ Log.prototype.write = function(entry, options, callback) {
...
-  if (!(entry instanceof Entry)) {
-    entry = this.entry(entry);
-  }
+Log.prototype.decorateEntries_ = function(entries, callback) {
+  var self = this;
 
-  var formattedEntry = entry.toJSON();
-  formattedEntry.logName = this.formattedName_;
-  return formattedEntry;
+  async.map(entries, function(entry, callback) {
```
callmehiphop: `async` is a dev dependency, so we'll either need to re-classify it OR we could probably just use promises here.

stephenplusplus: Phew, good catch. I updated it to be a non-dev dependency. I'm not opposed to promises, though, I just wasn't sure how to write it. How would it look?

stephenplusplus: Thanks for writing that up. I think let's just stick with the current way, then consider doing a `async.map` -> `Promise` sweep across the repo at a time in the future.

callmehiphop: Sounds good.

callmehiphop:
@stephenplusplus I think we need to stub the auth client or something, I'm seeing documentation failures in regards to grabbing the projectId.

stephenplusplus:
Pushed a fix that's a little bit sad. Let me know if you know a better way.

callmehiphop:
@stephenplusplus I'm ok with that if you are. Alternatively I think we could just stub `google-auto-auth` in the docs spec file. Something like..

```js
var FakeAuth = function() {
  var Auth = require('google-auto-auth');
  Auth.prototype.getProjectId = function(cb) {
    cb(null, 'grape-spaceship-123');
  };
  return Auth;
};
```

stephenplusplus:
I went that way originally, but couldn't get it to work :( That is definitely a better solution, but let me know if you know a better way to do it than:

```diff
--- a/test/docs.js
+++ b/test/docs.js
@@ -57,6 +57,14 @@ var JSHintError = createErrorClass('JSHintError', function(err) {
   this.code = err.code;
 });

+var FakeAuth = function() {
+  var Auth = require('google-auto-auth');
+  Auth.prototype.getProjectId = function(cb) {
+    cb(null, 'grape-spaceship-123');
+  };
+  return Auth;
+};
+
 var FakeConsole = Object.keys(console)
   .reduce(function(console, method) {
     console[method] = noop;
@@ -245,6 +253,7 @@ function createSnippet(mod, instantiation, method) {
       'keyFilename: \'/path/to/keyfile.json\'',
       'keyFilename: \'\''
     )
+    .replace('require(\'google-auto-auth\')', FakeAuth.toString())
     .replace('require(\'express\')', FakeExpress.toString())
     .replace('require(\'level\')', FakeLevel.toString())
     .replace('require(\'bluebird\')', FakeBluebird.toString())
```

callmehiphop:
Ah, right. We only swap out dependencies in the examples themselves.. I think if we wanted to go this route we'd probably have to run some code in the sandbox that uses proxyquire. So I think this is fine for now and maybe I'll try and take a crack at implementing proxyquire later.

callmehiphop:
Can we add <kbd>logging/metadata.js</kbd> to the ignore list within <kbd>scripts/docs/config.js</kbd>?

PR #1814, opened 2016-11-22
promisify Storage ACL methods

### LINKED ISSUE #1813: Storage ACL helper methods have not been promisified
e.g.

```js
// throws "Uncaught TypeError: callback is not a function"
bucket.acl.readers.addUser('foo@test.com')
  .then(() => {});
```

jmdobry:
Fixes #1813


callmehiphop:
@jmdobry I don't think we need to re-promisify here, so I added a commit with a different solution. PTAL!

callmehiphop:
@stephenplusplus PTAL

jmdobry:
Any ETA on when this will get released?

PR #1815, opened 2016-11-22
Log entries may be show up in the wrong order

ofrobots:
This PR changes an existing test to demonstrate the problem. (I haven't quite managed to figure out how to run the `system-test`s myself, but I think it should repro the problem that I encountered on a standalone app).

The core of the problem is that if the client doesn't provide a timestamp, the logging service will add the timestamps based on the time of receipt which may be subject to local and network latencies. This may mean that the log entries may be out of order w.r.t. to the source code they were submitted from.

The fix would be to add a timestamp in the `Entry` constructor or on `Entry.toJSON` (if one doesn't exist).

### DIFF
--- a/packages/logging/system-test/logging.js
+++ b/packages/logging/system-test/logging.js
@@ -348,24 +348,20 @@ describe('Logging', function() {
     });
 
     it('should preserve order of entries', function(done) {
-      var logEntries = [
-        log.entry('1')
-      ];
+      var entry1 = log.entry('1');
 
       setTimeout(function() {
-        logEntries.push(log.entry('2'));
+        var entry3 = log.entry('3');
+        var entry2 = log.entry({ timestamp: entry3.metadata.timestamp }, '2');
 
-        log.write([logEntries[1], logEntries[0]], options, function(err) {
+        // Re-arrange to confirm the timestamp is sent and honored.
+        log.write([entry2, entry3, entry1], options, function(err) {
           assert.ifError(err);
 
           setTimeout(function() {
-            log.getEntries({
-              pageSize: 2
-            }, function(err, entries) {
+            log.getEntries({ pageSize: 3 }, function(err, entries) {
               assert.ifError(err);
-
-              assert.deepEqual(entries.map(prop('data')), [ '2', '1' ]);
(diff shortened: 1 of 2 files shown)


ofrobots:
I can add the fix to the PR as well. Any preference on the fix? I think the constructor is the better place to add a timestamp in case people queue/batch entries.

stephenplusplus:
That sounds like a good plan to me, thanks!

ofrobots:
PTAL.

stephenplusplus:
@ofrobots I made a couple of changes to the tests & the way we cache the user's provided timestamp in the Entry constructor. PTAL.

packages/logging/test/entry.js
```diff
@@ -44,8 +44,23 @@ describe('Entry', function() {
   });
 
   describe('instantiation', function() {
-    it('should localize metadata and data', function() {
-      assert.strictEqual(entry.metadata, METADATA);
+    it('should assign timestamp to metadata', function() {
+      assert.deepEqual(entry.metadata, extend({
+        timestamp: new Date()
```
ofrobots: I suspect this might spuriously fail depending on timing / machine load. The new Date object may not equal the date captured by the constructor a few statements ago.

stephenplusplus: Yes, good call. I will add some breathing room.

packages/logging/system-test/logging.js
```diff
@@ -347,6 +347,32 @@ describe('Logging', function() {
       });
     });
 
+    it('should preserve order of entries', function(done) {
+      var logEntries = [
+        log.entry('1')
+      ];
+
+      setTimeout(function() {
+        logEntries.push(log.entry('2'));
```
ofrobots: I would suggest adding a third entry here perhaps with the same timestamp as '2'. In such a case the retrieval order should match the insertion order.

stephenplusplus:
@ofrobots I've updated the tests, ptal.

ofrobots:
Thanks of updating! LGTM.

stephenplusplus:
Thanks for the fix!

Issue #1816, opened 2016-11-23
Regenerate GAPIC files


jmuk:
This issue tracks the regeneration process of gapic (auto-generated files). The new auto-generated files have some updates, including:
- class name suffix has changed. Previously `-Api`, now `-Client` (like `speechClient`).
- revised paged iterations style -- previously a stream, now same as what GCN has.
- bug fix on enum values.

Google-cloud-node repository has the the generated files in the following APIs.

- [x] language
- [x] logging
- [x] monitoring
- [x] pubsub
- [x] speech
- [x] vision

stephenplusplus:
Woo!

PR #1829, opened 2016-11-28
Regenerate monitoring API client.

jmuk:
Updates #1816

### DIFF
--- a/packages/monitoring/src/v3/index.js
+++ b/packages/monitoring/src/v3/index.js
@@ -19,7 +19,6 @@ var groupServiceClient = require('./group_service_client');
 var metricServiceClient = require('./metric_service_client');
 var gax = require('google-gax');
 var extend = require('extend');
-var union = require('lodash.union');
 
 function v3(options) {
   options = extend({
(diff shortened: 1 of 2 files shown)


packages/monitoring/src/v3/index.js
```diff
@@ -15,21 +15,25 @@
...
-  extend(result, groupServiceApi(gaxGrpc));
-  extend(result, metricServiceApi(gaxGrpc));
+  extend(result, groupServiceClient(gaxGrpc));
+  extend(result, metricServiceClient(gaxGrpc));
   return result;
 }
-v3.SERVICE_ADDRESS = groupServiceApi.SERVICE_ADDRESS;
-v3.ALL_SCOPES = [];
+v3.SERVICE_ADDRESS = groupServiceClient.SERVICE_ADDRESS;
+v3.ALL_SCOPES = union(
```
stephenplusplus: This has occurred a couple of other times. `union` is an unnecessary dependency, as `ALL_SCOPES` are the same between the services.

jmuk: Done

PR #1830, opened 2016-11-28
Add winston logging transport

ofrobots:
* Pre-req: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1808. (commits from that PR will show up here too until it lands.)
* `npm run docs` parser doesn't like my docs yet (is there any info/docs on what the custom parser requires?)

This PR adds a new module `@google-cloud/logging-winston` that provides simple, minimal config way of logging to Stackdriver Logging from applications using winston.

The code is fairly small, but it should be a standalone module because we need to depend upon winston (the transport must be an instance of `winston.Transport`). Adding this dependency to `@google-cloud/logging` doesn't make sense (we should not install winston for people who are not asking for winston support).

/cc @cristiancavalli

### DIFF
--- a/scripts/docs/config.js
+++ b/scripts/docs/config.js
@@ -59,7 +59,7 @@ module.exports = {
     logging: {
       title: 'Google Cloud Logging'
     },
-    loggingWinston: {
+    'logging-winston': {
       title: 'Google Cloud Logging Winston'
     },
     monitoring: {


packages/logging-winston/src/index.js
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  // Translate `level` to a Stackdriver logging level.
+  var numericLevel = this.levels_[level];
+  var stackdriverLevel = levelNames[numericLevel];
+
+  var metadata = this.metadata_;
+  if (meta) {
+    // Logging proto requires that the label values be strings, so we convert
+    // using util.inspect.
```
stephenplusplus: Nevermind! I was mistaken while reading the code how the user-provided `metadata` and the cached `this.metadata_` were used together, but I'm with it now 👍

stephenplusplus: Do you think we should use `JSON.stringify` instead of `util.inspect`?:

```js
> JSON.stringify({ hey: 'there' })
'{"hey":"there"}'
> util.inspect({ hey: 'there' })
'{ hey: \'there\' }'
```

That's what we do in other places in the library, but I'm not sure if `util.inspect` is better here.

stephenplusplus:
@ofrobots I've added a commit. The main differences are,

- code style (in docs & JSDocs)
- examples (using `new` to instantiate winston & the transport -> `winston.add`)

Can you take a quick look and make sure it still behaves as intended?

packages/logging-winston/src/index.js
```diff
@@ -16,174 +16,153 @@
...
  *     appropriate Stackdriver logging severity level.
  * @param {string} msg - The message to be logged.
- * @param {object=} meta - Winston provided metadata that should be attached to
- *     log entry. Each property will be converted to a string using
- *     `util.inspect`.
- * @param {function=} callback A callback that is invoked when the logging agent
- *     either succeeds or gives up writing the log entry to the remote server.
+ * @param {object=} metadata - Winston-provided metadata that should be attached
+ *     to the log entry. Each property will be converted to a string using
+ *     `JSON.stringify`.
```
ofrobots: I think we should use `util.inspect` as it is more idiomatic with the built-in winston loggers, e.g. File and Console. See https://cloud.google.com/compute/docs/authentication#using

ofrobots: As a follow-on, I think we should parameterize this to allow users to pass in a custom string conversion function. For now, I think we should use `util.inspect`.

packages/logging-winston/src/index.js
```diff
@@ -16,174 +16,153 @@
...
+ * @param {function=} callback - A callback that is invoked when the logging
+ *     agent either succeeds or gives up writing the log entry to the remote
+ *     server.
  */
-StackdriverTransport.prototype.log = function(level, msg, meta, callback) {
-  if (_.isFunction(meta) && !callback) {
-    callback = meta;
-    meta = null;
+LoggingWinston.prototype.log = function(levelName, msg, metadata, callback) {
```
ofrobots: I find it really confusing that `metadata` new refers to objects of two very different types. One is a bag of properties provided by winston, but later in this function we change it to the `metadata` object as defined by the @google-cloud/logging library. I guess `meta` wasn't a lot better. Perhaps we can use `winstonMetadata` here and use the term `metadata` to refer to the concept from @google-cloud/logging?

packages/logging-winston/src/index.js
```diff
@@ -16,174 +16,153 @@
...
-  var numericLevel = this.levels_[level];
-  var stackdriverLevel = levelNames[numericLevel];
+  var levelCode = this.levels_[levelName];
+  var stackdriverLevel = STACKDRIVER_LOGGING_LEVEL_CODE_TO_NAME[levelCode];
+
+  var labels = {};
 
-  var metadata = this.metadata_;
-  if (meta) {
+  if (metadata) {
```
ofrobots: `metadata` guaranteed to be truthy here.

stephenplusplus: Will it always be an object?

packages/logging-winston/src/index.js
stephenplusplus: That was confusing indeed. I went the other way around and instead of changing `metadata` to an object expected by `log.entry()`, I created a new var named `entryMetadata`. PTAL.

packages/logging-winston/src/index.js
stephenplusplus: Gotcha, switched it back.

packages/logging-winston/src/index.js
```diff
@@ -16,174 +16,153 @@
...
+ *   level: 'warning', // log at 'warning' and above
+ *   resource: {
+ *     type: 'global'
+ *   }
  * });
- * logger.emerg('antimatter containment field collapse in 10 seconds');
+ *
+ * winston.emerg('antimatter containment field collapse in 10 seconds');
  */
+function LoggingWinston(options) {
```
stephenplusplus: That seemed wrong to me, because as independent modules, we usually try not to overwrite anything. E.g., if some other module writes their own StackdriveLogging implementation with the same name, they could end up overwriting ours from a dependency of a dependency of a dependency that a user doesn't even know they have.

If it is in fact recommended to add on to `winston.transports`, or I'm just misunderstanding the implications, let me know and I'll switch it back.

ofrobots: It is recommended to add the transport to the list of transports. See:  https://github.com/winstonjs/winston#adding-custom-transports. Yes, it is possible that another module may overwrite ours – and I don't have an easy answer for this.

packages/logging-winston/src/index.js
ofrobots: Good point.

packages/logging-winston/src/index.js
stephenplusplus: Okay, I'll add it back.

packages/logging-winston/README.md
```diff
@@ -1,44 +1,79 @@
...
-### Optional Configuration
+## Authentication
+
+It's incredibly easy to get authenticated and start using Google's APIs. You can set your credentials on a global basis as well as on a per-API basis. See each individual API section below to see how you can auth on a per-API-basis. This is useful if you want to use different accounts for different Google Cloud services.
 
-If your app is running on Google Cloud Platform, we handle all configuation & authentication for you.  We also auto-detect the appropriate [resource descriptor](todo://add/link) to report the log entries against.
+### On Google Compute Engine
 
-If you are running your application in a different environment, for example, locally, on-premise, or on another cloud provider, you may need to provide some configuration.
+If you are running this client on Google Compute Engine, we handle authentication for you with no configuration. You just need to make sure that when you [set up the GCE instance][gce-how-to], you add the correct scopes for the APIs you want to access.
```
ofrobots: I thought I added a comment for this earlier.. not sure where it went... oh well. (apologies if this is a duplicate)

Any specific reason to change 'Google Cloud Platform' to 'Google Compute Engine'. The instructions do work across all of Google Cloud Platform, including AppEngine, GKE, Functions and Compute Engine.

Secondly, I am not sure if what is meant by 'add the correct scopes'; users don't know what scopes are and how to add them. The linked page doesn't mention the term 'scopes' at all. Regardless, the cloud logging API is enabled by default, so this statement cam simply be removed.

stephenplusplus: Oops, I didn't mean to squash that. In fact, I changed it from GCE to GCP across all the readmes in a PR today.

When you create an instance through the console, one of the options presented is setting the scopes. It's at that time I expect the user to understand the connection. But thanks for pointing out that it's not required here, I'll remove it. And suggestions for how to improve the docs are welcome as well, as the other APIs do require a user to enable the correct scopes.

stephenplusplus:
I'll finish up the unit tests for this. Are we including this in the google-cloud bundle, e.g.:

```js
var gcloud = require('google-cloud')();
var transport = gcloud.loggingWinston;
```

Also, is there a better name than `logging-winston`? Generally, var names coincide with the module name, so this isn't ideal:

```
var stackdriverTransport = require('@google-cloud/logging-winston')
```

ofrobots:
/cc @JustinBeckwith @omaray for more opinions on the name. 

This library isn't a client library but rather an integration library. I think the general rule of var names coinciding doesn't have to apply here. Module names are primarily for discoverability of purpose, so I still prefer `@google-cloud/logging-winston`. The fact that this returns a winston transport would make complete sense to winston users.

Perhaps we should change the examples to the following to make things more elegant:

```js
var transport = require('@google-cloud/logging-winston');
```

I don't have very strong opinions on bundling this with `google-cloud` as I am not too familiar with the precedents in this library. My personal preference would be to *not* include integration libraries in the default bundle. The users who need to use it can add it explicitly.

ofrobots:
I've improved the examples.

ofrobots:
@stephenplusplus I've add some more unit tests. PTAL.

stephenplusplus:
I made some alterations from your last commit with regards to the input validation and testing strategy, keeping it more in line with the conventions here-- letting validation happen "last-minute", by an upstream module or API request, meaning we try to only throw if the next line of code in our script wouldn't work if they gave us something invalid. And with the tests, we stub everything to make sure the tests are as isolated as possible from logic that exists anywhere else.

PTAL!

ofrobots:
Sorry for the delay. LGTM.

ofrobots:
When should we expect this to land?

ofrobots:
Thanks for landing. Can we publish a package for this too?

stephenplusplus:
Published as `0.1.0`.

PR #1836, opened 2016-11-29
language: remove sentiment and salience range conversions

### LINKED ISSUE #1834: language: Sentiment score range is inconsistent with API documentation
This isn't a bug per se, but an unnecessary inconsistency IMO between the `sentiment.score` returned by the Node client and the actual response returned by the Natural Language API.

According to the [Sentiment Analysis Response Fields API documentation](https://cloud.google.com/natural-language/docs/basics#sentiment_analysis_response_fields):

> `score` of the sentiment ranges between `-1.0` (negative) and `1.0` (positive)

However, the Node client currently takes the API response and converts the `score` to a range of `-100` to `100`, as [described here](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.44.0/language/document?method=detectSentiment).

IMO this inconsistency between the API and the Node client adds unnecessary friction to the development process and opens up the possibility of misuse if the client documentation isn't read carefully. (Ie. A developer who has read the API documentation writes code using the Node client that marks text with a `sentiment.score` greater than `0.8` as "Positive", as mentioned [here in the API docs](https://cloud.google.com/natural-language/docs/basics#interpreting_sentiment_analysis_values))

I'd propose that the Node client _not_ change the expected `score` range and leave it as `-1.0` to `1.0`. I'd be happy to make a pull request with the [offending line](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/ad730415b2272b77120169f69e8630ce1ca533b8/packages/language/src/document.js#L1055) changed, but wanted to bring this up here first since this would be a semi-breaking change. (Good thing it's still in beta!)

sawyerh:
Closes #1834.

While I was in there I noticed that the `salience` range was also being changed from `0.0`-`1.0` to `0`-`100`, so I changed that as well to be consistent with the [API documentation](https://cloud.google.com/natural-language/docs/basics#entity_analysis_response_fields).

🚨 This will break anyones' implementation that used `salience` or `sentiment`. For example, previously a `sentiment` of `1` would be interpreted as "Negative", however now a score of `1` is  "Positive". Sorry. In the long run this should prevent confusion and accidental bugs, so IMO it's worth the pain while this is still under the "beta" status.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -23,6 +23,7 @@ Jesse Friedman <jesse@jesse.ws>
 Johan Euphrosine <proppy@google.com>
 Marco Ziccardi <marco.ziccard@gmail.com>
 Patrick Costello <pcostell@google.com>
+Sawyer Hollenshead <hi@sawyerh.com>
 Silvano Luciani <silvano@google.com>
 Stephen Sawchuk <sawchuk@gmail.com>
 Thomas Rognon <tcrognon@gmail.com>
--- a/packages/language/src/document.js
+++ b/packages/language/src/document.js
@@ -252,7 +252,7 @@ Document.PART_OF_SPEECH = {
  * @param {string} callback.annotation.language - The language detected from the
  *     text.
  * @param {number} callback.annotation.sentiment - A value in the range of
- *     `-100` to `100`. Large numbers represent more positive sentiments.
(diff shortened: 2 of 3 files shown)


stephenplusplus:
Thank you!

PR #1844, opened 2016-11-29
add node@7 to CI

stephenplusplus:


stephenplusplus:
Just when you thought builds couldn't get any slower...

screen shot 2016-11-29 at 4 02 25 pm

PR #1847, opened 2016-11-30
common: emit response event on correct stream

### LINKED ISSUE #1846: [BIGTABLE] Insert callback called multiple times if err
#### Environment details
  - OS: macOS Sierra 10.12.11
  - Node.js version: 6.9.1
  - npm version: 3.10.8
  - google-cloud-node version: "@google-cloud/bigtable": "^0.6.0"

#### Steps to reproduce
I use "insert" function to insert an array of 2 entries, for the example I force error with "wrong" timestamp (yes timestamp defined in milliseconds is invalid...)
```javascript
let i = 0;
const INSTANCE_NAME = 'myInstance';

require('@google-cloud/bigtable')().instance(INSTANCE_NAME).table('test').insert([{
  key: 'alincoln',
  data: {
    follows: {
      gwashington: {
        value: 1,
        timestamp: 1480517318949 // Invalid timestamp (note 1480517318 in seconds is OK, 1480517318949000 in microseconds is OK)
      }
    }
  }
},{
  key: 'alincoln',
  data: {
    follows: {
      helloworld: {
        value: 1,
        timestamp: 1480517318949 // Invalid timestamp (note 1480517318 in seconds is OK, 1480517318949000 in microseconds is OK)
      }
    }
  }
}], err => {
  i++;
  console.log('--');
  console.log(`CALLBACK ${i}`);
  if (err) console.error(err.message);
  console.log('--');
});

/*
--
CALLBACK 1
--
--
CALLBACK 2
Error in field 'entries' : Error in element #0 : Error in field 'Mutation list' : Error in element #0 : Timestamp granularity mismatch. Expected a multiple of 1000 (millisecond granularity), but got 1480517318949.
--
*/
````
In source file "table.js" I just add some logs to find source of error, callback is called multiple times, first time with end event, and last with error event
If I use promise syntax I just never see that an error occurred because promise is resolved with end event without error

```javascript
// table.js - line 868
this.requestStream(grpcOpts, reqOpts)
  .on('error', (err) => {
    console.log('event:error', err); // LAST_LOG: event:error { Error: Error in field 'entries' : Error in element #0 : Error in field 'Mutation list' : Error in element #0 : Timestamp granularity mismatch. Expected a multiple of 1000 (millisecond granularity), but got 1480517318949.
    callback(err);
  })
  .on('data', function (obj) {
    obj.entries.forEach(function (entry) {
      // Mutation was successful.
      if (entry.status.code === 0) {
        return;
      }

      var status = common.GrpcService.decorateStatus_(entry.status);
      status.entry = entries[entry.index];

      mutationErrors.push(status);
    });
  })
  .on('end', function () {
    var err = null;

    if (mutationErrors.length > 0) {
      err = new common.util.PartialFailureError({
        errors: mutationErrors
      });
    }

    console.log('event:end', err); // FIRST_LOG: event:end null
    callback(err);
  });
```

stephenplusplus:
Fixes #1846 

We were incorrectly emitting the `response` event on the gRPC request stream. As for why this fixes #1846... not so sure. Possibly a quirk of the inner-workings of the native Stream event handlers.

### DIFF
--- a/packages/common/src/grpc-service.js
+++ b/packages/common/src/grpc-service.js
@@ -361,7 +361,7 @@ GrpcService.prototype.requestStream = function(protoOpts, reqOpts) {
           // https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1444#discussion_r71812636
           var grcpStatus = GrpcService.decorateStatus_({ code: 0 });
 
-          this.emit('response', grcpStatus);
+          stream.emit('response', grcpStatus);
         });
     }
   };
--- a/packages/common/test/grpc-service.js
+++ b/packages/common/test/grpc-service.js
@@ -1178,14 +1178,12 @@ describe('GrpcService', function() {
           return options.request();
         };
 
-        fakeStream
-          .on('error', done)
+        grpcService.requestStream(PROTO_OPTS, REQ_OPTS)
           .on('response', function(resp) {
             assert.deepEqual(resp, GrpcService.GRPC_ERROR_CODE_TO_HTTP[0]);
             done();
           });
 
-        grpcService.requestStream(PROTO_OPTS, REQ_OPTS);
         fakeStream.emit('metadata');
       });
 


stephenplusplus:
@harscoet can you test this out? Find `node_modules/@google-cloud/bigtable/src/table.js` and find/replace:

```diff
- this.emit('response', grcpStatus);
+ stream.emit('response', grcpStatus);
```

harscoet:
It fixes when error occured but now success callback is never called :/

stephenplusplus:
Confirmed, the solution in this PR was a mistake. The original behavior was correct. I'll re-open your original issue and keep digging. Sorry about this!

PR #1848, opened 2016-11-30
Fix Bigtable docs references to "instances".

mbrukman:
Remove confusion between creating a "Compute Instance" (which is a GCE VM
instance) vs. a Cloud Bigtable instance and cluster (which is what we need
here).

Updated the link to point to the correct page in the documentation, as Bigtable
also has documentation for creating GCE VM instances with the correct scopes to
access Bigtable instances, but that's not what we want here, and not what the
code sample is trying to accomplish.

### DIFF
--- a/packages/bigtable/src/index.js
+++ b/packages/bigtable/src/index.js
@@ -71,10 +71,10 @@ var Instance = require('./instance.js');
  * //-
  *
  * //-
- * // <h3>Creating a Compute Instance</h3>
+ * // <h3>Creating a Bigtable Instance and Cluster</h3>
  * //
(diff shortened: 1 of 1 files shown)


mbrukman:
Note: this is a comment-only change; it would be nice, if possible, to make the 0.6.0 docs reflect this as well, unless there's an imminent release of the library coming soon (though this technically affects all releases of the Bigtable Node library docs, as it's just a terminology confusion, not a change in the API).

callmehiphop:
Thanks @mbrukman, I hadn't realized this changed. I think we might have a release for Bigtable come up, so I'm going to hold off on re-generating the docs.

PR #1861, opened 2016-12-06
Update speech to use new GAPIC LRO functionality.

landrito:
cc @jmuk.

### DIFF
--- a/packages/speech/test/index.js
+++ b/packages/speech/test/index.js
@@ -958,10 +958,41 @@ describe('Speech', function() {
         }
       };
 
+      it('should format the results', function(done) {
+        speech.api.Speech = {
+          asyncRecognize: function(reqOpts, callback) {
+            var operation = through.obj();
+            callback(null, operation, apiResponse);
+          }
+        };
+
+        var result = {
+          results: []
+        };
+
+        var formattedResults = [];
+        Speech.formatResults_ = function(results, verboseMode) {
+          assert.strictEqual(results, result.results);
+          assert.strictEqual(verboseMode, false);
+          return formattedResults;
+        };
+
+        speech.startRecognition(FILE, CONFIG, function(err, operation) {
+          assert.ifError(err);
+
+          operation.emit('complete', result, null, null, function(err, resp) {
+            assert.ifError(err);
+            assert.strictEqual(resp, formattedResults);
+            done();
+          });
+        });
+      });
+
       it('should format results in verbose mode', function(done) {
         speech.api.Speech = {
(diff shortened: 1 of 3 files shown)


landrito:
@stephenplusplus This should be ready now (depending on the status of CI).

I removed some tests who's granularity is covered by the `Operation` tests in gax-nodejs.

[googleapis/toolkit#823](https://github.com/googleapis/toolkit/pull/823) and [googleapis/gax-nodejs#82](https://github.com/googleapis/gax-nodejs/pull/82) have been merged. This is no longer blocked.

stephenplusplus:
Thanks! I gave it a quick sweep for style & tests. PTAL.

Also, is there any way to get an `Operation` from the generated layer? We have `speech.operation('operation-name')` that returns our `GrpcOperation` type. The user can call `getMetadata()`, `on('complete')`, etc from this.

landrito:
There is not a way for creating an operation from the generated client or through gax-node other than through the api call itself. I suppose I could expose the [constructor](https://github.com/googleapis/gax-nodejs/blob/master/lib/longrunning.js#L131) for the `Operation` in gax-nodejs. This leads to another problem in that another `OperationsClient` will need to be provided.

stephenplusplus:
Yes, that's true. It's not something we have to solve for this PR, but going forward, that will help us eliminate `GrpcOperation`.

If this PR looks good to you, I'll go ahead and merge pending the outcome of the tests.

packages/speech/test/index.js
```diff
@@ -958,10 +958,41 @@ describe('Speech', function() {
...
+        };
+
+        speech.startRecognition(FILE, CONFIG, function(err, operation) {
+          assert.ifError(err);
+
+          operation.emit('complete', result, null, null, function(err, resp) {
+            assert.ifError(err);
+            assert.strictEqual(resp, formattedResults);
+            done();
+          });
```
landrito: @stephenplusplus I'm a bit confused whats going on here. Why is a callback being passed into the emit?

I'm thinking this check would be written as such:
```js
operation.on('complete', function(resp) {
  assert.strictEqual(resp, formattedResults);
  done();
});
operation.on('error', function(err) {
  assert.ifError(err);
});
operation.emit('complete', result, null, null);
```

landrito:
I see. I'll file an [issue](https://github.com/googleapis/gax-nodejs/issues/83) to gax-nodejs to track exposing the creation of Operations.

The PR LGTM! Thanks!

stephenplusplus:
Great, thank you!

PR #1865, opened 2016-12-06
Add system-test for GAPIC files for logging API.

jmuk:
This is hand-written for now. Eventually, we will replace this
by automatically generated ones.

### DIFF
--- a/packages/logging/system-test/v2.js
+++ b/packages/logging/system-test/v2.js
@@ -57,9 +57,8 @@ describe('loggingV2', function() {
       var resourceName = client.parentPath(projectId);
 
       var myLogMessage = 'My new test message';
-      var resource = {'type': 'global'};
+      var resource = {type: 'global'};
       var entries = [{textPayload: myLogMessage, logName: logName}];
-      var start = new Date();
       var filter = 'logName="' + logName + '"';
       client.writeLogEntries({
         entries: entries,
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Some linting errors:

```

packages/logging/system-test/v2.js: line 62, col 11, 'start' is defined but never used.
packages/logging/system-test/v2.js: line 113, col 37, Strings must use singlequote.
packages/logging/system-test/v2.js: line 124, col 7, Missing semicolon.
packages/logging/system-test/v2.js: line 186, col 78, 'resp' is defined but never used.
``

jmuk:
Fixed lint errors.

stephenplusplus:
@jmuk is it possible these tests are better suited for the gax repo? As I understand this, these tests are certifying that the gax library works. But, we should always be able to assume it works, because there wouldn't have been a gax release unless tests like these passed. Does this make sense?

jmuk:
This is a bit different from gax-nodejs repo because it tests API-agnostic logic, while our code generator tests the generated files, character-by-character. This PR verifies the behaviors of generated files -- i.e. whether it's syntactically correct or not, or whether it can issue API calls correctly or not.

However we have another repository `api-client-staging` to store the autogen results (and actually I put the baseline of this PR already there, as https://github.com/googleapis/api-client-staging/tree/master/test/nodejs/logging). If that's good enough, I'll keep using that repository for those tests for autogen results.

Feel free to close this if that sounds good.

stephenplusplus:
Thanks, `api-client-staging` sounds like the perfect place 👍

PR #1867, opened 2016-12-06
Fix in Speech System Test.

landrito:
I misunderstood what it.only does (I thought it was adding more context to the test not to cause mocha to only run the test). I removed it to allow all of the rest of the system tests to run.


stephenplusplus:
Haha, I'm surprised it wasn't me who did this. It's my signature move.

PR #1868, opened 2016-12-07
bigquery: don't promisify date methods & add tests

stephenplusplus:
This adds tests as suggested here: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1854#discussion_r91230014

It also caught a bug where we were promisifying the `bigquery.date()` (and related) methods.

PR #1869, opened 2016-12-07
bigquery: fix structs for SQL parameter queries

stephenplusplus:
https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1868#issuecomment-265532500

Structs weren't being prepared correctly. We now have system tests for every type.

(Thanks, @c0b!)

### DIFF
--- a/packages/bigquery/src/index.js
+++ b/packages/bigquery/src/index.js
@@ -250,50 +250,51 @@ BigQuery.prototype.timestamp = function(value) {
  * @return {string} - The type detected from the value.
  */
 BigQuery.getType_ = function(value) {
+  var typeName;
+
   if (value instanceof BigQuery.date) {
-    return 'DATE';
+    typeName = 'DATE';
+  } else if (value instanceof BigQuery.datetime) {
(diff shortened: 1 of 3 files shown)


packages/bigquery/src/index.js
```diff
@@ -250,50 +250,51 @@ BigQuery.prototype.timestamp = function(value) {
...
+      'This value could not be translated to a BigQuery data type.',
+      value
+    ].join('\n'));
   }
 
-  throw new Error([
-    'This value could not be translated to a BigQuery data type.',
-    value
-  ].join('\n'));
+  return {
```
callmehiphop: Out of curiosity why do we prefer this style over the previous?

stephenplusplus: We used to early return with the string value. Now that we are returning an object, there was too much repetition.

callmehiphop:
Do we want to document this for the users?

stephenplusplus:
Do you mean show how to use a struct in a query? I don't think we need to show every type, since we also have arrays, Booleans, etc.

PR #1873, opened 2016-12-08
common: only set projectId if it is available

stephenplusplus:


packages/common/src/grpc-service.js
```diff
@@ -757,7 +757,9 @@ GrpcService.prototype.getGrpcCredentials_ = function(callback) {
       grpc.credentials.createFromGoogleCredential(authClient)
     );
 
-    self.projectId = authClient.projectId;
+    if (authClient.projectId) {
```
stephenplusplus: Yep, definitely. One sec.

PR #1876, opened 2016-12-09
Introducing a logger for common use

matthewloring:
This is a trimmed down version of the logger used by the cloud
diagnostics common repository with the debug agent specific operations
removed.

### DIFF
--- a/packages/common/src/logger.js
+++ b/packages/common/src/logger.js
@@ -1,5 +1,5 @@
-/**
- * Copyright 2014, 2015 Google Inc. All Rights Reserved.
+/*!
+ * Copyright 2016 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


ofrobots:
@stephenplusplus This adds a really simple logging functionality that we want to use in some upcoming PRs for Stackdriver diagnostics libraries. PTAL.

stephenplusplus:
Thanks! What about using something like https://github.com/cainus/logdriver?

ofrobots:
One key feature we want to have is the ability to include a 'sub-system' or module name in the log output. Since we are logging to the user's stdout, it would be nice to indicate where the log is coming from. Example:

```
ERROR:@google-cloud/debug: Unable to re-register debuggee, statusCode 503.
```

The code, otherwise, is fairly similar.

stephenplusplus:
You can use the `format` function to control exactly what gets printed:

```js
var gcloudLogger = require('log-driver')({
  levels: [
    'error',
    'warn',
    'info',
    'debug',
    'silly'
  ],

  format: function() {
    return [
      arguments[0].toUpperCase(),
      ':@google-cloud/debug: ',
      arguments[1]
    ].join('')
  }
})

gcloudLogger.error('Error!')
gcloudLogger.warn('Warning!')
gcloudLogger.info('Info!')
gcloudLogger.debug('Debug!')
gcloudLogger.silly('Silly!')
```
```sh
ERROR:@google-cloud/debug: Error!
WARN:@google-cloud/debug: Warning!
INFO:@google-cloud/debug: Info!
DEBUG:@google-cloud/debug: Debug!
SILLY:@google-cloud/debug: Silly!
```

ofrobots:
Great! Perhaps that little bit of boilerplate can move to common, with a tweak to allow arbitrary number of arguments?

stephenplusplus:
Sure, that should work 👍

ofrobots:
@stephenplusplus PTAL. Not sure if the code is any smaller.

stephenplusplus:
@matthewloring @ofrobots I gave it a once-over to adjust to the conventions we use in the library. The functionality should be the same, with these exceptions:

```diff
- common.logger.create(...)
+ common.logger(...)
```

And arguments:

```diff
- common.logger.create('level', 'prefix')
+ common.logger({
+   level: 'level',
+   tag: 'tag'
+ })
```

If there is just a level without needing a tag, there's a shorthand:

```diff
- common.logger.create('level')
+ common.logger('level')
```

I renamed `prefix` to `tag`, because the supplied prefix wasn't actually prefixing the entire log line. I thought tag might be more appropriate, but let me know if this is poor terminology on my part.

ofrobots:
LGTM, w/ a suggestion.

packages/common/src/logger.js
```diff
@@ -16,31 +16,53 @@
...
+    ],
+
+    level: options.level || 'error',
+
+    format: function() {
+      var args = [].slice.call(arguments);
 
+      return format('{level}{tag} {message}', {
+        level: args.shift().toUpperCase(),
+        tag: options.tag ? ':' + options.tag + ':' : '',
```
stephenplusplus: Yes, and I did originally have it that way, but that is a bit of a micro-optimization, where it affected logical, behavioral placement more than I liked. At the time `logger()` is invoked, `tag` doesn't necessarily need to be formatted, and it's formatting there has no relevance. For now, we're only using `tag` for a single purpose, but theoretically, it could be required in multiple places throughout the function, where its original form is expected. It's only when the entire message is formatted that the tag's formatting is important.

Basically, it just came down to this for me: if someday down the road, I'm looking to change the formatting of the printed log line, I would expect to find all formatting logic to be in the `format` function.

ofrobots:
Would it be possible to make this available through a release of `@google-cloud/common`? This being a mono-repo, I don't think I can npm install `common` using a github url – or do you have a work-around for this?

stephenplusplus:
No, I wish there was a better way. Just published the updates under `0.10.0`.

PR #1887, opened 2016-12-14
#1881 vision: include original Likelihood score

### LINKED ISSUE #1881: vision: include original Likelihood score
It would be great if the likelihood of emotions was send back in the response from the visionClient.detectFaces function. It is useful to the client code so that it can make decisions on the comparison of emotions.

It's intentionally removed replaced with Booleans in the gteLikelihood_ function.

Is there a reason this is done?  Do you want me to update it and submit a pull request?

rtw:
Fixes #1881

Add the original likelihood value into the response from the vision api
alongside the boolean values

### DIFF
--- a/packages/vision/src/index.js
+++ b/packages/vision/src/index.js
@@ -960,12 +960,19 @@ Vision.prototype.detectCrops = function(images, options, callback) {
  *   //     },
  *   //     confidence: 56.748849,
  *   //     anger: false,
+ *   //     angerLikelihood: 1,
  *   //     blurred: false,
+ *   //     blurredLikelihood: 1,
  *   //     headwear: false,
+ *   //     headwearLikelihood: 1,
  *   //     joy: false,
+ *   //     joyLikelihood: 1,
  *   //     sorrow: false,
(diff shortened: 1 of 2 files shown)


rtw:
I signed it!

stephenplusplus:
Thank you very much :) I think the only thing missing before we merge is an update to the docs (in the JSDocs snippet above the method) and a test to confirm we're returning the original value. If you prefer, I'm happy to do these at a later time.

packages/vision/src/index.js
```diff
@@ -960,12 +960,19 @@ Vision.prototype.detectCrops = function(images, options, callback) {
  *   //     },
  *   //     confidence: 56.748849,
  *   //     anger: false,
+ *   //     angerLikelihood: 1,
```
stephenplusplus: @rtw do you think we should go with the score here (`1`) or the string value, i.e. `UNLIKELY`?

stephenplusplus: cc @callmehiphop for thoughts

PR #1888, opened 2016-12-17
common: Add log levels field to logger

kjin:
There should be a way to get the list of possible log levels. Here it's just a member of `logger.js` exports.

### DIFF
--- a/packages/common/src/logger.js
+++ b/packages/common/src/logger.js
@@ -26,9 +26,9 @@ var logDriver = require('log-driver');
 
 /**
  * A list of log levels.
- * @type {Array}
+ * @type {string[]}
  */
-var logLevels = [
+var LEVELS = [
   'error',
   'warn',
   'info',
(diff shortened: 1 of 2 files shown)


packages/common/src/logger.js
```diff
@@ -25,6 +25,18 @@ var is = require('is');
 var logDriver = require('log-driver');
 
 /**
+ * A list of log levels.
+ * @type {Array}
+ */
+var logLevels = [
```
stephenplusplus: Can we rename this `LEVELS` and export it under the same name?

packages/common/src/logger.js
```diff
@@ -25,6 +25,18 @@ var is = require('is');
 var logDriver = require('log-driver');
 
 /**
+ * A list of log levels.
+ * @type {Array}
```
stephenplusplus: `@type {string[]}`

stephenplusplus:
LGTM. Thanks!

PR #1915, opened 2017-01-10
Speech API: fix createRecognizeStream timeout.

### LINKED ISSUE #1894: Speech SDK returns Client GRPC deadline too short
Hi,

Sometime the speech nodejs SDK returns the following error:

> Client GRPC deadline too short. Should be at least: 3 * audio-duration + 5 seconds. Current deadline is: 59.99929095065454 second(s). Required at least: 61 second(s).

And the audio stream is destroyed after that:

> stream was destroyed","stack":"Error: stream was destroyed\n    at /root/sip.rogervoice.com/speech_recognition/node_modules/pump/index.js:45:21\n    at call (/root/sip.rogervoice.com/speech_recognition/node_modules/pump/index.js:50:3)\n    at Array.forEach (native)\n    at /root/sip.rogervoice.com/speech_recognition/node_modules/pump/index.js:72:16\n    at f (/root/sip.rogervoice.com/speech_recognition/node_modules/once/once.js:17:25)\n    at /root/sip.rogervoice.com/speech_recognition/node_modules/pump/index.js:31:5\n    at f (/root/sip.rogervoice.com/speech_recognition/node_modules/once/once.js:17:25)\n    at DestroyableTransform.onfinish (/root/sip.rogervoice.com/speech_recognition/node_modules/pump/node_modules/end-of-stream/index.js:30:18)\n    at emitNone (events.js:91:20)\n    at DestroyableTransform.emit (events.js:185:7)","level":"error","timestamp":"2016-12-21T11:07:37.170Z

There is no information about this duration on your documentation. I don't know how to change it.

Here how I use the SDK:

```
createRecognizeStream({
            config: {
                encoding: 'LINEAR16',
                sampleRate: 16000,
		languageCode: 'en-US',
                maxAlternatives: 1,
	        profanityFilter: true
            },
            singleUtterance: true,
            interimResults: true,
            verbose: true
})
```

The audio stream is a continuous realtime stream.

jmuk:
This PR sets the following things:
- Increase the default timeout for createRecognizeStream to
  190 seconds, which will be long enough for 60 seconds recognition.
- Add some comments and links to the content limit page which says
  that the audio duration for streaming recognition is limited to
  ~60 seconds.
- Add "timeout" parameter to the createRecognizeStream which
  allows users to customize the timeout if the defualt timeout
  does not fit well into the use case.

Fixes #1894

### DIFF
--- a/packages/speech/src/index.js
+++ b/packages/speech/src/index.js
@@ -364,8 +364,9 @@ Speech.formatResults_ = function(resultSets, verboseMode) {
  *     property, a `confidence` score from `0` - `100`, and an `alternatives`
  *     array consisting of other transcription possibilities.
  *
- * Note that speech API sets the limits for the audio duration.
- * See [Content Limits]{@link https://cloud.google.com/speech/limits#content} for the details.
+ * Google Cloud Speech sets the limits for the audio duration. For more
+ * information, see
+ * [Content Limits]{@link https://cloud.google.com/speech/limits#content}.
  *
  * @resource [StreamingRecognize API Reference]{@link https://cloud.google.com/speech/reference/rpc/google.cloud.speech.v1beta1#google.cloud.speech.v1beta1.Speech.StreamingRecognize}
  * @resource [StreamingRecognizeRequest API Reference]{@link https://cloud.google.com/speech/reference/rpc/google.cloud.speech.v1beta1#google.cloud.speech.v1beta1.StreamingRecognizeRequest}
@@ -373,12 +374,12 @@ Speech.formatResults_ = function(resultSets, verboseMode) {
  *
  * @param {object} config - A `StreamingRecognitionConfig` object. See
  *     [`StreamingRecognitionConfig`](https://cloud.google.com/speech/reference/rpc/google.cloud.speech.v1beta1#google.cloud.speech.v1beta1.StreamingRecognitionConfig).
+ * @param {number=} config.timeout - In seconds, the amount of time before the
+ *     underlying API request times out. The default value, `190`, is sufficient
+ *     for audio input of 60 seconds or less. If your input is longer, consider
+ *     using a higher timeout value.
  * @param {boolean=} config.verbose - Enable verbose mode for a more detailed
  *     response. See the examples below. Default: `false`.
- * @param {number=} config.timeout - Customize the timeout number in
- *     milliseconds for underlying gRPC call. This will affect the input audio
- *     duration; longer timeout needs to be specified when the audio input is
- *     longer. By default, it will use the timeout enough for 60 seconds audio.
  *
  * @example
  * var fs = require('fs');
@@ -478,8 +479,9 @@ Speech.prototype.createRecognizeStream = function(config) {
   delete config.verbose;
 
   var gaxOptions = {};
-  if ('timeout' in config) {
-    gaxOptions.timeout = config.timeout;
+
+  if (is.number(config.timeout)) {
+    gaxOptions.timeout = config.timeout * 1000;
     delete config.timeout;
   }
 
--- a/packages/speech/test/index.js
+++ b/packages/speech/test/index.js
@@ -580,14 +580,16 @@ describe('Speech', function() {
     });
 
     it('should delete verbose option from request object', function(done) {
+      var stream = speech.createRecognizeStream({
+        verbose: true
+      });
+
       speech.api.Speech = {
         streamingRecognize: function() {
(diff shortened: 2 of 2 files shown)


jmuk:
@stephenplusplus @callmehiphop Please take a look. (the failures of appveyor / travis are for grpc installation issue on node7).

packages/speech/src/index.js
```diff
@@ -469,10 +477,16 @@ Speech.prototype.createRecognizeStream = function(config) {
   var verboseMode = config.verbose === true;
   delete config.verbose;
 
+  var callOptions = null;
+  if ('timeout' in config) {
+    callOptions = {timeout: config.timeout};
+    delete config.timeout;
+  }
```
stephenplusplus: Can we do something like this?

```js
var gaxOptions = {};

if (config.timeout) {
  gaxOptions.timeout = config.timeout;
  delete config.timeout;
}
```

packages/speech/test/index.js
```diff
@@ -599,6 +620,7 @@ describe('Speech', function() {
         }
       };
 
+      var stream = speech.createRecognizeStream({timeout: timeout});
```
stephenplusplus: Just a style thing:

```diff
- var stream = speech.createRecognizeStream({timeout: timeout});
+ var stream = speech.createRecognizeStream({
+  timeout: timeout
+ });
```

packages/speech/test/index.js
```diff
@@ -580,17 +580,38 @@ describe('Speech', function() {
...
+      stream.emit('writing');
+    });
+
+    it('should accept the optional timeout parameter', function(done) {
+      var timeout = 200000;
 
       speech.api.Speech = {
-        streamingRecognize: function() {
+        streamingRecognize: function(opts) {
+          assert.deepEqual(opts.timeout, timeout);
```
stephenplusplus:
```diff
- assert.deepEqual(...
+ assert.strictEqual(...
```

packages/speech/src/index.js
jmuk: Done

packages/speech/test/index.js
jmuk: Done

packages/speech/test/index.js
jmuk: Done

packages/speech/src/index.js
```diff
@@ -364,13 +364,21 @@ Speech.formatResults_ = function(resultSets, verboseMode) {
...
+ * @resource [Content Limits]{@link https://cloud.google.com/speech/limits#content}
  *
  * @param {object} config - A `StreamingRecognitionConfig` object. See
  *     [`StreamingRecognitionConfig`](https://cloud.google.com/speech/reference/rpc/google.cloud.speech.v1beta1#google.cloud.speech.v1beta1.StreamingRecognitionConfig).
  * @param {boolean=} config.verbose - Enable verbose mode for a more detailed
  *     response. See the examples below. Default: `false`.
+ * @param {number=} config.timeout - Customize the timeout number in
+ *     milliseconds for underlying gRPC call. This will affect the input audio
+ *     duration; longer timeout needs to be specified when the audio input is
+ *     longer. By default, it will use the timeout enough for 60 seconds audio.
```
stephenplusplus: Do we know what the actual default `timeout` number value is?

jmuk: It's written in the json file https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/speech/src/v1beta1/speech_client_config.json (this PR updates the json file to specify 190 seconds). Do you want to explicitly say the number here?

stephenplusplus: Sorry for missing the connection. Yes, my thought is that we should state here so the user has a point of reference for when they have input longer than 60 seconds. I will add this in for us real quick though, thanks!

stephenplusplus:
@jmuk PTAL at the latest commit. I added the explanation of the default in the docs, and also made our `timeout` argument expect input as seconds instead of milliseconds.

jmuk:
@stephenplusplus thank you for doing that. LGTM.

eljefedelrodeodeljefe:
Don't want to cahse you guys, but is there a possibility to land and publish this with the next 24hrs? Thanks in advance.

stephenplusplus:
I will try to release as soon as possible.

eljefedelrodeodeljefe:
Many thanks @stephenplusplus

PR #1918, opened 2017-01-11
fix(util): broken URL to auth docs

Splaktar:
```
Error: Sorry, we cannot connect to Google Cloud Services without a project ID. You may specify one with an environment variable named "GCLOUD_PROJECT". See https://googlecloudplatform.github.io/google-cloud-node/#//docs/guides/authentication for a detailed guide on creating an authenticated connection.
``` 
The [current link](https://googlecloudplatform.github.io/google-cloud-node/#//docs/guides/authentication) in this message opens https://googlecloudplatform.github.io/google-cloud-node/#/ and not the auth docs.

Removing the extra `/` properly routes to https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.45.0/guides/authentication

Relates to #1405

### DIFF
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -48,7 +48,7 @@ var errorMessage = format([
   'connection.'
 ].join(' '), {
   baseUrl: 'https://googlecloudplatform.github.io/google-cloud-node/#',
-  path: '/docs/guides/authentication'
+  path: 'docs/guides/authentication'
 });
 
 var missingProjectIdError = new Error(errorMessage);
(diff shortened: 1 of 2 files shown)


stephenplusplus:
Thank you! Which browser are you using? The original link works for me, as well as the new one, so it shouldn't be a problem to merge. Just want to make sure we're solving the real problem.

Splaktar:
Chrome, sorry for the delay.

Splaktar:
Added the [current link](https://googlecloudplatform.github.io/google-cloud-node/#//docs/guides/authentication) as a clickable link in the PR description to make sure that you were testing with the right link.

stephenplusplus:
Sorry for my delay, too. And thanks for the fix!

PR #1931, opened 2017-01-19
Fix pub/sub with 0 nanos

### LINKED ISSUE #1925: publishTime not visible on Pub/Sub message from emulator
I am testing out `publishTime` and it seems to not be present, at least while using the simulator. The `nanos` field is 0 which makes this check return false: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/pubsub/src/subscription.js#L367

arbesfeld:
Closes #1925 

Sometimes seconds/nanos are a string, so I decided to use `!= null` check instead.

### DIFF
--- a/packages/pubsub/src/subscription.js
+++ b/packages/pubsub/src/subscription.js
@@ -364,7 +364,7 @@ Subscription.formatMessage_ = function(msg, encoding) {
     if (innerMessage.publishTime) {
       var publishTime = innerMessage.publishTime;
 
-      if (publishTime.seconds !== undefined && publishTime.nanos !== undefined) {
+      if (is.defined(publishTime.seconds) && is.defined(publishTime.nanos)) {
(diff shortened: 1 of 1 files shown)


arbesfeld:
I signed it!

packages/pubsub/src/subscription.js
```diff
@@ -364,7 +364,7 @@ Subscription.formatMessage_ = function(msg, encoding) {
     if (innerMessage.publishTime) {
       var publishTime = innerMessage.publishTime;
 
-      if (publishTime.seconds != null && publishTime.nanos != null) {
+      if (publishTime.seconds !== undefined && publishTime.nanos !== undefined) {
```
stephenplusplus: I think just 'is.defined(...)' will work.

arbesfeld: done!

stephenplusplus:
Thank you!

PR #1935, opened 2017-01-24
extract gRPC functionality from common into common-grpc

### LINKED ISSUE #1822: grpc issues with alpine linux, complete incompatibility with node 7.x.x
#### Environment details

  - OS: alpine linux
  - Node.js version: 6.x.x, 7.x.x
  - npm version: 3.10.x
  - @google-cloud/storage version: <= 0.5.0

#### Steps to reproduce

This module relies on @google-cloud/common, which, in turn, relies on grpc. This causes problems for alpine linux, since it uses muslc. On 6.x.x it is possible to compile by providing glibc wrapper around muslc, so that compatible APIs are available, on node 7.x.x it just crashes during compilation (missing 7.x.x support)

Questions are:

a) why common depends on grpc? I haven't seen a single call using it in the sub-package
b) if I'm right with (a) - could somebody refactor @google-cloud/common and make grpc an optional dependency?
c) if (b) is doable, it's possible to load optional dependencies using Object.defineProperty on "demand", which, as a result, will allow google-cloud/storage API to be freely used on any system without any effort to maintain grpc module compatibility with anything

Thanks for considering this, looking forward to your feedback

stephenplusplus:
Fixes #1822

As discussed in #1822, the weight of including the gRPC module where it is not necessary is rather costly. Both in terms of the download size & time of the compilation step, but it can actually cause a complete installation failure for some systems where gRPC is incompatible. These compatibilities are being addressed in gRPC, but regardless, failing the installation because an unneeded dependency couldn't be installed isn't acceptable.

This takes the gRPC functionality outside of `@google-cloud/common` and into a new module `@google-cloud/common-grpc`.

### DIFF
--- a/scripts/docs/config.js
+++ b/scripts/docs/config.js
@@ -23,6 +23,7 @@ module.exports = {
   TOC: 'toc.json',
   IGNORE: [
     'common',
+    'common-grpc',
     'datastore/src/entity.js',
     'datastore/src/request.js',
     'logging/src/metadata.js',


stephenplusplus:
cc @ofrobots

ofrobots:
Thanks for putting this together! LGTM.

stephenplusplus:
No problem. @callmehiphop is pretty wrapped up in other business, and I'd ideally like his input before merging and releasing this. How much time do we have before you have to roll your own fix for these issues?

ofrobots:
We do need a fix for this fairly quickly to address https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/issues/351 and https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/issues/355. What's your ETA on when this can be merged/release?

stephenplusplus:
I would guess we can review/ship after Friday and before Tuesday.

callmehiphop:
@stephenplusplus overall LGTM, I noticed a docs error on CI, we're going to want to add this to the ignore list in <kbd>scripts/docs/config.js</kbd>.

stephenplusplus:
This is already shipped! What module are you using?

rajiff:
I am using +-- @google-cloud/speech@0.7.0

stephenplusplus:
I think there's some sort of issue with your environment and grpc, the module. This issue simply removed the `grpc` dependencies from packages that don't need it, but `@google-cloud/speech` isn't one of them, as it uses `grpc` to send requests.

Can you open your issue over on the [grpc](http://gitnpm.com/grpc) repo?

PR #1947, opened 2017-01-31
Make sure to pass options when encoding values more than 1 level deep

AdriVanHoudt:
I hope the styling is somewhat correct, if not let me know. EDIT: I found the linting command 🎆 

This is also breaking our app atm so anything I can do to help :shipit:  would be great ;)

### DIFF
--- a/packages/common-grpc/src/service.js
+++ b/packages/common-grpc/src/service.js
@@ -519,7 +519,7 @@ GrpcService.encodeValue_ = function(value, options) {
     convertedValue = {
       listValue: {
         values: value.map(function(value) {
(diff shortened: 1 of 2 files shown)


stephenplusplus:
Awesome, thanks! Just going to confirm the CI is relatively happy with it before merging (CI is never 100% happy around here). We should be rolling out some new releases very shortly, hopefully today.

AdriVanHoudt:
@stephenplusplus cool! 🔥

AdriVanHoudt:
@stephenplusplus since I'm to familiar with the project structure and it contains multiple modules it seem could you tell me when this lands in what version? I specifically need this for `logging` btw

PR #1070, opened 2016-01-19
pubsub: convert to grpc

stephenplusplus:
Fixes #1104

RE: #1049 

This adds support for gRPC APIs where necessary (Service & ServiceObject). This PR also converts Pub/Sub to use gRPC.
## To Dos
- [x] Get gRPC help from @tbetbetbe on the questions below and [here](https://github.com/GoogleCloudPlatform/gcloud-node/pull/1070#discussion_r50558085) (opened issue: https://github.com/grpc/grpc/issues/4962)
- ~~Use grpc.status error maps~~ - doesn't work, need HTTP status codes.
- [x] Use `deadline` for PubSub calls (https://github.com/GoogleCloudPlatform/gcloud-node/pull/1070#discussion_r51283736)
- [x] Leave comments in the code about why we're using snakeize/camelize (https://github.com/grpc/grpc/issues/5005)
- [x] Figure out how to shut off logs (https://github.com/grpc/grpc/issues/5004)
  - 0.13 released with a fix
- [x] Tests
  - [x] iam
  - [x] index
  - [x] topic
  - [x] subscription
  - [x] GrpcService
  - [x] GrpcServiceObject
- [x] Get to work with PUBSUB_EMULATOR_HOST
- [x] Confirm Pub/Sub with gRPC is generally available (https://github.com/GoogleCloudPlatform/gcloud-node/pull/1070#issuecomment-183392124)
  - No response. Assuming it works for all
## `mockery` -> `mockery-next`

Mockery wasn't holding up for our use of gRPC in this module. When trying to mock it in two different test suites, mockery couldn't handle it. I found [mockery-next](https://github.com/cspotcode/mockery-next) handles it quite well: https://github.com/cspotcode/mockery-next/issues/6.

Mockery Next requires a more logical way of specifying the modules you want to stub (https://github.com/mfncooper/mockery/pull/25/files). That's why the enormous changes to just filepaths in almost every test file.

PR #1961, opened 2017-02-03
pubsub: Fix broken abortion of outstanding requests

ctavan:
This restores the fix from GoogleCloudPlatform/google-cloud-node#1105
which was broken after GoogleCloudPlatform/google-cloud-node#1070

It seems like the introduction of gRPC in GoogleCloudPlatform/google-cloud-node#1070 broke the feature of aborting outstanding http requests when pubsub subscription listeners are removed.

This patch doesn't really add a suitable test-coverage to the pubsub testsuite yet, it just ensures that `GrpcService.prototype.request` returns an object which has an `abort()` method.

### DIFF
--- a/packages/common-grpc/src/service.js
+++ b/packages/common-grpc/src/service.js
@@ -280,7 +280,9 @@ GrpcService.prototype.request = function(protoOpts, reqOpts, callback) {
     }
   };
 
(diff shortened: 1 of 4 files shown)


stephenplusplus:
Thank you for this! I had to make some other changes, so please check out PR #1973 and let me know if it looks okay.

PR #1973, opened 2017-02-07
common-grpc: return requests to allow abort()

stephenplusplus:
As brought up in #1961, we weren't allowing active gRPC requests to be aborted.


ctavan:
Looks good to me! Thanks a lot for taking this over so quickly!

Issue #1877, opened 2016-12-12
Cloud Logging Error serialization


dsimmons:
#### Environment details

  - OS: `OSX 10.12.1`
  - Node.js version: `v7.0.0`
  - npm version: `3.10.8`
  - google-cloud-node version: `0.44.2`

#### Steps to reproduce

  1. require `google-cloud`
  2. Use something like [log.error](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.45.0/logging/log?method=error).

```Javascript
const entry = log.entry('gce_instance', new Error('oh no!'));
log.error(entry);
```

**NOTE:** For what it's worth, I created the above snippet on-the-fly, so it may contain mistakes or not run as-is -- my existing code is intertwined with project-specific configuration and business logic, so it's not feasible to copy/paste directly. The above snippet sufficiently illustrates my perspective though! 😄 

#### Description

Per a bit of trial-and-error, it appears as if there's not a default serializer of some kind for Javascript `Error` objects when using Cloud Logging through [log#write](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.45.0/logging/log?method=write) or any of the helper functions (eg. `error`, `alert`, `critical`).

While a new logging entry is indeed created within the GCP Logging administrative view, it doesn't include anything beyond the default metadata for each entry written. In other words, I know that an error occurred, but the extent to which I'm able to troubleshoot it from there is pretty close to nil. In a sense, it effectively skips over the "payload" of the log entry (the `Error` itself) and creates said entry with the body omitted.

Specifically, the scenario in which this is pertinent is with catching and logging errors. Typically, my callback function or promise error handler would invoke `logger.error(err)` -- in fact, that's what led me to discover this existing handling in the first place! For many of my production logging entries, I was seeing that errors were being thrown, but the log entry itself wasn't very useful apart from that 😄 

#### TL;DR / Proposal

I think slightly more intuitive behavior would be to have some sort of reasonable default in place as a serialization mechanism for the `Error` type/class, like using [Error.message](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) / `toString()`.

Currently, I've switched my code to doing something like the following in the interim: `(err && err.stack)`.

stephenplusplus:
@dsimmons that's a pretty sane idea. Right now, I believe you'll have the expected behavior if you just nest the Error object inside of an object, i.e.:

[...]

Issue #1354, opened 2016-06-01
logging: Serialization of circular structs


zbjornson:
Think I've hit all the edge cases now...! :)

```js
var circ = {a: 1};
circ.circ = circ;
var entry = log.entry(resource, circ);
log.write(entry...) // RangeError: Maximum call stack size exceeded
```

I think this should be handled in this library, not by the consumer?

stephenplusplus:
I appreciate your help finding these edge cases! We can use https://www.npmjs.com/package/is-circular to check if there is a circular reference in the object. What do you think we should do if we catch one?

zbjornson:
Rather than rejecting the payload, I think it would be useful to replace the circular reference with something like "[Circular]", which is what node does:

```js
> var x = {a: 1}
undefined
> x.x = x
{ a: 1, x: [Circular] }
```

Maybe https://www.npmjs.com/package/destroy-circular (haven't tested myself -- all of the unit tests are simple).

stephenplusplus:
Node does that for the sake of the output, but if you try to use that object (e.g. JSON.stringify), it will throw. The fact that a circular reference is there is likely an error, though I personally don't mind throwing to communicate that error to the user. It's possible where there is a circular reference, the user meant there to be something else entirely. If we delete it, it might be an unexpected bit of magic, making it harder for the user to catch their mistake.

@callmehiphop what do you think?

zbjornson:
All true.

BTW I didn't hit this issue in production, just when trying to find edge cases. The node http request object has circular references in it, but the source I'm pulling from [picks out some of the non-circular properties](https://github.com/trentm/node-bunyan/blob/master/lib/bunyan.js#L1083-L1098). Maybe detecting and throwing is sufficient.

stephenplusplus:
@callmehiphop should we add the `is-circular` dep and throw if we get a circular reference, or let the user's script break with a `RangeError: Maximum call stack size exceeded`?

callmehiphop:
@stephenplusplus I like the `is-circular` route, should make the error more specific/easier to debug, WDYT?

zbjornson:
[...]

Issue #1353, opened 2016-06-01
logging: Serialization of undefined values


zbjornson:
Also failing:

```js
var entry = log.entry(resource, {
  someValue: undefined
});
log.write(entry...) // Error: Value of type undefined not recognized
```

stephenplusplus:
Just going to lump this in with #1352.

PR #1977, opened 2017-02-09
Initial import of logging-bunyan

ofrobots:
This PR adds a new module @google-cloud/logging-bunyan that provides simple, minimal config way of logging to Stackdriver Logging from applications using bunyan.

### DIFF
--- a/packages/logging-bunyan/README.md
+++ b/packages/logging-bunyan/README.md
@@ -29,9 +29,9 @@ logger.info('shields at 99%');
 
 It's incredibly easy to get authenticated and start using Google's APIs. You can set your credentials on a global basis as well as on a per-API basis. See each individual API section below to see how you can auth on a per-API-basis. This is useful if you want to use different accounts for different Google Cloud services.
 
-### On Google Compute Engine
+### On Google Cloud Platform
 
-If you are running this client on Google Compute Engine, we handle authentication for you with no configuration. You just need to make sure that when you [set up the GCE instance][gce-how-to], you add the correct scopes for the APIs you want to access.
+If you are running this client on Google Cloud Platform, we handle authentication for you with no configuration. You just need to make sure that when you [set up the GCE instance][gce-how-to], you add the correct scopes for the APIs you want to access.
 
 ``` js
 var loggingBunyan = require('@google-cloud/logging-bunyan')();
--- a/packages/logging-winston/README.md
+++ b/packages/logging-winston/README.md
@@ -31,9 +31,9 @@ winston.verbose('sheilds at 99%');
 
 It's incredibly easy to get authenticated and start using Google's APIs. You can set your credentials on a global basis as well as on a per-API basis. See each individual API section below to see how you can auth on a per-API-basis. This is useful if you want to use different accounts for different Google Cloud services.
 
-### On Google Compute Engine
+### On Google Cloud Platform
 
-If you are running this client on Google Compute Engine, we handle authentication for you with no configuration. You just need to make sure that when you [set up the GCE instance][gce-how-to], you add the correct scopes for the APIs you want to access.
+If you are running this client on Google Cloud Platform, we handle authentication for you with no configuration. You just need to make sure that when you [set up the GCE instance][gce-how-to], you add the correct scopes for the APIs you want to access.
 
 ``` js
 var winston = require('winston');


packages/logging-bunyan/test/index.js
```diff
@@ -0,0 +1,228 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+    var assert = require('assert');
```
DominicKramer: nit: unindent

zbjornson:
Hello! There's an unofficial, moderately popular Bunyan transport as well: https://github.com/mlazarov/bunyan-stackdriver

If it helps, the key differences I see are:
- Batching of API calls, which is critical to avoid API errors. (I think the limits in the readme are incorrect, but they're low enough to matter.).
- Handling of log entries with circular references. (Discussed in #1354.)
- Handling of log entries with types that are not strict JSON types (Date, Error, RegExp, undefined). (#1877, #1353, #1352.) For these types and circular references, the resolution was that it's the consumer's responsibility to deal with it, which I think means it should be handled in this module?
- This implementation doesn't provide a true `stream.Writable`, which means Bunyan's `reemitErrorEvents` won't work.

ofrobots:
Hello @zbjornson! Thanks for chiming in. We wanted to start officially support a module for working with bunyan (and winston: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1830). It would be great to have your collaboration.

This is an initial version of the library here, but I do have some thoughts on the points you highlighted:

* Batching: It would good to have batching indeed, and we need it for `logging-winston` too, but ideally batching should be built into the base logging library. This is something we have to hash out still. /cc @stephenplusplus 
* We do need to deal with circular log entries and non JSON values in this module (and in `logging-winston`).
* I was not familiar with Bunyan's `reemitErrorEvents`. Thanks for pointing it out.

We can start addressing these in follow-ons. Again, collaboration would be most welcome!

zbjornson:
Sounds great! I'm a bit short on time for the next few days to actually code, but happy to review and discuss anything. Let me know how I can help.

packages/logging-bunyan/test/index.js
ofrobots: Done.

ofrobots:
@GoogleCloudPlatform/node-team, @stephenplusplus This is ready for review. PTAL.

ofrobots:
Yes. No reason not to.

packages/logging-bunyan/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@google-cloud/logging-bunyan",
+  "version": "0.0.0",
```
matthewloring: I've usually seen 0.0.1 as the first released version of a module. This may be fine though, I don't think semver has a restriction.

packages/logging-bunyan/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+  var entryMetadata = {resource: this.resource_, timestamp: rec.time};
+
+  var entry = this.log_.entry(entryMetadata, rec);
+
+  // Pass a no-op function because otherwise we get an unnecessary promise
+  // allocated and returned to us.
+  this.log_[level](entry, function() {});
+};
+
+module.exports = LoggingBunyan;
```
matthewloring: '\n'

packages/logging-bunyan/system-test/logging-bunyan.js
```diff
@@ -0,0 +1,96 @@
...
+                test.verify(entry);
+              });
+
+              done();
+            });
+      }, WRITE_CONSISTENCY_DELAY_MS);
+    });
+
+  });
+});
```
matthewloring: '\n'

packages/logging-bunyan/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+ * @param {object} rec - Bunyan log record.
+ * @private
+ */
+LoggingBunyan.prototype.write = function(rec) {
+  if (typeof rec === 'string') {
+    throw new Error(
+        '@google-cloud/logging-bunyan only works as a raw bunyan stream type.');
+  }
+  var level = BUNYAN_TO_STACKDRIVER[rec.level];
+  var entryMetadata = {resource: this.resource_, timestamp: rec.time};
```
matthewloring: Should `resource` be reported if it is null? I'm not sure how the API handles this.

packages/logging-bunyan/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+ *
+ */
+function LoggingBunyan(options) {
+  if (!(this instanceof LoggingBunyan)) {
+    return new LoggingBunyan(options);
+  }
+
+  options = options || {};
+
+  this.logName_ = options.logName || 'bunyan_log';
```
matthewloring: Do we want the default name to make it clear that our module is generating the logs.

packages/logging-bunyan/package.json
stephenplusplus: We have a release task (`npm run publish-module [version number]`) which for the first release, we'll start at `0.1.0`

matthewloring: Ah cool, makes sense.

packages/logging-bunyan/src/index.js
ofrobots: I am not sure I understand the concern. The module isn't generating the logs, it is simply routing the user's generated logs to Stackdriver Logging. When they show up in Stackdriver Logging, they should be distinct from stdout/stderr logs caught by fluentd or other mechanisms. The log name 'bunyan_log' conveys to the user that this is their bunyan logs (unless they provide custom names).

The name here matches what we do for winston as well.

packages/logging-bunyan/src/index.js
ofrobots: If the resource is null or undefined, @google-cloud/logging will try to autodetect the resource and apply it. The API rejects requests if they don't contain a valid resource.

packages/logging-bunyan/system-test/logging-bunyan.js
ofrobots: done.

packages/logging-bunyan/src/index.js
ofrobots: done.

packages/logging-bunyan/src/index.js
matthewloring: Yup, that makes sense.

stephenplusplus:
Once functionality is approved by the reviewers, I'll do some simple linting/style stuff.

ofrobots:
@zbjornson for the issues/feature requests you raised:

* Batching: will track through https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1296.
* Circular log entires: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1987
* Non JSON log entries: `@google-cloud/logging` can now 'deal with' non JSON types. That is to say, they are converted to strings before being sent to the API. I don't think anything further is needed in the higher level layers. Please correct me if I am wrong here.
* Proper Writable: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1988
* Better readability in Logging UI, interop with Error Reporting: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1989

zbjornson:
bueno, thanks! Looks like everything is addressed here or tracked in other issues.

stephenplusplus:
@ofrobots I went over it for lint/nit/style changes. The biggest change was probably that I removed tests that were using the Bunyan module inside of the unit test file for `index.js`. I saw the system tests were using the real module, so if more testing with Bunyan is needed, please pop them in there.

ofrobots:
LGTM. The removed tests are not that significant test of the interop to be really worth keeping.

packages/logging-bunyan/README.md
```diff
@@ -35,19 +29,12 @@ logger.info('shields at 99%');
 
 It's incredibly easy to get authenticated and start using Google's APIs. You can set your credentials on a global basis as well as on a per-API basis. See each individual API section below to see how you can auth on a per-API-basis. This is useful if you want to use different accounts for different Google Cloud services.
 
-### On Google Cloud Platform
+### On Google Compute Engine
```
ofrobots: Any reason for the edit? 'Compute Engine' exclude GAE, GKE and Cloud Functions. I think the instructions here are valid for all these environments.

stephenplusplus: Fixed.

ofrobots:
Let us know if there is anything more needed from our side before this can land. Once this lands, we can start work on some follow-on PRs.

stephenplusplus:
Just waiting on me?

ofrobots:
Yep.

stephenplusplus:
@google-cloud/logging-bunyan@0.1.0 published.

Issue #1982, opened 2017-02-10
system test setup instructions out of date?


ofrobots:
From https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/.github/CONTRIBUTING.md:

```
$ gcloud preview datastore create-indexes system-test/data/index.yaml
```

1. I think `preview` can be dropped now.
2. I cannot find the file `system-test/data/index.yaml`. I am guessing that this isn't the same as `./packages/vision/system-test/data/index.yaml`?

stephenplusplus:
That file should actually be moved to the `packages/datastore/system-test/data/` directory. It looks like it got misplaced when we modularized the repo. PR? :)

ofrobots:
Sure. But even after running the `create-indexes`, I get the following error with the `logging` system tests:

```
  1) Logging sinks should create a sink with a Bucket destination:
     Uncaught Error: The caller does not have permission
      at node_modules/grpc/src/node/src/client.js:434:17
```

I do have $GCLOUD_TESTS_PROJECT_ID and $GCLOUD_TESTS_KEY set to valid values.

stephenplusplus:
I think the create indexes command will only relate to the Datastore tests.

For the Logging test, that's a new issue. I can reproduce that as well. That used to happen for me when we were developing it, but it hasn't since. @callmehiphop can you see if you get that error?

PR #1983, opened 2017-02-10
`Log.prototype[severity]` functions cannot deal with circular references

ofrobots:
Right now the PR adds a test that shows this. What are your thoughts on the fix? The simplest fix would be to fix `Log.assignSeverityToEntries_` but am concerned that the problem may exist in other places too.

### DIFF
--- a/packages/logging/src/log.js
+++ b/packages/logging/src/log.js
@@ -118,8 +118,13 @@ util.inherits(Log, commonGrpc.ServiceObject);
  */
 Log.assignSeverityToEntries_ = function(entries, severity) {
   return arrify(entries).map(function(entry) {
-    var metadata = extend(true, {}, entry.metadata, { severity: severity });
-    return extend(new Entry(), entry, {metadata: metadata });
+    var metadata = extend(true, {}, entry.metadata, {
+      severity: severity
+    });
+
+    return extend(new Entry(), entry, {
+      metadata: metadata
+    });
   });
 };
 


stephenplusplus:
We chose to throw the error instead of hiding it and removing part of the user's object. My thoughts on that were explained [in this issue](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1354). Please discuss in that issue if you have reasons that's a bad idea on our part.

```js
1) Logging logs should write a single entry to a log:
     Error: The JSON data for this entry has a circular reference.
```

ofrobots:
Either we should test with `is-circular` uniformly, or not test `is-circular` at all.. Elsewhere in the repo we use `is-circular` to test for circular objects and report a nice error.

A large stack overflow in `extend` is not as useful to users.

I think this is a distinct issue.

stephenplusplus:
We throw a small error as I pasted above. Maybe I'm not seeing what you're seeing. Can you paste the output you're seeing?

ofrobots:
I guess it depends on test ordering. I bypassed some of the tests because the system tests aren't working for me (https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1982).

Try `mocha system-test/*.js --no-timeouts --bail -g 'logs'` with my test.

stephenplusplus:
Okay, I reproduced. We definitely should check with `is-circular` the same was we do in Entry#toJSON.

ofrobots:
Here's a better fix. @stephenplusplus PTAL.

packages/logging/src/log.js
```diff
@@ -118,11 +118,10 @@ util.inherits(Log, commonGrpc.ServiceObject);
  */
 Log.assignSeverityToEntries_ = function(entries, severity) {
   return arrify(entries).map(function(entry) {
-    return extend(true, new Entry(), entry, {
-      metadata: {
-        severity: severity
-      }
-    });
+    var entry = extend(new Entry(), entry);
```
stephenplusplus: Is removing `true` okay? I believe without it, modifying `entry.metadata` will affect the input `entry.metadata`.

ofrobots: Ack. Fixed.

stephenplusplus:
Cool, thanks!

PR #1985, opened 2017-02-11
logging-winston: avoid printing to console during test

ofrobots:
Don't use the default winston logger, which also prints to the console.


ofrobots:
@cristiancavalli @stephenplusplus @kjin PTAL.

stephenplusplus:
It's all good news, @googlebot!

PR #1997, opened 2017-02-14
logging-{winston,bunyan}: format logs to work better with Log Viewer and Error Reporting

ofrobots:
Stackdriver Logs Viewer picks up the summary line from the 'message' field
of the jsonPayload or from the textPayload. See:
https://cloud.google.com/logging/docs/view/logs_viewer_v2#expanding

Furthermore, for error messages at severity 'error' and higher, Stackdriver
Error Reporting will pick up error messages if the full stack trace is
included in the textPayload or in the `message` property of the
jsonPaylod. See:
https://cloud.google.com/error-reporting/docs/formatting-error-messages

The commit also fixes some non-determinism in the winston test caused by the sending logs through `async`.

### DIFF
--- a/packages/logging-bunyan/package.json
+++ b/packages/logging-bunyan/package.json
@@ -28,7 +28,7 @@
     "winston"
   ],
   "dependencies": {
-    "@google-cloud/logging": "^0.7.0",
+    "@google-cloud/logging": "^0.8.0",
     "extend": "^3.0.0"
   },
   "devDependencies": {
--- a/packages/logging-bunyan/test/index.js
+++ b/packages/logging-bunyan/test/index.js
@@ -155,8 +155,8 @@ describe('logging-bunyan', function() {
     });
 
     it('should rename the msg property to message', function(done) {
-      var recordWithMsg = { msg: 'msg' };
-      var recordWithMessage = { message: 'msg' };
+      var recordWithMsg = extend({ msg: 'msg' }, RECORD);
+      var recordWithMessage = extend({ message: 'msg' }, RECORD);
 
       loggingBunyan.log_.entry = function(entryMetadata, record) {
         assert.deepStrictEqual(record, recordWithMessage);
@@ -167,19 +167,19 @@ describe('logging-bunyan', function() {
     });
 
     it('should inject the error stack as the message', function(done) {
-      var record = {
+      var record = extend({
         msg: 'msg',
         err: {
           stack: 'the stack'
         }
-      };
-      var expectedRecord = {
+      }, RECORD);
+      var expectedRecord = extend({
         msg: 'msg',
         err: {
           stack: 'the stack'
         },
         message: 'the stack'
-      };
+      }, RECORD);
 
       loggingBunyan.log_.entry = function(entryMetadata, record_) {
         assert.deepStrictEqual(record_, expectedRecord);
@@ -190,13 +190,13 @@ describe('logging-bunyan', function() {
     });
 
     it('should leave message property intact when present', function(done) {
-      var record = {
+      var record = extend({
         msg: 'msg',
         message: 'message',
         err: {
           stack: 'the stack'
         }
-      };
+      }, RECORD);
 
       loggingBunyan.log_.entry = function(entryMetadata, record_) {
         assert.deepStrictEqual(record_, record);


ofrobots:
@cristiancavalli @kjin PTAL.

packages/logging-bunyan/src/index.js
```diff
@@ -131,6 +131,26 @@ LoggingBunyan.prototype.write = function(record) {
...
+  // Error Reporting will pick up error messages if the full stack trace is
+  // included in the textPayload or in the `message` property of the
+  // jsonPaylod. See:
+  // https://cloud.google.com/error-reporting/docs/formatting-error-messages
+  //
+  // We only do this if the user hasn't provided a message property themselves.
+  //
+  if (!record.message) {
+    if (record.err && record.err.stack) {
+      record.message = record.err.stack;
```
kjin: At [https://cloud.google.com/error-reporting/docs/formatting-error-messages](formatting-error-messages) I see that there are some other fields (namely, the `serviceContext`) field that are marked as 'Required' in order for the log to be recognized by Stackdriver Error Reporting. Do we expect the application to have supplied the `serviceContext` field as part of their log object?

packages/logging-winston/src/index.js
```diff
@@ -156,9 +156,20 @@ LoggingWinston.prototype.log = function(levelName, msg, metadata, callback) {
...
-  // using util.inspect.
-  if (is.object(metadata)) {
+  // For error messages at severity 'error' and higher, Stackdriver
+  // Error Reporting will pick up error messages if the full stack trace is
+  // included in the textPayload or the message property of the jsonPayload.
+  // https://cloud.google.com/error-reporting/docs/formatting-error-messages
+  //
+  if (is.error(metadata)) {
+    if (msg.length === 0) {
+      msg = metadata.stack;
```
kjin: This seems to have different behavior than in Bunyan, since here 'msg' is assigned while in Bunyan 'message' is assigned. The docs say that it should be 'message', so should this be changed to match that?

packages/logging-winston/system-test/logging-winston.js
```diff
@@ -68,34 +71,48 @@ describe('LoggingWinston', function() {
...
+        args: [ 'fifth message', new Error('fifth') ],
+        level: 'error',
+        verify: function(entry) {
+          assert.ok(entry.data.startsWith('fifth message: Error: fifth'));
+        }
+      },
     ];
 
     it('should properly write log entries', function(done) {
-      async.each(testData, function(test, callback) {
```
kjin: This seems to be the only reference to `async` so I think it should be removed from the top and `devDependencies`.

packages/logging-winston/system-test/logging-winston.js
```diff
@@ -68,34 +71,48 @@ describe('LoggingWinston', function() {
...
+        verify: function(entry) {
+          assert.ok(entry.data.startsWith('Error: forth'));
+        }
+      },
+
+      {
+        args: [ 'fifth message', new Error('fifth') ],
+        level: 'error',
+        verify: function(entry) {
+          assert.ok(entry.data.startsWith('fifth message: Error: fifth'));
```
kjin: I don't see any references to `msg` (or `message`, if `msg` is not the right field) in these tests. Should this (and the previous element) read `entry.data.msg.startsWith('...')`?

packages/logging-bunyan/src/index.js
ofrobots: I am trying to figure out under what scenarios `serviceContext` is really necessary.

packages/logging-winston/src/index.js
ofrobots: For bunyan, `message` is a property of the log entry's data field. The data field gets formatted as a jsonPayload in the protobuf.

For winston, `msg` *was* the log entry's data field – since winston gives us a plain JS string as the message to be recorded. This gets formatted as a textPayload in the jsonPayload.

However, this all has changed now given that Error Reporting seems to prefer jsonPayload, specially if the application is running off of GCP. The latest commits always format winston log entries as a jsonPayload, so the code should be more consistent between winston and bunyan.

packages/logging-winston/system-test/logging-winston.js
ofrobots: The other comment should explain this.

packages/logging-winston/system-test/logging-winston.js
ofrobots: done.

ofrobots:
This is currently blocked on https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2014.

packages/logging-winston/src/index.js
kjin: Ok, re-reading it now I realize `msg` is not a field but a variable name. Sounds good.

packages/logging-bunyan/src/index.js
kjin: Not sure what the status of this is at this point in time... but it seems to me like you'd need both the winston and bunyan transports to accept this as a config option and initialize it the same way as in the error reporting agent.

ofrobots: I have reduced the scope of this PR to just better formatting for the Log Viewer. Error Reporting is no longer in scope.

Yes, the Error Reporting service does seem to require a `serviceContext` to properly pick up errors. But given the large conceptual overlap between Logging's `resource` and `serviceContext` I do not think it is a good idea for us to expose both of these documents to the users of these libraries – not to mention the extra bytes on the wire to send both these pieces of information on log entries.

I am working with the Logging/Error Reporting  to get some clarity on why both of these needed.

cristiancavalli: Maybe that's why it's not showing up in errors?

Here's what errors defaults to for service context (this was previously vetted by Steren):
```JS
{service : 'node', version : undefined};
```

cristiancavalli:
LGTM once CI's pass

packages/logging-bunyan/src/index.js
ofrobots: Okay, so got some clarity on `serviceContext`. It is only needed when the `resource.type` is `global`. We do not need to provide it when running on GAE/GKE/GCE/GCF as the `resource` type would be set appropriately.

I am going to increase the scope to the PR to support auto-ErrorReporting when running in these environments. Supporting auto-ErrorReporting for applications running elsewhere can be handled in a follow on.

ofrobots:
@stephenplusplus this is good to go from our side. PTAL.

packages/logging-winston/src/index.js
```diff
@@ -169,7 +169,28 @@ LoggingWinston.prototype.log = function(levelName, msg, metadata, callback) {
...
+  //
+  // TODO(ofrobots): when resource.type is 'global' we need to additionally
+  // provide serviceContext.service as part of the entry for Error Reporting to
+  // automatically pick up the error.
+  var data = {
+    message:
+        (metadata && metadata.stack) ?
+            (msg.length === 0 ? metadata.stack : msg + ': ' + metadata.stack) :
+            msg
+  };
```
stephenplusplus: Can we do:

```js
if (metadata && metadata.stack) {
  // Append stack trace
  msg += (msg ? ': ' : '') + metadata.stack;
}

var data = {
  message: msg
};
```

packages/logging-bunyan/src/index.js
```diff
@@ -131,6 +131,29 @@ LoggingBunyan.prototype.write = function(record) {
...
+  // provide serviceContext.service as part of the entry for Error Reporting to
+  // automatically pick up the error.
+  if (!record.message) {
+    if (record.err && record.err.stack) {
+      record.message = record.err.stack;
+    } else {
+      record.message = record.msg;
+    }
+  }
+
```
stephenplusplus: Just to save some thought-steps, I think this makes it a bit more clear.

```js
var message = record.message || record.msg || record.err && record.err.stack;

record.message = message;
```

Also, what's up with `record.message` and `record.msg`? Do we support both, and if so, can we choose just one?

packages/logging-bunyan/src/index.js
```diff
@@ -131,6 +131,29 @@ LoggingBunyan.prototype.write = function(record) {
...
+  // included in the textPayload or in the `message` property of the
+  // jsonPaylod. See:
+  // https://cloud.google.com/error-reporting/docs/formatting-error-messages
+  //
+  // We only do this if the user hasn't provided a message property themselves.
+  //
+  // TODO(ofrobots): when resource.type is 'global' we need to additionally
+  // provide serviceContext.service as part of the entry for Error Reporting to
+  // automatically pick up the error.
+  if (!record.message) {
```
stephenplusplus: This behavior needs unit tests.

packages/logging-winston/src/index.js
```diff
@@ -169,7 +169,28 @@ LoggingWinston.prototype.log = function(levelName, msg, metadata, callback) {
...
+  // included in the textPayload or the message property of the jsonPayload.
+  // https://cloud.google.com/error-reporting/docs/formatting-error-messages
+  // We prefer to format messages as jsonPayload (by putting it as a message
+  // property on an object) as that works is accepted by Error Reporting in
+  // for more resource types.
+  //
+  // TODO(ofrobots): when resource.type is 'global' we need to additionally
+  // provide serviceContext.service as part of the entry for Error Reporting to
+  // automatically pick up the error.
+  var data = {
```
stephenplusplus: I believe the behavior of appending a stack trace (both with and without the `: ` prefix) is not being unit tested.

packages/logging-bunyan/src/index.js
ofrobots: Thanks, that's a lot more readable, albeit if we wrote it like this: `record.message = record.message || (record.err && record.err.stack) || record.msg`. However, you raise a good point about `record.msg` and `record.message` being duplicated values in some cases – correctly encoding the logic I want here puts me back to an if-conditional rather than the assignment you suggested:

```
  if (!record.message) {
    if (record.err && record.err.stack) {
      record.message = recored.err.stack;
    } else {
      record.message = record.msg;
      delete record.msg;
    }
  }
```

I'm also going to add a clone of `record` at the top.

ofrobots: That is to say, if we added the stack as the `message`, we should keep the `msg` property around.

If there is no stack to add, just rename `msg` to `message.

stephenplusplus: That's cool, but why do we get `msg` sometimes and not `message`?

packages/logging-winston/src/index.js
ofrobots: Sorry about that. Done.

packages/logging-winston/src/index.js
ofrobots: Done.

packages/logging-bunyan/src/index.js
ofrobots: Done.

packages/logging-bunyan/src/index.js
ofrobots: PTAL.

stephenplusplus: Sorry to be getting stuck on something that is probably simple, but I'm still missing the origination of `msg`. Is that just a Bunyan thing?

ofrobots: Bunyan will always pass the log message as `msg`.

Bunyan also allows the user to provide additional properties to be attached to the log record. The user may add a property named `message`.

packages/logging-bunyan/src/index.js
```diff
@@ -133,6 +134,30 @@ LoggingBunyan.prototype.write = function(record) {
...
+    // Error Reporting to pick up errors automatically (for severity 'error' or
+    // higher). In this case we leave the 'msg' property intact.
+    // https://cloud.google.com/error-reporting/docs/formatting-error-messages
+    //
+    // TODO(ofrobots): when resource.type is 'global' we need to additionally
+    // provide serviceContext.service as part of the entry for Error Reporting
+    // to automatically pick up the error.
+    if (record.err && record.err.stack) {
+      record.message = record.err.stack;
+    } else if (record.msg) {
```
stephenplusplus: Is it not possible for `record` to have both `err` and `msg`?

ofrobots: Yes, it is possible. In that case we use `err.stack` as the `message` and leave `msg` intact.

packages/logging-bunyan/src/index.js
ofrobots: Yes, it is a bunyan thing. Examples:

```js
log.info('hi'); // bunyan will format record as { msg: 'hi', /* other properties */}
log.info({ message: 'bar' }, 'hello'); // -> { msg: 'hello', message: 'bar', /* other properties */}.
```

packages/logging-bunyan/src/index.js
stephenplusplus: Is it weird to give them two message-like messages as apposed to appending to a single message as in the other case?

ofrobots: I don't think we should necessarily extrapolate between winston and bunyan. Bunyan is a lot more structured.

I don't think it is weird at all – if the user wanted to attach a property named `message` to their log in addition to the log message, then we should faithfull report both these to the log service. We certainly don't have any business appending these two properties.

ofrobots: Perhaps it would have been better for me to make separate PRs for winston and bunyan.

In case you're uncomfortable about winston having different behaviour, let's look at what winston's own built-in transports do for ambiguous cases:

```
winston.info('foo', new Error('bar'));
//info: foo Error: bar
//   at Object.<anonymous> (/Users/ofrobots/tmp/log-test/index.js:16:21)
// ->  simple concatenation.

winston.info(new Error('goo'))
//info:  Error: goo
//    at Object.<anonymous> (/Users/ofrobots/tmp/log-test/index.js:15:14)
// -> a simple concatenation: an empty `msg` is concatenated with the Error
```

I can make our concatenation even simpler (i.e. get rid of the ': ') to match this behaviour.

I still have strong feelings that Bunyan should have different behaviour.

Do you still have reservations?

stephenplusplus: No reservations at all, I appreciate the elaboration. I'm okay with the behavior as it is currently, but I'll let you decide if we should remove the `: `.

packages/logging-winston/system-test/logging-winston.js
```diff
@@ -41,61 +40,74 @@ describe('LoggingWinston', function() {
...
+        verify: function(entry) {
+          assert.ok(entry.data.message.startsWith('Error: forth'));
+        }
+      },
+
+      {
+        args: ['fifth message', new Error('fifth')],
+        level: 'error',
+        verify: function(entry) {
+          assert.ok(
```
stephenplusplus: Just ditch the `.ok(` and this should fit on one line.

packages/logging-winston/system-test/logging-winston.js
```diff
@@ -41,61 +40,74 @@ describe('LoggingWinston', function() {
...
           );
         }
-      }
+      },
+
+      {
+        args: [new Error('forth')],
+        level: 'error',
+        verify: function(entry) {
+          assert.ok(entry.data.message.startsWith('Error: forth'));
```
stephenplusplus: Just `assert` instead of `assert.ok`.

packages/logging-bunyan/src/index.js
ofrobots: `:` removed. PTAL.

packages/logging-winston/src/index.js
```diff
@@ -169,7 +169,29 @@ LoggingWinston.prototype.log = function(levelName, msg, metadata, callback) {
...
+  // https://cloud.google.com/error-reporting/docs/formatting-error-messages
+  // We prefer to format messages as jsonPayload (by putting it as a message
+  // property on an object) as that works is accepted by Error Reporting in
+  // for more resource types.
+  //
+  // TODO(ofrobots): when resource.type is 'global' we need to additionally
+  // provide serviceContext.service as part of the entry for Error Reporting to
+  // automatically pick up the error.
+  if (metadata && metadata.stack) {
+    msg += ' ' + metadata.stack;
```
stephenplusplus: I think we should only include the space if there is already a message:

```js
msg += (msg ? ' ' : '') + metadata.stack;
// or
msg = (msg + ' ' + metadata.stack).trim();
```

packages/logging-winston/test/index.js
```diff
@@ -167,20 +167,51 @@ describe('logging-winston', function() {
...
+        assert.deepStrictEqual(data, {
+          message: MESSAGE + ' ' + error.stack
+        });
+        done();
+      };
+
+      loggingWinston.log(LEVEL, MESSAGE, error, assert.ifError);
+    });
+
+    it('should append the stack when metadata is an error, empty message',
```
stephenplusplus:
```js
it('should append stack when metadata is err with empty msg', function(done) {
```

packages/logging-winston/system-test/logging-winston.js
ofrobots: Done.

packages/logging-winston/system-test/logging-winston.js
ofrobots: Done.

packages/logging-winston/src/index.js
ofrobots: Done.

stephenplusplus:
`npm test`:

```
  1) logging-bunyan write should rename the msg property to message:
     TypeError: this.log_[level] is not a function
      at LoggingBunyan.write (C:\Users\sawch\dev\google-cloud-node\packages\logging-bunyan\src\index.js:170:19)
      at Context.<anonymous> (C:\Users\sawch\dev\google-cloud-node\packages\logging-bunyan\test\index.js:166:21)
      at callFnAsync (C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runnable.js:349:8)
      at Test.Runnable.run (C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runnable.js:301:7)
      at Runner.runTest (C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runner.js:422:10)
      at C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runner.js:528:12
      at next (C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runner.js:342:14)
      at C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runner.js:352:7
      at next (C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runner.js:284:14)
      at C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runner.js:315:7
      at done (C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runnable.js:287:5)
      at callFn (C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runnable.js:344:7)
      at Hook.Runnable.run (C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runnable.js:319:7)
      at next (C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runner.js:298:10)
      at Immediate.<anonymous> (C:\Users\sawch\dev\google-cloud-node\node_modules\mocha\lib\runner.js:320:5)
```

ofrobots:
@stephenplusplus I am unable to reproduce that locally. Is that windows only failure? Are the appveyor results to be trusted?

stephenplusplus:
That was actually on my computer. I'll have to check that I didn't have the repo in some weird state and report back.

ofrobots:
Actually, I see it to be failing on CircleCI as well.

ofrobots:
I cannot reproduce on an ssh session on CircleCI either. Also, it seems that the first run on the test on Circle (with Node.js v4) passes. The second test (Node.js v6) fails. Running the test individually with either v4 or v6 passes. 

Somehow this is dependent on how the test is run from the script and/or dependent on something else non-deterministic.

ofrobots:
Let's see if ce8caf7 fixes things.

ofrobots:
@stephenplusplus Fixed the problem. Does this LGTY?

PR #2001, opened 2017-02-15
storage: support per-object storage classes

### LINKED ISSUE #1937: Add destinationPredefinedAcl to copy and move
Would be nice to be able to do something like:

```js
await tmpFile.copy(dstFile, { destinationPredefinedAcl: 'publicRead' })
```

Instead of:

```js
await tmpFile.copy(dstFile)
await dstFile.acl.readers.addAllUsers()
```

This isn't the only option missing, of course, but I'm just reporting my usecase.

### LINKED ISSUE #1975: Storage: add support for Per-Object Storage Class
We need to add support for Per-Object Storage Class (POSC).

Here some extra information to help:

- [POSC info from Cloud Storage documentation](https://cloud.google.com/storage/docs/per-object-storage-class)
- [POSC example with JSON API](https://cloud.google.com/storage/docs/changing-storage-classes)
- [Storage Class info from Cloud Storage documentation](https://cloud.google.com/storage/docs/storage-classes)

On top of that, google-cloud-dotnet has made an implementation which can be used as reference. 

- [Please view the Pull Request to understand the changes](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/766)

If you have any implementation questions, please direct them at jskeet

/cc @jskeet @danoscarmike

stephenplusplus:
Fixes #1937 
Fixes #1975

```js
file.setStorageClass('regional', function(err, apiResponse) {
  if (err) {
    // Error handling omitted.
  }
  // The storage class was updated successfully.
});

//-
// If the callback is omitted, we'll return a Promise.
//-
file.setStorageClass('regional').then(function() {});
```

### DIFF
--- a/packages/storage/src/file.js
+++ b/packages/storage/src/file.js
@@ -1826,8 +1826,9 @@ File.prototype.setStorageClass = function(storageClass, callback) {
   storageClass = storageClass
     .replace(/-/g, '_')
     .replace(/([a-z])([A-Z])/g, function(_, low, up) {
-      return low + '_' + up.toLowerCase();
-    });
+      return low + '_' + up;
+    })
+    .toUpperCase();
 
   this.copy(this, {
     storageClass: storageClass
--- a/packages/storage/test/file.js
+++ b/packages/storage/test/file.js
@@ -2342,7 +2342,7 @@ describe('File', function() {
       file.copy = function(newFile, options) {
         assert.strictEqual(newFile, file);
         assert.deepEqual(options, {
-          storageClass: STORAGE_CLASS
+          storageClass: STORAGE_CLASS.toUpperCase()
         });
         done();
       };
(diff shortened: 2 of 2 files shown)


packages/storage/src/file.js
```diff
@@ -1785,6 +1795,55 @@ File.prototype.save = function(data, options, callback) {
...
+ */
+File.prototype.setStorageClass = function(storageClass, callback) {
+  var self = this;
+
+  // In case we get input like `storageClass`, convert to `storage_class`.
+  storageClass = storageClass
+    .replace(/-/g, '_')
+    .replace(/([a-z])([A-Z])/g, function(_, low, up) {
+      return low + '_' + up.toLowerCase();
+    });
```
jmuk: Might be safer to append `.toLowerCase()` at the end?

Not sure, but the system-test case below said the response is uppercase (like `MULTI_REGIONAL`), some users might reuse the same values for further requests.

stephenplusplus: The upstream API allows lower case, the camelcase -> snakecase is just a helper we have in case the user couldn't remember. However, I've changed this to `toUpperCase()` all the input that we get.

PR #2005, opened 2017-02-16
spanner: fix lint errors

callmehiphop:
Looks like a few lint related issues were introduced via Spanner that are causing all the CI jobs to fail.

### DIFF
--- a/packages/spanner/src/admin/database/v1/database_admin_client.js
+++ b/packages/spanner/src/admin/database/v1/database_admin_client.js
@@ -30,7 +30,6 @@
 var configData = require('./database_admin_client_config');
 var extend = require('extend');
 var gax = require('google-gax');
-var path = require('path');
 
 var SERVICE_ADDRESS = 'spanner.googleapis.com';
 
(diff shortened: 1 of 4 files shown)


packages/spanner/src/admin/database/v1/database_admin_client.js
```diff
@@ -30,7 +30,6 @@
 var configData = require('./database_admin_client_config');
 var extend = require('extend');
 var gax = require('google-gax');
-var path = require('path');
```
stephenplusplus: @jmuk does this need to be addressed somewhere upstream?

jmuk: it was added manually to load a non public proto files and probably I've forgotten to remove them; no need to fix on the code generator, but that sounds like we should progress more automation around this area.

stephenplusplus: Ah right, that was probably my mistake when I took out the manual loading.

PR #2006, opened 2017-02-17
pubsub: fire simultaneous pull requests

### LINKED ISSUE #1974: Experiencing Extreme (10s+) Latencies with Pub/Sub
I described my situation in this GCP Pub/Sub forum post:
https://groups.google.com/forum/#!topic/cloud-pubsub-discuss/xhDvw63IVI8

> I'm using (or trying to) GCP Pub/Sub by publishes a message on an existing subscription and topic every 5 minute and then acknowledging it on the same topic. I get monitoring alerts that the process is failing after waiting for 10 seconds and timing out a few times a day. My monitoring reports that there have been 109 timeouts over the past 24 hours. I think 10 seconds is a very high threshold for waiting for a message and would love to know what is causing this.

Kamal Aboul-Hosn said to create an issue on this repo and commented that it may be a perf issue. Here's what he said:
> It looks like the Node.js Pub/Sub library has a performance issue. The way it looks to work is that it creates a pull request, waits for a response for that request, and then 10 milliseconds later creates a new pull request. This is not going to be the best way to minimize message end-to-end latency. To minimize that, one needs to have multiple outstanding pull requests simultaneously. As soon as a response comes, another request needs to be created.  I recommend entering an issue in the google-cloud-node Github repository. 
> 
> Even with that change, I expect that you will still occasionally see 10-second latencies for some messages.

#### Environment details

  - OS: Debian 8
  - Node.js version: 6.9.5
  - npm version: 4.x.x
  - google-cloud-node version: 0.8.0

#### Steps to reproduce

  1. require `@google-cloud/pubsub`
  2. create topic 
  3. create subscription
  4. send a message on topic created above (this happens every 5 minutes)
  5. immediately try to receive the same message on the same host
  6. randomly experience 10s+ timeouts

Image from monitoring tool showing frequency of timeouts over the past 24 hours:
image

stephenplusplus:
Fixes #1974 

#### To Dos

- [ ] Unit tests

### DIFF
--- a/packages/pubsub/package.json
+++ b/packages/pubsub/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/pubsub",
-  "version": "0.8.0",
+  "version": "0.9.0-0",
   "author": "Google Inc.",
   "description": "Google Cloud Pub/Sub Client Library for Node.js",
   "contributors": [
@@ -53,10 +53,12 @@
     "@google-cloud/common": "^0.12.0",
     "@google-cloud/common-grpc": "^0.1.1",
     "arrify": "^1.0.0",
+    "async": "^2.1.4",
     "extend": "^3.0.0",
     "google-gax": "^0.10.0",
     "google-proto-files": "^0.9.1",
     "is": "^3.0.1",
+    "methmeth": "^1.1.0",
     "modelo": "^4.2.0",
     "propprop": "^0.3.0",
     "uuid": "^3.0.1"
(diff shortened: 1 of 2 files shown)


packages/pubsub/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/pubsub",
-  "version": "0.8.0",
+  "version": "0.9.0-0",
```
stephenplusplus: Made a pre-release so it could be tested.

stephenplusplus:
Closed for now, as the Pub/Sub team's feedback could squash the need for this / force big API changes that would be too tiring to keep this PR up to date with. We can use this as a reference if we need to re-implement after the API makeover.

PR #2009, opened 2017-02-18
spanner: transaction docs

### LINKED ISSUE #2008: Documentation for Transaction.deleteRows() does not match behavior
Docs claim that if a callback is omitted, a Promise is returned (see: https://googlecloudplatform.github.io/google-cloud-node/#/docs/spanner/0.1.0/spanner/transaction?method=deleteRows).

When invoked without a callback, nothing is returned. See: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/7ba12d080b499e5461ba51f9e08d559d6c41aaaa/packages/spanner/src/transaction-

stephenplusplus:
Fixes #2008

The <kbd>transaction-request.js</kbd> were incorrect in a few places. This corrects those, and shows more examples for the promise workflow.

### DIFF
--- a/packages/spanner/src/database.js
+++ b/packages/spanner/src/database.js
@@ -571,7 +571,7 @@ Database.prototype.run = function(query, options, callback) {
  * database.runStream(query)
  *   .on('error', function(err) {})
  *   .on('data', function(row) {})
- *   .on('end', function() {})
+ *   .on('end', function() {});
  *
  * //-
  * // If you anticipate many results, you can end a stream early to prevent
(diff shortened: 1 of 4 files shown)


callmehiphop:
@stephenplusplus changes look fine, however upon checking CI I noticed 2 things

1. gRPC no longer builds for 0.12 (should we remove it from the test matrix?)
2. There looks to be some doc bugs in the top level object - https://travis-ci.org/GoogleCloudPlatform/google-cloud-node/jobs/202985688#L1969

stephenplusplus:
> gRPC no longer builds for 0.12 (should we remove it from the test matrix?)

Sure, was going to just punt until we get on Circle, but it couldn't hurt to remove in the meantime.

> There looks to be some doc bugs in the top level object

D'oh, fixed, and it caught a bunch of stuff.

Issue #2019, opened 2017-02-24
New x-goog-api-client information for gRPC-based APIs


jmuk:
(sorry not filed an issue so far); we are going to add x-goog-api-client header info to the gRPC-based APIs in certain format:
- common-grpc based: `"gl-node/(node version) gccl/(library version) grpc/(grpc version)"`
- APIs on top of gapic: `"gl-node/(node version) gccl/(library version) gapic/(autogen version) gax/(gax version) grpc/(grpc version)"`

Also the gapic autogen files will generate its own http header -- so that if the user chooses to use gapic layer directly, it has the information, but not with `gccl` part. This means we need to regenerate the autogen files too.

This issue tracks its progress.

The following APIs will be affected:
- [x] bigtable: common-grpc based (#2020)
- [x] datastore: common-grpc based (#2020)
- [x] language: on top of gapic (#2022)
- [x] language/gapic (#2022, #2037)
- [x] logging: common-grpc based (#2020)
- [x] logging/gapic (#2031)
- [x] monitoring/gapic (#2059)
- [x] pubsub: common-grpc based (#2020)
- [x] pubsub/gapic (#2032)
- [x] spanner: on top of gapic (#2066)
- [x] spanner/gapic (#2066)
- [x] speech: on top of gapic (#2025)
- [x] speech/gapic (#2025, #2037)
- [x] vision: on top of gapic (#2026)
- [x] vision/gapic (#2026, #2037)

jmuk:
cc: @bjwatson @omaray @danoscarmike

bjwatson:
Can we track the PRs for each API, the same way as we're doing for [Python](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3049) and [Ruby](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1268)?

bjwatson:
@stephenplusplus @jmuk Can we track the PRs for each API, the same way as we're doing for [Python](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3049) and [Ruby](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1268)?

bjwatson:
@jmuk Please note that this ordering is wrong:
> APIs on top of gapic: `"gl-node/(node version) gccl/(library version) gax/(gax version) gapic/(autogen version) grpc/(grpc version)"`

The `gapic/` token should appear before the `gax/` token:
[...]

PR #2020, opened 2017-02-24
Creates the new x-goog-api-client header.

jmuk:
This will send x-goog-api-client with the following format:
gl-node/(node version) gccl/(library version) grpc/(grpc version)

We will use this information for future tracking.

The GAPIC autogen files will have a slightly different format,
and that will be addressed later. This PR only changes the
APIs using gRPC without the GAPIC.

Updates #2019

### DIFF
--- a/packages/bigtable/src/index.js
+++ b/packages/bigtable/src/index.js
@@ -309,6 +309,10 @@ function Bigtable(options) {
     return new Bigtable(options);
   }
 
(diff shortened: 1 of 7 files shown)


stephenplusplus:
@jmuk - sorry, can you PR this again? Each gRPC API using `GrpcService` already passes in its `package.json`. The GrpcService class has access to it under `options.packageJson`.

Issue #2014, opened 2017-02-22
logging may send entries in incorrect order


ofrobots:
Two successive calls to `Log.prototype.write` should preserve the order of the log entries, as written to the Logging service. I am noticing my log entries showing up out of order.

I think the problem is due to the async behaviour in `Log.prototype.decorateEntries_`:

```js
Log.prototype.decorateEntries_ = function(entries, callback) {
  var self = this;

  async.map(entries, function(entry, callback) {
    if (!(entry instanceof Entry)) {
      entry = self.entry(entry);
    }

    var decoratedEntry = entry.toJSON();
    decoratedEntry.logName = self.formattedName_;

    self.metadata_.assignDefaultResource(decoratedEntry, function(err, entry) {
      // Ignore errors (the API will speak up if it has an issue).
      callback(null, entry || decoratedEntry);
    });
  }, callback);
};
```

Two successive calls to the above, within the timer resolution of JavaScript `Date.now()` (i.e. having the same timestamp), may get written to the network out of order depending on which log entry gets called back first from `assignDefaultResource`.

stephenplusplus:
Should we switch to `async.mapSeries`?

ofrobots:
I don't think the problem is with `async.map`. In my test-case, `entries` is always a single element array.

stephenplusplus:
Ah, right. I misunderstood the first read over. So, probably better to capture the timestamp at the top of the function, and manually assign to all created entries within `async.map`?

ofrobots:
I don't think it is related to the capture of the timestamp. Imagine two entries that are *supposed* to get the same timestamp because `log.write` was called successively twice. They *should* have the same timestamp.

It is the fetch of the default resource potentially from the network that adds non-determinism here. I think we will need to remove this non-determinism from the path by somehow.

stephenplusplus:
As a developer, I would not *expect* these to have the timestamp:

```js
log.write('...', function() {})
log.write('...', function() {})
```

If `log.write` was completely sync, it's still possible to have different timestamps on two different lines of code. For cases that require that certainty, they should remove the uncertainty and manually assign the appropriate timestamp onto their entries. Please correct me if I'm missing something.

ofrobots:
BTW, here's the test: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1997/files#diff-4becf23148fef29d25428567c8d33fc7.

ofrobots:
[...]

PR #2021, opened 2017-02-24
Add insertId on log entries

ofrobots:
JS timestamp values have a millisecond resolution. The Log service does
not guarantee to preserve insertion order for log entries with the same
timestamp. We add a globally unique `insertId` that's used as a way to
order entries with the same timestamp.

I do have some concerns about performance here. Would love to to get other's opinions.

Fixes: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2014

EDIT: unit tests can be added once I have some comments from others on the general methodology.

### DIFF
--- a/packages/logging/system-test/logging.js
+++ b/packages/logging/system-test/logging.js
@@ -351,8 +351,8 @@ describe('Logging', function() {
       var entry1 = log.entry('1');
 
       setTimeout(function() {
-        var entry3 = log.entry('3');
-        var entry2 = log.entry({ timestamp: entry3.metadata.timestamp }, '2');
+        var entry2 = log.entry('2');
+        var entry3 = log.entry({ timestamp: entry2.metadata.timestamp }, '3');
 
         // Re-arrange to confirm the timestamp is sent and honored.
         log.write([entry2, entry3, entry1], options, function(err) {


packages/logging/src/entry.js
```diff
@@ -87,6 +114,29 @@ function Entry(metadata, data) {
...
+    // UUID in our case) along with an incrementing 'event counter'. The other
+    // alternative would be to use `process.hrtime()` as the event id/timestamp,
+    // but that would be more overhead.
+    insertCounter++;
+
+    // We encode in base36 to make the string compact. We left pad with zeros to
+    // make it sortable. We left pad upto a width of 13 as a max 64-bit integer
+    // would fit in 13 digits in base36.
+    var counter = leftPad(insertCounter.toString(36), 13, '0');
+    var insertId = SERVICE_ID + '-' + counter;
```
matthewloring: The overall approach looks good. If performance ends up being a problem, I think things can be sped up substantially (~2 orders of magnitude on a simple throughput benchmark) by avoiding string concatenation and incrementing in place:

```javascript
// Initialize
var buffer = Buffer.allocUnsafe(size);
uuid.v4(null, buffer);

// Increment
for (var i = buffer.length - 1; i >= 0; i--) {
  if (buffer[i] !== 255) {
    buffer[i]++;
    break;
  } else {
    buffer[i] = 0;
  }
}

// Serialize
buffer.toString('base64');
```

stephenplusplus:
LGTM 👍

packages/logging/src/entry.js
ofrobots: +1. The one issue is that base64 doesn't preserve lexicographic order which we need here. Also, I think doing additions on `Uint32` would make the addition faster, and get rid of the loop.

I am starting to think it might be a good idea to split it out into a standalone module for the `EventId` concept.

stephenplusplus:
Just labeling this so I don't forget we still need unit tests before I merge.

ofrobots:
* Extracted the EventId concept into a module of its own: [eventid](https://www.npmjs.com/package/eventid).
* Added unit tests.

PTAL.

stephenplusplus:
@ofrobots pushed a commit with style-only changes, but noticed a system test failing, "it should preserve order of entries" -- could you take a look?

packages/logging/system-test/logging.js
```diff
@@ -369,6 +369,22 @@ describe('Logging', function() {
...
+
+      messages.forEach(function(message) {
+        log.write(log.entry(message));
+      });
+
+      setTimeout(function() {
+        log.getEntries({ pageSize: messages.length }, function(err, entries) {
+          assert.ifError(err);
+          assert.deepEqual(entries.reverse().map(prop('data')), messages);
+          done();
```
stephenplusplus: This `done()` wasn't in place earlier, but don't worry-- after adding, the test still passed :)

packages/logging/system-test/logging.js
```diff
@@ -351,8 +351,8 @@ describe('Logging', function() {
       var entry1 = log.entry('1');
 
       setTimeout(function() {
-        var entry3 = log.entry('3');
-        var entry2 = log.entry({ timestamp: entry3.metadata.timestamp }, '2');
+        var entry2 = log.entry('2');
+        var entry3 = log.entry({ timestamp: entry2.metadata.timestamp }, '3');
```
stephenplusplus: Embarrassed I didn't catch this... thanks for the fix!

ofrobots:
Fixed said test. With the given changes in PR, we now correctly order the Log Entry objects. Objects constructed earlier are now correctly recognized as being earlier.

packages/logging/system-test/logging.js
ofrobots: I should have caught this earlier too – I had been running `mocha` with `--grep` with my own test...

PR #2022, opened 2017-02-24
Update language client autogen.

jmuk:
Updates #2019

### DIFF
--- a/packages/language/src/v1/index.js
+++ b/packages/language/src/v1/index.js
@@ -1,11 +1,11 @@
-/*!
- * Copyright 2016 Google Inc. All Rights Reserved.
+/*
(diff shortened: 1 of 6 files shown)


packages/language/src/v1/index.js
```diff
@@ -1,11 +1,11 @@
-/*!
- * Copyright 2016 Google Inc. All Rights Reserved.
+/*
```
stephenplusplus: We need the `!` which marks this as a block that doesn't get parsed by JSdocs.

stephenplusplus: Then again, this is a generated file, isn't it? Might not be a big deal.

PR #2024, opened 2017-02-24
Revert "Creates the new x-goog-api-client header."

stephenplusplus:
Reverts GoogleCloudPlatform/google-cloud-node#2020

### DIFF
--- a/packages/bigtable/src/index.js
+++ b/packages/bigtable/src/index.js
@@ -309,10 +309,6 @@ function Bigtable(options) {
     return new Bigtable(options);
   }
 
-  options = extend({}, options, {
-    libVersion: require('../package.json').version
-  });
-
   var baseUrl = 'bigtable.googleapis.com';
   var adminBaseUrl = 'bigtableadmin.googleapis.com';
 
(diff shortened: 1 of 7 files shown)


jmuk:
Ugh, I didn't notice this. Why reverted? This means lacking the header info for non-gapic APIs.

cc: @bjwatson @lukesneeringer

stephenplusplus:
https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2020#issuecomment-282335050

> @jmuk - sorry, can you PR this again? Each gRPC API using GrpcService already passes in its package.json. The GrpcService class has access to it under options.packageJson.

I was suggesting we not do any code changes in `packages/src/{service}/index.js`, rather move them to `packages/common-grpc/index.js`, since it should have all the information it needs from the `package.json` it is sent from the extending service class.

Looking at the other PRs, I don't think you went with that style (which makes sense, if you never saw the comment), and I must have forgotten about that request, so now we're in a half-n-half state.

jmuk:
Created https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2164

PR #2028, opened 2017-02-25
(logging-winston) Fix example in README.md

ofrobots:
Also removed the resource section in the config as it is not needed in
common usage.

### DIFF
--- a/packages/logging-winston/README.md
+++ b/packages/logging-winston/README.md
@@ -17,7 +17,7 @@ var transport = require('@google-cloud/logging-winston');
 winston.add(transport, {
   projectId: 'grape-spaceship-123',
   keyFilename: '/path/to/keyfile.json',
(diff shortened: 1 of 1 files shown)


packages/logging-winston/README.md
```diff
@@ -17,10 +17,7 @@ var transport = require('@google-cloud/logging-winston');
 winston.add(transport, {
   projectId: 'grape-spaceship-123',
   keyFilename: '/path/to/keyfile.json',
-  level: 'warning', // log at 'warning' and above
-  resource: {
-    type: 'global'
-  }
+  level: 'warn', // log at 'warn' and above
```
stephenplusplus: Can you remove the trailing comma?

ofrobots: Done.

Issue #2027, opened 2017-02-24
Race condition in ServiceObject.prototype.get


jcking:
#### Environment details

  - OS: Ubuntu 14.04
  - Node.js version: v6.9.5
  - npm version: 3.10.10
  - google-cloud-node version: google-cloud@0.46.1 @google-cloud/pubsub@0.8.0

#### Steps to reproduce

This can be reproduced under specific conditions when using the "get or create" functionality described in `ServiceObject.protoype.get()` with `{ autoCreate: true }`. I discovered it while using the pubsub API. If you run 2 `ServiceObject.protoype.get()` at the same time with the same arguments for a resource that does not exist prior, it is possible for one to succeed and the other to fail. This does not seem like intended behavior to me.

1. Request 1 -> get
2. Request 2 -> get
3. Request 1 <- 404 Not Found
4. Request 2 <- 404 Not Found
5. Request 1 -> create
6. Request 1 <- 200 OK
7. Request 2 -> create
8. Request 2 <- 409 Conflict

If these requests occur in different orders it is possible for both to pass or one to fail and the other pass. 

Thoughts?

I have a proposed fix that modifies the `ServiceObject.prototype.get()` when `{ autoCreate: true }` that wraps the call to `ServiceObject.prototype.create()` to re-run `ServiceObject.prototype.get()` with `{ autoCreate: false }` if a `409 Already Exists` error is detected on the call to `ServiceObject.prototype.create()`.

stephenplusplus:
Sounds good!

jcking:
@stephenplusplus I also noticed [service-object.js#L221](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/5f88e99cba5932fd10dc59d6762eed66b6943e47/packages/common/src/service-object.js#L221) and [service-object.js#L225](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/5f88e99cba5932fd10dc59d6762eed66b6943e47/packages/common/src/service-object.js#L225) pass `metadata` as the third argument, however the documentation states the third argument to the callback should be the apiResponse. Are these equivalent or do you want me to modify it to be the apiResponse as well?

stephenplusplus:
Yep, `metadata` is actually the `apiResponse`. Thanks for the help!

danoscarmike:
@stephenplusplus can this one be closed now?

PR #2030, opened 2017-02-27
Fix race condition in ServiceObject.prototype.get

jcking:
Prospective fix for #2027. This could possibly change specific API behavior for users, and may effect all other child packages like pubsub as it modifies the parent class of all service objects, however it makes sure that the "get or create" method does what is expected.


stephenplusplus:
Thank you for doing this! I added some unit tests. If CI is happy, I will merge and release soon.

PR #2031, opened 2017-02-28
Regenerate logging API client

jmuk:
Updates #2019

### DIFF
--- a/packages/logging/src/v2/index.js
+++ b/packages/logging/src/v2/index.js
@@ -26,11 +26,11 @@ function v2(options) {
     scopes: v2.ALL_SCOPES
   }, options);
   var gaxGrpc = gax.grpc(options);
-  var result = {};
-  extend(result, loggingServiceV2Client(gaxGrpc));
-  extend(result, configServiceV2Client(gaxGrpc));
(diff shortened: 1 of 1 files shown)


stephenplusplus:
(`Don't merge` until ordering issue is resolved: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2019#issuecomment-283230446)

packages/logging/src/v2/index.js
```diff
@@ -15,24 +15,30 @@
...
-  extend(result, configServiceV2Client(gaxGrpc));
   extend(result, loggingServiceV2Client(gaxGrpc));
+  extend(result, configServiceV2Client(gaxGrpc));
   extend(result, metricsServiceV2Client(gaxGrpc));
   return result;
 }
 
 v2.SERVICE_ADDRESS = loggingServiceV2Client.SERVICE_ADDRESS;
-v2.ALL_SCOPES = loggingServiceV2Client.ALL_SCOPES;
+v2.ALL_SCOPES = union(
```
stephenplusplus: Same dealio as the other APIs-- all of the `ALL_SCOPES` values are the same, so the `lodash.union` dependency is unnecessary.

PR #2032, opened 2017-02-28
Regenerate pubsub autogen API client

jmuk:
updates #2019

### DIFF
--- a/packages/pubsub/src/v1/index.js
+++ b/packages/pubsub/src/v1/index.js
@@ -19,7 +19,6 @@ var publisherClient = require('./publisher_client');
 var subscriberClient = require('./subscriber_client');
 var gax = require('google-gax');
 var extend = require('extend');
-var union = require('lodash.union');
 
 function v1(options) {
   options = extend({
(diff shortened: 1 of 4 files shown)


stephenplusplus:
(`Don't merge` until ordering issue is resolved: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2019#issuecomment-283230446)

packages/pubsub/src/v1/index.js
```diff
@@ -32,5 +33,9 @@ function v1(options) {
 }
 
 v1.SERVICE_ADDRESS = publisherClient.SERVICE_ADDRESS;
-v1.ALL_SCOPES = publisherClient.ALL_SCOPES;
+v1.ALL_SCOPES = union(
```
stephenplusplus: This has a tendency to creep back on when clients are regenerated. So far, all scopes have been the same, so the `lodash.union` dependency isn't necessary. Can we remove?

packages/pubsub/src/v1/publisher_client.js
```diff
@@ -827,10 +855,11 @@ function PublisherClientBuilder(gaxGrpc) {
   }]);
   extend(this, publisherClient.google.pubsub.v1);
 
-  var grpcClients = {
-    iamPolicyClient: iamPolicyClient,
-    publisherClient: publisherClient
-  };
+  var grpcClients = merge(
```
stephenplusplus: You should get the same results with `extend(true, {}, iamPolicyClient, publisherClient)`.

PR #2033, opened 2017-03-01
Rename the optional parameters for annotate

### LINKED ISSUE #1998: language: allow raw request objects to `annotate()`
From internal feedback from the API team:

https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/language/src/document.js#L235

`annotate` has options for features, named as `entities`, `sentiment`, and `syntax`. They are different from the terms in the API reference, thus, we should put the references in the comments of each features (`entities` -> `extractEntities`, `syntax` -> `extractSyntax`, `sentiment` -> `extractDocumentSentiment`). Also better to have a link to https://cloud.google.com/natural-language/docs/reference/rest/v1/documents/annotateText#features

cc: @monattar

jmuk:
Fixes #1998

### DIFF
--- a/packages/language/src/document.js
+++ b/packages/language/src/document.js
@@ -232,18 +232,18 @@ Document.PART_OF_SPEECH = {
  *
  * @param {object=} options - Configuration object. See
  *     [documents.annotateText](https://cloud.google.com/natural-language/reference/rest/v1/documents/annotateText#request-body).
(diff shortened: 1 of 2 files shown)


stephenplusplus:
I don't think this was the solution. We still want to have our aliases, we simply need to honor the raw API terminology as well.

jmuk:
Sorry for my misunderstanding, closing this for now.

Issue #1987, opened 2017-02-13
logging-winston and logging-bunyan should deal with circular objects


ofrobots:
Originally raised here: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1977#issuecomment-278743604

Related: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1983, https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1354

ofrobots:
@stephenplusplus  I don't have the power to self-assign this.

ofrobots:
To be clear, unlike @google-cloud/logging, the higher level layers for winston and bunyan should erase circularity.

ofrobots:
I have been thinking of a way to do this that wouldn't be highly inefficient. As things stand today, `Entry.prototype.toJSON` already does two walks of the full object graph:

1. Upfront check for`isCircular`, and throw if it is.
2. The serialization via `Service.objToStruct_`.

Doing another object walk in the higher level `winston` and `bunyan` layers seems highly suboptimal.

What would be nicer, would be if I could customize the `objToStruct_` walk when the Entry has been constructed through the higher level layers.

There are two ways I can see for doing this:

1. Add a new property to entry that encodes behaviour (throw vs. erase) upon encountering a circular reference.
2. Given me a mechanism where I can hand you a pre-serialized Entry object with a `jsonPayload`.

Thoughts?

zbjornson:
I feel like it's common for users to want logging to "just work" even if it's on a best-effort basis; I don't think many people want to add try/catch and error handling for logging. So, from a general perspective, option 1 sounds nice.

ofrobots:
@stephenplusplus thoughts?

stephenplusplus:
Option 1 works for me. `options.removeCircular`, default false. (better options name welcome, especially if there is prior art)

PR #2038, opened 2017-03-02
common-grpc: objToStruct: detect or erase circular

ofrobots:
The object graph walk to encode an object as a protobuf now has the
capability to detect or remove circular references. This will be used
in follow-on PRs in `logging`.

Ref: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1987

### DIFF
--- a/packages/common-grpc/test/service.js
+++ b/packages/common-grpc/test/service.js
@@ -1924,7 +1924,7 @@ describe('GrpcService', function() {
     });
   });
 
-  describe('ObjectToStruct', function() {
+  describe('ObjectToStructConverter', function() {
     var objectToStructConverter;
 
     beforeEach(function() {


packages/common-grpc/src/service.js
```diff
@@ -696,25 +748,9 @@ GrpcService.shouldRetryRequest_ = function(response) {
...
-      if (is.undefined(value)) {
-        continue;
-      }
-
-      convertedObject.fields[prop] = GrpcService.encodeValue_(value, options);
-    }
-  }
-
-  return convertedObject;
+  console.log('common-grpc.Service.objToStruct');
```
matthewloring: Errant log.

packages/common-grpc/src/service.js
```diff
@@ -652,6 +702,8 @@ GrpcService.shouldRetryRequest_ = function(response) {
  *
  * @param {object} obj - An object to convert.
  * @param {object=} options - Configuration object.
+ * @param {boolean} options.removeCircular - Remove circular references in the
+ * object with a placeholder string.
```
stephenplusplus: Just needs some indentation:

```diff
- * object with...
+ *     object with...
```

packages/common-grpc/src/service.js
```diff
@@ -537,27 +577,37 @@ GrpcService.encodeValue_ = function(value, options) {
...
     };
   } else if (is.object(value)) {
-    convertedValue = {
-      structValue: GrpcService.objToStruct_(value, options)
-    };
+    if (this.seenObjects.has(value)) {
+      // Circular reference.
+      if (!this.removeCircular) {
+        throw new Error(
+            'Object has a circular reference – cannot encode as a struct');
```
stephenplusplus: Since I believe this error will be returned to the user, maybe:

"This object contains a circular reference. To automatically remove it, set `removeCircular` to true."

packages/common-grpc/src/service.js
```diff
@@ -496,24 +496,64 @@ GrpcService.decodeValue_ = function(value) {
 };
 
 /**
- * Convert a raw value to a type-denoted protobuf message-friendly object.
+ * A class that can be used tp convert an object to a struct. Optionally this
```
stephenplusplus: to* convert...

stephenplusplus:
Could you check if there's something already out there we can use? https://www.npmjs.com/search?q=circular%20references&page=1&ranking=optimal

ofrobots:
> Could you check if there's something already out there we can use? https://www.npmjs.com/search?q=circular%20references&page=1&ranking=optimal

The objective of this PR is to re-use the existing object graph in `objToStruct_` walk to find/erase circular references. Using alternative library (e.g. `is-circular` and `circular`) would incur multiple walks of the object graph. `circular` and kin are specifically is for JSON stringification and not applicable directly.

After this lands, we should be able to drop the `is-circular` dependency from `@google-cloud/logging`.

packages/common-grpc/src/service.js
ofrobots: Done.

packages/common-grpc/src/service.js
ofrobots: Fixed.

packages/common-grpc/src/service.js
ofrobots: Done.

packages/common-grpc/src/service.js
ofrobots: Done.

stephenplusplus:
@ofrobots  I pushed a commit which is mostly a code style / test style refactor, along with:

- Rename `ObjectToStruct` class to `ObjectToStructConverter`
- Default `stringify` and `removeCircular` to `false`

LMKWYT

packages/common-grpc/src/service.js
```diff
@@ -886,7 +768,143 @@ GrpcService.prototype.getService_ = function(protoOpts) {
...
+
+      if (is.undefined(value)) {
+        continue;
+      }
+
+      convertedObject.fields[prop] = this.encodeValue_(value);
+    }
+  }
+
+  this.seenObjects.clear();
```
ofrobots: Is the change from `delete(obj)` to `clear()` intentional? I think this changes behaviour.

stephenplusplus: Yes, thanks for catching. I put it in before I noticed that `encodeValue_` calls `convert`. Changed it back.

packages/common-grpc/test/service.js
```diff
@@ -2046,4 +1923,208 @@ describe('GrpcService', function() {
       assert.strictEqual(service, fakeService);
     });
   });
+
+  describe('ObjectToStruct', function() {
```
ofrobots: nit: ObjectToStructConverter would be a better description now?

Issue #1979, opened 2017-02-10
Pubsub - Set timeout on ack method


askmon:
#### Environment details

  - OS: Ubuntu 14.04
  - Node.js version: 4.4.2
  - npm version: 2.15.9
  - google-cloud-node version: 0.46.2

For some reason, this past week, I started to see that some of my messages were not being acked/delivered, and I have to wait for the ack deadline, as in this image: 
![image](https://cloud.githubusercontent.com/assets/711732/22829823/811bd462-ef8b-11e6-80e8-aa5b77ba57e4.png)

I tried updating pubsub but wasn't successful.

I'd like to be able to set a timeout on `ack` request, so after some time it'd give me an error and I'd be able to retry the ack.

I found this on stackoverflow: http://stackoverflow.com/questions/41901108/acked-messages-are-lingering-at-the-google-pubsub it's very similar to what happens to me, but I don't know how I'd change the grpc behavior like this....

stephenplusplus:
Thanks for reporting. As you've noticed, this currently isn't possible. To support, we would just need to allow an `options.timeout` property to the `ack` method, which we would pass through to gax. PR welcome if you're up for it!

stephenplusplus:
Fixed in #2042.

PR #2042, opened 2017-03-02
pubsub: add ability to pass timeout to ack

jcking:
See #1979.

Something like this "should" work assuming the `protoOpts` are passed as is to gax, granted I haven't tested it yet. This adds an options argument to `Subscription.prototype.ack` that forwards `options.timeout` to gax. It preserves existing behavior and replicates the style of arguments orders that is consistent with the rest of the library.

Feel free to nit pick as this was quick and dirty while investigating another issue, I was in the neighborhood.

### DIFF
--- a/packages/pubsub/src/subscription.js
+++ b/packages/pubsub/src/subscription.js
@@ -437,14 +437,13 @@ Subscription.prototype.ack = function(ackIds, options, callback) {
     ].join(''));
   }
 
-  if (!callback) {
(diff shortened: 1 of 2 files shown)


jcking:
This is related to #1974 as well.

stephenplusplus:
Sorry for the long delay on this one. I've added tests -- let's see what CI thinks!

stephenplusplus:
Thanks @jcking!

Issue #2015, opened 2017-02-22
Feature request: Add "wait for VM running" function


NickZ:
The AWS EC2 sdk has this really handy feature "waitFor": http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#waitFor-property

It calls the callback once the specified instances are in the specified state.

It would be handy to have this in the compute API, something like `vm.waitFor("RUNNING").then(...)`, instead of having to carry around the operation object returned from `createVM()` and wait for the event to be emitted.

stephenplusplus:
Thanks for the idea! What should happen if the user is accessing a VM that already exists? Do we call `getMetadata()` in the background and execute the callback if it's running? If it's not running, do we start it for them?

NickZ:
> What should happen if the user is accessing a VM that already exists? Do we call getMetadata() in the background and execute the callback if it's running? 

It should return consistent results, so if the VM already exists and is running, it should return the same results as if the VM was just spun up, so I'm guessing that would require a `getMetadata()` call in the background. I know the reason that I would call this is because I'm waiting for the VM to be in a running state so that I know that I can get the metadata containing the external IP.

> If it's not running, do we start it for them?

It shouldn't by default, as the user is just waiting for it to be running under the assumption that it already 
been started. Perhaps it could take in an "autostart" option, similar to `vm.resize()`, and set it to `false` by default?

stephenplusplus:
I like all of these ideas so far. The only thing I'd change is sticking with the `.on('event')` style of registering event listeners.

Would you like to start an implementation? I can help along the way as much or as little as necessary.

NickZ:
I can certainly give it a shot. I'll take a look this weekend.

NickZ:
I've gotten familiar with how the API internals work, and I've started implementing this. I have a few questions:

@stephenplusplus: You said you would like to stick to events for this. Do you mean that you would like for the vm object itself to emit an event (i.e. listening for the event by `vm.on('RUNNING', function(...))`)? 

Also, would it be alright if I did both event emitting and a function with a callback? Say, user could register a listener which would initiate polling and then emit when the desired state is achieved, but could also call `waitFor('RUNNING')`, which would do the same thing and just call the callback and/or resolve the promise.

stephenplusplus:
[...]

PR #2047, opened 2017-03-04
compute: VM.waitFor()

NickZ:
See issue #2015

This pull request implements `VM.waitFor()`. This function waits for the VM to be in a specified state, then calls the callback (or resolves the promise).

example:
```es6
vm.waitFor("RUNNING", options, (err, metadata) =>{
  if (!err)
    //VM is running.
});
```

It performs this by calling `getMetadata()` in the background at a specified interval (currently `2000` milliseconds).

It takes in an `options` object, with a parameter `timeout`. This is by default set to 300, but it can be set to a maximum of `600` or a minimum of `0`.

### Regarding Events
It was requested in the issue that I implement this using events. After some thought, I've decided against using events for this. I think in this case they offer too many opportunities for unexpected behavior and/or hammering the API.

Events work on things like Operations because the event flow goes like this:

	running->complete OR running->error.

Once the Operation is completed (or returned with an error), it will stay like that, forever. If a user calls `operation.on("complete")`, there's no problem with having to keep polling for statuses after it's reached "complete" or "error", because it will stay there. 

However, with a VM, it can go from RUNNING to SUSPENDED to STOPPED and back any number of times. The state is transient, and isn't as simple to handle as an Operation.

Say a user calls `vm.once("RUNNING")`. This is fine because we will wait until it enters the state "RUNNING", and the event triggers, and stops polling because there are no more listeners.

However, say that a user calls `vm.on("RUNNING")`. It would be a reasonable assumption for the user to expect this to trigger every time the VM enters the "RUNNING" state, since they know that the VM state is transient. Because of this, this will require us to keep polling for status, forever, until the user removes the listener. This means a lot of superfluous API requests, and angry Google engineers.

We could just treat every "on" as "once", and explicitly say so in the API documentation. However, I don't know of many things that do this, and it would be unexpected by the user who is used to "on" meaning "on", unless they carefully read the documentation.

We could keep the number of api requests to a minimum to not hammer the server, say, every 15 seconds, but it's possible for the VM to transition states and back in that amount of time, and the event wouldn't get triggered.

We could carry around every single operation that is performed on the VM and emit when those have emitted, but that would require a large amount of architecting, and we want this to work independent of any operation executed in the script (say, a user stops and starts the VM in the web console).

This is not to say that events should not be implemented, I just couldn't find satisfactory solutions for these issues, and that's why I decided not to implement them. If you come up with some solutions for these issues, let me know and I can take another crack at implementing this with events.

### Design Considerations.
So anyway, this is what I kept in mind when I designed `VM.waitFor()`. 

I didn't want to passively enable people to ping the API continously, so I added an `options` field with a `timeout` parameter. By default, this is 300 seconds, but the maximum is 600 seconds.

Since it's a one-time event, it's predictable: it will callback when it reaches the specified state with the metadata, or it will return an error. It's independent of any operation, and will work whether or not the user had performed any operations on the VM in the script.

### Issues
I'm not quite sure that I have error handling correct. If it times out, it passes back an error with a `name` field set to "WaitForTimeout". I'm not sure if this is the correct procedure, so if you could please take a look.

### DIFF
--- a/packages/compute/src/vm.js
+++ b/packages/compute/src/vm.js
@@ -882,7 +882,7 @@ VM.prototype.waitFor = function(status, options, callback) {
     timeout = Math.min(Math.max(options.timeout, 0), 600);
   }
 
-  if (!VALID_STATUSES.includes(status)) {
+  if (VALID_STATUSES.indexOf(status) === -1) {
     throw new Error('Status passed to waitFor is invalid.');
   }
 


stephenplusplus:
Hey @NickZ, just wanted to let you know I appreciate this PR very much, and your level of thought into it. We've been on a bit of a release-after-release streak lately, so it's been hard to prioritize the review for this, but I will get to it, and your help has not been taken for granted. Thanks for your patience!

packages/compute/src/vm.js
```diff
@@ -45,6 +45,24 @@ var DetachDiskError = createErrorClass('DetachDiskError', function(message) {
   this.message = message;
 });
 
+/**
+ * Interval for polling during waitFor.
+ */
+var WAIT_FOR_POLLING_INTERVAL = 2000;
```
lukesneeringer: Should this be a cadence, or a backoff algorithm?
There is no API or rate limit involved, so a candence might be entirely fine.

packages/compute/src/vm.js
```diff
@@ -809,6 +829,139 @@ VM.prototype.request = function(reqOpts, callback) {
...
+ * @param {string} status - The status to wait for. This can be:
+ *  - "PROVISIONING"
+ *  - "STAGING"
+ *  - "RUNNING"
+ *  - "STOPPING"
+ *  - "SUSPENDING"
+ *  - "SUSPENDED"
+ *  - "TERMINATED"
+ * @param {object=} options - Configuration object.
+ * @param {number} options.timeout - The number of seconds to wait until
```
lukesneeringer: At the point that you are taking a `timeout` option, you could also take an `interval` option, and default it to `2000`.

packages/compute/src/vm.js
```diff
@@ -809,6 +829,139 @@ VM.prototype.request = function(reqOpts, callback) {
...
+  options = options || {};
+  // If timeout not set, set to default of 300 seconds.
+  if (!options.timeout) {
+    options.timeout = 300;
+  }
+  // If timeout is more than 600 seconds, set to 600 seconds.
+  else if (options.timeout > 600) {
+    options.timeout = 600;
+  }
+  // If timeout is less than 0, set to 0 seconds.
```
lukesneeringer: This is probably not what you want. A value of `0` on line 934 above was changed to five minutes; presumably any value below zero should be set to the default, also.

That actually makes this code easier:

```javascript
// The timeout should default to five minutes, and be less than
// or equal to 10 minutes.
options.timeout = Math.max(options.timeout || 0, 0);
options.timeout = options.timeout || 300;
options.timeout = Math.min(options.timeout, 600);
```

lukesneeringer: Alternatively, skip the negative number check entirely. If someone goes out of their way and explicitly puts in a negative timeout, just let them get a timeout. :-p

packages/compute/src/vm.js
```diff
@@ -809,6 +829,139 @@ VM.prototype.request = function(reqOpts, callback) {
...
+  }
+  var now = new Date();
+  this.waiters.push({
+    status: status,
+    timeout: options.timeout,
+    startTime: now.getTime() / 1000,
+    callback: callback
+  });
+
+  if (this.hasActiveWaiters === false) {
```
lukesneeringer: What happens if someone calls `VM.waitFor` in two separate concurrent threads?

packages/compute/src/vm.js
```diff
@@ -809,6 +829,139 @@ VM.prototype.request = function(reqOpts, callback) {
...
+ * };
+ *
+ * //-
+ * // If the callback is omitted, we'll return a Promise.
+ * //-
+ * vm.waitFor('RUNNING', options).then(function(data) {
+ *   var metadata = data[0];
+ * });
+ */
+VM.prototype.waitFor = function(status, options, callback) {
```
lukesneeringer: One thing this function seems not to do is actually check if the status is _already_ the desired status. If it is, we should just call the callback and short-circuit the rest of the logic.

packages/compute/src/vm.js
lukesneeringer: Addendum: Yeah, cadence is fine.

packages/compute/src/vm.js
NickZ: Yeah, this probably isn't concurrent thread-safe. However, does google-cloud-node support javascript implementations that use concurrent threading? Looking through the rest of the code, it doesn't seem like it was designed to be. For example, I designed this based off listenForEvents_ in common/operation.js, and it behaves the same way (see here: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/common/src/operation.js#L120)

NickZ:
@lukesneeringer what do you think about the error messages? Do you think they look OK?

packages/compute/src/vm.js
NickZ: We would still have to make a getMetadata call in order to check the status, and that happens immediately when we call startPolling_?

packages/compute/src/vm.js
lukesneeringer: To be honest, I am a new maintainer, and not sure. Summoning @stephenplusplus for the answer to this. (I am fine being told this is not a concern.)

packages/compute/src/vm.js
lukesneeringer: Sold.

lukesneeringer:
Am I blind? I do not _see_ error messages. :boggle:

packages/compute/src/vm.js
```diff
@@ -809,6 +829,139 @@ VM.prototype.request = function(reqOpts, callback) {
...
+        self.waiters.splice(index, 1);
+      } else if (metadata && metadata.status === self.waiters[index].status) {
+        self.waiters[index].callback(null, metadata);
+        self.waiters.splice(index, 1);
+      } else {
+        if ((now.getTime() / 1000) - self.waiters[index].startTime >=
+              self.waiters[index].timeout) {
+          self.waiters[index].callback({
+            name: 'WaitForTimeout'
+          },null);
```
NickZ: This is an error message here. It gets called when the timeout is triggered.

packages/compute/src/vm.js
```diff
@@ -809,6 +829,139 @@ VM.prototype.request = function(reqOpts, callback) {
...
+  }
+  // If timeout is less than 0, set to 0 seconds.
+  else if (options.timeout < 0) {
+    options.timeout = 0;
+  }
+
+  if (VALID_STATUSES.indexOf(status) === -1) {
+    callback({
+      name: 'StatusNotValid'
+    }, null);
```
NickZ: This is another error message, which gets triggered if the user passes a status to check that isn't valid.

NickZ:
@lukesneeringer Apologies, I added comments to point out where they were.

packages/compute/src/vm.js
NickZ: Wait, sorry, this won't work. I only have one polling interval running at any one time to minimize the number of API requests, and it checks every waiter on each API response. If we add this, we'd have to either :
a) change the polling interval for every `waitFor()` currently waiting after each `waitFor()` call with an `options.timeout`, which would confuse users, or 
b) change it so it polls for each individual `waitFor()` call, which would increase the amount of superfluous API requests.
Neither of these is desirable. Besides, we probably don't want to go any lower than 2 seconds for this, and user will probably not want to set the interval for anything longer than they have to.

lukesneeringer:
Ah, okay. I was looking for sentences when I was scanning.

I am fine with the codes, but send an `Error` object, not a plain dictionary-like object.

packages/compute/src/vm.js
lukesneeringer: I am convinced.

packages/compute/src/vm.js
lukesneeringer: @stephenplusplus Can you speak to this, and then this PR is probably good to go.

NickZ: Maybe he doesn't know either? 😛

stephenplusplus: Sorry guys, I completely missed this thread (ha). This is a fine way to do it-- we do not have any management for concurrent Node processes.

NickZ: @lukesneeringer It sounds like we're ready to go?

lukesneeringer:
@stephenplusplus This looks good to me. Do you see anything?

stephenplusplus:
Tests! I'll tag in and get on that. Thanks @NickZ!

stephenplusplus:
@NickZ please take a look at my latest commit to see that all functionality is as intended.

NickZ:
Looks good to me. I am ok with these changes.

stephenplusplus:
Awesome, thanks for the idea and all of the work on this!

PR #2054, opened 2017-03-06
language: re-work encoding property

### LINKED ISSUE #2018: fix "encoding" parameter for language API
Also from internal feedback;

Currently, the "encoding" parameter for Language API belongs to a Document object, but the encoding specifies the data for the server response, conceptually it does not belong to Document. This is indeed confusing though.

The encoding parameter has the effect on the offset values (the character position) in the request.

Think that you have a text file encoded in UTF-8, and you want to analyze the file. In Node.JS world, the code will look like:
```javascript
const fs = require('fs');
const language = require('@google-cloud/language')();
fs.readFile('./test.txt', {encoding: 'utf8'}, function(err, content) {
  language.text(content, {encoding: 'utf8'}).annotate(..., function(err, resp) {
    ...
  });
});
```

This is actually wrong. JavaScript strings are always UTF16, the server will successfully annotate the content, but the response contains the character position as if the content is encoded as utf8, that can mismatch with the `content`.

Things will be more complicated if the file is somewhere else, like GCS. Think that the UTF-8 file is on a GCS bucket, like:
```javascript
const gcs = require('@google-cloud/storage')();
const language = require('@google-cloud/language')();
let file = gcs.bucket('my-bucket').file('test.txt');
language.text(file, {encoding: 'utf8'}).annotate(..., function(err, resp) {
   ...
});
```

The `resp` still contains the data regarding the position based on UTF8, this can mismatch with the position when the Node.JS code reads the content of `file` into a string. When the JS programmer will read the `file` content into a string, `language.text(file, {encoding: 'utf16'})` will still succeed and get the right value. (although, if the JS programmer wants to treat the `file` content as a buffer or some byte stream, UTF8 encoding is still correct).

I think this is quite confusing, and the proper handling of encoding is really difficult -- but I'd say, the current structure doesn't capture the API concept properly. It's more like a concept of how JS programmer will handle the content, not how the content is encoded.

Suggestions:
- remove encoding parameter from the document. Add it as the optional parameter of methods like `analyze`
- introduce default encoding -- I think the default encoding would be UTF16, since it's the encoding of a string. Might be worth switching the default encoding based on the type of the content, like UTF16 for string while UTF8 for buffer.

some personal thoughts:
- add aliases of encoding; `'string'` for an alias of UTF16, and `'buffer'` for UTF8, since many JS programmers are not aware of text encoding of JS string.
- use a different word for the `encoding` option in the JS API surface; I guess users will easily misunderstand and wrongly set the parameters.

Cc: @monattar

stephenplusplus:
Fixes #2018

- [x] remove encoding parameter from the document. Add it as the optional parameter of methods like analyze
- [x] introduce default encoding -- I think the default encoding would be UTF16, since it's the encoding of a string. Might be worth switching the default encoding based on the type of the content, like UTF16 for string while UTF8 for buffer.
- [x] add aliases of encoding; 'string' for an alias of UTF16, and 'buffer' for UTF8, since many JS programmers are not aware of text encoding of JS string.
- [x] use a different word for the encoding option in the JS API surface; I guess users will easily misunderstand and wrongly set the parameters. **NOTE** I kept both, `encoding` and `encodingType` are allowed.

#### To Dos

- [x] @monattar review
- [x] Tests

@monattar could you please take a look?

#2018 explained that `encodingType` was not a value attached to a document. But, the "Document" object that we have in our API is just meant to cache content that will be used again for future analysis. So, since the "Document" holds the value, it seems like a fine place to set the encoding type (by the user or by us, using defaults). By not allowing users to set it at the time of Document creation, we are forcing them to set it each time they make a method call, even though the encodingType should be the same each time-- right?

### DIFF
--- a/packages/language/src/document.js
+++ b/packages/language/src/document.js
@@ -67,7 +67,7 @@ function Document(language, config) {
   var content = config.content || config;
 
   this.api = language.api;
-  this.encodingType = 'UTF16';
+  this.encodingType = this.detectEncodingType_(config);
 
   this.document = {};
 
@@ -232,7 +232,8 @@ Document.PART_OF_SPEECH = {
  * @param {object=} options - Configuration object. See
  *     [documents.annotateText](https://cloud.google.com/natural-language/docs/reference/rest/v1/documents/annotateText#features).
  * @param {string} options.encoding - `UTF8` (also, `buffer`), `UTF16` (also
- *     `string`), or `UTF32`. (Alias for `options.encodingType`). See
+ *     `string`), or `UTF32`. (Alias for `options.encodingType`). Default:
+ *     'UTF8' if a Buffer, otherwise 'UTF16'. See
  *     [`EncodingType`](https://cloud.google.com/natural-language/reference/rest/v1/EncodingType)
  * @param {boolean} options.entities - Detect the entities from this document.
  *     By default, all features (`entities`, `sentiment`, and `syntax`) are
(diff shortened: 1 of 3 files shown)


packages/language/src/document.js
```diff
@@ -44,8 +44,6 @@ var prop = require('propprop');
  *     object to specify the encoding and/or language of the document, use this
  *     property to pass the inline content of the document or a Storage File
  *     object.
- * @param {string} options.encoding - `UTF8`, `UTF16`, or `UTF32`. See
- *     [`EncodingType`](https://cloud.google.com/natural-language/reference/rest/v1/EncodingType).
```
monattar: Should we add comments explaining the encoding defaults?

stephenplusplus: 👍 Done

stephenplusplus:
@monattar thanks for the review!

@callmehiphop PTAL.

Issue #1782, opened 2016-11-09
language: verbose mode


callmehiphop:
We've received some feedback for our Natural Language implementation from the API team. They have some concerns about our `verbose` option. 

If you're unfamiliar with said option, it essentially collapses data that was deemed non-crucial. Using the following text, please refer to the examples below - *Google is an American multinational technology company specializing in Internet-related services and products.*

<details>
  <summary>Non-verbose Entities</summary>

  ```js
  entities = {
    organizations: [
      'Google'
    ],
    places: [
      'American'
    ]
  }
  ```
</details>

<details>
  <summary>Verbose Entities</summary>

  ```js
  entities = {
    organizations: [
      {
        name: 'Google',
        type: 'ORGANIZATION',
        metadata: {
          wikipedia_url: 'http: *   //en.wikipedia.org/wiki/Google'
        },
        salience: 65.137446,
        mentions: [
          {
            text: {
              content: 'Google',
              beginOffset: -1
            }
          }
        ]
      }
    ],
    places: [
      {
        name: 'American',
        type: 'LOCATION',
        metadata: {
          wikipedia_url: 'http: *   //en.wikipedia.org/wiki/United_States'
        },
        salience: 13.947371,
        mentions: [
          {
            text: {
              content: 'American',
              beginOffset: -1
            }
          }
        ]
      }
    ]
  }
  ```
</details>
<br>

I believe the intention for `verbose` was to supply the user with the information returned from the API that covered the most common use cases, however we may actually be supplying the user with too little information making the non-verbose mode impractical. I'd like to use this issue to open a line of dialogue to see if we can improve our implementation in this area and to address any concerns from the API team.

jmdobry:
Wait, slow down. How did you do that accordion thing?

bjwatson:
@callmehiphop Do you have a PR for the NL work so far?

callmehiphop:
@jmdobry [details!](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) (@stephenplusplus showed me that one)

@bjwatson I still need to send it through! However this issue shouldn't affect the outcome of that PR, we agreed that this was sorta outside the scope of what we should try and accomplish.

bjwatson:
@callmehiphop Agreed.

stephenplusplus:
[...]

PR #2056, opened 2017-03-06
language: remove verbose option

callmehiphop:
Closes https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1782

---

#### TODOs

- [x] Docs
- [x] Tests
  - [x] System
  - [x] Unit

### DIFF
--- a/packages/language/system-test/language.js
+++ b/packages/language/system-test/language.js
@@ -315,24 +315,11 @@ describe('Language', function() {
       try {
         assert.ifError(err);
 
-        assert.strictEqual(annotation.language, 'en');
-
-        assert(is.object(annotation.sentiment));
-        assert(is.number(annotation.sentiment.score));
-        assert(is.number(annotation.sentiment.magnitude));
-
-        assert(is.array(annotation.entities));
-        assert.strictEqual(annotation.entities.length, 3);
-        assert(is.object(annotation.entities[0]));
-
-        assert(is.array(annotation.sentences));
-        assert(is.object(annotation.sentences[0]));
-
-        assert(is.array(annotation.tokens));
-        assert(is.object(annotation.tokens[0]));
-        assert.strictEqual(annotation.tokens[0].text.content, 'Hello');
-
-        assert(is.object(apiResponse));
+        assert.strictEqual(annotation.language, apiResponse.language);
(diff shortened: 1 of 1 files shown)


callmehiphop:
@stephenplusplus before I go in and do it, my plan for the docs was to remove a lot of the params detailing the objects we return to the user in favor of just linking them to the upstream docs. Using `document.annotate()` as [an example](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/language/src/document.js#L251), I'm not sure we need to document this object so heavily if it doesn't really vary from the raw API response too much.

WDYT?

stephenplusplus:
I think if we're returning the raw API response, linking is good. But, if we do any re-mapping, we have to document it.

However, I think @jgeewax wanted to kill the grouping of types, which I *think* means we are only ever returning the raw API response. So, no special docs would be needed, if we do that.

callmehiphop:
Got it, could you PTAL at my latest changes to see if what we'll now be providing in the callback looks ok? If so, I'm going to get started on tests and docs.

stephenplusplus:
LGTM.

callmehiphop:
@stephenplusplus I think I'm done here if you would kindly take a look!

packages/language/system-test/language.js
```diff
@@ -339,66 +310,20 @@ describe('Language', function() {
...
-  function validateAnnotationVerbose(callback) {
+  function validateAnnotation(callback) {
     return function(err, annotation, apiResponse) {
       try {
         assert.ifError(err);
 
         assert.strictEqual(annotation.language, 'en');
 
         assert(is.object(annotation.sentiment));
+        assert(is.number(annotation.sentiment.score));
```
stephenplusplus: For these, let's just check that `annotation` === apiResponse.sentiment` (or whatever). Keep anything that checks that the correct amount of results are returned, but type-checking isn't necessary anymore.

stephenplusplus:
All looks good to me, just the one thing about simplifying the system test checks now that we're just passing back the raw response.

packages/language/system-test/language.js
callmehiphop: Done!

stephenplusplus: Woo!

PR #2071, opened 2017-03-10
spanner: fix keep alive implementation

### LINKED ISSUE #2061: Spanner: Error: Session not found
#### Environment details

  - OS: Ubuntu 16.04 LTS
  - Node.js version: 7.7.1
  - npm version: 4.4.0
  - google-cloud-node version: {"@google-cloud/spanner": "^0.1.1"}

#### Steps to reproduce

  1. require `@google-cloud/spanner`
  2. Connect to Spanner
  3. Select an instance
  4. Select a database
  5. Select a table
  6. Call function table.read
  7. Get following error message:

```
Error: Session not found: projects/bt-bsp/instances/test-instance-asia/databases/test/sessions/AOJoGSnIR9eLi-hXzRBqrp0ufJKnafwK7FvtRcOo8Zn2ac1WgqUGx_IqQlbi                                                                                           
    at ClientReadableStream._emitStatusIfDone (/root/breaktime-bsp-api/node_modules/grpc/src/node/src/client.js:201:19)
    at ClientReadableStream._readsDone (/root/breaktime-bsp-api/node_modules/grpc/src/node/src/client.js:169:8)
    at readCallback (/root/breaktime-bsp-api/node_modules/grpc/src/node/src/client.js:242:12)
```

I tried to search this error message on Google but get nothings. I think this problem isn't make by Spanner client but happen when I use it.

Thanks!

callmehiphop:
Closes #2061

### DIFF
--- a/packages/spanner/test/session-pool.js
+++ b/packages/spanner/test/session-pool.js
@@ -1563,7 +1563,7 @@ describe('SessionEvictor', function() {
       sessionEvictor.evict(CONFIG, RESOURCE, AVAILABLE);
     });
 
-    it('should run a query & update evicted status on session', function(done) {
+    it('should set evicted to true when a session expires', function(done) {
       sessionEvictor.evictor.evict = function() {
         return true;
       };
@@ -1573,6 +1573,23 @@ describe('SessionEvictor', function() {
 
         setImmediate(function() {
           assert.strictEqual(RESOURCE.obj.evicted_, true);
+          done();
+        });
+      };
+
+      sessionEvictor.evict(CONFIG, RESOURCE, AVAILABLE);
+    });
+
+    it('should set evicted to false when sessions dont expire', function(done) {
+      sessionEvictor.evictor.evict = function() {
+        return true;
+      };
+
+      RESOURCE.obj.keepAlive = function(callback) {
+        callback(null);
+
+        setImmediate(function() {
+          assert.strictEqual(RESOURCE.obj.evicted_, false);
           done();
         });
       };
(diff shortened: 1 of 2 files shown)


packages/spanner/system-test/spanner.js
```diff
@@ -396,6 +396,10 @@ describe('Spanner', function() {
         done();
       });
     });
+
+    it('should keep the session alive', function(done) {
+      session.keepAlive(done);
```
stephenplusplus: Is there a more elaborate way to test that it's working at keeping the session alive? For example, if a session normally expires after `x` ms, can we test that keepAlive keeps a session still alive after `x` ms?

packages/spanner/test/session-pool.js
```diff
@@ -1533,55 +1533,55 @@ describe('SessionEvictor', function() {
...
         return true;
       };
 
-      var session = extend({}, SESSION, {
-        run: function(query, callback) {
-          assert.strictEqual(query, 'SELECT 1');
-
-          callback(new Error('Error.'));
+      RESOURCE.obj.keepAlive = function(callback) {
+        callback(new Error('Error.'));
```
stephenplusplus: I think we also need a test to see if the value is set properly when an error is not returned.

packages/spanner/test/session.js
```diff
@@ -231,6 +231,28 @@ describe('Session', function() {
     });
   });
 
+  describe('keepAlive', function() {
+    it('should correctl call and return the gax API', function() {
```
stephenplusplus: correctly*

packages/spanner/system-test/spanner.js
callmehiphop: So sessions last an hour, I'm not really sure this is something we want to system test.

PR #2073, opened 2017-03-11
Correct how custom metadata properties are specified in options object

almstrand:
The current way of specifying custom metadata properties, per the source code documentation, will not cause the upload to fail. However if you call `getMetadata` on the uploaded file, the custom metadata properties will be missing from the response.

Custom metadata properties need to be nested one level deeper in the JSON document, i.e.:
```
let options = {
    metadata: {
        metadata: {
            a: 'b'
        }
    }
};
```
instead of
```
let options = {
    metadata: {
        a: 'b'
    }
};
```


stephenplusplus:
Thanks @almstrand!

PR #2082, opened 2017-03-15
Introduce autogenerated Datastore client.

jmuk:
RE: #1859


packages/datastore/src/index.js
```diff
@@ -545,3 +545,4 @@ Datastore.Query = Query;
 Datastore.Transaction = Transaction;
 
 module.exports = Datastore;
+module.exports.v1 = require('./v1');
```
stephenplusplus: Can you add a test that this is exported?

PR #2083, opened 2017-03-15
log: accept removeCircular as an option to write

ofrobots:
Callers of write (and related functions) can now pass removeCircular in
the options. This allows log entries to have circular references that
are replaced with '[Circular]' placeholder strings when encoded.

### DIFF
--- a/packages/logging/src/entry.js
+++ b/packages/logging/src/entry.js
@@ -133,8 +133,9 @@ Entry.fromApiResponse_ = function(entry) {
 /**
  * Serialize an entry to the format the API expects.
  *
- * @param {object} options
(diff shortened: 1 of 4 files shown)


ofrobots:
Do not land yet.

Turns out that this may not be the best way to do this – the `options` param to `Log.prototype.write` is expected to passed to the service (ie. is a service object). It is not acceptable for me to add a new property to it.

My preference would be to add a options object to the Log constructor. Thoughts?

ofrobots:
Oh well, the browser hadn't refreshed. This is already landed. This isn't a big deal as-landed since no one is passing in removeCircular as an argument yet. Let me put together a follow-on.

PR #2084, opened 2017-03-15
logging: fix jsdoc for Log constructor

ofrobots:


packages/logging/src/log.js
```diff
@@ -52,7 +52,8 @@ var Metadata = require('./metadata.js');
  * @alias module:logging/log
  * @constructor
  *
- * @param {object} options - [Configuration object](#/docs).
+ * @param {module:logging} logging - parent Logging instance.
```
stephenplusplus: We don't advertise this, because it has no meaning to a user. Instead, we put it in a private "Developer Documentation" block above this one. (Obviously, not this time since we made a mistake)

If you're willing, please snoop around the other files to see how we denote that.

packages/logging/src/log.js
```diff
@@ -52,7 +52,8 @@ var Metadata = require('./metadata.js');
...
+ * @param {string} name - name of the log.
```
stephenplusplus: "Name" (casing) please.

PR #2090, opened 2017-03-15
Make removeCircular a property of the Log object

ofrobots:
Follow-on from https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2083.

Thoughts?

### DIFF
--- a/packages/logging/test/log.js
+++ b/packages/logging/test/log.js
@@ -145,8 +145,6 @@ describe('Log', function() {
     });
 
     it('should accept and localize options.removeCircular', function() {
-      assert.strictEqual(log.removeCircular_, false);
-
       var options = { removeCircular: true };
       var log = new Log(LOGGING, LOG_NAME_FORMATTED, options);
       assert.strictEqual(log.removeCircular_, true);


packages/logging/src/log.js
```diff
@@ -57,13 +57,19 @@ var Metadata = require('./metadata.js');
  * @constructor
  *
  * @param {string} name - Name of the log.
+ * @param {object} options - Configuration object.
+ * @param {boolean} options.removeCircular - Replace circular references in an
```
stephenplusplus: `(Default: `false`)`

packages/logging/src/log.js
```diff
@@ -57,13 +57,19 @@ var Metadata = require('./metadata.js');
  * @constructor
  *
  * @param {string} name - Name of the log.
+ * @param {object} options - Configuration object.
```
stephenplusplus: `object=`

packages/logging/src/log.js
```diff
@@ -520,8 +526,6 @@ Log.prototype.warning = function(entry, options, callback) {
  * @param {object[]} options.labels - Labels to set on the log.
  * @param {object} options.resource - A default monitored resource for entries
  *     where one isn't specified.
- * @param {boolean} options.removeCircular - Replace circular references in an
```
stephenplusplus: This is still a useful option on the individual write. I think we should keep this, and if not specified, default to the value assigned on the Log.

packages/logging/test/log.js
```diff
@@ -139,6 +139,14 @@ describe('Log', function() {
         }
       });
     });
+
+    it('should accept and localize options.removeCircular', function() {
+      assert.strictEqual(log.removeCircular_, false);
+
+      var options = { removeCircular: true };
+      log = new Log(LOGGING, LOG_NAME_FORMATTED, options);
```
stephenplusplus: var log

packages/logging/src/log.js
```diff
@@ -57,13 +57,19 @@ var Metadata = require('./metadata.js');
...
  * @example
  * var log = logging.log('syslog');
  */
-function Log(logging, name) {
+function Log(logging, name, options) {
+  options = options || {};
+
   this.formattedName_ = Log.formatName_(logging.projectId, name);
   this.name = this.formattedName_.split('/').pop();
+  this.removeCircular_ = !!options.removeCircular;
```
stephenplusplus: Need a test to confirm this defaults to false.

packages/logging/src/log.js
```diff
@@ -619,24 +621,16 @@ Log.prototype.write = function(entry, options, callback) {
...
-    options = {};
-  }
-
   async.map(entries, function(entry, callback) {
     if (!(entry instanceof Entry)) {
       entry = self.entry(entry);
     }
 
-    var decoratedEntry = entry.toJSON(options);
+    var decoratedEntry = entry.toJSON({ removeCircular: self.removeCircular_ });
```
stephenplusplus:
```js
{
  removeCircular: self.removeCircular_
}
```

(formatting)

packages/logging/src/log.js
ofrobots: That's do-able, but I don't think there is a strong use-case for this. It will be marginally unsightly: we'll have to remove this property before adding `options` to `reqOpts`.

stephenplusplus: I'll take your word for it on the use case. If it's not common, then let's keep it out.

packages/logging/src/log.js
ofrobots: Done.

packages/logging/src/log.js
ofrobots: Done.

packages/logging/src/log.js
ofrobots: Done.

packages/logging/test/log.js
ofrobots: Done.

packages/logging/src/index.js
```diff
@@ -452,13 +452,16 @@ Logging.prototype.getSinksStream = common.paginator.streamify('getSinks');
  * @resource [Log Overview]{@link https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/projects.logs}
  *
  * @param {string} name - Name of the existing log.
+ * @param {object} options - Configuration object.
+ * @param {boolean} options.removeCircular - Replace circular references in an
```
stephenplusplus: Sorry I didn't see this before, but for this line and where it's repeated, there's a grammar error: "in an logged objects"

stephenplusplus: This is also missing the default: false part.

packages/logging/test/log.js
```diff
@@ -139,6 +143,14 @@ describe('Log', function() {
         }
       });
     });
+
+    it('should accept and localize options.removeCircular', function() {
+      assert.strictEqual(log.removeCircular_, false);
```
stephenplusplus: Let's just ditch this check.

ofrobots: Done.

PR #2096, opened 2017-03-17
common: allow promises to resolve with a single arg

callmehiphop:


packages/common/src/util.js
```diff
@@ -687,13 +687,18 @@ util.getUserAgentFromPackageJson = getUserAgentFromPackageJson;
  * Wraps a callback style function to conditionally return a promise.
  *
  * @param {function} originalMethod - The method to promisify.
+ * @param {object=} options - Promise options.
+ * @param {boolean} options.singular - Resolve the promise with single arg
```
stephenplusplus: Is there a standard name for this configuration?

packages/common/src/util.js
```diff
@@ -723,7 +728,11 @@ function promisify(originalMethod) {
           return reject(err);
         }
 
-        resolve(callbackArgs);
+        if (options.singular && callbackArgs.length === 1) {
```
stephenplusplus: I think we also need a test for when they specified `singular`, but `callbackArgs.length` is > 1.

packages/common/src/util.js
callmehiphop: Well, I think the closest would be spread, but that would imply that we're passing multiple arguments which this does not. This is unique behavior for us since we initially decided to only ever pass back arrays.

packages/common/src/util.js
callmehiphop: done!

Issue #2068, opened 2017-03-09
Preserve missing project ID error stack trace


gajus:
I am referring to:

https://github.com/GoogleCloudPlatform/google-cloud-node/blob/1939deaa32298b3cbe123686e5c52f480c1d7ee2/packages/common/src/util.js#L55

Error stack trace is captured at the time of constructing the `Error` instance.

As a result, the error stack trace does not relate to the origin site of the error, e.g.

```
Error: Sorry, we cannot connect to Cloud Services without a project ID. You may specify one with an environment variable named "GCLOUD_PROJECT". See https://googlecloudplatform.github.io/google-cloud-node/#/docs/guides/authentication for a detailed guide on creating an authenticated connection.
    at Object.<anonymous> (/Users/gajus/Documents/dev/applaudience/movie-editor/node_modules/google-cloud/node_modules/@google-cloud/common/src/util.js:55:29)
    at Module._compile (module.js:571:32)
    at Module._extensions..js (module.js:580:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/gajus/Documents/dev/applaudience/movie-editor/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/gajus/Documents/dev/applaudience/movie-editor/node_modules/google-cloud/node_modules/@google-cloud/common/src/service-object.js:32:12)

```

Consider using the following approach to construct errors:

https://github.com/gajus/xfetch/blob/0d8cd012354b0bb8b67162f74b2a9e126d3bdabb/src/errors.js

eoogbe:
Thanks for filing this issue. We agree that this would be a good issue to fix, and we invite you or another community member to contribute a pull request (PR), which we would be happy to review.

stephenplusplus:
This issue was moved to https://github.com/googleapis/nodejs-common/issues/12.

PR #2100, opened 2017-03-17
core: only recurse relevant request objects

### LINKED ISSUE #1908: Project ID required even if running script from Compute Engine
```js
var datastore = require('@google-cloud/datastore')();
```

The line above used on compute engine (not app engine) will still throw the error below -

```js
[Error: Sorry, we cannot connect to Google Cloud Services without a project ID. You may specify one with an environment variable named "GCL
OUD_PROJECT". See https://googlecloudplatform.github.io/google-cloud-node/#//docs/guides/authentication for a detailed guide on creating an
 authenticated connection.]
```

Yet in the documentation it says - 

> If you are running your app on Google Compute Engine, you won't need to worry about supplying connection configuration options to @google-cloud/datastore— we figure that out for you.

That is not the case.

To reproduce create a google Compute engine instance and run a node.js script without a supplied project ID it will fail.

stephenplusplus:
Fixes #1908

Previously, we took an input object that would go to `request` and iterated over all of the properties. This turned out to be an issue when the request object had properties containing circular references. What we really want to do is loop a select number of properties which may include the projectId token:

- `reqOpts.qs` (an object which gets serialized into a query string)
- `reqOpts.json` (the JSON payload)
- `reqOpts.uri` (the URL the request is sent to, which always exists)

### DIFF
--- a/packages/common-grpc/src/service.js
+++ b/packages/common-grpc/src/service.js
@@ -260,7 +260,7 @@ GrpcService.prototype.request = function(protoOpts, reqOpts, callback) {
   }
 
   try {
-    reqOpts = util.decorateRequest(reqOpts, { projectId: this.projectId });
+    reqOpts = this.decorateRequest_(reqOpts);
   } catch(e) {
     callback(e);
     return;
@@ -362,7 +362,7 @@ GrpcService.prototype.requestStream = function(protoOpts, reqOpts) {
   }
 
   try {
-    reqOpts = util.decorateRequest(reqOpts, { projectId: this.projectId });
+    reqOpts = this.decorateRequest_(reqOpts);
   } catch(e) {
     setImmediate(function() {
       stream.destroy(e);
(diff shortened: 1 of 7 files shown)


callmehiphop:
LGTM - my only request would be to add a (probably system) test that passes in something with a circular reference.

stephenplusplus:
That's a fair request, but hopefully you can help me figure out the right way to handle this, because it might be more complex than it seems at first glance.

The error reported from #1908 is indeed from a recursion issue. And this PR does fix it, but not by handling the circular reference directly, but by narrowing the scope of the input to these methods that we loop over. In effect, that addresses the circular reference, because we aren't looping over unknown values on the `reqOpts` object, which seemed to be affected by the environment. This will now only loop over the parts of the request that could contain the project ID token, which we control.

But, the project ID token could be alongside user input, and that user input could have a circular reference, and when we iterate over it looking for the project ID token, the code would blow up the same way as in #1908. The complicated part is, each API-- or even, each method-- could have its own preferred way of handling circular references-- maybe it's okay, maybe it should be replaced with `[Circular]`, maybe it should be removed, maybe it *should* blow up the code (granted, with a more helpful error, which will [happen after #2068).

So, the question is-- do we handle circular references, and if so, how?

Issue #2102, opened 2017-03-17
Docs (README) should warn customers about the implications of requiring the all-in-one module


jasonpolites:
On Cloud Functions, require("google-cloud") will cause the smallest (default) function shape to run out of memory, and it's also very heavy-weight so affects cold-start time fairly dramatically.

The preferred approach for Cloud Functions is to only require the specific submodule you need.  Can we add relevant warnings to the README (and anywhere else for that matter) that "require("google-cloud")" will be slower and take more memory? (and that they should use the submodule variant)

Or perhaps only every recommend the submodule, then have a separate section about the all-in-one version with the appropriate caveats

jmuk:
Agreed. Related issue is https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1890

Assigning @omaray for aligning -- please tell us what we should do.

jmuk:
also fyi: @lukesneeringer

omaray:
That's a good point. Now that we start to have a lot of modules, it's true that having the "kitchen sink" will be more _expensive_ (or take more memory too it seems).

Let's chat about this in our weekly. We can definitely add the warning in the main ReadMe - and suggest considering individual modules. The nice thing though is that the ReadMe for each individual module is clear about installing only the module for that API. For example [BigQuery](https://github.com/GoogleCloudPlatform/google-cloud-node/tree/master/packages/bigquery) and [Datastore](https://github.com/GoogleCloudPlatform/google-cloud-node/tree/master/packages/datastore).

@bjwatson (fyi since related to other issue).

bjwatson:
I can create issues for Python and Ruby. @jskeet @garrettjonesgoogle This might also be relevant to your languages.

jskeet:
Not relevant to .NET - we don't have an "all-in-one" package.

lukesneeringer:
No need to create an issue for Python.  While I hardly think a kitchen sink practice is a best practice there, Python imports lazily and so the cost is not super high. Also, I already have a documentation restructure planned soon.

You should definitely open up an issue on Ruby; aggressive `require` resolution is standard practice in Sinatra and other places, and use of the meta-package should be discouraged there.

stephenplusplus:
Fixed in #2119

bjwatson:
I created the Ruby issue: https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1347

quartzmo:
[...]

PR #2119, opened 2017-03-21
Discourage the meta-package.

lukesneeringer:
RE #2102

This updates the docs to discourage metapackage use.


lukesneeringer:
README only, so merging.

PR #2124, opened 2017-03-22
Throw TypeError if we get a Number to spanner.date.

### LINKED ISSUE #2097: Inserting Date values into Spanner date column fails
What I want to do:

```js
myTable.insert({id: 1, start_date: new Date(1976, 4, 1)}).then(function (data) {
  console.log(data);
});
```

What I have to do instead :(

```js
myTable.insert({id: 1, start_date: '1976-04-01'},).then(function (data) {
  console.log(data);
});
```

Can we make dates do "the right thing" ? Or is that not possible?

lukesneeringer:
Fixes #2097.

### DIFF
--- a/packages/spanner/test/codec.js
+++ b/packages/spanner/test/codec.js
@@ -43,6 +43,17 @@ describe('codec', function() {
   });
 
   describe('SpannerDate', function() {
+    it('should choke on multiple arguments', function() {
+      var expectedErrorMessage = [
+        'The spanner.date function accepts a Date object or a',
+        'single argument parseable by Date\'s constructor.'
+      ].join(' ');
+
+      assert.throws(function() {
+        new codec.SpannerDate(2012, 3, 21);
+      }, new RegExp(expectedErrorMessage));
+    });
+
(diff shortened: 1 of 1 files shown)


packages/spanner/src/codec.js
```diff
@@ -26,6 +26,10 @@ var commonGrpc = require('@google-cloud/common-grpc');
 var is = require('is');
 
 function SpannerDate(value) {
+  if (is.number(value)) {
```
jgeewax: I don't think this is what we want... I think this should choke on more than one argument. For example:

```js
> new Date(1490195461000)
2017-03-22T15:11:01.000Z
```

I'm worried about people passing the arguments to a "date":

```js
> new Date(2000, 99, 41)
2008-05-11T04:00:00.000Z 
```

So this is really about anything more than one argument.

lukesneeringer: Sold. I thought timestamps did not work properly, but they do.

stephenplusplus:
Thank you, LGTM, I just need to make sure my code style commits didn't break the tests before merging.

PR #2127, opened 2017-03-22
spanner: refactor pooling

### LINKED ISSUE #2113: Spanner queries run from inside transaction callback via non-txn object silently fail
If I'm inside a transaction callback (`runTransaction()`) and I try to execute some query on the database *outside* the transaction, things silently fail.

They should probably either say "you can't do that!" or ... work. Here's a snippet that shows what's happening...

```js
var database = spanner.instance('test-instance').database('test-database');

database.runTransaction().then(function (data) {
  var txn = data[0];

  var query = 'SELECT * FROM Singers';

  // This works just fine.
  txn.run(query).then(function (results) {
    console.log('txn.run() row count:', results[0].length);
  }).catch(console.log);

  // This should either work or throw an error...
  database.run(query).then(function (results) {
    console.log('database.run() row count:', results[0].length);
  }).catch(console.log);

  // This works just fine.
  var testdb = spanner.instance('test-instance').database('test-database');
  testdb.run(query).then(function (results) {
    console.log('testdb.run() row count:', results[0].length);
  }).catch(console.log);

});
```

I'm not sure what's going on, but the `database.run()` callback never runs:

```
$ node test.js 
txn.run() row count: 3
testdb.run() row count: 3
=
```

Which is super duper weird. Can we either make it throw an error saying "you can't do that outside of a transaction" or make it work as expected? I'd prefer it just work rather than throw an error, but if there's a reason it can't, then I guess we'll survive...

/cc @callmehiphop

callmehiphop:
Closes #2113

#### TODO

- [x] Docs
- [x] Tests

### DIFF
--- a/packages/spanner/src/database.js
+++ b/packages/spanner/src/database.js
@@ -73,7 +73,7 @@ var TransactionRequest = require('./transaction-request.js');
  * @param {object=} options - Session pool configuration options.
  * @param {number} options.acquireTimeout - Time in milliseconds before giving
  *     up trying to acquire a session. If the specified value is `0`, a timeout
- *     will not occur.
+ *     will not occur. (Default: `0`)
  * @param {boolean} options.fail - If set to true, an error will be thrown when
  *     there are no available sessions for a request. (Default: `false`)
  * @param {number} options.max - Maximum number of resources to create at any
--- a/packages/spanner/src/session-pool.js
+++ b/packages/spanner/src/session-pool.js
@@ -37,7 +37,7 @@ var util = require('util');
  * @param {object=} options - Configuration options.
  * @param {number} options.acquireTimeout - Time in milliseconds before giving
  *     up trying to acquire a session. If the specified value is `0`, a timeout
- *     will not occur.
+ *     will not occur. (Default: `0`)
  * @param {boolean} options.fail - If set to true, an error will be thrown when
  *     there are no available sessions for a request. (Default: `false`)
  * @param {number} options.max - Maximum number of resources to create at any


packages/spanner/src/session-pool.js
```diff
@@ -45,6 +45,8 @@ var util = require('util');
  *     at any given time. (Default: `0`)
  * @param {number} options.keepAlive - How often to ping idle sessions, in
  *     minutes. Must be less than 1 hour.
+ * @param {number} options.acquireTimeout - Time before giving up trying to
```
stephenplusplus: This should either be `acquireTimeoutMs` or just mention we expect this in milliseconds.

callmehiphop:
@stephenplusplus I've updated the unit tests + docs if you could be so kind as to PTAL!

packages/spanner/test/session-pool.js
```diff
@@ -30,7 +30,7 @@ var fakeGenericPool = {
 };
 var originalFakeGenericPool = extend({}, fakeGenericPool);
 
-describe('SessionPool', function() {
+describe.only('SessionPool', function() {
```
lukesneeringer: Remove `.only` before merge. You knew that. :-p

packages/spanner/src/session-pool.js
```diff
@@ -35,18 +35,18 @@ var util = require('util');
  *
  * @param {module:spanner/database} database - The DB instance.
  * @param {object=} options - Configuration options.
+ * @param {number} options.acquireTimeout - Time in milliseconds before giving
+ *     up trying to acquire a session.
  * @param {boolean} options.fail - If set to true, an error will be thrown when
  *     there are no available sessions for a request. (Default: `false`)
  * @param {number} options.max - Maximum number of resources to create at any
- *     given time. (Default: `1`)
+ *     given time. (Default: Infinity)
```
lukesneeringer: I see this documentation change but not the corresponding code.

packages/spanner/src/session-pool.js
```diff
@@ -59,6 +61,9 @@ function SessionPool(database, options) {
   this.maxIdle = options.maxIdle || 1;
   this.fail = !!options.fail;
 
+  this.pendingAcquires = [];
```
stephenplusplus: Quick test needed to make sure we initialize this.

packages/spanner/src/session-pool.js
```diff
@@ -35,10 +35,12 @@ var util = require('util');
  *
  * @param {module:spanner/database} database - The DB instance.
  * @param {object=} options - Configuration options.
+ * @param {number} options.acquireTimeout - Time in milliseconds before giving
+ *     up trying to acquire a session.
```
stephenplusplus: Add a mention for the meaning of `0` and that it is the default value.

packages/spanner/src/session-pool.js
```diff
@@ -35,10 +35,12 @@ var util = require('util');
...
  * @param {boolean} options.fail - If set to true, an error will be thrown when
  *     there are no available sessions for a request. (Default: `false`)
  * @param {number} options.max - Maximum number of resources to create at any
- *     given time. (Default: `1`)
+ *     given time. (Default: Infinity)
```
stephenplusplus: This should say `Number.MAX_SAFE_INTEGER`, I believe.

packages/spanner/src/session-pool.js
```diff
@@ -141,7 +146,7 @@ SessionPool.getPoolOptions_ = function(userOptions) {
   var poolOptions = {
     idleTimeoutMillis: 59 * 60000,
     testOnBorrow: true,
-    max: userOptions.max || 1,
+    max: userOptions.max || Number.MAX_SAFE_INTEGER,
     min: userOptions.min || 0,
     numTestsPerRun: Infinity
```
stephenplusplus: The tests are expecting this to be `Number.MAX_SAFE_INTEGER`

packages/spanner/test/session-pool.js
```diff
@@ -1376,23 +1345,143 @@ describe('SessionPool', function() {
...
-          assert.strictEqual(
-            onAvailable(),
-            sessionPool.getNextAvailableSession_(OPTIONS, assert.ifError)
-          );
+    beforeEach(function() {
+      setInterval_ = global.setInterval;
+      global.setInterval = function(cb) {
+        global.setInterval.calledWith_ = arguments;
+        setImmediate(cb);
+        return 1;
```
stephenplusplus: Can you make this something like `SET_INTERVAL_ID` for clarity when used in the later tests?

packages/spanner/src/session-pool.js
```diff
@@ -489,9 +492,58 @@ SessionPool.prototype.getNextAvailableSession_ = function(options, callback) {
...
   });
+
+  if (this.acquireInterval) {
+    return;
+  }
+
+  var self = this;
+  var intervalSpeed = 30000;
+
+  this.acquireInterval = setInterval(checkForSession, intervalSpeed);
```
stephenplusplus: Can you rename to `this.acquireIntervalId`?

packages/spanner/test/session-pool.js
```diff
@@ -1376,23 +1345,143 @@ describe('SessionPool', function() {
...
+      var intervalArgs = global.setInterval.calledWith_;
+
+      assert.strictEqual(sessionPool.acquireInterval, 1);
+      assert.strictEqual(typeof intervalArgs[0], 'function');
+      assert.strictEqual(intervalArgs[1], 30000);
+    });
+
+    it('should call getNextAvailableSession_ if a read free', function(done) {
+      sessionPool.getNextAvailableSession_ = function(callback) {
+        callback();
```
stephenplusplus: Can you leave a comment here and anywhere else `callback === done` that this is the value of `done`, typically `// done()`

packages/spanner/test/session-pool.js
```diff
@@ -1376,23 +1345,143 @@ describe('SessionPool', function() {
...
+        sessionPool.pendingAcquires = [];
+      };
+
+      global.clearInterval = function(handle) {
+        assert.strictEqual(handle, sessionPool.acquireInterval);
+
+        setImmediate(function() {
+          assert.strictEqual(sessionPool.acquireInterval, null);
+
+          global.clearInterval = clearInterval_;
```
stephenplusplus: Should pop this up to the first line after `global.clearInterval` is re-assigned. (or just capture the real value in the `before` and re-assign in the beforeEach and afterEach)

packages/spanner/src/session-pool.js
```diff
@@ -489,9 +492,58 @@ SessionPool.prototype.getNextAvailableSession_ = function(options, callback) {
...
+    if (!self.pendingAcquires.length) {
+      clearInterval(self.acquireInterval);
+      self.acquireInterval = null;
+    } else if (self.acquireTimeout) {
+      var err = new Error('Unable to acquire Session, timeout occurred.');
+      var acquire;
+
+      for (var i = self.pendingAcquires.length - 1; i > -1; i--) {
+        acquire = self.pendingAcquires[i];
+        acquire.timeout -= intervalSpeed;
```
stephenplusplus: I don't believe this is tested.

packages/spanner/src/session-pool.js
```diff
@@ -489,9 +492,58 @@ SessionPool.prototype.getNextAvailableSession_ = function(options, callback) {
...
+
+        if (acquire.timeout < 0) {
+          acquire.callback(err);
+          self.pendingAcquires.splice(i, 1);
```
stephenplusplus: I don't believe this is tested.

packages/spanner/src/session-pool.js
callmehiphop: I think the change you're looking for is here - https://github.com/callmehiphop/gcloud-node/blob/4e63fbf477716b45794e8669ba1c0ca6c21ee567/packages/spanner/src/session-pool.js#L149

packages/spanner/test/session-pool.js
callmehiphop: It's a sad day when one of these sneaks by, haha.

packages/spanner/src/session-pool.js
callmehiphop: Done!

packages/spanner/src/session-pool.js
callmehiphop: Right, fixed!

packages/spanner/src/session-pool.js
callmehiphop: Done!

packages/spanner/src/session-pool.js
callmehiphop: Fixed!

packages/spanner/src/session-pool.js
callmehiphop: Sure can, done!

packages/spanner/src/session-pool.js
callmehiphop: Fixed

packages/spanner/test/session-pool.js
callmehiphop: Sure can, done!

packages/spanner/test/session-pool.js
callmehiphop: Sure, done.

packages/spanner/test/session-pool.js
callmehiphop: Sure, done!

packages/spanner/src/session-pool.js
callmehiphop: Fixed!

stephenplusplus:
I'm not sure what's up with coverage decreasing. From the Coveralls UI, it looks like, while we are missing coverage in some files, they aren't the ones you edited. Is there an explanation for the coverage dropping that I'm not thinking of?

packages/spanner/src/database.js
```diff
@@ -71,10 +71,13 @@ var TransactionRequest = require('./transaction-request.js');
  *
  * @param {string} name - Name of the database.
  * @param {object=} options - Session pool configuration options.
+ * @param {number} options.acquireTimeout - Time in milliseconds before giving
+ *     up trying to acquire a session. If the specified value is `0`, a timeout
+ *     will not occur.
```
callmehiphop: Done!

callmehiphop:
Not sure, might have something to do with bumping our istanbul version - since we don't really have nightly builds setup, I'd guess this could be the first time coverage has been tested for spanner since upgrading?

stephenplusplus:
Good enough for me, I'll merge after CI's happy.

lukesneeringer:
> I'm not sure what's up with coverage decreasing. From the Coveralls UI, it looks like, while we are missing coverage in some files, they aren't the ones you edited. Is there an explanation for the coverage dropping that I'm not thinking of?

Total and covered lines probably decreased linearly.

Issue #2125, opened 2017-03-22
logging miscategorizes logs from GKE


dennismartensson:
We did try the new logging-bunyan with our bunyan logging node.js application today. 

When we put it under some load the cpu when to 100% and left the application crashing. It is logging fine with just bunyan. 

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS: Docker kubernetes from node:6.9
  - Node.js version: node 6.9
  - npm version:
  - google-cloud-node version: 0.1.0

#### Steps to reproduce

  1. require `google-cloud`
  2. Was setting it up like this:

```js
const Logger = require('bunyan')
const loggingBunyan = require('@google-cloud/logging-bunyan')({
  keyFilename: 'keyfile.json',
  projectId: 'project-id',
})


const log = new Logger({
  name: 'name',
  src: true,
  streams: [
    loggingBunyan.stream('info'),
  ] })
```

Thanks!

dennismartensson:
Hi @ofrobots and @stephenplusplus let me know how I can help.

Do you think that 0.2.0 fixes some of this?

Thank you

Dennis

ofrobots:
Hi @dennismartensson. 0.2.0 fixes issues with objects with circular references. This would have manifested as high CPU usage and a stack overflow. Can you try 0.2.0 and report back?

dennismartensson:
@ofrobots I will do that.

dennismartensson:
So @ofrobots now we are trying it a bit under some load. The cpu is now better but we get no logs at all on https://console.cloud.google.com/logs

```
const Logger = require('bunyan')

const loggingBunyan = require('@google-cloud/logging-bunyan')({
  keyFilename: 'keyfile.json',
  projectId: 'project-id',
})

const log = new Logger({
  name: 'name',
  src: true,
  streams: [
    loggingBunyan.stream('info'),
  ] })
```

But there is no loggs.

dennismartensson:
Did also try to change it to be like the example but its not showing any loggs in the logs viewer...

What can I do to help resolve this?

```
const bunyan = require('bunyan')

const loggingBunyan = require('@google-cloud/logging-bunyan')({
  keyFilename: './keyfile.json',
  projectId: 'project-id',
})

const log = bunyan.createLogger({
  name: 'name',
  streams: [
    loggingBunyan.stream('info'),
  ],
})
```

ofrobots:
[...]

PR #2130, opened 2017-03-23
logging: avoid unnecessary promises

ofrobots:
Calls to Log#write from Log.prototype[severity], when the options
parameter is omitted, were always allocating a promise despite a
callback being available. This will result in unexpected (unhandled)
promise rejections.

TODO: add test for `makeSeverityFunction_`, but opening now to get opinions.

### DIFF
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -706,13 +706,17 @@ function promisify(originalMethod, options) {
   var slice = Array.prototype.slice;
 
   var wrapper = function() {
-    var args = slice.call(arguments);
-    var hasCallback = is.fn(args[args.length - 1]);
-    var context = this;
-
-    if (hasCallback) {
-      return originalMethod.apply(context, args);
+    var last;
+    for (last = arguments.length - 1; last >= 0; last--) {
+      var arg = arguments[last];
+      if (arg === undefined) continue;  // skip trailing undefined.
+      if (!is.fn(arg)) break;           // non-callback last argument found.
+      return originalMethod.apply(this, arguments);
     }
+
+    // peel trailing undefined.
+    var args = slice.call(arguments, 0, last + 1);
+    var context = this;
 
     var PromiseCtor = Promise;
 
--- a/packages/common/test/util.js
+++ b/packages/common/test/util.js
@@ -1648,6 +1648,32 @@ describe('common/util', function() {
 
       assert.strictEqual(FakeClass.prototype.methodName, method);
     });
+
+    describe('trailing undefined arguments', function() {
+      it('should not return a promise in callback mode', function(done) {
+        var func = util.promisify(function(optional, callback) {
+          assert(is.fn(optional));
+          optional(null);
+        });
+
+        var returnVal = func(function() {
+          assert(!returnVal);
+          done();
+        });
+      });
+
+      it('should return a promise when callback omitted', function(done) {
+        var func = util.promisify(function(optional, callback) {
+          assert.strictEqual(arguments.length, 1);
+          assert(is.fn(optional));
+          optional(null);
+        });
+
+        var returnVal = func(undefined, undefined).then(function() {
+          done();
+        });
+      });
+    });
   });
 
   describe('promisify', function() {


ofrobots:
The other option would be to peel off training undefined arguments in `util.promisify`. That would be slow, and potentially error prone – if there are valid uses where a trailing undefined is passed? @stephenplusplus

stephenplusplus:
I'll let the Promise expert handle this one!

lukesneeringer:
> The other option would be to peel off trailing undefined arguments in util.promisify. That would be slow, and potentially error prone – if there are valid uses where a trailing undefined is passed intentionally?

Checking for the `callback` being `undefined` is sufficient in my mind.

lukesneeringer:
> Calls to Log#write from Log.prototype[severity], when the options
> parameter is omitted, were always allocating a promise despite a
> callback being available. This will result in unexpected (unhandled)
> promise rejections.

Can you point to the code doing the wrong thing? I do not see it in the PR.

ofrobots:
@lukesneeringer The bug is because of the interplay between functions with optional parameters and `util.promisify`.

Original implementation of `Log.prototype.alert`:

```js
Log.prototype.alert = function(entry, options, callback) {
   this.write(Log.assignSeverityToEntries_(entry, 'ALERT'), options, callback);		
}
```

`this.write` is a promisified function, but note that we are calling it with three arguments. Since `options` is optional, we will call `write` with `(entry, callback, undefined)` when caller of `alert` omits options.

This is the relevant code in `util.promisify`:

```js
  var wrapper = function() {
    var args = slice.call(arguments);
    var hasCallback = is.fn(args[args.length - 1]);
```

`args[args.length - 1]` would be undefined. `hasCallback` is set to `false` even though a callback is present.

stephenplusplus:
> The other option would be to peel off trailing undefined arguments in util.promisify. That would be slow, and potentially error prone – if there are valid uses where a trailing undefined is passed intentionally? 

I like this better, because this is indeed a shortcoming of `promisify` as it's making a miscalculation which should be fixed.

ofrobots:
> I like this better, because this is indeed a shortcoming of promisify as it's making a miscalculation which should be fixed.

I think I agree. I am concerned that we are silently creating promises elsewhere too – not just logging. Some of the folks in node-code want to make unhandled promise rejections a process terminator in the future – which is going to be catastrophic. (Rest assured, I'm against such a change in node core, but we still need to fix this).

lukesneeringer:
> Original implementation of Log.prototype.alert:

> ```javascript
> Log.prototype.alert = function(entry, options, callback) {
>    this.write(Log.assignSeverityToEntries_(entry, 'ALERT'), options, callback);		
> }
> ```
> 
> this.write is a promisified function, but note that we are calling it with three arguments. Since options is optional, we will call write with (entry, callback, undefined) when caller of alert omits options.

Actually we would call `write` with `(entry, undefined, callback)` in the implementation you show above, which is not the same.

> Some of the folks in node-code want to make unhandled promise rejections a process terminator in the future – which is going to be catastrophic. (Rest assured, I'm against such a change in node core, but we still need to fix this)

Oh, I am _totally_ on the side of "some of the folks in node-core" on this. The fact that promise rejections essentially amount to silently swallowed exceptions is one of the single biggest most broken things in Node.

ofrobots:
> Actually we would call write with (entry, undefined, callback) in the implementation you show above, which is not the same.

Yes, because this PR is supposed to be a potential fix for the problem :laughing: – but only for this one promisified function – `Log.prototype.write`. I agree with @stephenplusplus that the shortcoming is in `promisify` and should be fixed there instead.

What solution would you like to see @lukesneeringer?

> Oh, I am totally on the side of "some of the folks in node-core" on this. The fact that promise rejections essentially amount to silently swallowed exceptions is one of the single biggest most broken things in Node.

In the interest of keeping this thread on-topic, let's take this up elsewhere/offline. Apologies for the original digression.

lukesneeringer:
I am not arguing against the solution; it is more that I do not understand it. I think that is based on a lack of context, though, so do not treat me as a blocker. :-)

ofrobots:
The code with the bug is here:
https://github.com/GoogleCloudPlatform/google-cloud-node/blob/0ff331200c36a590e8f51a3ac0a26f82e33b0ac9/packages/common/src/util.js#L706

Specifically
```js
  var wrapper = function() {
    var args = slice.call(arguments);
    var hasCallback = is.fn(args[args.length - 1]);
```

`hasCallback` results in the wrong value when reached from this line:

```js
this.write(Log.assignSeverityToEntries_(entry, 'ALERT'), options, callback);	
```

lukesneeringer:
Aha, okay. I am now tracking. Thanks @ofrobots

stephenplusplus:
@callmehiphop take a read over the thread-- we're thinking of moving the fix into the `promisify` method itself-- wdyt?

ofrobots:
@callmehiphop that's a good optimization, and should be made independently. I may be confusing things, but I don't think it will actually fix the problem at hand. The problem is due to the promisification of `write` rather than the severity functions (`alert`, etc.). These functions will still need to callback sniffing.

This bug gave me a lot of grief yesterday in the investigation of #2125 with the unnecessary promise swallowing my actual promises 😢 .

stephenplusplus:
I think we should just morph this PR into the fix in `promisify` for now-- @ofrobots are you interested in tackling that one, or want me to take it?

ofrobots:
I can tackle this one.

ofrobots:
PTAL.

packages/common/src/util.js
```diff
@@ -706,14 +706,18 @@ function promisify(originalMethod, options) {
   var slice = Array.prototype.slice;
 
   var wrapper = function() {
-    var args = slice.call(arguments);
-    var hasCallback = is.fn(args[args.length - 1]);
-    var context = this;
-
-    if (hasCallback) {
-      return originalMethod.apply(context, args);
+    var last;
```
stephenplusplus: Can we rename to `lastElementIndex`?

packages/common/src/util.js
```diff
@@ -706,14 +706,18 @@ function promisify(originalMethod, options) {
...
+    for (last = arguments.length - 1; last >= 0; last--) {
+      var arg = arguments[last];
+      if (arg === undefined) continue;  // skip trailing undefined.
```
stephenplusplus: Can you use `is.undefined` or `typeof arg === 'undefined'`?

packages/common/test/util.js
```diff
@@ -1648,6 +1648,32 @@ describe('common/util', function() {
 
       assert.strictEqual(FakeClass.prototype.methodName, method);
     });
+
+    describe('trailing undefined arguments', function() {
```
stephenplusplus: I think these tests need to go down in the `promisfy` block, this is the `promisifyAll` describe block.

stephenplusplus:
Before merging, I will wait for @callmehiphop's review on the promise stuff.

callmehiphop:
LGTM

stephenplusplus:
@ofrobots if you still wanted to make those changes I had from my review, please feel free :D

lukesneeringer:
Sorry, I thought the LGTM was the merge catalyst. :-)

ofrobots:
Yep. I will follow up with the suggestions. At a team off site so I have
been slow in reacting.

stephenplusplus:
No problem, and say hi to everyone for me! 👋 I took care of the leftover issues, so no worries.

PR #2132, opened 2017-03-23
Fixed insert in bigtable that would set incorrect timestamps.

### LINKED ISSUE #2120: Bigtable Date system test broken
After merging #2116, I noticed we had a system error:

https://github.com/GoogleCloudPlatform/google-cloud-node/blob/e353ce0c9010dde1a85edd2610dbca68a5f0619a/packages/bigtable/system-test/bigtable.js#L895

```
Uncaught AssertionError: false == true
      + expected - actual

      -false
      +true
```

@callmehiphop could you ptal?

rFlex:
Timestamps were also saved as milliseconds on save, this converts it to microseconds.

Fixes #2120.

I apologize for the system tests breaking...

### DIFF
--- a/packages/bigtable/test/mutation.js
+++ b/packages/bigtable/test/mutation.js
@@ -185,7 +185,7 @@ describe('Bigtable/Mutation', function() {
         setCell: {
           familyName: 'follows',
           columnQualifier: 'gwashington',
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Thank you!

PR #2134, opened 2017-03-24
Improve our coverage configuration.

lukesneeringer:
I want to get started fixing our coverage. It annoys me that it is at 99% and coveralls gives us a big fat red :x:.

This is the first step, which is making the reporting better. This PR:

  * Upgrades Istanbul to the current version (0.3.5 is quite old; 0.4 has been out for months, and has much prettier HTML reports).
  * Adds an `npm run cover-html` command to get an HTML report.
  * Adds an `.istanbul.yml` config file, which notably changes the reporting threshholds. Now we get red below 95% and yellow below 100%. The defaults on Istanbul (50% and 80%) are ridiculous and show up green even when you are missing a lot of coverage.
  * Moves coverage reports into the `.coverage` directory (note the leading dot), because I am OCD and really hate seeing it on a default `ls`. The `.gitignore` and `npm run coveralls` are updated accordingly.

### DIFF
--- a/scripts/circle/test.sh
+++ b/scripts/circle/test.sh
@@ -42,7 +42,7 @@ npm run lint
 node ./scripts/build.js
 
 export COVERALLS_REPO_TOKEN="kEhKxvu3L9Z4GiClA9LcKYk7G9i9dl83k"
-export CIRCLE_ARTIFACTS="$(pwd)/coverage"
+export CIRCLE_ARTIFACTS="$(pwd)/.coverage"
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Woo! Just need an edit here too: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/scripts/circle/test.sh#L45

lukesneeringer:
Edited. Everything else look good to you?

PR #2136, opened 2017-03-24
[WIP] Vision curation methods.

lukesneeringer:
This is a refactor based on my attempt to get vision's test coverage to 100%.

The ultimate goal is to move the functions referenced in this code block:

```javascript
      var detections = foundImages
        .map(groupDetectionsByImage)
        .map(removeExtraneousAnnotationObjects)
        .map(assignTypeToEmptyAnnotations)
        .map(removeDetectionsWithErrors)
        .map(flattenAnnotations)
        .map(decorateAnnotations);
```

...into their own module, with their own tests. This will make these functions (which are essentially atomic) easier to test.

The first commit moves `removeExtraneousAnnotationObjects` into a new `curation` module. Follow-up commits will remove the rest (unless you tell me on review that I should abandon this refactor for whatever reason).

### DIFF
--- a/packages/vision/src/curation.js
+++ b/packages/vision/src/curation.js
@@ -25,6 +25,27 @@
 
 var arrify = require('arrify');
 
+
+/**
+ * Swap the keys and values of a dictionary like object.
+ * If there are duplicate values, use the `resolve` callback to choose one.
+ */
+var swap = function(dict, resolve) {
+  var answer = {};
+  for (var key in dict) {
+    var value = dict[key];
+    if (typeof answer[value] === 'undefined') {
+      answer[value] = key;
+    }
+    else {
+      answer[value] = resolve(value, key, answer[value]);
+    }
+  }
+  return answer;
+}
+
+// A mapping between the short names used within the client library,
+// and the long names used in the API.
 var TYPE_SHORT_NAME_TO_RESPONSE_NAME = {
   crop: 'cropHintsAnnotation',
   crops: 'cropHintsAnnotation',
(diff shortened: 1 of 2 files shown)


packages/vision/src/curation.js
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * A module containing stand-alone functions for curating responses that
+ * come back from the Vision API.
+ */
+
+'use strict';
+
+var arrify = require('arrify');
+
+var TYPE_SHORT_NAME_TO_RESPONSE_NAME = {
```
lukesneeringer: My goal is to have this be needed _only_ in this file, and not in `index.js`.

packages/vision/src/index.js
```diff
@@ -410,7 +412,9 @@ Vision.prototype.detect = function(images, options, callback) {
 
       var detections = foundImages
         .map(groupDetectionsByImage)
-        .map(removeExtraneousAnnotationObjects)
+        .map(function(image) {
+          return curation.removeExtraneousAnnotations(image, options.types);
```
lukesneeringer: I am confused about this code. Are these _images_ or _annotations_? The variable is named `foundImages` but the old method expected _annotations_. Is `image` the right thing for me to be sending here, or should I be sending an annotations property?

The unit tests still pass with this code change, but I am not sure the code change is _actually_ correct.

stephenplusplus: It starts the chain with `foundImages` because we're ultimately giving back a result for each image.

So, `foundImages` could be 2 images. Its values are completely ignored, because in the first step of the `map` transformation ([`groupDetectionsByImage`](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/688ac58e72f950c64aac7999b57c247a0db793f5/packages/vision/src/index.js#L433)), we return from that the apiResponse, *split* into as many annotation types that were requested based on how many images.

So it's basically only used as a way to split the annotations the right amount of times. More obvious solutions are very welcome.

lukesneeringer: Okay, so essentially, we start with `images` but the `groupDetectionsByImage` function just the annotations. Got it.

lukesneeringer: As far as how to make it more obvious, most likely I will just use an intervening variable. Even just doing:

```javascript
var detections = foundImages.map(groupDetectionsByImage);
detections = dections.map(...).map(...) // and so on
```

...would actually help clarity a lot.

stephenplusplus:
All LGTM so far 👍

stephenplusplus:
Quick tip, you can save some CI cycles if you put `[ci skip]` in your commit message 👍

lukesneeringer:
Yeah, good call. I need to circle back to this later this week.

The mapping functions actually have quite a few side effects, which is making them harder than expected to port.

lukesneeringer:
Closing this for now. I have gotten a much better understanding of this code and plan to return to it, but it might not be with this particular PR.

PR #2141, opened 2017-03-27
spanner: deleteRows() + createReadStream() accept multiple keys

### LINKED ISSUE #2023: Spanner - Deleting rows with a composite key
This line seems to be the cause of the problem - https://github.com/GoogleCloudPlatform/google-cloud-node/blob/bc0503e46b7fcea0ce3274b6d8dd31c7eef4dd1d/packages/spanner/src/transaction-request.js#L284

IMHO, given that monotonically increasing keys are discouraged, the lack of table operations upon composite keys (`String[][]`, not just `String[]`!) is a _major_ limitation. Even a quick-fix for this will be highly appreciated

### LINKED ISSUE #2098: Deleting multiple rows from a table fails
What I want to do:

```js
table.deleteRows(['1', '2', '3']);
```

(My ID field is an `INT64` type, but I get that I have to use strings here because 64-bit integers don't nicely fit in Node's world....)

What happens:

```
table.deleteRows(['1', '2', '3'])
Promise { <pending> }
> (node:111838) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 10): Error: Wrong number of key parts for mytable. Expected: 1. Got: ["1","2","3"]
```

To check that I'm not crazy, the following works as expected.

```js
table.deleteRows(['1']);
```

I'm guessing that this has to do with the hierarchical key stuff, and that what's actually happening here is the list is a *single* key, which has many parts -- not a list of keys.

To double check, I tried doing each as their own list, which similarly failed:

```js
table.deleteRows([ ['1'], ['2'], ['3'] ]);
```

stephenplusplus:
Fixes #2023
Fixes #2098 

#### To Dos

- [x] Docs
- [x] Tests
  - [x] System
  - [x] Unit

This allows multiple keys to be sent to `deleteRows()`, `createReadStream()`, and `read()`.

### DIFF
--- a/packages/spanner/src/table.js
+++ b/packages/spanner/src/table.js
@@ -117,7 +117,7 @@ Table.prototype.create = function(schema, callback) {
  * @param {*[]|*[][]} query.keys - The primary keys of the rows in this table to
  *     be yielded. If using a composite key, provide an array within this array.
  *     See the example below.
- * @param {string[]=} query.index - The name of an index on the table.
+ * @param {string=} query.index - The name of an index on the table.
  * @param {number} query.limit - The number of rows to return.
  * @param {object=} options - [Transaction options](https://cloud.google.com/spanner/docs/timestamp-bounds).
  * @param {number} options.exactStaleness - Executes all reads at the timestamp
@@ -251,7 +251,7 @@ Table.prototype.delete = function(callback) {
  * table.deleteRows(keys, function(err, apiResponse) {});
  *
  * //-
- * // Provide an array for `keys` to read with a composite key.
+ * // Provide an array for `keys` to delete rows with a composite key.
  * //-
  * var keys = [
  *   [
@@ -346,7 +346,7 @@ Table.prototype.insert = function(keyVals, callback) {
  * @param {*[]|*[][]} query.keys - The primary keys of the rows in this table to
  *     be yielded. If using a composite key, provide an array within this array.
  *     See the example below.
- * @param {string[]=} query.index - The name of an index on the table.
+ * @param {string=} query.index - The name of an index on the table.
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this
  *     request.
--- a/packages/spanner/src/transaction-request.js
+++ b/packages/spanner/src/transaction-request.js
@@ -114,7 +114,7 @@ TransactionRequest.formatTimestampOptions_ = function(options) {
  * @param {*[]|*[][]} query.keys - The primary keys of the rows in this table to
  *     be yielded. If using a composite key, provide an array within this array.
  *     See the example below.
- * @param {string[]=} query.index - The name of an index on the table.
+ * @param {string=} query.index - The name of an index on the table.
  * @param {number=} query.limit - The number of rows to yield.
  * @return {Stream}
  *
@@ -272,7 +272,7 @@ TransactionRequest.prototype.createReadStream = function(table, query) {
  * });
  *
  * //-
- * // Provide an array for `keys` to read with a composite key.
+ * // Provide an array for `keys` to delete rows with a composite key.
  * //-
  * var keys = [
  *   [
@@ -432,7 +432,7 @@ TransactionRequest.prototype.insert = function(table, keyVals, callback) {
  * @param {*[]|*[][]} query.keys - The primary keys of the rows in this table to
  *     be yielded. If using a composite key, provide an array within this array.
  *     See the example below.
- * @param {string[]=} query.index - The name of an index on the table.
+ * @param {string=} query.index - The name of an index on the table.
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this
  *     request.


stephenplusplus:
@richtera @angadn would you mind taking a look to see if I did this correctly?

packages/spanner/src/transaction-request.js
```diff
@@ -203,11 +203,11 @@ TransactionRequest.prototype.createReadStream = function(table, query) {
 
   if (query.keys) {
     reqOpts.keySet = {
-      keys: [
-        {
-          values: arrify(query.keys).map(codec.encode)
-        }
-      ]
+      keys: arrify(query.keys).map(function(key) {
```
lukesneeringer: This might be a mistake in the other direction? If I read this right, now _every_ key is treated as composite.
But maybe that is okay in Spanner? (Honestly asking.)

packages/spanner/src/transaction-request.js
```diff
@@ -279,11 +279,11 @@ TransactionRequest.prototype.deleteRows = function(table, keys, callback) {
   mutation['delete'] = {
     table: table,
     keySet: {
-      keys: [
-        {
-          values: arrify(keys).map(commonGrpc.Service.encodeValue_)
-        }
-      ]
+      keys: arrify(keys).map(function(key) {
```
lukesneeringer: Same question as above.

packages/spanner/src/transaction-request.js
vkedia: Yes this should be fine.

vkedia: Above we document query.keys as:
```
  * @param {string[]} query.keys - The primary keys of the rows in this table to
  *     be yielded.
```

That seems incorrect to me.

packages/spanner/src/transaction-request.js
vkedia: The documentation and the example for keys seems to be incorrect.

packages/spanner/system-test/spanner.js
```diff
@@ -511,6 +511,46 @@ var spanner = new Spanner(env);
...
+          Name: name
+        },
+        {
+          SingerId: id2,
+          Name: name
+        }
+      ], function(err) {
+        assert.ifError(err);
+
+        table.deleteRows([id, id2], function(err) {
```
vkedia: Can you add a test case for deleting multiple rows where each row has a composite key and atleast one of the key should be non string.

packages/spanner/src/transaction-request.js
angadn: @lukesneeringer That is the idea! Since we don't use monotonically increasing integers for data-distribution reasons, a composite key is _the_ way to do it.

angadn:
@stephenplusplus looks good! Not able to pull and test this quickly - I ended up porting out to Golang to get around this issue (and friends) :( Will give it a real-world test and tell you how it goes

richtera:
Looks good but I had to shutdown Spanner due to cost, sorry.

callmehiphop:
@stephenplusplus does this also close #2098?

stephenplusplus:
I think it does. We weren't previously testing with multiple keys passed to `deleteRows` and I now have two tests which do that, so I'm inclined to believe that it works

Also, PTAL everyone-- docs and tests corrected/added.

packages/spanner/src/table.js
```diff
@@ -232,6 +251,20 @@ Table.prototype.delete = function(callback) {
  * table.deleteRows(keys, function(err, apiResponse) {});
  *
  * //-
+ * // Provide an array for `keys` to read with a composite key.
```
vkedia: Provide an array for `keys` to delete rows with a composite key.

packages/spanner/src/table.js
```diff
@@ -310,8 +343,9 @@ Table.prototype.insert = function(keyVals, callback) {
  *     the table.
  * @param {string[]} query.columns - The columns of the table to be returned for
  *     each row matching this query.
- * @param {string[]} query.keys - The primary keys of the rows in this table to
- *     be yielded.
+ * @param {*[]|*[][]} query.keys - The primary keys of the rows in this table to
+ *     be yielded. If using a composite key, provide an array within this array.
+ *     See the example below.
  * @param {string[]=} query.index - The name of an index on the table.
```
vkedia: shouldnt this just be string and not string[]

packages/spanner/src/transaction-request.js
```diff
@@ -254,6 +272,20 @@ TransactionRequest.prototype.createReadStream = function(table, query) {
  * });
  *
  * //-
+ * // Provide an array for `keys` to read with a composite key.
```
vkedia: Provide an array for `keys` to delete rows with a composite key

stephenplusplus:
@vkedia thanks for catching those, they are fixed now.

packages/spanner/src/transaction-request.js
lukesneeringer: Convinced. Thank you.

packages/spanner/system-test/spanner.js
lukesneeringer: +1. Feel free to link to it if it is already present.

stephenplusplus: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2141/files#diff-aaea4d158497e38a9c5306e1834c5442R554

PR #2144, opened 2017-03-28
spanner: refactor transaction retry logic

### LINKED ISSUE #2118: Spanner overlapping transactions don't fail when they should?
My understanding is that if you read a cell (say the `name` column from a specific row in an `employees` table) in two different transactions, then commit a change to that cell in one of them, then a subsequent commit to that same cell from the other transaction should fail. Is this correct?

(I'm pulling this info from our docs, and making sure that these are not 'blind writes' by doing a read first -- see https://cloud.google.com/spanner/docs/transactions#rw_transaction_performance)

If so, our client library doesn't appear to do that...?

```js
var instance = spanner.instance('test-instance');
var database = instance.database('test-database', {max: 5});
var table = database.table('employees');

var nameQuery = {keys: ['1'], columns: ['name']};

Promise.all([database.runTransaction(), database.runTransaction()]).then(
  function (txns) {
    var txn1 = txns[0][0], txn2 = txns[1][0];
    var txn = {1: txns[0][0], 2: txns[1][0]};

    function readNameFromTransaction(number) {
      return function () {
        return txn[number].read('employees', nameQuery).then(function (results) {
          console.log('txn' + number + ':', results[0][0]);
        });
      };
    }

    function changeNameFromTransaction(number) {
      return function () {
        txn[number].update('employees', {
          employee_id: '1',
          name: 'Updated by txn' + number
        });
        return txn[number].commit().then(function (results) {
          console.log('txn' + number + ':', results);
        });
      }
    }

    function readNameFromTable() {
      return table.read(nameQuery).then(function (results) {
        console.log('table:', results[0][0]);
      });
    }

    readNameFromTable()
      .then(readNameFromTransaction(1))
      .then(readNameFromTransaction(2))
      .then(changeNameFromTransaction(1))
      .then(readNameFromTable)
      .then(changeNameFromTransaction(2))
      .then(readNameFromTable)
      .catch(console.log);
  }
);
```

The output from this script looks totally successful:

```
table: [ { name: 'name', value: 'Updated by txn2' } ]
txn1: [ { name: 'name', value: 'Updated by txn2' } ]
txn2: [ { name: 'name', value: 'Updated by txn2' } ]
txn1: [ { commitTimestamp: { seconds: '1490106991', nanos: 173288000 } } ]
table: [ { name: 'name', value: 'Updated by txn1' } ]
txn2: [ { commitTimestamp: { seconds: '1490106991', nanos: 269254000 } } ]
table: [ { name: 'name', value: 'Updated by txn2' } ]
```

I expected that `txn2` commit result to be a failure because the data changed out from under us... Is my understanding wrong? or the client library wrong? or a bug in my code?

/cc @callmehiphop

callmehiphop:
closes #2118

### DIFF
--- a/packages/spanner/src/database.js
+++ b/packages/spanner/src/database.js
@@ -721,16 +721,15 @@ Database.prototype.runStream = function(query, options) {
  *   }
  *
  *   // Run a transactional query.
- *   return transaction.run('SELECT * FROM Singers')
- *     .then(function() {
- *       // Queue a mutation (note there is no callback passed to `insert`).
- *       transaction.insert('Singers', {
- *         SingerId: 'Id3b',
- *         Name: 'Joe West'
- *       });
- *       // Commit the transaction.
- *       return transaction.commit();
+ *   return transaction.run('SELECT * FROM Singers').then(function() {
+ *     // Queue a mutation (note there is no callback passed to `insert`).
+ *     transaction.insert('Singers', {
+ *       SingerId: 'Id3b',
+ *       Name: 'Joe West'
  *     });
+ *     // Commit the transaction.
+ *     return transaction.commit();
+ *   });
  * })
  * .then(function() {
  *   // Transaction committed successfully.


packages/spanner/src/transaction.js
```diff
@@ -223,25 +228,26 @@ Transaction.prototype.begin = function(callback) {
...
  * //-
- * // If the callback is omitted, we'll return a Promise.
+ * // To use promises, simply return a Promise.
  * //-
- * database.runTransaction()
- *   .then(function(data) {
- *     var transaction = data[0];
+ * database.runTransaction(function(err, transaction) {
+ *   if (err) {
+ *     // Error handling omitted.
```
stephenplusplus: What type of err would this be? Would the user's `.catch` block get executed with it, too, which I think is where they'd expect to handle it?

As a promise user, if you were using this method, would you think it would be easier to handle the success/failure of a commit *on the commit call itself*? I know we had some tricky requirements for this method with the retry rules, but given the chance to re-think it for a second, what about:

```js
database.runTransaction(function(err, transaction) {
  // handle err

  transaction.insert(...)

  transaction.commit()
    .then(handleSuccess, handleFailure)
})
```

This way, I don't think they'd have to return the promise, so the whole thing would feel more predictable. I believe `commit` (or any other transaction-terminating method) would still be able to retry the function passed to `runTransaction`. Because if it failed, it would not execute the callback, which would block the promise from resolving. Their promise pipeline would halt and we'd run the `runFn` again, and it would do this until it succeeds or hits a max-attempt limit.

WDYT? Is that possible?

packages/spanner/src/database.js
```diff
@@ -713,45 +713,44 @@ Database.prototype.runStream = function(query, options) {
...
- *
- *         // Commit the transaction.
- *         return transaction.commit();
+ * database.runTransaction(function(err, transaction) {
+ *   if (err) {
+ *     // Error handling omitted.
+ *   }
+ *
+ *   // Run a transactional query.
+ *   return transaction.run('SELECT * FROM Singers')
```
lukesneeringer: Generally I dislike style discussions in code reviews, but I had to re-read this a couple times because of the odd nesting and inconsistently aligned `.then`s.

Can this be changed to:
```javascript
return transaction.run('SELECT * FROM Singers').then(function() {
```

If that exceeds 80 characters, change `transaction` to `txn`. It will still end up more readable.

packages/spanner/src/transaction-request.js
```diff
@@ -201,6 +202,12 @@ TransactionRequest.prototype.createReadStream = function(table, query) {
     table: table
   }, query);
 
+  if (this.transaction && this.id) {
```
lukesneeringer: Is this the entire bugfix? :open_mouth: 
As someone unfamiliar with the code, can you explain why this was the solution? It is not really clear from reading the changelog.

packages/spanner/src/transaction.js
callmehiphop: Actually the error is most likely to occur from session pooling, say they had no sessions free and configured the pool to throw an error in that specific event.

I feel like the example you are proposing says to me "do not support promises in `runTransaction`".  I guess my decision to agree or disagree heavily depends on whether or not the expectation is to inform the user whether or not a Transaction was aborted while still retrying. I assumed that since we're retrying, we would only inform the user when we either got the final failure or a success (like any other API call retry strategy).

packages/spanner/src/database.js
callmehiphop: Sure, I can dig it!

packages/spanner/src/transaction-request.js
callmehiphop: Haha, all the code within `TransactionRequest` was originally baked into our `Table` class - we didn't tackle transactions until later so a lot of that code was refactored out into what is now `TransactionRequest`, which serves as an abstract class for `Table` and `Transaction`. I think this simply just got missed while refactoring the code originally.

packages/spanner/src/transaction.js
stephenplusplus: > do not support promises in runTransaction

I just don't understand how it works, and it doesn't match our pattern for the rest of our API in how we use callbacks vs promises. If this is how we supported promise usage, then I'd understand:

```js
database.runTransaction() // no callback? promise mode!
  .then(function([transaction]) {
    transaction.insert(...)
  })
  .catch(function([err]) {})
```

But `runTransaction` already take a function that gets executed *even in promise mode* to communicate if it failed. As I mentioned so far, the goal is maximum comprehension, intuitiveness, and predictability. As I think of this, `runTransaction` creates a sandbox. You can execute all the promises / non-promises you want in there, just like you can outside of it. But the walls of the sandbox are purposefully built so that there's predictability of where related behavior stays.

>  whether or not the expectation is to inform the user whether or not a Transaction was aborted while still retrying. I assumed that since we're retrying, we would only inform the user when we either got the final failure or a success (like any other API call retry strategy).

Yes, we would just "try to make it work" once, as far the user sees. Then, after all of our attempts, we communicate that it failed.

packages/spanner/src/database.js
stephenplusplus: > Generally I dislike style discussions in code reviews,

We value that all the same. If the docs make something confusing, it doesn't matter how good our code is, nobody will ever know how to use it :)

packages/spanner/src/transaction.js
callmehiphop: I think those are fair points, I am hesitant because I already see this being an annoyance for Promise users. I'm guessing if we were to just not support Promises here, we would see a bunch of users doing something like this..

```js
function runTransaction() {
  return new Promise((resolve, reject) => {
    database.runTransaction((err, transaction) => {
      if (err) return reject(err);

      // do a bunch of transactional chaining..
      transaction.commit().then(resolve, reject);
    });
  });
}
```

This makes me sad. It might be because we make the user manually call `commit`, but I feel like we should be able to offer Promise users something similar to this - I'm just not sure what the ideal way to go about this is.

packages/spanner/src/transaction-request.js
lukesneeringer: That makes sense, thank you.

packages/spanner/src/transaction.js
stephenplusplus: You are a better representative of our promise users, so if what I'm suggesting is more annoying to you than what we have now, then let's just call it case closed and keep what we have. If you think it's worth talking about more, please open us an issue or just wait to see if an annoyed user brings it up for us, thus validating the discussion.

lukesneeringer: My vote: As we are in "promise mode", we should avoid ever having an `(err, desiredArg)` style function signature.

Looking at the code more carefully now.

lukesneeringer: Could I see a quick usage example (in this thread) of both callback style and promise style? I feel like the doc example I am reading _mixes_ them, which is bizarre to me.

packages/spanner/src/database.js
```diff
@@ -713,45 +713,43 @@ Database.prototype.runStream = function(query, options) {
...
+ *
+ *   // Run a transactional query.
+ *   return transaction.run('SELECT * FROM Singers').then(function() {
+ *     // Queue a mutation (note there is no callback passed to `insert`).
+ *     transaction.insert('Singers', {
+ *       SingerId: 'Id3b',
+ *       Name: 'Joe West'
+ *     });
+ *     // Commit the transaction.
+ *     return transaction.commit();
```
vkedia: I know we have discussed this before but I am trying to job my memory. What is the reason to call transaction.commit. Why cant the completion of the returned Promise indicate that the transactions needs to be committed (if the promise resolved) or rolled back (if the promise rejectes)?

packages/spanner/src/database.js
```diff
@@ -713,45 +713,43 @@ Database.prototype.runStream = function(query, options) {
...
  *   });
+ * })
+ * .then(function() {
+ *   // Transaction committed successfully.
```
vkedia: what happens if the transaction fails? Can we show example of that as well.

packages/spanner/src/transaction.js
vkedia: I admit that I don't completely understand what is being suggested/discussed here. But my naive take would be that the runfunction should not take an error argument. We should just be able to handle that by rejecting the Promise returned. Also I am not sure I understand why does user need to call commit. Even in the non promise case, i dont undertstand the need to pass error to run function.

packages/spanner/system-test/spanner.js
```diff
@@ -976,17 +976,18 @@ var spanner = new Spanner(env);
...
+          return Promise.reject(err);
+        }
 
-          return transaction.commit();
+        transaction.insert('Singers', {
+          SingerId: id,
+          Name: name
         });
+
+        return transaction.commit();
```
vkedia: I dont see any assertions here?

packages/spanner/system-test/spanner.js
```diff
@@ -1034,9 +1035,56 @@ var spanner = new Spanner(env);
 
             transaction.commit(function(err) {
               assert.ifError(err);
+              firstTransaction.commit(assert.ifError);
```
vkedia: I dont understand why do you need to call commit here? On line 1013, instead of returning, you should go ahead with the commit and make sure that commit fails.
Also in the second attempt of first transaction, we should assert that we read the values that were inserted by second transaction.

packages/spanner/src/database.js
lukesneeringer:
```javascript
.catch(function(err) {
```

Anyone using this syntax would intuitively know that.

packages/spanner/src/database.js
lukesneeringer: Because then how would someone run a transaction with multiple statements?

callmehiphop: @stephenplusplus might be table to speak to this?

packages/spanner/src/database.js
callmehiphop: Sure!

vkedia: I think it is not clear in this case since the function also takes an error. Will the function get called with the final error as well?

packages/spanner/src/database.js
vkedia: @lukesneeringer I don't understand your comment. Can you please explain with an example?

packages/spanner/system-test/spanner.js
callmehiphop: Our testing framework allows you to return a promise and if the promise is rejected the test will fail. Essentially here we're just checking to see if we can run/commit a transaction via promises.

packages/spanner/src/database.js
lukesneeringer: We could, at some future import, introduce some sort of "context manager promise" for a transaction that added this way (it would require a clever curry):

```javascript
transaction.atomic(txn => new Promise((res, rej) => {
  txn.insert(...);
  txn.run(...);
  other_code_here();
  txn.run(...);
});
```

That is outside of the scope of this, though.

stephenplusplus: @lukesneeringer's statement is true, but also because we didn't want to mask anything to do with committing. We did this with Datastore, "just call `done()` when your transaction is ready to be committed" and it confused many users. We ended up exposing the explicit API, and haven't received any complaints.

packages/spanner/src/transaction.js
lukesneeringer: @vkedia I am not 100% sure I understand either, but I _think_ the issue is that we may or may not actually be in a Promise context.

packages/spanner/src/database.js
stephenplusplus: Previous convo: https://github.com/GoogleCloudPlatform/google-cloud-node-private/pull/16/files/6ee316c44ae6ddd8ac720bde34b3bdd69b0ee49b#issuecomment-271366227

vkedia: @lukesneeringer Unfortunately I still dont understand what going on. @callmehiphop can you add a test case which involves invoking multiple statements on a single transaction in both Promise and non Promise mode.

packages/spanner/system-test/spanner.js
callmehiphop: > I dont understand why do you need to call commit here?

This could be a misunderstanding by me - I was working with the notion that I needed to wrap up a second transaction before committing the first in order to get an aborted error? I was just passing the first transaction. Since everything is async, I need to pass some sort of callback in order to ensure the commit order is correct.

packages/spanner/src/database.js
vkedia: @stephenplusplus But this is not exactly the same situation. IIUC you are saying that in datastore users had to call `done()` but here what I am proposing is that users should just resolve the Promise when they are done with the transaction. To me that would seem pretty idiomatic but I am a node.js  noob so I might be wrong.

packages/spanner/system-test/spanner.js
vkedia: Your understanding is correct. The second transaction does need to commit for the first one to fail. I see your point about async. But I think the code would be easier to understand if we did not break the usual flow by calling commit somewhere else as you are doing now. Maybe you can some other mechanism to communicate to the first callback that the second txn has been committed, some sort of semaphore.

packages/spanner/src/database.js
stephenplusplus: That's just the promise flavor of calling done(), but the similarity is we are trying to associate an action to commit as a side effect, whereas what we found was users just wanted to call "commit()" themselves. But there are technical limitations, anyway-- it's not possible with non-promise use cases, as the user has to call a function to signal that they're ready to commit the transaction due to the async nature of JS.

packages/spanner/system-test/spanner.js
vkedia: I think we should also check that the data actually got inserted.

packages/spanner/src/database.js
vkedia: Ok. So to be clear, in the Promise use case technically we don't need the user to call commit. But to be consistent with the non Promise api, we want them to call commit there as well.  That is fine.

lukesneeringer: > @lukesneeringer's statement is true, but also because we didn't want to mask anything to do with committing. We did this with Datastore, "just call done() when your transaction is ready to be committed" and it confused many users. We ended up exposing the explicit API, and haven't received any complaints.

I bow to user data. Something like what I showed above would work well in Python where context managers are a well-understood thing. But you are completely right that it is not nearly as prevalent of a pattern in JavaScript (and it is hard to do).

packages/spanner/src/database.js
lukesneeringer: Yeah, the function taking an `err` object is why this thread is still open.

packages/spanner/system-test/spanner.js
lukesneeringer: I agree with @vkedia here.

packages/spanner/src/database.js
lukesneeringer: I just realized I did not reply to something that @vkedia asked:

> @lukesneeringer I don't understand your comment. Can you please explain with an example? 

An example of multiple statements in a transaction is:

```es6
transaction.runTransaction()
  .then(txn => txn.insert('...data to insert...'))
  .then(txn => txn.update('...data to change...'))
  .then(txn => txn.insert('...other data to insert...'))
  .then(txn => txn.commit());
```

> The Promise that the function returns should resolve only after they have done all they want with the transaction.

I think you may misunderstand what promises are for. They are not a tool for "wrapping" arbitrary amounts of behavior like a context manager; they are a way of idiomatically waiting on _a single_ asychronous event. Chaining a bunch of `.then` calls is common (see my example).

So, there is absolutely no guarantee that the resolution of a specific promise means that you are done with the transaction. In fact, the opposite is true -- it would be entirely idiomatic for the promise to resolve when the transaction is _successfully opened_, before you execute a single statement (whether the code actually does that I am unsure).

vkedia: @lukesneeringer In  your example you are not passing a callable to runTransaction. As far as I can tell the API is that the runTransaction takes in a callable that returns a Promise. User should resolve that Promise only when he is done with the txn. So your example would look like this:

```es6
transaction.runTransaction(function(error, txn) {
  // Note that insert and updates just buffer the mutations in memory and do not involve any IO so 
 // I think they dont return a Promise.
  txn.insert('...data to insert...'))
  txn.update('...data to change...'))
  txn.insert('...other data to insert...'))
  // If you wanted to also do multiple queries you could do this.
  return txn.run("Select * from Singer")
    .then(txn => txn.run("Select * from Singers")
    .then(txn => txn.commit());
})
  .catch(err => handle error /* this runs if ANYTHING above fails */);
```

Also note that user does not call rollback himself, it automatically gets rolled back by the client library if there is some non retryable error

lukesneeringer: My understanding is that our callback-promise functions return a promise if it gets no callback, and run the callback if they get one. I am not sure _this function_ works like this (see other parts of the discussion). but I was answering the in-principle question of whether the first promise resolution should commit the transaction (which is: "no").

lukesneeringer: @callmehiphop Are `txn.insert` and `txn.update` in our examples above async?

(If they are not, then my example code is wrong. If they are, then Vikas' example code is wrong.)

vkedia: @lukesneeringer take a look at this test:
https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2144/files#diff-aaea4d158497e38a9c5306e1834c5442R979

packages/spanner/src/transaction.js
stephenplusplus: Just link-dropping-- posted the question to Twitter and asked around for advice. Dave and I agreed this sums up our two best options:

https://twitter.com/stephenplusplus/status/846822914490073088
https://gist.github.com/stephenplusplus/3cb49529ee537c640a4670edf900e294

packages/spanner/src/database.js
callmehiphop: @lukesneeringer they aren't async - all those mutations can only be sent via `commit()` so we let the user do sync mutations that essentially just get bundled up and sent at the end of the transaction.

PR #2147, opened 2017-03-28
Coverage branch: codec.js

lukesneeringer:
Quick addition of a test to catch an `if` branch we missed.


stephenplusplus:
Did we expect coverage to stay the same? Merge away If so!

lukesneeringer:
Yeah, it fixed a missing _branch_, not a missing _line_.

stephenplusplus:
Ah, got it. Thanks!

PR #2148, opened 2017-03-28
spanner: add txn id to reads and remove promise support in runTransaction

### LINKED ISSUE #2118: Spanner overlapping transactions don't fail when they should?

callmehiphop:
Closes #2118 
Closes #2144

Per https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2118#issuecomment-289919717 I'm extracting out the critical parts while we decide on how best to integrate Promises within `runTransaction`.

### DIFF
--- a/packages/spanner/system-test/spanner.js
+++ b/packages/spanner/system-test/spanner.js
@@ -1069,23 +1069,6 @@ var spanner = new Spanner(env);
       });
     });
 
-    it('should run a transaction in promise mode', function() {
-      var id = generateName('id');
-      var name = generateName('name');
-
-      return database.runTransaction()
-        .then(function(data) {
-          var transaction = data[0];
-
-          transaction.insert('Singers', {
-            SingerId: id,
-            Name: name
-          });
-
-          return transaction.commit();
-        });
-    });
-
     it('should retry an aborted transaction', function(done) {
       var id = generateName('id');
       var name = generateName('name');
@@ -1105,7 +1088,9 @@ var spanner = new Spanner(env);
           });
 
           if (attempts < 2) {
-            runOtherTransaction(transaction);
+            runOtherTransaction(function() {
+              transaction.commit(assert.ifError);
+            });
             return;
           }
 
@@ -1117,7 +1102,7 @@ var spanner = new Spanner(env);
         });
       });
 
-      function runOtherTransaction(firstTransaction) {
+      function runOtherTransaction(callback) {
         database.runTransaction(function(err, transaction) {
           assert.ifError(err);
 
@@ -1131,8 +1116,7 @@ var spanner = new Spanner(env);
 
             transaction.commit(function(err) {
               assert.ifError(err);
-
-              firstTransaction.commit(done);
+              callback();
             });
           });
         });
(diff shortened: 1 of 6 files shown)


packages/spanner/system-test/spanner.js
```diff
@@ -1105,7 +1088,9 @@ var spanner = new Spanner(env);
           });
 
           if (attempts < 2) {
```
vkedia: I dont undertstand why was this test passing so far, if transaction id was not being specified?

packages/spanner/system-test/spanner.js
```diff
@@ -1105,7 +1088,9 @@ var spanner = new Spanner(env);
...
-            runOtherTransaction(transaction);
+            runOtherTransaction(function() {
```
vkedia: Can you also add another test, exactly like this, except that instead of sql query you are doing read by specifying keys?

packages/spanner/system-test/spanner.js
vkedia: I wonder can it be that both these transactions end up using the same session, so the first one is actually failing due to that and not because of conflicts?

PR #2153, opened 2017-03-29
bigquery: table.insert: correct docs

### LINKED ISSUE #2149: BigQuery table insert documentation is incorrect and insufficient
#### Environment details

  - OS: OS X El Capitan
  - Node.js version: 6.9.x
  - npm version: 3.10.9
  - google-cloud-node version: 0.49

#### Issues

- Incorrect Promise API array indexes in documentation
- Incomplete `table.insert` `PartialFailureError` handling example in Promise API documentation

#### Steps to reproduce

1. Follow [documentation here](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.49.0/bigquery/table?method=insert) and write the following code:
```
    this.table.insert(rows, { raw: false }).then((data) => {
      let insertErrors = data[0];
      let apiResponse = data[1];
    }
```
2. Debug the code and notice that the API Response is in `data[0]` and that the insertion errors are in `data[1]`. This is the opposite of what the docs state.

The code makes sense because I don't want to be checking for the `apiResponse` in index 0 on success and index 1 on failure. '`apiResponse` should be in index 0 as it is now and then index 1 will be empty or have the insertion errors if they exist. So this seems to be a docs issue rather than a code issue.

My working code is similar to this:
```
    this.table.insert(item, { raw: false }).then((data) => {
      let insertErrors = data[1];

      if (insertErrors) {
        logger.info(`insertErrors: ${JSON.stringify(insertErrors)}`);
        // Some rows failed to insert, while others may have succeeded.
        insertErrors.map((insertError) => {
          insertError.errors.map((error) => {
            logger.error(`PartialFailureError: BigQuery insert failed due to: ${JSON.stringify(error)}`);
          });
        });
      }
      return message;
    }).catch((error) => {
      throw new Error(`Error inserting into bigQuery for id: "${item.id}": ${JSON.stringify(error)}`);
    });
```

3. Now the callback docs for `table.insert` mention that you can do something like `if (err.name === 'PartialFailureError')` but using the Promise API, I never get back any objects with a `name` field. The `apiResponse` is of this format:
```
  "kind": "bigquery#tableDataInsertAllResponse",
```
And the `insertErrors` object is of this format:
```
  [
    {
      "index": 1,
      "errors": [
        {
          "reason": "invalid",
          "location": "name",
          "debugInfo": "generic::invalid_argument: Conversion from int64 to string is unsupported.",
          "message": "Conversion from int64 to string is unsupported."
        }
      ]
    },
    {
      "index": 0,
      "errors": [
        {
          "reason": "stopped"
        }
      ]
    }
  ]
```
So it's unclear why the docs indicate that you should check against the `err.name` field.

At minimum, if there is a significant difference for handling the response from the Promise API, then that needs to be documented but currently the docs for using the `table.insert` Promise API are bare bones and incomplete.

From [this comment](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2052#issuecomment-287203354) it looks like the tests are only written against the callback API and not the Promise API.

stephenplusplus:
Fixes #2149

### DIFF
--- a/packages/bigquery/src/table.js
+++ b/packages/bigquery/src/table.js
@@ -1051,10 +1051,22 @@ Table.prototype.import = function(source, metadata, callback) {
  * //-
  * // If the callback is omitted, we'll return a Promise.
  * //-
- * table.insert(rows).then(function(data) {
- *   var insertErrors = data[0];
- *   var apiResponse = data[1];
- * });
+ * table.insert(rows)
+ *   .then(function(data) {
+ *     var apiResponse = data[0];
+ *   })
+ *   .catch(function(err) {
+ *     // An API error or partial failure occurred.
+ *
+ *     if (err.name === 'PartialFailureError') {
+ *       // Some rows failed to insert, while others may have succeeded.
+ *
+ *       // err.errors (object[]):
+ *       // err.errors[].row (original row object passed to `insert`)
+ *       // err.errors[].errors[].reason
+ *       // err.errors[].errors[].message
+ *     }
+ *   });
  */
 Table.prototype.insert = function(rows, options, callback) {
   if (is.fn(options)) {


Splaktar:
I would love to see a test that verifies that this example actually works in v0.49.0 or is it only for v0.50.0? I know that the code I wrote two weeks ago was working fine in the `then` block which means that the `catch` block wasn't being thrown for `PartialFailureError`s.

stephenplusplus:
I just tried causing a failure using promises, both on master and from the git tag when we released 0.49.0, and the success handler did not trigger, while the catch was executed with the expected PartialFailureError.

The reason why we're not adding a unique promise test is because this method is no exception from how *all* of our others behave, in terms of how callbacks/promises are handled. By checking that the callback style works, we are simultaneously confirming it works with promises, as we know the promise behavior is consistent across our API.

I'm not sure how to explain this scenario you mentioned in #2149:

```js
  this.table.insert(item, { raw: false }).then((data) => {
      let insertErrors = data[1];
      // --------------------
      // data[1] should be undefined, we only executed this with one argument, the apiResponse
      // --------------------

      if (insertErrors) {
        logger.info(`insertErrors: ${JSON.stringify(insertErrors)}`);
        // Some rows failed to insert, while others may have succeeded.
        insertErrors.map((insertError) => {
          insertError.errors.map((error) => {
            logger.error(`PartialFailureError: BigQuery insert failed due to: ${JSON.stringify(error)}`);
          });
        });
      }
      return message;
    }).catch((error) => {
      throw new Error(`Error inserting into bigQuery for id: "${item.id}": ${JSON.stringify(error)}`);
    });
```

Could you provide an isolated, reproducible example in that issue if you're still seeing an apiResponse in the success handler that has insertErrors?

Splaktar:
OK, that makes a lot of sense. Thank you. I will run some more tests on my end.

PR #2155, opened 2017-03-29
Additional Spanner tests.

lukesneeringer:
This commit provides tests for every missed line in the spanner package.
Additionally, it brings branch coverage to 100% in codec.js and partial-result-stream.js.

### DIFF
--- a/packages/spanner/test/codec.js
+++ b/packages/spanner/test/codec.js
@@ -453,13 +453,15 @@ describe('codec', function() {
 
     it('should only encode public properties of objects', function() {
       var obj = {
-        hasOwnProperty: function(key) { return key === 'public' },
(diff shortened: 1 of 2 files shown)


packages/spanner/test/codec.js
```diff
@@ -453,13 +453,15 @@ describe('codec', function() {
 
     it('should only encode public properties of objects', function() {
       var obj = {
-        hasOwnProperty: function(key) { return key === 'public' },
+        hasOwnProperty: function(key) {  // jshint ignore:line
```
lukesneeringer: I dislike ignoring things in the linter, but this is a spot where it really is the right thing to do.
We _want_ the linter to complain if we define `hasOwnProperty` arbitrarily, but in this spot it is the entire point of the test.

PR #2156, opened 2017-03-29
Improve coverage in common.

lukesneeringer:
This gets line coverage in common to 100% and improves branch coverage.


stephenplusplus:
LGTM, just a few linting issues.

lukesneeringer:
In a meeting; will knock them when I get out.
I will also install a linter plugin in Atom so I catch those sooner.

PR #2162, opened 2017-03-30
common: update google-auto-auth to 0.6.0

ofrobots:


stephenplusplus:
Just leaving the note, `0.6.0` adds the ability to detect if we're in a Container Engine environment.

(https://github.com/stephenplusplus/google-auto-auth/commits/master)

PR #2169, opened 2017-03-31
Mark a single line of Vision uncovered.

lukesneeringer:
I am declaring bankruptcy on the last line of coverage for now. I just want coveralls to stop giving Xes. I will go back and refactor that code later (it will be a non-trivial effort).


stephenplusplus:
OMG!

PR #2173, opened 2017-03-31
ci: doc scripts

callmehiphop:
Previously the `prepare-ghpages` task would create a submodule of gh-pages and update it with the latest doc for a single module. This change essentially allows you to update the same submodule with multiple modules docs without having to cleanup the old submodule and re-check it out.


stephenplusplus:
Can you add a little detail on what this PR did?

Issue #1930, opened 2017-01-19
[autogen] add auto-generated unit tests


jmuk:
@shinfan has been working for the auto-generation of unit tests (for the autogen layers), the results are seen under https://github.com/googleapis/api-client-staging/tree/master/generated/nodejs/google-cloud-node/packages right now.

It would be beneficial to merge them into the GCN repository.

stephenplusplus:
I believe we discussed this before, and thought `api-client-staging` was the winner for hosting these tests. I don't feel too strongly about this if there aren't better options, but would prefer the repo that generates the files should test those files they generate. Since that development doesn't happen here, it feels out of place.

stephenplusplus:
@jmuk are there APIs left that need to have tests auto-generated?

jmuk:
I've been detached from the progress here. reassigning to @swcloud ?

stephenplusplus:
These are coming in as we're bringing in the partial APIs and full-GAPIcs. I'm not sure this issue is necessary, but to any interested parties, please re-open if that's an incorrect assumption.

PR #2175, opened 2017-04-01
speech: update to v1

stephenplusplus:
Update the Speech API to v1.


stephenplusplus:
Note sure: #1930

lukesneeringer:
For now, my vote is to leave them included (they may or may not _actually run_). We are working on a longer-term testing strategy for these.

It is worth noting that while the _development_ does not happen here, this is more or less the canonical location for GAPICs. :-/

I think the best way to do that ultimately is to have the generation process make a PR to this repo, and have the tests run that way, and rely less on api-client-staging as a canonical location.

stephenplusplus:
That's cool.

packages/speech/src/index.js
```diff
@@ -492,6 +432,12 @@ Speech.prototype.createRecognizeStream = function(config) {
     throw new Error('A recognize request requires a configuration object.');
   }
 
+  config = extend(true, {
+    config: {
+      languageCode: 'en-US'
```
lukesneeringer: This is bad. The client library needs to not assume `en-US`. The user must provide this explicitly. (The backend API changed this for a reason.)

PR #2179, opened 2017-04-03
docs: show correct module name in deprecation notice

stephenplusplus:
Looks like I wasn't finding the module name correctly in the deprecation notice: https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.50.0/bigquery.

The docs basically say "don't use `google-cloud`, use `google-cloud`".

@callmehiphop does this look right?

### DIFF
--- a/scripts/docs/templates/overview.html
+++ b/scripts/docs/templates/overview.html
@@ -6,10 +6,10 @@
 <div class="notice">
   <strong><code>google-cloud</code> is deprecated</strong>
   <p>
-    Due to the scope of our API rapidly increasing, you may notice a decrease in performance if you continue to use the bundled module, `google-cloud`.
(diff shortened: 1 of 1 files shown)


callmehiphop:
@stephenplusplus I think `className` will be camel cased. You *might* need to add a new field to the [template params](https://github.com/stephenplusplus/gcloud-node/blob/7f621cbf6dda9734eee002649fac67687ccca6ee/scripts/docs/parser.js#L391) for the correct formatting

stephenplusplus:
Thanks for that link-- I think that actually means we're good with `className` because its value is `moduleName`, and `moduleName` is used during the require, but let me know if you get a different read: https://github.com/stephenplusplus/gcloud-node/blob/7f621cbf6dda9734eee002649fac67687ccca6ee/scripts/docs/parser.js#L387

stephenplusplus:
Woo!

PR #2017, opened 2017-02-23
Stage errors for repo integration

cristiancavalli:
TODOS:
- [x] Refactor test directory to follow repo conventions
- [x] Unit and system tests pass
- [x] Fix linting errors
- [x] Switch from diagnostics-common to gcp-common

PR #2181, opened 2017-04-03
docs: update script parser for error-reporting

callmehiphop:
RE: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2017


stephenplusplus:
I asked if the ER PR could just move `index.js` to `src/` so we wouldn't need any special handling.

callmehiphop:
@stephenplusplus nice, I'll update this PR once that gets resolved then!

callmehiphop:
@stephenplusplus I've updated the PR.

PR #2183, opened 2017-04-04
Generate gapic client for language v1beta2

landrito:


lukesneeringer:
@landrito I think the coverage change is because you are out of date on your local master. Can you merge master into your PR and push again?

landrito:
Done!

stephenplusplus:
Is there documentation on this versioning scheme, I couldn't find any. I would have thought v1beta2 preceded v1, but we already have v1. Just want to confirm that we didn't make a mistake somewhere, and our v1 was actually a beta as well.

Also, will we drop the current v1 when we switch our code to use v1beta2, or do we leave the versions around for users to access directly (not using the handwritten layer)?

lukesneeringer:
> Is there documentation on this versioning scheme, I couldn't find any. I would have thought v1beta2 preceded v1, but we already have v1. Just want to confirm that we didn't make a mistake somewhere, and our v1 was actually a beta as well.

There is mass confusion on this question. Yes, `v1beta2` _follows_ `v1`. We are trying to determine how we are going to handle this for client libraries because you are about the 74,819th person to ask this.

stephenplusplus:
Gotcha, thanks! Just some questions about how to handle the version update, because this PR changes `v1` source content as well as adds a new directory for `v1beta2` content.

- should we drop `v1` the directory and the access points in our code to it?
- are we switching our handwritten code to use v1beta2 now?

lukesneeringer:
The answers to that are: No, and no.
What we landed on was just dropping v1beta2 in and not making any changes to the manual layer at all _except_ to add it to `module.exports` as a separate thing people can get it.

I ended up doing this in #2188 (d'oh, sorry @landrito), and that PR supercedes this one.

PR #2185, opened 2017-04-04
Support proper query strings on getJobs.

### LINKED ISSUE #2182: bigquery jobs list api: the stateFilter need useQuerystring in the request options
the jobs list API says I can filter jobs by state: `done` `pending` or `running` but in some cases I want to know all pending and running jobs, instead of run jobs list twice, I found the bq server accepts if I provide the stateFilter twice in qs, that's by [httpie](https://pypi.python.org/pypi/httpie); it needs `stateFilter=pending&stateFilter=running&...` in the GET request parameters;
however with current `google-cloud-node` code, if I provide `stateFilter: [ 'pending', 'running' ]` in the [`getJobs`](https://googlecloudplatform.github.io/google-cloud-node/#/docs/bigquery/0.9.0/bigquery?method=getJobs) call, the actual sent on the wire is `stateFilter[0]=pending&stateFilter[1]=running&...` which bq server just ignored

https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/list

when I research where did that `[0]` `[1]` comes from, it turns to be a default behavior in request options, users need to turn on `useQuerystring: true` to avoid that,
https://www.npmjs.com/package/request#requestoptions-callback

So here comes a diff I made it working locally, I think request interceptors may also be able to get it working; but you guys may want to integrate this `useQuerystring: true` to be the default

```js
--- ./node_modules/@google-cloud/bigquery/src/index.js.orig	2017-03-23 11:48:53.730426569 -0700
+++ ./node_modules/@google-cloud/bigquery/src/index.js	2017-04-03 03:13:23.873907848 -0700
@@ -613,15 +613,16 @@ BigQuery.prototype.getJobs = function(op
     options = {};
   }
 
   options = options || {};
 
   this.request({
     uri: '/jobs',
-    qs: options
+    qs: options,
+    useQuerystring: true,
   }, function(err, resp) {
     if (err) {
       callback(err, null, null, resp);
       return;
     }
 
     var nextQuery = null;
```

lukesneeringer:
This PR adds `useQueryString: true` to the request on the BigQuery `getJobs` method, allowing `options.stateFilter` to filter by more than one job state.

Closes #2182.

### DIFF
--- a/packages/bigquery/src/index.js
+++ b/packages/bigquery/src/index.js
@@ -569,7 +569,8 @@ BigQuery.prototype.getDatasetsStream =
  *     of selected fields. Acceptable values are "full", for all job data, and
  *     "minimal", to not include the job configuration.
  * @param {string=} options.stateFilter - Filter for job state. Acceptable
- *     values are "done", "pending", and "running".
+ *     values are "done", "pending", and "running". Sending an array to this
+ *     option performs a disjunction.
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request
  * @param {module:bigquery/job[]} callback.jobs - The list of jobs in your
@@ -617,7 +618,8 @@ BigQuery.prototype.getJobs = function(options, callback) {
 
   this.request({
     uri: '/jobs',
-    qs: options
+    qs: options,
+    useQueryString: true
   }, function(err, resp) {
     if (err) {
       callback(err, null, null, resp);


lukesneeringer:
Can I get a review on this? It should only take a couple minutes. :-)

callmehiphop:
LGTM - I think we'll want to update our unit tests to assert that this option is set properly.

packages/bigquery/src/index.js
```diff
@@ -617,7 +618,8 @@ BigQuery.prototype.getJobs = function(options, callback) {
 
   this.request({
     uri: '/jobs',
-    qs: options
+    qs: options,
+    useQueryString: true
```
c0b: I found the problem of https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2182#issuecomment-294312078 `@google-cloud/bigquery@0.9.2` is broken because of here; wonder if @lukesneeringer actually read my #2182 ? or read the [`request`](https://www.npmjs.com/package/request) doc?

```diff
-    useQueryString: true
+    useQuerystring: true
```

image

PR #2186, opened 2017-04-04
Require languageCode to be explicitly specified.

lukesneeringer:
We need to require our users to explicitly require `languageCode`. Defaulting to English is not acceptable. (The API changed this for a reason; it was a common point of failure for users.)

### DIFF
--- a/packages/speech/src/index.js
+++ b/packages/speech/src/index.js
@@ -597,6 +597,7 @@ Speech.prototype.operation = function(name) {
  * //-
  * var config = {
  *   encoding: 'LINEAR16',
+ *   languageCode: 'en-US',
  *   sampleRateHertz: 16000,
  *   verbose: true
  * };
@@ -752,6 +753,7 @@ Speech.prototype.recognize = function(file, config, callback) {
  * //-
  * var config = {
  *   encoding: 'LINEAR16',
+ *   languageCode: 'en-US',
  *   sampleRateHertz: 16000,
  *   verbose: true
  * };


packages/speech/src/index.js
```diff
@@ -372,6 +372,8 @@ Speech.formatResults_ = function(resultSets, verboseMode) {
  *
  * @param {object} config - A `StreamingRecognitionConfig` object. See
  *     [`StreamingRecognitionConfig`](https://cloud.google.com/speech/reference/rpc/google.cloud.speech.v1beta1#google.cloud.speech.v1beta1.StreamingRecognitionConfig).
+ * @param {string} config.languageCode - The language of the supplied audio as
```
stephenplusplus: Can you turn BCP-47 into a link to a definition and list of acceptable values?

lukesneeringer: That seems reasonable. :-)

packages/speech/src/index.js
```diff
@@ -432,11 +435,19 @@ Speech.prototype.createRecognizeStream = function(config) {
...
+  //
+  // This is expected within a nested config, but in the interest of user
+  // sanity, accept it in the outer config object.
+  config.config = config.config || {};
+  if (config.languageCode) {
+    config.config.languageCode = config.languageCode;
+    delete config.languageCode;
+  }
+  if (is.undefined(config.config.languageCode)) {
+    throw new Error('A `languageCode` is required in the config object.');
```
pbogle: The server will return an error if the language code isn't provided.  Therefore it's not necessary to duplicate that logic here.   (Letting the server do error checking ensures consistency when accessing it through the veneer.  Also, if we ever added automatic language detection on the server and relaxed the constraint, we wouldn't need to update all of the client libraries to let users take advantage of the change.)

packages/speech/src/index.js
```diff
@@ -620,9 +634,11 @@ Speech.prototype.recognize = function(file, config, callback) {
     throw new Error('A recognize request requires a configuration object.');
   }
 
-  config = extend({
-    languageCode: 'en-US'
-  }, config);
+  // As of Speech v1, a language code is required; throw an exception if we
+  // did not receive one.
+  if (is.undefined(config.languageCode)) {
+    throw new Error('A `languageCode` is required in the config object.');
```
pbogle: Likewise here we can rely on the server to validate parameters.

packages/speech/src/index.js
lukesneeringer: Hmm.

Not entirely true. The other client libraries have hard language code requirements, because we made `language_code` or `languageCode` a required argument to the methods. In other words, the Python or Ruby client will never _get to_ the server because you would get an error message saying, "`recognize` has one required argument, you sent zero" (paraphrase).

At minimum, the notion that you can relax this in the future and most clients will require no changes is untrue. It is technically feasible for _this_ one since all arguments in JavaScript are optional.

packages/speech/src/index.js
lukesneeringer: Done.

packages/speech/src/index.js
pbogle: I see your point. No need to change this in this instance.

lukesneeringer:
@stephenplusplus Anything else? (Other than CI, of course.)

stephenplusplus:
Nope, LGTM.

lukesneeringer:
Can you flip the approve bit, then? :-D

Still looks like there are CI issues though. Investigating that.

Issue #1296, opened 2016-05-06
automatic batching interface for cloud logging client


faithseed:
The logging API support batching mode manually, where you can put multiple entry in the log.write()

`
log.write([entry, secondEntry]);
`

Is that possible to implement an batching interface to support this automatically?

for example:

```
log.add(entry);
...
log.add(another_entry);
```

In the background, logging client does periodic flush internally sending all buffered logs to cloud logging automatically. The flush is triggered when enough entries are buffered and/or at a fixed interval.

This would make the API better performance wise by making less APIs calls to cloud logging?

stephenplusplus:
I think this is better resolved in the application layer and not our library, since every use case would likely have custom requirements around what action triggers a flush.

@filipjs do you think this is something we should try to support?

Also to consider, we use gRPC for our Logging API, which maintains an open connection to send multiple requests through. That eliminates the overhead of making several HTTP/1 calls, so it's possible the performance wins could be negligible.

faithseed:
Thanks @stephenplusplus.

What is performance difference between

`log.write([entry, secondEntry]);`

and

```
log.write([entry]);
log.write([secondEntry]);
```

If they are considered equally efficient, maybe we should document that in the developer doc and encourage users to choose either one of them.

stephenplusplus:
We haven't measured the performance difference, and I think it might be micro-optimization if we did or recommended an approach one way or the other. I think it's best left to the application and its concerns to make the best decision for their use.

filipjs:
Our documentation recommends batching as it's more efficient and doesn't consume QPS quota that quickly (https://cloud.google.com/logging/quota-policy)
I'm not sure how gRPC is counted for QPS purposes.

stephenplusplus:
@jgeewax any insight into how QPS are measured when using gRPC?

jgeewax:
@rimey : Any idea? Basically is "1 write log entry" equivalent to a single gRPC call? or is it the number of log entries written in the gRPC call?

Or in more specific terms, is `log.write([entry1, entry2])` counted as 1 log write entry? or 2?

stephenplusplus:
Friendly ping @rimey -- or anyone else who can help.

cc @ofrobots

ofrobots:
This thread conflates performance implications of batching with quota implications.

> Or in more specific terms, is log.write([entry1, entry2]) counted as 1 log write entry? or 2?

I think the more interesting question is this:

```
log.write(foo);
[...]

PR #2187, opened 2017-04-05
datastore | logging: promote to GA

stephenplusplus:
#### Blockers

- [ ] [automatic batching interface for cloud logging client](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1296)
- [x] [gRPC memory leak](https://github.com/grpc/grpc/issues/10445)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,16 +12,16 @@
 
 This client supports the following Google Cloud Platform services at a [General Availability (GA)](#versioning) quality level:
 
+* [Cloud Datastore](#cloud-datastore-ga) (GA)
 * [Cloud Storage](#cloud-storage-ga) (GA)
+* [Google Stackdriver Logging](#google-stackdriver-logging-ga) (GA)
 
 This client supports the following Google Cloud Platform services at a [Beta](#versioning) quality level:
 
-* [Cloud Datastore](#cloud-datastore-beta) (Beta)
 * [Cloud Natural Language](#cloud-natural-language-beta) (Beta)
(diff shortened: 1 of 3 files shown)


lukesneeringer:
@ofrobots Is this good now that the memory leak is fixed?

ofrobots:
sgtm from my side.

lukesneeringer:
@stephenplusplus You have the go-ahead to do this tomorrow morning.

stephenplusplus:
Just going over the issues left for Logging and Datastore. All but one issue will involve backwards-compatible fixes. The one I'm not sure of is a feature request for a batching interface in Logging: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1296

I think we should hold off on going GA for Logging until I have a better understanding of what's required for that issue. I've labeled it as <kbd>status: release blocking</kbd> for now.

lukesneeringer:
Uggggh. This will become urgent extremely quickly, so let's discuss in our Monday meeting how to knock these out as quickly as possible.

stephenplusplus:
I think we might be able to move forward with the release if @ofrobots finds an extra minute to let us know if it's not as serious as I'm making it out to be: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1296#issuecomment-293908332. If it's just a feature that won't affect backwards-compatibility (which I *think* it is), then we can get on with the release.

ofrobots:
I will try to take a look tomorrow.

PR #2192, opened 2017-04-07
common-grpc: correctly define default service options

### LINKED ISSUE #2190: pubsub: gRPC max recv set to 4MB
Reproduce steps:
---------------------
@google-cloud/pubsub: ~~0.3.0~~ 0.9.0
Publish message > 4MB to a topic
Retrieve message through subscription: Error: Received message larger than max (6374241 vs. 4194304)

#2007 was pushed to fix the limits but the existing PubSub client seems to be still defaulting to 4MB. It was originally posted about on #1991 which was closed after #2007 was merged but still throwing that error.

stephenplusplus:
Fixes #2190 

According to [this comment](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2190#issuecomment-292610265), I was improperly defining gRPC default options.

@murgatroid99 does this look better? Thanks!

### DIFF
--- a/packages/common-grpc/src/service.js
+++ b/packages/common-grpc/src/service.js
@@ -125,7 +125,7 @@ var GRPC_ERROR_CODE_TO_HTTP = {
 };
 
 /**
- * The default configuration for all gRPC requests.
+ * The default configuration for all gRPC Service instantions.
  *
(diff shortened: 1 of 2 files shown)


murgatroid99:
I think this should work

PR #2193, opened 2017-04-07
language: analyzeEntitySentiment support

callmehiphop:
This adds support for the `analyzeEntitySentiment` rpc.

### DIFF
--- a/packages/language/src/document.js
+++ b/packages/language/src/document.js
@@ -530,7 +530,7 @@ Document.prototype.detectEntities = function(options, callback) {
 
 /**
  * Detect the entities from a block of text, similar to
- * {module:language/document#detectEntities}, and analyzes sentiment associated
+ * {module:language/document#detectEntities}, and analyze sentiment associated
  * with each entity and its mentions.
  *
  * @param {object=} options - Configuration object.
--- a/packages/language/src/index.js
+++ b/packages/language/src/index.js
@@ -236,7 +236,7 @@ Language.prototype.detectEntities = function(content, options, callback) {
 
 /**
  * Detect the entities from a block of text, similar to
- * {module:language#detectEntities}, and analyzes sentiment associated with
+ * {module:language#detectEntities}, and analyze sentiment associated with
  * each entity and its mentions.
  *
  * NOTE: This is a convenience method which doesn't require creating a


stephenplusplus:
Is this going to replace `detectSentiment`?

packages/language/src/document.js
```diff
@@ -531,6 +529,86 @@ Document.prototype.detectEntities = function(options, callback) {
 };
 
 /**
+ * Detect the entities from a block of text, similar to
+ * {module:language/document#detectEntities}, and analyzes sentiment associated
```
callmehiphop: Done

packages/language/src/index.js
```diff
@@ -234,6 +235,92 @@ Language.prototype.detectEntities = function(content, options, callback) {
 };
 
 /**
+ * Detect the entities from a block of text, similar to
+ * {module:language#detectEntities}, and analyzes sentiment associated with
```
callmehiphop: Done

stephenplusplus:
- detectEntitySentiment = detect entities; include their sentiment scores
- detectEntities = detect entities; don't include their sentiment scores

Would things be simpler if we could just use an option to distinguish between intent, as opposed to multiple methods?

```js
language.detectEntities({ sentiment: false }) // default
language.detectEntities({ sentiment: true }) // calls detectEntitySentiment
```

Better yet, how about we just *always* include sentiment? Is there something expensive about this method that it needs to be separated from the one that doesn't include sentiment? The API response is exactly the same, with the exception of the `sentiment` object being present or not.

callmehiphop:
> Better yet, how about we just always include sentiment? Is there something expensive about this method that it needs to be separated from the one that doesn't include sentiment? The API response is exactly the same, with the exception of the sentiment object being present or not.

@stephenplusplus I think it would be better to get some feedback from googlers on this.

/cc @monattar @landrito @lukesneeringer

monattar:
Wrt question about why not always including sentiment: This feature is more expensive than all other features. We shouldn't include the sentiment by default, as it will incur high costs that the user might not intend.

We thought quite a bit about where to expose this API (within analyzeEntity, as a separate method, etc), and separate method has been the best. Please do not change the surface.

stephenplusplus:
Thanks for the information.

> and separate method has been the best.

Why?

lukesneeringer:
We do not want to merge this for a couple of reasons.

  * We do not want to opt people in to the 1.1 beta API. Users should get v1 unless they explicitly request otherwise.
  * We have decided to only expose the new method on the GAPIC for now (and DPEs are handling documenting this).

Sorry, @callmehiphop -- I did not realize you were working on this last week, or I would have updated you sooner.

PR #2201, opened 2017-04-11
logging-bunyan: fix system-test

ofrobots:
1. `BUNYAN_TO_STACKDRIVER` is private, so this can be semver-patch.
1. The _writev function gets a list of `WriteReq` objects.


stephenplusplus:
Thank you!

PR #2203, opened 2017-04-11
Language: Expose an apiVersion parameter

lukesneeringer:
This commit adds an `apiVersion` parameter to the `Language` constructor to provide access to the v1beta2 GAPIC client.

Instantiation looks like:

```javascript
var language = new Language({apiVersion: 'v1beta2'});
```

The default value is `'v1'`, and anything other than `'v1'` or `'v1beta2'` is an error.
Note that using the `'v1beta2'` version will route _all_ calls made to the object through that endpoint, and so existing methods will also get `v1beta2` behavior.

### DIFF
--- a/packages/language/src/index.js
+++ b/packages/language/src/index.js
@@ -26,7 +26,7 @@ var is = require('is');
 var v1 = require('./v1');
 var v1beta2 = require('./v1beta2');
 
-var _GAPICS = {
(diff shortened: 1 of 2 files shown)


stephenplusplus:
@lukesneeringer could you take a quick look at the changes? Feel free to merge if they look good 👍

PR #2210, opened 2017-04-11
Fixed missing production dependency: methmeth

tehsenaus:
This is used in src/session-pool.js

### DIFF
--- a/packages/spanner/package.json
+++ b/packages/spanner/package.json
@@ -46,6 +46,7 @@
     "lodash.chunk": "^4.2.0",
     "lodash.snakecase": "^4.1.1",
     "merge-stream": "^1.0.1",
+    "methmeth": "^1.1.0",
     "split-array-stream": "^1.0.0",
     "stream-events": "^1.0.1",
     "string-format-obj": "^1.1.0",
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Amazing, thank you. I'm going to release separately due to the CLA thing, and release immediately. 

Just wondering, how did you run into this? It's actually been this way since the beginning-- `methmeth` must have been loading as an indirect dependency.

stephenplusplus:
Fixed in 6d090ee800e004a07ed180df2eaa513bdd1db603 and released under `@google-cloud/spanner@0.4.1` -- thanks again!

tehsenaus:
I think this was working before due to the shrinkwrap.json in the root project.

I had to install with --no-shrinkwrap due to being behind a corporate proxy (shrinkwrap forces it to go to public NPM and we have a private instance, but that's another issue), and then the issue manifested itself.

stephenplusplus:
Ah, interesting. Thanks for the info. Can you try now to see if it's resolved? That would save us a release of `google-cloud` if it's not required.

tehsenaus:
Yes ```google-cloud@0.51.1``` is pulling in the fixed version.

Thanks for the quick fix!

Issue #2209, opened 2017-04-11
Does loggingServiceV2Client.ALL_SCOPES need cloud-platform scope?


ofrobots:
I am a bit surprised by [all the scopes](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/87312f904e8f5ba0b5202e5c4df1bf0dc5ff107a/packages/logging/src/v2/logging_service_v2_client.js#L60) that `@google-cloud/logging` requests by default:

```js
/**
 * The scopes needed to make gRPC calls to all of the methods defined in
 * this service.
 */
var ALL_SCOPES = [
  'https://www.googleapis.com/auth/cloud-platform',
  'https://www.googleapis.com/auth/cloud-platform.read-only',
  'https://www.googleapis.com/auth/logging.admin',
  'https://www.googleapis.com/auth/logging.read',
  'https://www.googleapis.com/auth/logging.write'
];
```

Are these all – and specifically `cloud-platform` – necessary? @lukesneeringer @jmuk.

jmuk:
These are from the service configuration yaml file in https://github.com/googleapis/googleapis/blob/master/google/logging/logging.yaml

I am not sure about the choice of the scopes, we simply make the union of the all scopes of the rules.

ofrobots:
I am not the authority on this, but my expectation would be that the following implies that *either* `cloud-platform` *or* `logging.admin` scopes would work for the specified RPCs.

```yaml
      canonical_scopes: |-
        https://www.googleapis.com/auth/logging.admin,
        https://www.googleapis.com/auth/cloud-platform
```

/cc @munangst: can you confirm?

jmuk:
To be clear, `cloud-platform` appears on all three rules, so it needs to be eliminated from all of them if you want. Also please keep in mind that this configuration comes from our internal repository. You need to modify there.

Reviewing them, it's indeed weird to me too; for example, the second section (mostly for retrieving data) has both `cloud-platform.read-only` and `cloud-platform`. I don't know why.

munangst:
@ofrobots Yes, you are correct, an OAuth token with any of the entries in the `canonical_scopes` list will be accepted. You can see this in the documentation for [`google.api.OAuthRequirements`](https://cloud.google.com/service-management/reference/rpc/google.api#google.api.OAuthRequirements), which is the message used for this config.

[...]

PR #2211, opened 2017-04-11
logging-winston: default to minimal scopes

ofrobots:
By default the logging library defaults to all admin, read and write
scopes. For this library we only need to write by default.

Ref: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2209

I will follow up with a similar change for bunyan once this lands.

### DIFF
--- a/packages/logging-winston/test/index.js
+++ b/packages/logging-winston/test/index.js
@@ -18,7 +18,6 @@
 
 var assert = require('assert');
 var extend = require('extend');
-var isSubset = require('is-subset');
 var nodeutil = require('util');
 var proxyquire = require('proxyquire');
 var util = require('@google-cloud/common').util;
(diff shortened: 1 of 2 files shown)


packages/logging-winston/test/index.js
```diff
@@ -119,12 +136,12 @@ describe('logging-winston', function() {
 
       new LoggingWinston(optionsWithLogName);
 
-      assert.strictEqual(fakeLoggingOptions_, optionsWithLogName);
+      assert(isSubset(fakeLoggingOptions_, optionsWithLogName));
```
stephenplusplus: The only thing not cool about this is, if it is not a subset, we don't see a meaningful explanation of that in the npm test output. Can we stick to `deepEqual` to see that the object is in there?

```js
var loggingOptions = extend({}, fakeLoggingOptions_);
delete loggingOptions.scope;

assert.deepEqual(loggingOptions, optionsWithLogName);
```

packages/logging-winston/test/index.js
```diff
@@ -111,6 +112,22 @@ describe('logging-winston', function() {
       });
     });
 
+    it('should default to logging.write scope', function() {
+      assert.deepEqual(fakeLoggingOptions_.scopes,
```
stephenplusplus: Style nit:
```js
assert.deepEqual(fakeLoggingOptions_.scopes, [
  'https://www...'
]);
```

packages/logging-winston/test/index.js
ofrobots: Done.

packages/logging-winston/test/index.js
ofrobots: Done.

Issue #2213, opened 2017-04-11
Incorrectly formatted FQDN's to googleapis.com


ericuldall:
I've noticed all of the references to the Fully Qualified Domain Name "googleapis.com" are incorrectly formatted. Per [RFC 1034](http://www.ietf.org/rfc/rfc1034.txt) FQDN's should have a trailing dot in order to remove any ambiguity and unnecessary lookups.

Explanatory documentation [HERE](http://www.dns-sd.org/trailingdotsindomainnames.html)

We were pointed to this doc by a Kubernetes network engineer at Google after having many DNS lookup issues inside of our GKE clusters. After running some tests we noticed a minimum 100ms improvement per request after adding the trailing dot.

I have a PR ready that I will post for review.

lukesneeringer:
Hi @ericuldall,
Thanks for reporting.

I am not going to make the change suggested in #2214 for the reasons I covered there. However, you can continue to use this in your own implementation by specifying your `baseUrl` directly:

```javascript
new BigTable({ baseUrl: '...' });
```

ericuldall:
I understand your concerns. I would highly suggest putting a warning for people using this library in a kubernetes cluster with a large number of http requests as the internal kube-dns will drop lots of requests without the trailing dot. This is because in the kubernetes resolve.conf they use a large number of ndots to check for internal services.

*resolv.conf sample*:
```
search default.svc.cluster.local svc.cluster.local cluster.local google.internal c.project.internal
nameserver 10.20.30.40
options ndots:5
```

One of the main problems is the setting: `sysctl -w net.netfilter.nf_conntrack_max=65535`
That table is of a finite size per host machine and if it overflows requests get dropped. Using the trailing dot seems to be an easy remedy to avoid all the internal lookups and bypass the problem completely.

Another option is to set the pod dnsPolicy, `DnsPolicy: Default` in the pod spec.

We will move forward creating instances with the overridden baseUrl for our needs. Thanks for looking into this.

lukesneeringer:
Hmm, I understand. Let me think on this more and discuss it and see if my position changes. (It might.)

stephenplusplus:
Due to an issue brought up in #2249, we're going to revert this change.

PR #2214, opened 2017-04-11
updated FQDN's to googleapis.com with a trailing dot

ericuldall:
** edit by @stephenplusplus **

Due to an issue brought up in #2249, we have reverted this change.

---

PR for Issue #2213

### DIFF
--- a/packages/logging/test/index.js
+++ b/packages/logging/test/index.js
@@ -1122,7 +1122,7 @@ describe('Logging', function() {
 
             assert.strictEqual(config, CONFIG);
 
-            var expectedDestination = 'storage.googleapis.com./' + bucket.name;
+            var expectedDestination = 'storage.googleapis.com/' + bucket.name;
             assert.strictEqual(config.destination, expectedDestination);
 
             callback(); // done()
@@ -1237,7 +1237,7 @@ describe('Logging', function() {
           it('should call createSink with string destination', function(done) {
             logging.createSink = function(name, config, callback) {
               var expectedDestination = [
-                'bigquery.googleapis.com.',
+                'bigquery.googleapis.com',
                 'projects',
                 dataset.parent.projectId,
                 'datasets',
@@ -1361,7 +1361,7 @@ describe('Logging', function() {
 
           it('should call createSink with string destination', function(done) {
             logging.createSink = function(name, config, callback) {
-              var expectedDestination = 'pubsub.googleapis.com./' + topic.name;
+              var expectedDestination = 'pubsub.googleapis.com/' + topic.name;
               assert.strictEqual(name, SINK_NAME);
               assert.strictEqual(config, CONFIG);
               assert.strictEqual(config.destination, expectedDestination);


ericuldall:
I caught an issue where scopes cannot have the trailing dot. I'm fixing that now.

ericuldall:
I think I got them all, per `grep -rni 'googleapis\.com\./auth' *` from root of rep

.jscsrc
```diff
@@ -16,7 +16,7 @@
   ],
   "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties",
   "maximumLineLength": {
-    "value": 80,
+    "value": 81,
```
lukesneeringer: We do not want to change this.

packages/bigquery/src/index.js
```diff
@@ -66,7 +66,7 @@ function BigQuery(options) {
   }
 
   var config = {
-    baseUrl: 'https://www.googleapis.com/bigquery/v2',
+    baseUrl: 'https://www.googleapis.com./bigquery/v2',
```
lukesneeringer: I do not think we want to do this either:

```
$ curl -I https://github.com./GoogleCloudPlatform
HTTP/1.1 301 Moved Permanently
Content-length: 0
Location: https://github.com/GoogleCloudPlatform
Connection: close
```

ericuldall: Just for the record, since this is a Google repo:

```
curl -I www.google.com.
HTTP/1.1 200 OK
Date: Wed, 12 Apr 2017 18:35:27 GMT
Expires: -1
Cache-Control: private, max-age=0
Content-Type: text/html; charset=ISO-8859-1
P3P: CP="This is not a P3P policy! See https://www.google.com/support/accounts/answer/151657?hl=en for more info."
Server: gws
X-XSS-Protection: 1; mode=block
X-Frame-Options: SAMEORIGIN
Set-Cookie: NID=101=j6R0GIlDZYi8xxjty0Zd3MBI7vcOJh8TKY64j9ah_JhkQKi7lNQc7j3N_PMK4uSrt1o7fNGLc32wrglaTZpxCYaQa17t86_LyOrLei2V_yg4rEFd8_0az9FtLCKAYu1zYNyBbzsJCJYhWsrnyQ; expires=Thu, 12-Oct-2017 18:35:27 GMT; path=/; domain=.google.com; HttpOnly
Transfer-Encoding: chunked
Accept-Ranges: none
Vary: Accept-Encoding
```

At least someone is doing it right ;)

ericuldall: I also disagree that the benefit is tiny, in our case. We see a large number of dropped requests due to this issue. See my response in the issue #2213 for more details on how it's affecting our organization.

lukesneeringer:
I discussed this with our team internally and have been persuaded. I do want you to change the maximum line length back to 80, though.

ericuldall:
Sure, I can certainly move that back to 80 chars. What is the suggested style choice for the baseUrl lines that have reached 81 chars in the line?

ericuldall:
Will something like this be acceptable:

```
  ¦ it('should set the default base URL', function() {
  ¦ ¦ assert.strictEqual(
  ¦ ¦ ¦ datastore.defaultBaseUrl_,
  ¦ ¦ ¦ 'datastore.googleapis.com.'
  ¦ ¦ );
  ¦ });
```

stephenplusplus:
Regarding style, whatever you think works best for the block the code is in, including renaming vars to be shorter, or:

```js
var url = [
  'http://www.googleapis.com./...',
  'other-part'
].join('/');
```

Anything you come up I'm sure will be just fine, and thanks very much for asking!

> Will something like this be acceptable:

Yep, that's perfect 👍

ericuldall:
Okay, changes made and rebased from upstream. Let me know if anything else needs to change.

stephenplusplus:
Just assigning to myself -- running the system tests locally had a couple failures, so I'll need to comb over and make sure we've got the dots in the right spots.

lukesneeringer:
@stephenplusplus No need to do an entire umbrella release for this, but maybe release BigTable for @ericuldall. The rest of them can just show up when we release for other reasons.

stephenplusplus:
I stacked on a commit which removes the changes from areas that weren't working as expected and in the files that are generated-- those will have to be edited upstream.

@landrito @jmuk in the generated files, would you be interested in taking on this change?

ericuldall:
Just for the record we're not using Big Table. Our most used api's are Big Query and PubSub.

stephenplusplus:
Due to an issue brought up in #2249, we have reverted this change.

Issue #2231, opened 2017-04-18
Google Speach API- @google-cloud/speech@0.9.0 failing on specifying the SpeechContext in the speech request config


nandums:
Speech API is failing on specifying the SpeechContext in the speech request config, see the below error

```
Error: .google.cloud.speech.v1.RecognitionConfig#speechContext is not a field: undefined
at Error (native)
at MessagePrototype.set (C:\test-google-asr\node_modules\protobufjs\dist\protobuf.js:2490:35)
at MessagePrototype.set (C:\test-google-asr\node_modules\protobufjs\dist\protobuf.js:2483:38)
at Message (C:\test-google-asr\node_modules\protobufjs\dist\protobuf.js:2411:34)
at Element.ProtoBuf.Reflect.ElementPrototype.verifyValue (C:\test-google-asr\node_modules\protobufjs\dist\protobuf.js:1925:28)
at T.ProtoBuf.Reflect.FieldPrototype.verifyValue (C:\test-google-asr\node_modules\protobufjs\dist\protobuf.js:3499:33)
at MessagePrototype.set (C:\test-google-asr\node_modules\protobufjs\dist\protobuf.js:2493:59)
at MessagePrototype.set (C:\test-google-asr\node_modules\protobufjs\dist\protobuf.js:2483:38)
at Message (C:\test-google-asr\node_modules\protobufjs\dist\protobuf.js:2411:34)
at Element.ProtoBuf.Reflect.ElementPrototype.verifyValue (C:\test-google-asr\node_modules\protobufjs\dist\protobuf.js:1925:28)
```

#### Environment details

  - OS: Windows 7
  - Node.js version: v6.7.0
  - npm version: 3.10.3
  - google-cloud/speech: 0.9.0

#### Steps to reproduce

Run the example code with the below configuration
```
speechClient.createRecognizeStream({
	config: {
		encoding: 'LINEAR16',
		//sampleRate: 16000,
		sampleRateHertz: 16000,
		maxAlternatives: 2,
		languageCode:"en-US",
		speechContext: {
			phrases: [
				"testing"
			]
		}
	},
	singleUtterance: false,
	interimResults: false
})
```

stephenplusplus:
Try `speechContexts`. Did you find this sample code somewhere?

stephenplusplus:
See: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2036#issuecomment-293898330

[...]

PR #2232, opened 2017-04-18
pubsub: add ability to pass timeout to publish

### LINKED ISSUE #2230: Pub/Sub Deadline Exceeded
#### Environment details

  - OS: Docker in GKE
  - Node.js version: 6.9
  - google-cloud-node version: "google-cloud": "^0.52.0",

#### Steps to reproduce

We are publishing messages on a number of topics, We hade memory leak problems if calling publish every time we got a message with would put it at around 1000 times per second. So we implemented a batching where we publish 225 messages in every batch. Is do solve the memory problem but we have started to get `Error: Deadline Exceeded` with the later versions.
```
{
code: 4, 
metadata: {
_internal_repr: {}
}, 
note: Exception occurred in retry method that was not classified as transient
}
```

```
/usr/src/app/node_modules/grpc/src/node/src/client.js in null.<anonymous> at line 434:17
              code: grpc.status.INTERNAL,
              details: 'Failed to parse server response'
            };
          }
        }
      }
      if (status.code !== grpc.status.OK) {
        error = new Error(status.details);
        error.code = status.code;
        error.metadata = status.metadata;
        args.callback(error);
      } else {
        args.callback(null, deserialized);
      }
      emitter.emit('status', status);
```

Cant figure out way its started to be thrown more often only change is the version. 

Let me know if we can provide anymore information to get this resolved.

Thanks!

dennismartensson:
Fixes #2230 

Is based on #2042 to bring the ability to pass timeout to publish aswel.

Please have a look!

### DIFF
--- a/packages/pubsub/src/topic.js
+++ b/packages/pubsub/src/topic.js
@@ -375,6 +375,8 @@ Topic.prototype.getSubscriptionsStream = function(options) {
  * @param {object=} options - Configuration object.
  * @param {boolean} options.raw - Enable if you require setting attributes on
  *     your messages.
+ * @param {number} options.timeout - Set a maximum amount of time in
+ *     milliseconds before giving up if no response is received.
  * @param {function=} callback - The callback function.
  *
  * @example
@@ -435,6 +437,7 @@ Topic.prototype.getSubscriptionsStream = function(options) {
  *   var messageIds = data[0];
  *   var apiResponse = data[1];
  * });
+ *
  */
 Topic.prototype.publish = function(messages, options, callback) {
   messages = arrify(messages);


packages/pubsub/src/topic.js
```diff
@@ -456,6 +456,10 @@ Topic.prototype.publish = function(messages, options, callback) {
     method: 'publish',
   };
 
+  if (options && is.number(options.timeout)) {
```
stephenplusplus: Just if `is.number` works here, because we know `options` is defaulted to an object above.

dennismartensson:
I signed it!

dennismartensson:
@stephenplusplus I added a test like the one ju linked to.

AUTHORS
```diff
@@ -11,3 +11,6 @@ Anand Suresh
 Brett Bergmann
 Jesse Friedman
 Zach Bjornson <zbbjornson@gmail.com>
+
+Greta.io
+Dennis Mårtensson <dennis@greta.io>
```
stephenplusplus: @lukesneeringer is this the way to mark up a company contribution?

stephenplusplus:
Looks awesome to me, thank you very much!

AUTHORS
dennismartensson: This is the instructions:

```
# Names should be added to this file as:
# Name or Organization <email address>
# The email address is not required for organizations
```
So I read it as adding the org and then the person in that org.

dennismartensson:
@stephenplusplus How long before you think could be out so we can start mitigating the timeouts?

stephenplusplus:
@dennismartensson I can merge and release as soon as @lukesneeringer can confirm the edit to the AUHTORS file is okay.

dennismartensson:
@stephenplusplus Thank you!

dennismartensson:
@stephenplusplus do you know what the default timeout value is?

packages/pubsub/src/topic.js
```diff
@@ -456,6 +456,10 @@ Topic.prototype.publish = function(messages, options, callback) {
     method: 'publish',
   };
 
+  if (is.number(options.timeout)) {
+    protoOpts.timeout = options.timeout;
```
stephenplusplus: Oh, one more thing I forgot about, sorry. Can you add a line to the JSDocs above this method that show we accept this option?

```js
 * @param {number} options.timeout - Set a maximum amount of time in
 *     milliseconds before giving up if no response is received.
```

stephenplusplus:
> do you know what the default timeout value is?

I'm not sure, this would be a default value that [grpc](http://www.gitnpm.com/grpc) sets.

Also, have you tested that this change solves the problem in #2231?

dennismartensson:
Okey, I have not tested for #2231 don't get how that is related? (might be a miss type) but looks like it can be used to solve #2230 throwing to early timeouts.

packages/pubsub/src/topic.js
dennismartensson: Will have a look at this

stephenplusplus:
Yes, it was a mis-type, sorry about that. I'm wondering if we should also default the timeout to a reasonable value, so users don't have to think about it.

packages/pubsub/src/topic.js
dennismartensson: Just pushed this

dennismartensson:
That might be a good idea but what should the default value be?

stephenplusplus:
I'm not sure, what have you had luck with?

dennismartensson:
I have only tried the default and 10 min just to make sure it went away.

dennismartensson:
Found this:
```
// The default timeout in seconds for RPCs sent to this method. This can
      // be overridden in code. If no reply is received in the specified amount
      // of time, the request is aborted and a deadline-exceeded error status
      // is returned to the caller.
      //
      // The actual deadline used will be the minimum of the value specified
      // here and the value set by the application via the gRPC client API.
      // If either one is not set, then the other will be used.
      // If neither is set, then the request has no deadline.
``` [Here](https://github.com/grpc/grpc/blob/master/doc/service_config.md)

So I guess that means it could be set by the Google cloud pub/sub endpoint?  With creates that error and maybe then default today is to follow the server?

stephenplusplus:
Interesting. I think we should still override it. I can try to play with it later, but if you beat me to it, we want the lowest value that is consistently working. I would start at 10ms, go to 100ms, 1000ms, etc.

AUTHORS
lukesneeringer: I all kinds of do not care. As long as it is readable, I am fine with it.

dennismartensson:
I think it depends a bit on many messages are posted in one go. We we where posting just one message every time the feeling is we did not have as many of this error as we do now. So should we tune it for one message or many?

My quick tests points to 225 messages around 500byte per message 10sec looks okey, If I do 450 message per publish we seam to need a higher time. But do not feel like a well thought out strategy.

Have you had the time to run any tests?

stephenplusplus:
This is good for now! I forgot we have a bit of a cheat sheet: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/pubsub/src/v1/publisher_client_config.json

I'll go through and see that we're setting these across the API in a follow-on PR. I'll release today so you can start using (at least) the change from this PR. Thank you again for doing this!

dennismartensson:
Thank you @stephenplusplus for helping getting this out. The #2242 looks interesting!

PR #2234, opened 2017-04-18
introduce common-gax

stephenplusplus:
We currently have:

- @google-cloud/common -- Core functionality and base classes for Services and ServiceObjects
- @google-cloud/common-grpc -- Base classes that extend common.Service & common.ServiceObject

And as we shift to more GAX APIs (#1859), a new common library will be necessary to handle some plumbing while interacting with GAX, such as:

- Instantiating clients
- Caching/re-using clients
- Handling automatic project ID insertion
- Preparing request streams

### DIFF
--- a/packages/common-gax/src/service.js
+++ b/packages/common-gax/src/service.js
@@ -38,10 +38,6 @@ var util = require('util');
  * @param {object} options - [Configuration object](#/docs/?method=gcloud).
  */
 function GaxService(config, options) {
-  if (global.GCLOUD_SANDBOX_ENV) {
-    return;
-  }
-
   config = extend({
     scopes: config.module.ALL_SCOPES
   }, config);


lukesneeringer:
Let's talk about this. I am not sure this is an ideal solution.

In particular, I do not really understand why `gax-nodejs` and `common-gax` both need to exist. Probably we should bundle the entirety of the common gax tooling in one of them.

stephenplusplus:
GAX doing auto-project ID insertion would be great. That means:

- detect the project ID from the environment
- iterate over all user-supplied request options and replace a placeholder value with it 

That would remove the need for 'request' and 'requestStream' from this API, as well as the need to lazily instantiate clients... so yeah, everything.

cc @jmuk @landrito

stephenplusplus:
@lukesneeringer #1859 (moving from our in-house GrpcService to GAX) has two choices:

- go forward, but duplicate the code and logic that this PR was going to squash
- use common-gax
- wait for google-gax to implement auto-project ID insertion

Can we prioritize *a decision* on this, so we can still move on with #1859 in some capacity?

packages/common-gax/src/service.js
```diff
@@ -0,0 +1,172 @@
...
+ * @param {function} config.method - The gax method to call.
+ * @param {object} config.reqOpts - Request options.
+ * @param {function=} callback - The callback function.
+ */
+GaxService.prototype.request = function(config, callback) {
+  if (global.GCLOUD_SANDBOX_ENV) {
+    return;
+  }
+
+  this.prepareRequest_(config, function(err, requestFn) {
```
callmehiphop: I am a little lost on why both `request` and `prepareRequest_` need to exist, it seems like `request` doesn't do much more than call `prepareRequest_` which makes me think we could probably just ditch this method and rename `prepareRequest_` to `request`.

stephenplusplus: It's because requestStream also needs a "prepared" request. To define prepared:

- Instantiates & caches GAX client
- Auto-inserts the project ID to the request options

callmehiphop: Nvm! Just noticed how it's used in streaming D:

stephenplusplus:
Leaving the `do not merge` label per @lukesneeringer's comment: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2234#issuecomment-294983483

lukesneeringer:
> Can we prioritize a decision on this, so we can still move on with #1859 in some capacity?

Let's discuss it in our Monday meeting. It is Friday, so that seems like soon enough if that is okay.

You have already done significant work in this direction, so I might go ahead and let it ride, but I will want a plan for merging the two libraries.

stephenplusplus:
We're going a different way (#2272).

PR #2235, opened 2017-04-18
logging: use common-gax

stephenplusplus:
#### Blockers

- [ ] Introduce @google-cloud/common-gax (#2234)

This converts the Logging API to use common-gax.


stephenplusplus:
We're going a different way (#2272).

PR #2237, opened 2017-04-18
Use WARN as the default `error-reporting` logging level

### LINKED ISSUE #2236: The default log level for the `error-reporting` library is not set to WARN
#### Environment details

  - OS: OS X
  - Node.js version: v6.9.4
  - npm version: 3.10.10
  - google-cloud-node version: @google-cloud/error-reporting@0.1.0

#### Steps to reproduce

Provided the `NODE_ENV` environment variable is not set to `production`, the following code should print a message to standard out stating that the `NODE_ENV` environment variable needs to be set to `production` or the `ignoreEnvironmentCheck` needs to be set to `true`.  However, nothing is printed to standard out.

```
var errors = require('@google-cloud/error-reporting')({
  serviceContext: {service: 'my-service'}
});
errors.report(new Error('Something broke!'));
```

DominicKramer:
Fixes #2236:  If the logging level is not specified when configuring the `error-reporting` library, warning messages are not displayed to the user.

### DIFF
--- a/packages/error-reporting/test/unit/testLogger.js
+++ b/packages/error-reporting/test/unit/testLogger.js
@@ -68,13 +68,13 @@ describe('logger', function() {
         text = undefined;
         console.log = oldLog;
       });
-      it('Should be able to WARN by default', function() {
+      it('Should print WARN logs by default', function() {
         var logger = createLogger();
         logger.warn('test warning message');
         assert.strictEqual(text,
           'WARN:@google/cloud-errors: test warning message');
       });
-      it('Should be able to ERROR by default', function() {
+      it('Should print ERROR logs by default', function() {
         var logger = createLogger();
         logger.error('test error message');
         assert.strictEqual(text,


packages/error-reporting/test/unit/testLogger.js
```diff
@@ -52,9 +52,33 @@ describe('logger', function() {
...
+        oldLog = console.log;
+        text = '';
+        console.log = function() {
+          oldLog.apply(this, arguments);
+          for (var i=0; i<arguments.length; i++) {
+            text += arguments[i];
+          }
+        };
+      });
+     afterEach(function() {
```
stephenplusplus: It looks like this is not indented the same as the other lines. I wonder if our linter is ignoring this file intentionally.

packages/error-reporting/test/unit/testLogger.js
```diff
@@ -52,9 +52,33 @@ describe('logger', function() {
...
+        text = undefined;
+        console.log = oldLog;
+      });
       it('Should be able to WARN by default', function() {
         var logger = createLogger();
         logger.warn('test warning message');
+        assert.strictEqual(text,
+          'WARN:@google/cloud-errors: test warning message');
+      });
+      it('Should be able to ERROR by default', function() {
```
ofrobots: nit: the test name can be improved.. how about 'Should print ERROR logs by default'. Likewise for the `WARN` message above.

DominicKramer: Fixed

PR #2240, opened 2017-04-19
pubsub: implement message retention

stephenplusplus:
#### Blockers

- [x] [Doesn't work](https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2240#issuecomment-295353404)

#### To Dos

- [x] Remove expiration concept and only accept a number to represent seconds.

This adds support for two new options when creating a subscription:

- **options.retainAckedMessages** - If set, acked messages are retained in the subscription's backlog for 7 days (unless overriden by `options.messageRetentionDuration`). Default: `false`
- **options.messageRetentionDuration** - Set this to override the default duration of 7 days. This value is expected in seconds (length of retention) or a Date (expiration of retention). Acceptable values are in the range of 10 minutes and 7 days.

Official docs via the proto files: https://github.com/googleapis/googleapis/blob/c75ebb5719066289f91d6b8a1f5372d484ac4f75/google/pubsub/v1/pubsub.proto#L358-L370

### DIFF
--- a/packages/pubsub/system-test/pubsub.js
+++ b/packages/pubsub/system-test/pubsub.js
@@ -18,7 +18,6 @@
 
 var assert = require('assert');
 var async = require('async');
-var moment = require('moment');
 var Subscription = require('../src/subscription.js');
 var uuid = require('uuid');
 


lukesneeringer:
@pongad Is this behind a whitelist? (I am not sure why this does not work for @stephenplusplus.)

pongad:
@stephenplusplus @lukesneeringer I don't think the whitelist is the problem. If the project is not white listed, you should not see the last two fields in the returned message.

stephenplusplus:
I haven't tested against the real system since I opened the PR, but I would assume if nothing has changed on the backend, we should see the same result. The CI for PRs only run unit tests, which mock the back-end API.

Is anything about the request incorrect? Are there permissions for my service account that might affect the result? Maybe something about the topic needs to be configured to allow its subscriptions to be customized in this way?

pongad:
I don't see anything wrong with the request. Perhaps @amordkov can shed some light?

stephenplusplus:
Just tested against the real system, and sadly the same result.

packages/pubsub/src/index.js
```diff
@@ -553,8 +553,15 @@ PubSub.prototype.getTopicsStream = common.paginator.streamify('getTopics');
  *     messages. (default: 10)
  * @param {number} options.maxInProgress - Maximum messages to consume
  *     simultaneously.
+ * @param {number|date} options.messageRetentionDuration - Set this to override
+ *     the default duration of 7 days. This value is expected in seconds (length
+ *     of retention) or a Date (expiration of retention). Acceptable values are
```
pongad: Thank you Alex for catching this. I don't think this makes sense.

The duration is from the moment each message is published, so different messages will have different "expiration dates". The doc seems to imply that all messages will expire at the same time which is not the case.

stephenplusplus: Thanks for helping me understand. I will remove this functionality and only accept a number to represent seconds.

Issue #2131, opened 2017-03-23
Deprecate google-cloud metapackage.


lukesneeringer:
We are deprecating the `google-cloud` metapackage.

This entails:

  * [x] updating our documentation to discourage it (#2119)
  * [x] deprecating it at npm

lukesneeringer:
Now if you `npm install google-cloud`:

```
lukesneeringer@lukesneeringer0:~/Code/Google/test$ npm install google-cloud
npm WARN deprecated google-cloud@0.49.0: The google-cloud meta-package is not recommended; install individual packages.
```

stephenplusplus:
Are we going to keep releasing `google-cloud`s, or nuh uh?

lukesneeringer:
For the time being.

jgeewax:
The docs still say that this is deprecated... Can we fix that?

(https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/v0.55.0/language)

stephenplusplus:
`google-cloud` is deprecated on npm. What do we want it to say?

jgeewax:
Then why do we have that as the docs on the main page ? E.g. if I click
into reading the docs, I'm brought to those.... I thought this ticket said
we weren't going to deprecate ... Did I misread?

stephenplusplus:
Yeah, the package is deprecated, but @lukesneeringer says we still want to make new releases of it. I would prefer we stop making releases, since we are discouraging its use. But as it stands now, `require('google-cloud')` is still an accurate way to use the module, if you are looking for the all-in-one package.

stephenplusplus:
Our last release of `google-cloud` was 3 months ago. We've shut down the gh-pages docs site, so I don't believe we are advertising `npm install google-cloud` anywhere anymore.

PR #2241, opened 2017-04-19
Do not claim that the metapackage will cease being updated.

lukesneeringer:
We need to continue to maintain the metapackage, even if we want to discourage its use. Therefore, we need to remove the line from the README that claims that it will cease being updated.


stephenplusplus:
RE: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2131#issuecomment-288824990

PR #2246, opened 2017-04-21
error-reporting: Remove Credential Validation Checks

### LINKED ISSUE #2245: `error-reporting` Only Supports Authentication Using a Service Account Key
The `@google-cloud/error-reporting@0.1.0` package does not work locally or on App Engine Flexible Environment if a service account key is not specified.  That is, neither automatic nor manually specified errors are reported in the Error Reporting console.

#### Environment details

  - OS: OS X 10.12.4
  - Node.js version: v6.9.4
  - npm version: 3.10.10
  - google-cloud-node version: HEAD

#### Steps to reproduce

  1. Create an app that uses the `error-reporting` library and reports errors.
  2. Enable the Error Reporting API for the current project.
  3. Use `gcloud beta auth application-default login` to authenticate on a development machine with `gcloud` initialized to use the project with the Error Reporting API enabled.
  4. Run `NODE_ENV=production GCLOUD_PROJECT=<projectId> npm start`.

#### Actual Results

Errors are not reported in the Error Reporting console.

#### Expected Results

Errors should be reported in the Error Reporting console.

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

DominicKramer:
Fixes #2245 - `error-reporting` Only Supports Authentication
Using a Service Account Key.

### DIFF
--- a/packages/error-reporting/src/configuration.js
+++ b/packages/error-reporting/src/configuration.js
@@ -22,8 +22,6 @@ var isObject = is.object;
 var isBoolean = is.boolean;
 var isString = is.string;
 var isNumber = is.number;
-var isEmpty = is.empty;
-var isNull = is.null;
 var version = require('../package.json').version;
 
 /**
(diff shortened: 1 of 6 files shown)


ofrobots:
For now, let's prefer @stephenplusplus land things.

PR #2252, opened 2017-04-24
error-reporting: Warn early if cannot access API

DominicKramer:
An attempt will be made to communicate with the API as soon as
the error reporting library is loaded so that the user will be
notified on application start if the error reporting library
cannot communicate with the API.

Without this change, the user will instead receive a warning
only when the error reporting library attempts to report its first
error.

### DIFF
--- a/packages/error-reporting/src/google-apis/auth-client.js
+++ b/packages/error-reporting/src/google-apis/auth-client.js
@@ -101,9 +101,9 @@ class RequestHandler extends common.Service {
     var that = this;
     this.authClient.getToken(function(err, token) {
       if (err) {
-        that._logger.warn([
-          'Unable to find credential information on instance. This library ' +
-          'will be unable to communicate with the Stackdriver API to save ' +
+        that._logger.error([
+          'Unable to find credential information on instance. This library',
+          'will be unable to communicate with the Stackdriver API to save',
           'errors.  Message: ' + err.message
         ].join(' '));
       }
--- a/packages/error-reporting/test/unit/testAuthClient.js
+++ b/packages/error-reporting/test/unit/testAuthClient.js
@@ -20,7 +20,7 @@ var proxyquire = require('proxyquire');
 
 var Configuration = require('../../src/configuration.js');
 
-function verifyWarningMessage(errToReturn, expectedWarn) {
+function verifyReportedMessage(errToReturn, expectedMessage) {
   class ServiceStub {
     constructor() {
       this.authClient = {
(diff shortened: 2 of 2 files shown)


packages/error-reporting/src/google-apis/auth-client.js
```diff
@@ -97,6 +97,17 @@ class RequestHandler extends common.Service {
...
     this._config = config;
     this._logger = logger;
+
+    var that = this;
+    this.authClient.getToken(function(err, token) {
+      if (err) {
+        that._logger.warn([
+          'Unable to find credential information on instance. This library ' +
+          'will be unable to communicate with the Stackdriver API to save ' +
+          'errors.  Message: ' + err.message
```
kjin: I think you should either use `+` or `join(' ')` to concatenate these strings, but not both.
Additional nit: leave single space after period on this line.

packages/error-reporting/test/unit/testAuthClient.js
```diff
@@ -0,0 +1,64 @@
...
+
+  var warnText = '';
+  var logger = {
+    warn: function(text) {
+      warnText += text;
+    }
+  };
+  var config = new Configuration({ ignoreEnvironmentCheck: true }, logger);
+  new RequestHandler(config, logger);
+  assert.strictEqual(warnText, expectedWarn);
```
kjin: Maybe you just want to make sure that warnText contains expectedWarn instead of being strictly equal?

packages/error-reporting/src/google-apis/auth-client.js
```diff
@@ -97,6 +97,17 @@ class RequestHandler extends common.Service {
     }, config);
     this._config = config;
     this._logger = logger;
+
+    var that = this;
+    this.authClient.getToken(function(err, token) {
+      if (err) {
+        that._logger.warn([
```
matthewloring: By the time we've gotten here, I think we've determined we will never be able to successfully publish errors. If that is the case, it would be good to log this at the most severe level.

packages/error-reporting/src/google-apis/auth-client.js
DominicKramer: Now only `join(' ')` is used to match the rest of the file.

packages/error-reporting/test/unit/testAuthClient.js
DominicKramer: I think being strictly equal shouldn't make the test too brittle.

packages/error-reporting/src/google-apis/auth-client.js
DominicKramer: I've updated it to use `error()` instead of `warn()`.

Issue #2254, opened 2017-04-25
Should retry on ECONNRESET


inlined:
*edit by @stephenplusplus*

## If you're just joining the conversation...

Here is where we stand: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2254#issuecomment-299609092

Original Post
---

#### Environment details

  - OS: Google Cloud Functions (not an OS. I know)
  - Node.js version: LTS
  - npm version: LTS
  - google-cloud-node version: HEAD

#### Steps to reproduce

  1. git clone git@github.com:firebase/functions-samples.git
  2. cd generate-thumbnail
  3. firebase init
  4. firebase deploy
  5. play with app for a while and observe that it gets into a mode that always fails with ECONNRESET

#### Theory

I've noticed this in my own Cloud Functions and only in those that use google-cloud-node. I think the socket is dying between requests when the Cloud Functions environment is throttled. This should be OK, but I suspect the google-cloud-node module needs to add an extra error case [here](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/common/src/util.js#L279). I'll write a pull request to that effect.

stephenplusplus:
_Continuing from https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2255#issuecomment-299329661_ (cc @s2young)

I'm not able to get the process mentioned in this issue to fail. Does anyone have a fool proof way I can get this script or another to break? I've been uploading 24 images at a time and the thumbnails are all generated successfully.

s2young:
My app is doing quite a bit more than the example, and I'm stripping some of this out to see if I can avoid the problem. But all is just standard Storage api stuff:

1.  Download the source file to tmp directory.
2. Call getMetadata (so I can get the byte size)- this I'm stripping out.
3. Resize the file (3 different sizes for different screens).
4. Upload the resized images to Storage.
5. Delete the source file.

s2young:
One more caveat @stephenplusplus:  If you follow the pattern I've given, make sure you ignore the upload of the copies.  In my case, when I upload a resized copy I am adding some metadata to the image so that the imageListener function can ignore those storage events.  I got myself into an infinite loop early on!

stephenplusplus:
[...]

PR #2255, opened 2017-04-25
Add retries in case of a socket hangup on HTTP

inlined:
Fixes https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2254

Should hopefully catch the issue customers are seeing in Google Cloud Functions where google APIs get into a permanent failed state.

### DIFF
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -293,9 +293,9 @@ function shouldRetryRequest(err) {
         }
       }
     }
-    
+
     // Underlying socket throws "Error: read ECONNRESET" for a TCP hangup.
-    if (err.toString().indexOf('ECONNRESET' !== -1) {
+    if (err.toString().indexOf('ECONNRESET' !== -1)) {
       return true;
     }
   }


stephenplusplus:
Thanks for this! I can't think of a reason why we shouldn't retry on this type of error, but before we merge, I just want to make sure we're consistent between our different API request transport layers (we have 3):

- `common.Service` (REST): (covered in this PR)
- `commonGrpc.Service` (gRPC): We would want to find a similar solution in this function: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/40abb0b0fc3868900cd342c84d13069c128cc5a5/packages/common-grpc/src/service.js#L602 -- AFAIK, gRPC itself does not do any retries.
- GAX: @jmuk @landrito - does GAX do any retries for this error?

And one more thought, since we're re-thinking this-- are there any other similar errors we should retry on?

s2young:
Is there any way to npm install @google-cloud/storage at this commit?  I cannot figure out how to get this into my firebase functions project.   Any help would be greatly appreciated!  I've tried every recommended syntax on the google machine and nothing is working.

stephenplusplus:
Sorry @s2young, there isn't a way to use this with npm due to our mono-repo for multiple packages.

cc @swcloud @lukesneeringer -- any thoughts on if we should retry on this error, and if there are any others?

s2young:
This module is pretty much unusable on Firebase Functions without some change.  I can confirm that after the timeout period the only way to get my functions working is to redeploy them.  Thanks!

s2young:
Sorry to pester, but this issue is REALLY a problem for me.  What about a way to programmatically reconnect?  A setting I can pass in when instantiating the @google-cloud/storage object? Anything?!

stephenplusplus:
The solution for this goes a bit deeper into a dependency module, `retry-request`. That is the module that the changes in this PR affect. It will take a custom "shouldRetryRequest" function, pass into it the HTTP response object, and use the returned value to determine if it should retry the request.

There is a bit of complexity to how this is solved. Currently, `retry-request` immediately errors out on ones such as "ECONNRESET" or other low-level HTTP/DNS errors ([here](https://github.com/stephenplusplus/retry-request/blob/576627f98be5c1991e7db17f11a60fc35974196a/index.js#L94)). It won't even give the user the chance to retry on that. If we want to support retrying from these types of errors, that gives us 2 options:

1) Return the error to the `shouldRetryRequest` function, allowing the user to specify they want retries on ECONNRESET errors, or
2) Automatically retry using an exponential backoff strategy on these low-level errors, considering it a "probably what you want" best practice. These retries would not count against the `maxRetries` value the user provides.

Any thoughts are welcome.

stephenplusplus:
cc @callmehiphop

s2young:
My personal vote is for number 2 (and soon).  I can continue developing the solution I'm working on, but cannot launch anything until this is resolved.  Thanks!

s2young:
With respect to option 1, why would a user not want to retry?  The user is interfacing with a method that is supposed to upload/setMetadata/getMetadata/etc.   What about going with option 2, and rejecting the promise after a user-configurable timeout?  A timeout option could give the user flexibility enough to move on with their program, while having the nature of the failure in the rejection.

stephenplusplus:
I prefer that option as well. I put out a release with [this change](https://github.com/stephenplusplus/retry-request/commit/e25c836835490d7c8187482aa7715ffeaf35d198). If you un- and re-install @google-cloud/storage, you should have the new code. Can you let me know if it works?

stephenplusplus:
I'm going to consider this resolved after the fix mentioned above. Just let me know if it's not working, and we'll figure out a new way to attack this. Thanks @inlined and @s2young for reporting the issue and your patience while we figure(d) this out.

s2young:
So far so good!  Thanks so much!

s2young:
Ok, maybe I spoke too soon.  It's definitely better but I'm still seeing the ECONNRESET.  My scenario is the following:

1. Client application uploads files to Firebase Storage.
2. Firebase Function listens for changes to Firebase Storage, downloads and resizes images where needed, and updates Firebase real-time db with some data about the uploaded files.
3. Client monitors changes to Firebase Db to figure out when all the processing is complete.

In my testing I'm uploading 16 images, and sometimes it works for all 16.  But equally often, I'll only get 14 or 15 out of 16 properly updated in the Db.

stephenplusplus:
Thanks for letting me know. The change to retry-request allows 3 total failures. Upping that number might be a naive solution, so I'm going to start with what was described in #2254, and modify as necessary to try to reproduce and hopefully become more aware of the underlying issue.

stephenplusplus:
Carried the conversation over to #2254.

PR #2256, opened 2017-04-25
Prevent decoding and then re-encoding to base64

foxdavidj:
Currently when `encoding === 'base64'`, node decodes and then re-encodes back to `base64`.

This PR prevents this redundant conversion.

### DIFF
--- a/packages/pubsub/src/subscription.js
+++ b/packages/pubsub/src/subscription.js
@@ -347,7 +347,7 @@ modelo.inherits(Subscription, commonGrpc.ServiceObject, events.EventEmitter);
  *
  * @private
  */
-Subscription.formatMessage_ = function(msg, encoding) {
+Subscription.formatMessage_ = function(msg, enc) {
   var innerMessage = msg.message;
   var message = {
     ackId: msg.ackId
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Thank you for this! Looks like we'll need work on the unit tests to cover using `base64`: https://coveralls.io/builds/11234899/source?filename=packages%2Fpubsub%2Fsrc%2Fsubscription.js#L362

It should be possible to just change the `binary` here to `base64`: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/40abb0b0fc3868900cd342c84d13069c128cc5a5/packages/pubsub/test/subscription.js#L345

obto:
1. Not sure how you'd test that portion of the code. Any ideas?

2. You could use `binary`, but its not a no-op (does the same as `latin1` encoding)

stephenplusplus:
I think all it will take is changing `binary` to `base64` in this test: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/40abb0b0fc3868900cd342c84d13069c128cc5a5/packages/pubsub/test/subscription.js#L345

It's kind of a trick to get the line coverage back up, since we are currently not specifically testing `base64` (before this PR, there was no special handling of `base64` in our code). The point of the test linked above is to ensure that if I want "x" encoding, I'll get back "x" encoding, so it's a good place to replace `binary` with `base64`.

obto:
Sorry, I misunderstood what you meant previously. Thought you were proposing this special handling wasn't needed and to just use `binary` encoding instead haha.

I've updated the test as you proposed.

obto:
Looks like it may be best to add another test just for base64

stephenplusplus:
Thanks @obto!

PR #2258, opened 2017-04-26
Add Google Drive scope to BigQuery

abetomo:
You can set up Google spreadsheets for BigQuery data sources.
However, executing the query resulted in an authorization error.

The following error occurred.
```
  code: 403,
  errors: 
   [ { domain: 'global',
       reason: 'accessDenied',
       message: 'Access Denied: BigQuery BigQuery: Permission denied while globbing file pattern.',
       locationType: 'other',
       location: '/gdrive/id/****' } ],
  response: undefined,
  message: 'Access Denied: BigQuery BigQuery: Permission denied while globbing file pattern.' }
```

### DIFF
--- a/packages/bigquery/src/index.js
+++ b/packages/bigquery/src/index.js
@@ -69,7 +69,7 @@ function BigQuery(options) {
     baseUrl: 'https://www.googleapis.com/bigquery/v2',
     scopes: [
       'https://www.googleapis.com/auth/bigquery',
-      'https://www.googleapis.com/auth/drive'
+      'https://www.googleapis.com/auth/drive.readonly'
     ],
     packageJson: require('../package.json')
   };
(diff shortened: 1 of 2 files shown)


abetomo:
I signed it!

stephenplusplus:
Thanks for this PR! @lukesneeringer what do you think about this?

stephenplusplus:
@swcloud for thoughts as well.

abetomo:
@stephenplusplus @lukesneeringer @swcloud 
How about this PR?
If there is a problem I will fix it so please point out.

fujio0000:
@lukesneeringer
I am having the same problem.
When is the merge around?

abetomo:
@lukesneeringer @swcloud
Currently, it is used frequently with Google spreadsheet as a data source.
Thank you for your review.

ifokeev:
The same issue here.
UP

c0b:
is there a drive.read_only role? for the permission concern, I see bigquery can only read from drive but cannot / no need to write to gdrive?

stephenplusplus:
The BigQuery code is now at https://github.com/googleapis/nodejs-bigquery. Thank you for this PR -- I'm not sure what the right thing to do is, so I can't merge or encourage you to resend the PR, since I'm not sure it will be merged there. Pinging @lukesneeringer again for input.

PR #2264, opened 2017-04-27
error-reporting: Add `errors.report` system test

DominicKramer:
This change adds a system test that verifies that the
`errors.report` method correctly reports an error to the error
reporting console.

### DIFF
--- a/packages/error-reporting/system-test/testAuthClient.js
+++ b/packages/error-reporting/system-test/testAuthClient.js
@@ -391,6 +391,11 @@ describe('error-reporting', function() {
   });
 
   it('Should correctly publish errors', function(done) {
+    // After an error is reported, this test waits TIMEOUT ms before
+    // verifying the error has been reported to ensure the system had
+    // enough time to receive the error report and process it.
+    // As such, this test is set to fail due to a timeout only if sufficiently
+    // more than TIMEOUT ms has elapsed to avoid test fragility.
     this.timeout(TIMEOUT * 2);
     var errorId = buildName('message');
     errors.report(new Error(errorId), function(err, response, body) {


packages/error-reporting/utils/errors-api-transport.js
```diff
@@ -0,0 +1,75 @@
...
+    self.getProjectId(function(err, id) {
+      if (err) {
+        return cb(err);
+      }
+
+      var options = {
+        uri: [API, id, 'events'].join('/'),
+        method: 'DELETE'
+      };
+      self.request_(options, function(err, /* jshint unused:false */ response,
```
stephenplusplus: Couldn't you just not name the parameters `response` and `body` to get around JSHint?

DominicKramer: My personal preference is to leave the parameters in the function for completeness but mark them as not being used to make it explicit that they were not accidentally omitted.  However, if you prefer removing them (in particular if that matches the style of code in other parts of the repo) just let me know.

stephenplusplus: I don't mind at all, I think that makes sense. I think you'd be better off finding a way to use your own .jshintrc / extending the default one so that you can just always ignore these errors. For contributors, not worrying about having to put these ignore in place is a more welcoming experience.

DominicKramer: That makes sense.  I noticed that no other packages had `.jshintrc` files, and so I thought the packages just directly used the `.jshhintrc` from the root of the repo.  However, if that is not the case, I can add a error-reporting level `.jshintrc` that extends the `.jshintrc` from the root of the repo.

stephenplusplus: >  I thought the packages just directly used the .jshhintrc from the root of the repo.

That is the case, but I think error-reporting is too large of a code base where the effort to conform to the parent style conventions is not worth it. The point of being consistent between them is to make maintenance and contribution easier, but:

  1. we're not doing a lot of cross-maintenance (GCN devs in error-reporting, or error-reporting in the other areas of GCN), and
  2. the potential outside contributors for error-reporting aren't necessarily direct users of the other libraries in this repo

As long as our CI can run the necessary commands (doc generation, linting, tests: unit, snippet, and system) in E-R, I think it's safe for you to operate under your own definition of best practices and conventions.

The parent `npm run lint` task *might* need to be re-engineered to accept a custom .jshintrc file for E-R, but I don't know. Feel free to make any necessary changes.

DominicKramer: Ok will do.

DominicKramer: I just wanted to let you know that I am planning on making the jshint changes in a follow-up PR.  Thus, after @matthewloring has reviewed and approved this PR, it will be ready to land (unless you think doing the jshint work should be done in this PR).

stephenplusplus: Follow up is fine with me 👍

packages/error-reporting/system-test/testAuthClient.js
```diff
@@ -357,3 +359,69 @@ describe('Expected Behavior', function() {
...
+      serviceContext: {
+        service: SERVICE_NAME,
+        version: SERVICE_VERSION
+      }
+    });
+    transport = new ErrorsApiTransport(errors._config, errors._logger);
+  });
+
+  after(function(done) {
+    transport.deleteAllEvents(function(err) {
```
matthewloring: Are events deleted by the time this call returns or does this just register their deletion with the service at some point in the future?

packages/error-reporting/system-test/testAuthClient.js
```diff
@@ -357,3 +359,69 @@ describe('Expected Behavior', function() {
...
+
+          // The error should have been reported exactly once
+          assert.strictEqual(matchedErrors.length, 1);
+          var errItem = matchedErrors[0];
+          assert.ok(errItem);
+          assert.equal(errItem.count, 1);
+          var rep = errItem.representative;
+          assert.ok(rep);
+          var context = rep.serviceContext;
+          assert.ok(context);
```
matthewloring: If time stamps are recorded when errors are published, those could be used to ensure we are observing the correct error.

packages/error-reporting/system-test/testAuthClient.js
DominicKramer: By the time the call returns the events have just been registered to be deleted, but will be deleted at some point in the future.  I wrote the code this way because the api only has publish capabilities and thus verifying deletion of the events is not necessary.

packages/error-reporting/system-test/testAuthClient.js
DominicKramer: I thought the timestamps could vary to an extent that could make it hard to determine if the error is the correct one, but I can look into it.

packages/error-reporting/system-test/testAuthClient.js
```diff
@@ -357,3 +359,69 @@ describe('Expected Behavior', function() {
...
+
+  after(function(done) {
+    transport.deleteAllEvents(function(err) {
+      assert.ifError(err);
+      done();
+    });
+  });
+
+  it('Should correctly publish errors', function(done) {
+    this.timeout(TIMEOUT * 2);
```
DominicKramer: I've added a comment in the code to make it clear why a scaled version of the timeout is used.

DominicKramer:
@stephenplusplus This is ready to land now.

Issue #2266, opened 2017-04-28
[Datastore] runQuery() method returns record values as "undefined"


kurapatijayaram:
#### Environment details

  - OS: Ubuntu
  - Node.js version: v6.9.1
  - npm version: 3.10.8
  - google-cloud-node version: 

#### Steps to reproduce

  1. require `google-cloud`
 
I am able to fetch the records but those values are displaying as "undefined".
Example code:
```javascript
const datastore = require('@google-cloud/datastore')({
  projectId: '<project_id>',
  keyFilename: '<json_credentials_path>'
});
var query = datastore.createQuery('Tasks');

datastore.runQuery(query, function(err, entities) {
  console.log(entities)
});
```

Output:
[ { name: undefined }, { name: undefined, some: undefined } ]

stephenplusplus:
Can you try a full insert and query routine?

kurapatijayaram:
@stephenplusplus I didn't get you?

stephenplusplus:
Sorry! I mean can you write some code that inserts one entity and then you query for it directly after to verify it inserts and reads successfully?

kurapatijayaram:
Data insertion works perfectly.

ajmcdermitt:
I am having the same problem.  This was working last night and now this morning it does not work.  There was no code change.

stephenplusplus:
@pcostell @lukesneeringer aware of any service changes or outages?

NicholasHoff:
I'm having the same problem.  I tried running the same simple test query using the python client library and it worked correctly.  I also tried making a cloud function (using the Node.js library of course) and I got the same problem, everything returned undefined.

Interestingly - I get the correct number of entities.  If three should be returned, three are returned, but in all cases, the actual values are undefined.

I can see the correct data in the web console, and the fact that the python library works shows that it's not a service outage.

I'll look through the repo to see if anything changed with the Node.js client library recently.

hudgins:
I ran into this this morning also. Reverting to an old version of the umbrella package fixed things ("google-cloud": "^0.45.2",)

ajmcdermitt:
@hudgins How do you revert back?

stephenplusplus:
Early debugging indicates grpc @ 1.3.0 (released last night) is the issue. I will release a quick fix that downgrades it to the last known working version while we figure out the underlying cause. I'll update when that's ready.

stephenplusplus:
[...]

PR #2267, opened 2017-04-28
common-grpc: update to grpc@1.3.0

stephenplusplus:
(RE: #2266)

## ⚠️ New Info

gRPC is reverting the upgrade to protobuf.js@6.0, which solves some problems. Others persist in the updated problem chart below.

- [x] camel case: https://github.com/grpc/grpc/pull/9004#discussion_r92059008
- [x] `message.encode` is not a function: https://github.com/googleapis/google-cloud-node-core/issues/488
- [x] Spanner system tests hang indefinitely
- [x] Speech system tests for `startRecognition` hangs indefinitely

Spanner seems to have issues with the GAX streaming methods. The specific method that is giving trouble is [executeStreamingSql](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/03621060782ce0f6fa5fdecc5d4bf8f181b73d3b/packages/spanner/src/v1/spanner_client.js#L562). The method is called, then the process immediately exits without any output, similar to if the code path lead to nowhere.

---

## ❌ Old Info

gRPC published 1.3.0 yesterday and it's giving our code some issues. This PR is putting out the fires that have popped up so far.

Related discoveries, and their resolution status:

- [x] camel case: https://github.com/grpc/grpc/pull/9004#discussion_r92059008
- [ ] `message.encode` is not a function: https://github.com/googleapis/google-cloud-node-core/issues/488
- [ ] Spanner system tests hang indefinitely
- [ ] Speech system tests for `startRecognition` hangs indefinitely

cc @murgatroid99 for awareness

### DIFF
--- a/packages/common-grpc/package.json
+++ b/packages/common-grpc/package.json
@@ -43,7 +43,7 @@
     "duplexify": "^3.5.0",
     "extend": "^3.0.0",
     "google-proto-files": "^0.11.0",
-    "grpc": "^1.3.1-pre1",
+    "grpc": "^1.3.1",
     "is": "^3.2.0",
     "modelo": "^4.2.0",
     "retry-request": "^2.0.0",


murgatroid99:
I would like to point out that grpc/grpc#9004 was not actually merged, and its contents do not exactly reflect what is actually in the new release. In particular, the default camel casing behavior should not change between gRPC 1.2.x and 1.3.x

stephenplusplus:
Maybe it's a change in a new version of protobuf that 1.3.0 uses? Before 1.3.0, snake case was expected, and in 1.3.0, camelcase is.

murgatroid99:
I see. Your call to `grpc.load` is explicitly setting the option `convertFieldsToCamelCase`, which is why all of the field names need to be in camel case.

stephenplusplus:
Yes, and even still in 1.2.x, we were getting some snake case property names in responses. In 1.3.0, the snake case property names come back as camel case.

stephenplusplus:
Found https://github.com/grpc/grpc/pull/9004#discussion_r93095975:

> Protobuf.JS spits out camel case now

So... not sure what happened. But something happened.

murgatroid99:
That comment is accurate. gRPC's `convertFieldsToCamelCase` option instructs Protobuf.js to use what is now its default behavior, and negating or omitting that option tells it to preserve the casing in the proto file.

stephenplusplus:
Can you explain more simply (for my benefit) why we needed the fix in this PR to use 1.3.0?

murgatroid99:
I can't tell exactly what the context is of the code changes you made there, but it looks like you were receiving (some) field names in snake case instead of camel case, even though you set the option `convertFieldsToCamelCase`. I would consider that to be a bug (though one I was unaware of), and it seems to have been fixed by the upgrade to Protobuf.js 6.

stephenplusplus:
Thanks for the clarity, @murgatroid99!

murgatroid99:
gRPC is switching its dependency on Protobuf.js back to version 5, which should fix at least some of the problems you are seeing here.

murgatroid99:
I have published grpc@1.3.1-pre1, which switches the dependency on Protobuf.js back to version 5. Can you check if that fixes these issues?

stephenplusplus:
Thanks, here are the results:

- [x] camel case (changes reverted): https://github.com/grpc/grpc/pull/9004#discussion_r92059008
- [x] `message.encode` is not a function: https://github.com/googleapis/gax-nodejs/issues/135
- [ ] Spanner system tests hang indefinitely
- [x] Speech system tests for `startRecognition` hangs indefinitely

Spanner seems to have issues with the GAX streaming methods. The specific method that is giving trouble is [executeStreamingSql](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/03621060782ce0f6fa5fdecc5d4bf8f181b73d3b/packages/spanner/src/v1/spanner_client.js#L562). The method is called, then the process immediately exits without any output, similar to if the code path lead to nowhere.

murgatroid99:
Thank you for checking that. I'm not really familiar with GAX, so I can't really tell what gRPC functionality that codepath should eventually touch.

stephenplusplus:
Ran tests again with 1.3.1 and all is well!

Issue #1972, opened 2017-02-07
double leading slash when uploading a file to a bucket


kilianc:
#### Environment details

  - OS: GKE
  - Node.js version: 7
  - npm version: 4.0.5
  - google-cloud-node version: "@google-cloud/storage": "^0.5.0"

#### Steps to reproduce

```js
const bucket = Storage().bucket('gs://my-name')
const file = bucket.file('/foo/bar/lol.txt')
const readStream = fs.createReadStream('./lol.txt')
const writeStream = file.createWriteStream({ metadata: { contentType: 'text/plain' } })
readStream.pipe(writeStream)
```
Given the above code it uploads a file at `gs://my-name//foo/bar/lol.txt`

stephenplusplus:
The bucket name should simply be `my-name`, not its URL:

```diff
- const bucket = Storage().bucket('gs://my-name')
+ const bucket = Storage().bucket('my-name')
```

And same for the file method:
```diff
- const file = bucket.file('/foo/bar/lol.txt')
+ const file = bucket.file('foo/bar/lol.txt')
```

With these changes, the file uploaded should land at: `gs://my-name/foo/bar/lol.txt`.

I don't think it'd be a big change if our library auto-removed any leading slashes from the input to `bucket.file()`, but given the explanation above of the correct way to use these methods, do you think it's necessary?

kilianc:
@stephenplusplus I was using only `my-name` indeed, I messed up while trying to extract the use case.

> given the explanation above of the correct way to use these methods, do you think it's necessary?

Since I am not feeding the library anywhere a double leading slash it is very strange that this is the outcome. I mean `my-name` + `/foo` -> `my-name/foo`. I understand the other way around where in case of a missing leading slash it's assumed the user means root.

In my opinion this is straight up a bug, I can't think of anybody that needs to upload a file to `something//file.txt` IMHO.

🍻

stephenplusplus:
You're right, that is weird. I'll mark as `help wanted`, so if anybody wants to strip leading `/` from within the `File` constructor, please help!

stephenplusplus:
Oh, and 🍻

kilianc:
I am looking at the code, I'll def try to help

PR #2269, opened 2017-05-01
storage: strip leading slash in File() constructor

cjihrig:
This commit prevents a double leading slash from being included in the uploaded file name when `bucket.file()` is called with a path starting with a slash. Only a single slash is removed, because it is assumed that if multiple leading slashes are provided, then the user may have done so on purpose.

Fixes: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1972

### DIFF
--- a/packages/storage/src/file.js
+++ b/packages/storage/src/file.js
@@ -101,13 +101,9 @@ function File(bucket, name, options) {
   this.bucket = bucket;
   this.storage = bucket.parent;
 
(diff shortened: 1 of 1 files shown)


cjihrig:
I signed it!

stephenplusplus:
Thank you @cjihrig! I made a small commit with a style change. If that doesn't break, I will merge straight away.

PR #2272, opened 2017-05-02
auto projectId insertion

stephenplusplus:
RE: #1859 
RE: https://github.com/googleapis/google-cloud-node-core/issues/489

This is a not-correctly-done version of what https://github.com/googleapis/google-cloud-node-core/issues/489 might evolve into. This PR changes generated code, so that will obviously need to be moved to the GAX repo.

However, this works! Kind of. I'm running into an issue, which I'll note in a code comment.

### DIFF
--- a/packages/logging/src/index.js
+++ b/packages/logging/src/index.js
@@ -24,7 +24,6 @@ var arrify = require('arrify');
 var common = require('@google-cloud/common');
 var extend = require('extend');
 var format = require('string-format-obj');
-var googleAuth = require('google-auto-auth');
 var is = require('is');
 var pumpify = require('pumpify');
 var streamEvents = require('stream-events');
@@ -69,14 +68,18 @@ function Logging(options) {
     return new Logging(options);
   }
 
-  var options_ = extend({
-    scopes: v2.ALL_SCOPES
+  options = extend({
+    projectId: '{{projectId}}'
   }, options);
 
-  this.api = {};
-  this.auth = googleAuth(options_);
-  this.options = options_;
-  this.projectId = options.projectId || '{{projectId}}';
+  this.projectId = options.projectId;
+
+  var gax = v2(options);
+
+  this.api = {
+    Config: gax.configServiceV2Client(),
+    Logging: gax.loggingServiceV2Client()
+  };
 }
 
 // jscs:disable maximumLineLength
@@ -163,12 +166,7 @@ Logging.prototype.createSink = function(name, config, callback) {
 
   delete reqOpts.sink.gaxOptions;
 
-  this.request({
-    client: 'configServiceV2Client',
-    method: 'createSink',
-    reqOpts: reqOpts,
-    gaxOpts: config.gaxOptions
-  }, function(err, resp) {
+  this.api.Config.createSink(reqOpts, config.gaxOptions, function(err, resp) {
     if (err) {
       callback(err, null, resp);
       return;
@@ -299,12 +297,7 @@ Logging.prototype.getEntries = function(options, callback) {
     autoPaginate: options.autoPaginate
   }, options.gaxOptions);
 
-  this.request({
-    client: 'loggingServiceV2Client',
-    method: 'listLogEntries',
-    reqOpts: reqOpts,
-    gaxOpts: gaxOptions
-  }, function() {
+  this.api.Logging.listLogEntries(reqOpts, gaxOptions, function() {
     var entries = arguments[1];
 
     if (entries) {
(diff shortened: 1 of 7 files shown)


packages/logging/src/v2/metrics_service_v2_client.js
```diff
@@ -99,27 +99,94 @@ function MetricsServiceV2Client(gaxGrpc, grpcClients, opts) {
...
-    self['_' + methodName] = gax.createApiCall(
-      metricsServiceV2Stub.then(function(metricsServiceV2Stub) {
-        return function() {
-          var args = Array.prototype.slice.call(arguments, 0);
-          return metricsServiceV2Stub[methodName].apply(metricsServiceV2Stub, args);
-        };
-      }),
-      defaults[methodName],
-      PAGE_DESCRIPTORS[methodName]);
+    self['_' + methodName] = function() {
```
stephenplusplus: This is the code @callmehiphop helped write.

packages/logging/src/v2/metrics_service_v2_client.js
```diff
@@ -99,27 +99,94 @@ function MetricsServiceV2Client(gaxGrpc, grpcClients, opts) {
...
+      return Promise.resolve(metricsServiceV2Stub);
+    }
+
+    metricsServiceV2Stub = gaxGrpc.createStub(
+      grpcClients.google.logging.v2.MetricsServiceV2, opts);
+
+    return metricsServiceV2Stub;
+  }
+
+  function replaceProjectIdToken(value, projectId) {
```
stephenplusplus: This should likely move to a static function somewhere inside GAX, as opposed to each client's constructor.

packages/logging/src/index.js
```diff
@@ -163,12 +166,7 @@ Logging.prototype.createSink = function(name, config, callback) {
 
   delete reqOpts.sink.gaxOptions;
 
-  this.request({
-    client: 'configServiceV2Client',
-    method: 'createSink',
-    reqOpts: reqOpts,
-    gaxOpts: config.gaxOptions
-  }, function(err, resp) {
+  this.api.Config.createSink(reqOpts, config.gaxOptions, function(err, resp) {
```
stephenplusplus: Interestingly, all `resp`s are coming back as arrays. I'm not sure why, but it's forcing me to unbox each. I think that's either a bug in GAX or the code you wrote, and I possibly butchered @callmehiphop. Please let me know if you have any thoughts.

packages/logging/src/v2/config_service_v2_client.js
```diff
@@ -100,27 +100,94 @@ function ConfigServiceV2Client(gaxGrpc, grpcClients, opts) {
...
-    self['_' + methodName] = gax.createApiCall(
-      configServiceV2Stub.then(function(configServiceV2Stub) {
-        return function() {
-          var args = Array.prototype.slice.call(arguments, 0);
-          return configServiceV2Stub[methodName].apply(configServiceV2Stub, args);
-        };
-      }),
-      defaults[methodName],
-      PAGE_DESCRIPTORS[methodName]);
+    self['_' + methodName] = function() {
```
stephenplusplus: @landrito this is where a big part of the changes begin. The function waits until it is called to do the `gax.createApiCall` behavior.

packages/logging/src/v2/metrics_service_v2_client.js
```diff
@@ -99,27 +99,94 @@ function MetricsServiceV2Client(gaxGrpc, grpcClients, opts) {
...
-      PAGE_DESCRIPTORS[methodName]);
+    self['_' + methodName] = function() {
+      var args = [].slice.call(arguments);
+      var cb;
+
+      if (typeof args[args.length - 1] === 'function') {
+        cb = args.pop();
+      }
+
+      var promise = getStub().then(function(stub) {
```
callmehiphop: Did this get linted? Seems like this should throw an error since we never use `stub` and we refer to the `metricsServiceV2Stub` instead.

stephenplusplus: Nope, these files are excluded from the linter. These won't merge in this PR, it is just reflecting the changes we require by the generator to make this work.

callmehiphop: Gotcha, thanks!

packages/logging/src/v2/config_service_v2_client.js
```diff
@@ -100,27 +100,94 @@ function ConfigServiceV2Client(gaxGrpc, grpcClients, opts) {
...
+
+              try {
+                reqOpts = replaceProjectIdToken(reqOpts, self.auth.projectId);
+              } catch(e) {
+                var callback = args[args.length - 1];
+                callback(e);
+                return;
+              }
+
+              return configServiceV2Stub[methodName].apply(configServiceV2Stub, args);
```
callmehiphop: @stephenplusplus I think this is where that response array is coming from? Since promises can only resolve with a single parameter, I'm guessing when we run the API method in promise mode, it boxes everything in

PR #2275, opened 2017-05-03
Update google-auto-auth dependency

kohlmannj:
This PR updates `google-auto-auth` to v0.7.0, which itself uses a newer version of `gcp-metadata`, whose dependencies no longer install `node-uuid`. This eliminates a `node-uuid is deprecated` message when installing `@google-cloud/common`.


stephenplusplus:
Thanks @kohlmannj! If you want to skip the CLA stuff, I'll just take care of this.

stephenplusplus:
This will be covered as part of our next release process. Thanks!

PR #2276, opened 2017-05-04
bigtable/table/createReadStream: allow early ending

### LINKED ISSUE #2271: Bigtable this.end() does not stop the stream?
Using Bigtable 0.9.1 - I seem to be getting more than 1 result after calling `this.end()`. Is this expected behavior?

stephenplusplus:
Fixes #2271

### DIFF
--- a/packages/bigtable/system-test/bigtable.js
+++ b/packages/bigtable/system-test/bigtable.js
@@ -379,37 +379,36 @@ describe('Bigtable', function() {
   });
 
   describe('rows', function() {
+    var ROWS = [{
+      key: 'gwashington',
+      data: {
+        follows: {
+          jadams: 1
+        }
+      }
+    }, {
+      key: 'tjefferson',
+      data: {
+        follows: {
+          gwashington: 1,
+          jadams: 1
+        }
+      }
+    }, {
+      key: 'jadams',
+      data: {
+        follows: {
+          gwashington: 1,
+          tjefferson: 1
+        }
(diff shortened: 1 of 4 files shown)


stephenplusplus:
That doesn't seem right, @coveralls.

stephenplusplus:
This is turning out to be a bit of a problem. Marking as `do not merge` until I can get to the bottom of this.

packages/bigtable/system-test/bigtable.js
```diff
@@ -510,7 +509,7 @@ describe('Bigtable', function() {
       it('should get rows', function(done) {
         TABLE.getRows(function(err, rows) {
           assert.ifError(err);
-          assert.strictEqual(rows.length, 4);
+          assert(rows.length >= ROWS.length);
```
jgeewax: Why >= from strict equal? Could the rows returned be more than the rows provided?

stephenplusplus: This PR will need reworking, but currently some tests insert rows, which affects the state outside of the test block. 4 came from "3 in the set up hook, and 1 from the last test". If this test was run in isolation, that last test wouldn't have inserted a row, so we would have 3 not 4. Testing for "at least as many in the set up hook" is an easier solution than creating a table for each test.

stephenplusplus:
Carried on in https://github.com/googleapis/nodejs-bigtable/pull/18

PR #2279, opened 2017-05-05
bigquery/table: add copyFrom()

### LINKED ISSUE #2278: Copying multiple source tables using `table.copy`
In the BigQuery Docs' REST reference, for a `copy` job, one may specify a single `sourceTable` object, or a list of tables as `sourceTables`. It looks like it's currently not possible to specify `sourceTables` using the NodeAPI's copy method. Instead, we have to post a much more verbose REST request like: 

```
const copyJob = await new Promise((resolve, reject) => {
	bigQuery.request({
		method: 'POST',
		uri: '/jobs',
		json: {
			configuration: {
				copy: {
					sourceTables,
					destinationTable: {
						datasetId: table.dataset.id,
						projectId: table.bigQuery.projectId,
						tableId: table.id,
					},
					writeDisposition: overwrite ? 'WRITE_TRUNCATE' : 'WRITE_APPEND',
				},
			},
		},
	}, (err, resp) => {
		if (err) {
			return reject(err);
		}

		const job = bigQuery.job(resp.jobReference.jobId);
		job.metadata = resp;

		resolve(job);
	});
});
```
I'd like to request a feature on the Node API's copy function that allows us to specify multiple source tables ala `sourceTables`. Cheers!

#### Environment details

  - OS: MacOs Sierra v. 10.12.4
  - Node.js version: 7.7.3
  - npm version: 4.1.2
  - google-cloud/storage: 1.1.0

stephenplusplus:
Fixes #2278

This adds a method on a Table object, `copyFrom`:

```js
table.copyFrom([
  dataset.table('source-table-1'),
  dataset.table('source-table-2')
], function(err, job) {});
```

### DIFF
--- a/packages/bigquery/src/table.js
+++ b/packages/bigquery/src/table.js
@@ -442,7 +442,7 @@ Table.prototype.copy = function(destination, metadata, callback) {
  * @throws {Error} If a source other than a Table object is provided.
  *
  * @example
- * var sourceTables = p
+ * var sourceTables = [
  *   dataset.table('your-table'),
  *   dataset.table('your-second-table')
  * ];


stephenplusplus:
Just a thought, would it be better to merge this functionality into `import`?

```js
table.import([
  sourceTable1,
  sourceTable2
], ...
```

callmehiphop:
> Just a thought, would it be better to merge this functionality into import?

Personally I like this! We'd definitely want to document very clearly what supplying a string would do for that method though.

stephenplusplus:
For any users familiar with the BQ concepts of copying as opposed to importing, the separation might be appreciated. And like you pointed out, `import` already has a couple ways of being used, so adding another might add more confusion. For those reasons, let's see if `copyFrom` is good enough and always keep the door open for refactoring and reconsideration in the future.

Issue #1699, opened 2016-10-17
Should our promises be documented as first class?


stephenplusplus:
Continuing https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1697#discussion_r83665859:

Should we be promoting promise use first, as in "this is the way to use our library", with a mention about callbacks? Since the demand for promises is so strong, I'm wondering if we might be better off changing our position to a promise library that supports callbacks instead of the other way around.

SimenB:
As a complete outsider, I vote yes! 
Especially with async await on the horizon, it's a way better way of writing async code

eoogbe:
This issue may no longer be relevant due to its age. Feel free to re-open.

eoogbe:
I am no longer on the team

PR #2281, opened 2017-05-05
bigquery: jsdoc optional paramteres for query

mentos1386:
Not all parameters are required when doing `bigquery.query(options)`.


stephenplusplus:
Thank you @mentos1386. We document and list the `options` argument itself as optional. It is implied that everything it accepts within it is also optional. Some of our methods accept an argument named `config`, which *is* required, but not all of the properties within it are required. We make sure to mark the optional items in those instances.

The callback is indeed optional when you are using promise mode, even though it isn't marked as such. However, it will take a big sweep of our library to change the documentation to be consistent there, and I think this would be covered by #1699.

PR #2286, opened 2017-05-09
Set the default value for max sessions to 100

vkedia:
There can only be 100 concurrent rpcs per grpc channel so it does not make sense for this to be greater than 100. Also keeping it bounded makes it less likely for a bugggy user code to create a tons of sessions, for eg is they do not release a session to the pool.

### DIFF
--- a/packages/spanner/src/session-pool.js
+++ b/packages/spanner/src/session-pool.js
@@ -149,7 +149,7 @@ SessionPool.getPoolOptions_ = function(userOptions) {
     testOnBorrow: true,
     max: userOptions.max || 100,
     min: userOptions.min || 0,
-    numTestsPerRun: Number.MAX_SAFE_INTEGER
+    numTestsPerRun: 100
   };
 
   poolOptions.numTestsPerRun = poolOptions.max;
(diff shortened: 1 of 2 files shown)


vkedia:
@callmehiphop @stephenplusplus can one of you approve this.

callmehiphop:
@vkedia would you mind updating the [unit tests](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/spanner/test/session-pool.js#L343) as well? Looks like CI is unhappy.

vkedia:
Fixed the test.

vkedia:
@callmehiphop  Its still failing since I also need to update numTestsPerRun. What does that parameter do?

vkedia:
Ah you already fixed it. Thanks. But what does that parameter do?

callmehiphop:
@vkedia that's part of the pooling library we use, it is used to limit the number of sessions to loop through when trying to see if an idle session has expired. I went ahead and made a commit to fix that

vkedia:
I see so it makes sense to cap it to maxSessions. Sounds good.

vkedia:
I am just curious. if we leave it unset wont the pooling library
automatically cap it to maxSessions since it does not make sense for it be
any more than that.

callmehiphop:
@vkedia nope, just double checked, for whatever reason the library defaults that value to `3`.

vkedia:
Ok. Sounds good. I will wait for CircleCI to pass before merging this.

Issue #2218, opened 2017-04-13
logging-winston: use custom serialization


stephenplusplus:
Could you take a look at this system test that is failing?

- [Build](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-node/765?utm_campaign=build-failed&utm_medium=email&utm_source=notification)

```
 1) LoggingWinston log should properly write log entries:

      Uncaught AssertionError: '2017-04-13T13:18:30.003Z' === 'Thu Apr 13 2017 13:17:08 GMT+0000 (UTC)'
      + expected - actual

      -2017-04-13T13:18:30.003Z
      +Thu Apr 13 2017 13:17:08 GMT+0000 (UTC)
      
      at Object.testData.verify (system-test/logging-winston.js:63:18)
      at system-test/logging-winston.js:104:20
      at Array.forEach (native)
      at system-test/logging-winston.js:102:14
      at node_modules/@google-cloud/logging/src/index.js:314:14
```

ofrobots:
Curious... seems to be passing for me.

stephenplusplus:
Alright, you can close this if you can't find anything flaky about the test, then if the problem shows up again on CI, we'll just re-open.

ofrobots:
Let me stare at the code for a bit.

ofrobots:
I am really puzzled by this. The test is:

```js
var testTimestamp = new Date();
//...
assert.strictEqual(
    entry.metadata.labels.testTimestamp,
    util.inspect(testTimestamp)
);
```

Even if it is a flake, I am quite puzzled how `util.inspect(testTimestamp)` managed to format it as `Thu Apr 13 2017 13:17:08 GMT+0000 (UTC)`. `util.inspect` should use the ISO form.

Do you have objections to pre-emptively adding a debug print to catch some more info if it flakes again?

stephenplusplus:
Looks like that's a new thing:

```
$ node
> process.version
'v4.8.2'
> util.inspect(new Date())
'Thu Apr 20 2017 12:30:34 GMT-0400 (Eastern Daylight Time)'
```

ofrobots:
Okay, it makes a lot more sense now.

It is definitely a flake in that case. It seems that we are retrieving the log entries from previous system-test run. That is, a node 4 run is fetching the entries from a preceding node 6/7 run.

The delay is already set to 90 seconds, should we increase it?

stephenplusplus:
[...]

PR #2288, opened 2017-05-10
logging-winston: metadata serialization

ofrobots:
This is a semver-major change that modifies how we serialize winston
provided log metadata. Since `util.inspect` behaviour is quite variable
between node releases, this used to give us different behaviour
depending on node version.

The choice to go with `util.inspect` was motivated by trying to match
the document for the built-in winston transports. It turns out that the
docs weren't quite precise and inspect isn't the default behaviour
anyway.

Previously we attached metadata as string labels. We now attach
metadata as a `metadata` property on the payload. If the config flag
`options.inspectMetadata` is provided, we will use `util.inspect` to
stringify.

We no longer use labels in either case.

Fixes: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2218

### DIFF
--- a/packages/logging-winston/test/index.js
+++ b/packages/logging-winston/test/index.js
@@ -85,11 +85,41 @@ describe('logging-winston', function() {
       });
     });
 
+    it('should default to logging.write scope', function() {
+      assert.deepEqual(fakeLoggingOptions_.scopes, [
+        'https://www.googleapis.com/auth/logging.write'
+      ]);
+    });
+
+    it('should initialize Log instance using provided scopes', function() {
+      var fakeScope = 'fake scope';
+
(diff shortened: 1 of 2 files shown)


stephenplusplus:
Thanks @ofrobots. I made a few style changes. If CI is happy, I'll merge after. Would you like a release to immediately follow that?

packages/logging-winston/test/index.js
```diff
@@ -85,13 +85,43 @@ describe('logging-winston', function() {
       });
     });
 
+    it('should default to logging.write scope', function() {
```
stephenplusplus: These unrelated blocks weren't changed, only re-sorted to match the order of the logic in the constructor.

PR #2298, opened 2017-05-11
Vision Partial Veneer

lukesneeringer:
This is the new surface with breaking changes.

This is an initial pass at the implementation of a "partial Veneer" for Vision.

It:

  * deprecates the entire manual layer
  * makes the GAPIC, with helpers, available at `require('@google-cloud/vision').v1()`.
  * provides the `annotateImage` and single-feature methods
  * has 100% test coverage on manual code

It does *not*:

  * ~provide a default partial Veneer on the versionless entry point (because it is taken by the deprecated manual layer)~ have backwards compatibility, at all, even remotely a little
  * [x] handle specifying a file handler for an image (needs to be done)
  * [x] provide documentation for the partial Veneer (needs to be done)
  * [x] have 100% test coverage on automated code (needs to be done)

Notes:

  * This introduces the use of ES6 syntax supported by Node 4.
  * This uses power-assert ~sinon and chai, which are not currently being used by other packages~

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -1,9 +1,9 @@
 {
   "report-dir": "./.coverage",
   "exclude": [
-    "src/*{/*,/**/*}.js",
-    "src/v*/*.js",
-    "test/**/*.js"
+    "packages/*/src/*{/*,/**/*}.js",
+    "packages/*/src/*/v*/*.js",
+    "packages/*/test/**/*.js"
   ],
   "watermarks": {
     "branches": [


stephenplusplus:
> provide a default partial Veneer on the versionless entry point (because it is taken by the deprecated manual layer)

If the manual layer wasn't there, what would the ideal solution have been?

> This uses sinon and chai, which are not currently being used by other packages

Instinctively, I would think we should stick to our testing style in the rest of the repo. Is it just because it's easier, prettier, had some better features? We use sinon in places for easier stubbing, the chai part would be new.

jmdobry:
> provide a default partial Veneer on the versionless entry point (because it is taken by the deprecated manual layer)

In which version of `@google-cloud/vision` do you plan to export the partial gapic as the versionless entrypoint?

jmdobry:
> If the manual layer wasn't there, what would the ideal solution have been?

The ideal solution would be:

```js
module.exports = vision_v1;
module.exports.v1 = vision_v1;
// fictional other version
module.exports.v1beta1 = vision_v1beta1;
```

right?

stephenplusplus:
I think so, but I can see being explicit worth considering, as well, if we intend to carry multiple versions.

lukesneeringer:
> > If the manual layer wasn't there, what would the ideal solution have been?

> The ideal solution would be:

> ```js
module.exports = vision_v1;
module.exports.v1 = vision_v1;
// fictional other version
module.exports.v1beta1 = vision_v1beta1;
```

> right?

Precisely.

If you ask for a version, you get that version.
If you do not, you get the latest GA.

**Comment by lukesneeringer** [comment]
> In which version of `@google-cloud/vision` do you plan to export the partial gapic as the versionless entrypoint?

I do not know; I was going to ask for feedback on that. :-) I did not realize I would run up against that problem until yesterday when I was writing the code.

There are a few options:

  * Move the old layer to a new namespace (backwards-incompatible change), and document that. Existing customers make a one-line change to "get working again" and then can go port their stuff.
  * The current form: For now, you need to use the version to get the new stuff. 
      * A hybrid of the above two: We do the current form and _also_ and old `manual` namespace; give people a month or so to make the one-line change, then they have a longer-deprecation period to update to the new surface.
  * Move fast and break things: Just wipe out the manual surface sooner in Node than in other languages. (In fairness, Node developers are likely _the most_ receptive to this option of any language community. :-))

**Comment by lukesneeringer** [comment]
> Instinctively, I would think we should stick to our testing style in the rest of the repo. Is it just because it's easier, prettier, had some better features? We use sinon in places for easier stubbing, the chai part would be new.

I really needed sinon to do the mocks I wanted. Since you are using it elsewhere, I assume this is not interesting.

I ended up adding chai at the last minute because `assert` was not giving me a useful error message for a particular test failure, and I knew that chai would. I switched it, got the error I needed, and fixed the test. Once I fixed the test, I did not bother to change it back. On the one hand, I could totally change it back now; on the other hand, I prefer more useful error messages for future test failures. I could go either way on this.

**Comment by lukesneeringer** [comment]
@stephenplusplus I would also like your thoughts on my use of ES6 syntax. Is that something you are okay with beginning to use now that we dropped Node 0.12 support, or would you prefer have the repo remain ES5 for the forseeable future?

**Comment by stephenplusplus** [comment]
Move fast and break things.

I also don't feel strongly about chai or no chai, if it helps us. My concern would be adding another tool to learn that contributors have to face.

ES6 syntax is great.

I'll do a deeper dive soon.

**Comment by jmdobry** [comment]
I agree, move fast and break things. It's better to remove the manual code pre-1.0 than to have to do a post-1.0 major version bump.

**Comment by jmdobry** [comment]
> I prefer more useful error messages

The most useful error message come not from using chai, but by using [power-assert](https://github.com/power-assert-js/power-assert).

**Comment by lukesneeringer** [comment]
> I agree, move fast and break things. It's better to remove the manual code pre-1.0 than to have to do a post-1.0 major version bump.

These are not the only options, though. We are not going 1.0 when we release this. We could deprecate the manual layer, and still remove it pre-1.0.

(Nonetheless, I am okay with "move fast and break things" in Node. I am not okay with it in Python and Ruby.)

**Comment by lukesneeringer** [comment]
> The most useful error message come not from using chai, but by using power-assert.

Oh hey, I like that way better. (That is similar to py.test in Python.)
I will swap to that later.

**Comment by lukesneeringer** [comment]
I moved over to power-assert. Holding off on scrapping the manual layer until that can have more discussion.

File: packages/vision/package.json [hunk 1]
```diff
@@ -70,10 +70,40 @@
     "multiline": "^1.0.2",
     "node-uuid": "^1.4.7",
     "normalize-newline": "^2.0.0",
+    "nyc": "^10.3.0",
+    "power-assert": "^1.4.2",
     "proxyquire": "^1.7.10",
+    "sinon": "^2.2.0",
     "tmp": "^0.0.31"
   },
+  "nyc": {
```
stephenplusplus: Maybe this can go into a .nycrc in the root directory and have the same effect. Or possibly better, we have a npm run-script, that is a small wrapper around the current `cover` task, that takes the module to test as an argument.

lukesneeringer: I took this from the existing root `package.json`.

stephenplusplus: The goal would be to not have duplicate places of configuration.

packages/vision/src/helpers.js
```diff
@@ -101,6 +106,19 @@ module.exports = apiVersion => {
...
+    // If there is no image, throw an exception.
+    if (is.undefined(request.image)) {
+      throw new Error('Attempted to call `annotateImage` with no image.');
+    }
+
+    // If we got a filename for the image, open the file and transform
+    // it to content.
+    if (request.image.source && request.image.source.filename) {
+      request.image = {
+        content: fs.readFileSync(request.image.source.filename),
```
jmdobry: We should defer to performing I/O asynchronously as much as possible, as this will block the user's whole app.

packages/vision/package.json
lukesneeringer: I see.

packages/vision/src/helpers.js
```diff
@@ -0,0 +1,197 @@
...
+    //
+    // The method name should be camelCased (e.g. "FACE_DETECTION" becomes
+    // "faceDetection").
+    let featureValue = features[feature];
+    let methodName = feature.toLowerCase().replace(/_([\w])/, match => {
+      return match[1].toUpperCase();
+    });
+
+    // Assign the single-feature method to the `methods` object.
+    methods[methodName] = promisify(_createSingleFeatureMethod(featureValue));
```
stephenplusplus: Using a loop to define methods is a problem for the docs. I'm not sure if this will work, we might have to actually write out the methods in order to get them into the docs.

packages/vision/src/helpers.js
```diff
@@ -0,0 +1,197 @@
...
+   *   The second parameter to the callback is an object representing
+   *   [AnnotateImagesResponse]{@link AnnotateImagesResponse}.
+   * @return {Promise} - The promise which resolves to an array.
+   *   The first element of the array is an object representing
+   *   [AnnotateImagesResponse]{@link AnnotateImagesResponse}.
+   *   The promise has a method named "cancel" which cancels the ongoing
+   *   API call.
+   *
+   * @example
+   * var vision = require('@google-cloud/vision').v1();
```
stephenplusplus: Is it important to use the `v1()` here, or can we just use "require('@google-cloud/vision')()", since they are the same?

If we don't need the `v1()`, you can actually omit this line. Our docs script will always define and show the instantiation of "var moduleName = require('@google-cloud/module-name')` in the docs. This way, all methods are free to use the `moduleName` variable, without having to show how to instantiate the module in each block.

packages/vision/src/helpers.js
```diff
@@ -0,0 +1,197 @@
...
+   * var request = {
+   *   image: {source: {image_uri: 'gs://path/to/image.jpg'}},
+   *   features: [],
+   * };
+   * client.annotateImage(request).then(response => {
```
stephenplusplus: vision.annotateImage*

packages/vision/src/index.js
```diff
@@ -20,2097 +20,45 @@
...
+  opts.libVersion = VERSION;
 
-  if (!is.object(options)) {
-    options = {
-      types: options
-    };
+  // Create the image annotator client with the provided options.
+  var client = gapic.v1(opts).imageAnnotatorClient(opts);
+  if (is.undefined(client.annotateImage)) {
+    Object.assign(client.constructor.prototype, helpers('v1'));
```
stephenplusplus: Like the feature methods defined inside the helpers.js file, this way of programmatically defining methods is a problem for the docs. Off hand, I'm not exactly sure how to solve for this.

lukesneeringer:
The coverage change seems to actually be because so much code was removed as to materially affect the overall total. The Vision coverage is 100%.

lukesneeringer:
@stephenplusplus @jmdobry The ML folks want this delivered reasonably soon. Can you give it another look and see if everything looks good?

In particular, be looking at the _pattern_, since future GAPIC and partial Veneer releases will follow it.

package.json
```diff
@@ -22,9 +23,11 @@
     "multiline": "^1.0.2",
     "nyc": "^10.3.0",
     "package-json": "^2.4.0",
+    "power-assert": "^1.4.2",
     "propprop": "^0.3.1",
     "semver": "^5.3.0",
     "shelljs": "^0.7.3",
+    "sinon": "^2.2.0",
```
stephenplusplus: I don't think this is being used.

stephenplusplus:
@lukesneeringer I'm okay with it overall. The only issue is the docs (https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2298/files/6f3caf64a8887189591b4279e5370beb178e189b#diff-ef8f4754e676d8ad6a326bd2f9bb1d28, https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2298/files/6f3caf64a8887189591b4279e5370beb178e189b#diff-99cc4574db14fc26a3c7bf0b5e2e6ca8).

packages/vision/src/index.js
stephenplusplus: I resolved this part in my commit.

stephenplusplus:
I added a commit which resolved one of the docs issues -- how to merge the methods in `helpers.js` with the `visionV1` definition.



We are still stuck on this however: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2298/files/6f3caf64a8887189591b4279e5370beb178e189b#r121120113

swcloud:
@stephenplusplus @jmdobry Do you think this PR is able to be merged? DPE needs to work based on the partial veneer implementation as soon as possible.

packages/vision/src/helpers.js
swcloud: @stephenplusplus @lukesneeringer Two questions

(1) We define methods.annotateImage here and it has its comments as documentation. Is jsdoc able to surface this documentation?

(2) If (1) is true, then can we surface each individual single feature method in such a way (instead of using a loop)

/**
 *  Documentation for faceDetection
*/
methods.faceDetection = promisify(_createSingleFeatureMethod(features['FACE_DETECTION']));

/**
 *  Documentation for logoDetection
*/
methods.logoDetection = promisify(_createSingleFeatureMethod(features['LOGO_DETECTION']));

stephenplusplus: Yes, JSDocs will only work if there is a comment block for each method, above the method definition, as in your example.

swcloud: OK. I will push a commit to have individual methods with their documentation instead of a loop.

swcloud:
@lukesneeringer @stephenplusplus Please take a look at my latest commit https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2298/commits/ba50074a02acf945857d3c9ad4caf1c244926ee1

stephenplusplus:
LGTM.

packages/vision/package.json
```diff
@@ -1,42 +1,16 @@
...
-      "name": "Silvano Luciani",
-      "email": "silvano@google.com"
-    },
-    {
-      "name": "Stephen Sawchuk",
-      "email": "sawchuk@gmail.com"
-    }
-  ],
-  "main": "./src/index.js",
+  "version": "0.7.1",
```
stephenplusplus: The version should remain the same as it is, 0.11.3. When we publish after the merge, it will become 0.12.0.

swcloud: Done.

packages/vision/package.json
```diff
@@ -1,42 +1,16 @@
...
-      "name": "Silvano Luciani",
-      "email": "silvano@google.com"
-    },
-    {
-      "name": "Stephen Sawchuk",
-      "email": "sawchuk@gmail.com"
-    }
-  ],
-  "main": "./src/index.js",
+  "version": "0.12.0",
```
stephenplusplus: Sorry for the confusion-- this should not be edited in a PR. It is bumped during our release process.

stephenplusplus:
@lukesneeringer close this, since this is going in #2421?

stephenplusplus:
Made some changes.

Docs: http://stephenplusplus.github.io/gcloud-node/#/docs/vision/master/vision

lukesneeringer:
Looks good to me. Thanks @stephenplusplus

julien-c:
I'm using the Vision API to annotate an image with several features at the same time.

Unless I'm missing something obvious, the new `0.12` SDK makes it pretty hard to do this (I had to read the code to find the feature type enumeration)

Happy to open a new issue if that makes sense.

PR #2305, opened 2017-05-12
logging: omit logName from entries

ofrobots:
As per [1] we do not need to have logName assigned on each entry if we
are also providing it on call to entries.write.

This optimization reduces the on-the-wire packet size for LogEntries.

[1]: https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write

As a follow-on, I will make the same optimization for the `resource`.

### DIFF
--- a/packages/logging/src/log.js
+++ b/packages/logging/src/log.js
@@ -306,10 +306,6 @@ Log.prototype.entry = function(metadata, data) {
     metadata = {};
   }
 
-  metadata = extend({}, metadata, {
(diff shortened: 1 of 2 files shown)


ofrobots:
Calls to write insert the `logName` [here](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/b043cda5707df33bd500d1213bb71c1ae4f37920/packages/logging/src/log.js#L599).

PR #2309, opened 2017-05-15
logging: omit resource from each log entry

ofrobots:
Following from https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2305, we can do the same optimization for the resource. We don't need to attach the resource to each entry if we instead attach it to the `entries.write` call. This should reduce the on-the-wire packet size and will be marginally faster.

As a result, there is nothing async left in `Log.prototype.decorateEntries_` so I have converted it to a sync function.

It would be easier to review each commit individually.


packages/logging/src/log.js
```diff
@@ -589,12 +589,27 @@ Log.prototype.write = function(entry, options, callback) {
     options = {};
   }
 
-  this.decorateEntries_(arrify(entry), function(err, decoratedEntries) {
-    // Ignore errors (the API will speak up if it has an issue).
+  if (!options.resource) {
```
stephenplusplus: Why not do this in `decorateEntries_`?

stephenplusplus: This does pretty much what `metadata_.assignDefaultResource` does. Can we use that instead:

```js
this.metadata_.assignDefaultResource(function(err, entry) {
  // Ignore errors (the API will speak up if it has an issue).
  
  var decoratedEntries;

  try {
    // ...
});
```

stephenplusplus: This does pretty much what `metadata_.assignDefaultResource` does. Can we use that instead:

```js
this.metadata_.assignDefaultResource(function(err, entry) {
  // Ignore errors (the API will speak up if it has an issue).
  
  var decoratedEntries;

  try {
    // ...
});
```

stephenplusplus: Sorry, I see the difference now, of course. Since `assignDefaultResource` is no longer used, would you mind removing it?

ofrobots: Oops. I intended to remove it all along, but it seems like in my quest to make the commits look nice I lost that change!

Removed now.

packages/logging/src/log.js
```diff
@@ -591,64 +591,59 @@ Log.prototype.write = function(entry, options, callback) {
...
+ * @param {object[]} entries - array of entry objects
+ * @return {object[]} serialized entries.
+ * @throws if there is an error during serialization.
  */
-Log.prototype.decorateEntries_ = function(entries, callback) {
+Log.prototype.decorateEntries_ = function(entries) {
   var self = this;
 
-  async.map(entries, function(entry, callback) {
+  return entries.map(function(entry) {
```
ofrobots: Sounds like a good idea, but given that entry is not guaranteed to be a `Entry` instance, we have to conditionally construct the entry. I am not sure if it fits into the `methmeth` use-case.

packages/logging/src/log.js
```diff
@@ -591,64 +591,59 @@ Log.prototype.write = function(entry, options, callback) {
...
 
 /**
- * All entries are passed through here to make sure this log is attached to the
- * entry.
+ * All entries are passed through here in order to get them serialized.
  *
  * @private
  *
- * @param {object} entry - An entry object.
+ * @param {object[]} entries - array of entry objects
```
ofrobots: done.

packages/logging/src/log.js
```diff
@@ -591,64 +591,59 @@ Log.prototype.write = function(entry, options, callback) {
...
+ * @return {object[]} serialized entries.
```
ofrobots: done.

packages/logging/src/log.js
stephenplusplus: Ah, sorry. Missed that part.

stephenplusplus:
```
> google-cloud-node@ lint /home/ubuntu/google-cloud-node

packages/logging/src/log.js: line 24, col 5, 'async' is defined but never used.

packages/logging/test/log.js: line 338, col 8, Missing semicolon.

2 errors
```

packages/logging/src/log.js
```diff
@@ -21,7 +21,6 @@
 'use strict';
 
 var arrify = require('arrify');
-var async = require('async');
```
stephenplusplus: I believe `async` is only a devDependency now.

ofrobots: Indeed. Fixed.

PR #2313, opened 2017-05-16
spanner: various bug fixes + test coverage

### LINKED ISSUE #2176: spanner: Transaction was aborted. It was wounded by a higher priority transaction due to conflict on keys in range
#### Environment details

  - OS: Mac OS Siera 10.12.3
  - Node.js 6.9.3:
  - npm version: 3.19.19
  - google-cloud-node version:
"@google-cloud/spanner":"^0.3.0",
"@google-cloud/datastore":"^0.7.0",

I have Users Table:
```
CREATE TABLE Users (
	userId STRING(36) NOT NULL,
	contactName STRING(300) NOT NULL,
	created INT64 NOT NULL,
	eMail STRING(100) NOT NULL,
	environmentsIds ARRAY<STRING(36)> NOT NULL,
	envLanguage STRING(3) NOT NULL,
	facebookName STRING(300) NOT NULL,
	facebookPictureUrl STRING(2000) NOT NULL,
	facebookProfile STRING(2000) NOT NULL,
	googleName STRING(300) NOT NULL,
	googlePictureUrl STRING(2000) NOT NULL,
	googleProfile STRING(2000) NOT NULL,
	gssClientID STRING(36),
	gssGroupID STRING(36) NOT NULL,
	isActive BOOL,
	mobile STRING(30) NOT NULL,
	modifierId STRING(36) NOT NULL,
	modifierName STRING(300) NOT NULL,
	note STRING(MAX) NOT NULL,
	photoID STRING(36),
	purchasedServicesCasheExpiryDateString STRING(MAX),
	registered BOOL NOT NULL,
	signedInBy STRING(20) NOT NULL,
	timestamp INT64 NOT NULL,
	timeZoneID STRING(100) NOT NULL,
	webBrowserID STRING(36),
) PRIMARY KEY (userId)
```
first i check  user on database and then do upsert
```
database.runTransaction((err, tx) => {
    if (err) {
    ... error
    }

    tx.run(`SELECT * FROM Users WHERE eMail = '${user.eMail}' AND isActive = TRUE`,
        (err, rows) => {
            if (err) {
            ... log error
            }
            ...
            console.log('-- upsert - userId = ' + ola.userId + ' - isActive = ' + ola.isActive);
            tx.upsert("Users", user);
            ...
            tx.commit((err, apiResponse) => {
                if (err) {
                ... log error
                }
                console.log('commited - userId = ' + user.userId);
            }
```
When i call many requests at same time - a Error occurs:

Error occurred on sometime on - **select** sometime on - **commit**
log:
```
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = a3e85cdf-9a79-4940-8ac0-075a344a5db3 - isActive = true
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = ef62be27-b553-4b4f-ba45-11cdb4785713 - isActive = true
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = ac48ad39-45ce-4102-948b-11af2ddd4659 - isActive = true
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 04d9c1a2-66ff-4fa1-9388-17dfd20dac1c - isActive = true
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = e40881d1-e3be-438b-89a7-48124ccd12e5 - isActive = true
-- upsert - userId = a3e85cdf-9a79-4940-8ac0-075a344a5db3 - isActive = true
-- upsert - userId = ef62be27-b553-4b4f-ba45-11cdb4785713 - isActive = true
-- upsert - userId = 04d9c1a2-66ff-4fa1-9388-17dfd20dac1c - isActive = true
-- upsert - userId = e40881d1-e3be-438b-89a7-48124ccd12e5 - isActive = true
-- upsert - userId = ac48ad39-45ce-4102-948b-11af2ddd4659 - isActive = true
commited - userId = ef62be27-b553-4b4f-ba45-11cdb4785713
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 13a0f431-2451-4e20-a6bd-5e7e7bb644fe - isActive = true
-- upsert - userId = 13a0f431-2451-4e20-a6bd-5e7e7bb644fe - isActive = true
commited - userId = 13a0f431-2451-4e20-a6bd-5e7e7bb644fe
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = cd64b4be-217e-41b4-808a-61eb984e30b9 - isActive = true
-- upsert - userId = a3e85cdf-9a79-4940-8ac0-075a344a5db3 - isActive = true
-- upsert - userId = ac48ad39-45ce-4102-948b-11af2ddd4659 - isActive = true
commited - userId = a3e85cdf-9a79-4940-8ac0-075a344a5db3
-- upsert - userId = 04d9c1a2-66ff-4fa1-9388-17dfd20dac1c - isActive = true
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = ebadcf2f-7657-4d08-80f5-72af47c30d5e - isActive = true
-- upsert - userId = e40881d1-e3be-438b-89a7-48124ccd12e5 - isActive = true
commited - userId = e40881d1-e3be-438b-89a7-48124ccd12e5
-- upsert - userId = cd64b4be-217e-41b4-808a-61eb984e30b9 - isActive = true
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 126e9b3d-3f60-4731-9bcd-7b79f2b86480 - isActive = true
-- upsert - userId = ebadcf2f-7657-4d08-80f5-72af47c30d5e - isActive = true
commited - userId = cd64b4be-217e-41b4-808a-61eb984e30b9
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 9f9cec23-121d-466a-b747-89685acc6e8e - isActive = true
-- upsert - userId = 126e9b3d-3f60-4731-9bcd-7b79f2b86480 - isActive = true
-- upsert - userId = ac48ad39-45ce-4102-948b-11af2ddd4659 - isActive = true
commited - userId = ac48ad39-45ce-4102-948b-11af2ddd4659
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = e6ec38c9-24e5-4497-8303-90b994a7c916 - isActive = true
-- upsert - userId = 9f9cec23-121d-466a-b747-89685acc6e8e - isActive = true
commited - userId = 9f9cec23-121d-466a-b747-89685acc6e8e
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 889df848-9a2e-4ed9-8126-a45c0c33a101 - isActive = true
-- upsert - userId = ebadcf2f-7657-4d08-80f5-72af47c30d5e - isActive = true
commited - userId = ebadcf2f-7657-4d08-80f5-72af47c30d5e
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = f7405e25-e78c-4d81-b3ed-bace7490ef20 - isActive = true
-- upsert - userId = 04d9c1a2-66ff-4fa1-9388-17dfd20dac1c - isActive = true
-- upsert - userId = e6ec38c9-24e5-4497-8303-90b994a7c916 - isActive = true
commited - userId = 04d9c1a2-66ff-4fa1-9388-17dfd20dac1c
-- upsert - userId = 126e9b3d-3f60-4731-9bcd-7b79f2b86480 - isActive = true
-- upsert - userId = 889df848-9a2e-4ed9-8126-a45c0c33a101 - isActive = true
-- upsert - userId = f7405e25-e78c-4d81-b3ed-bace7490ef20 - isActive = true
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 1e62b2fd-5170-45a5-a139-bf58389cec46 - isActive = true
commited - userId = 126e9b3d-3f60-4731-9bcd-7b79f2b86480
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 25eadd95-0960-412f-9706-c57c05c1f903 - isActive = true
-- upsert - userId = 1e62b2fd-5170-45a5-a139-bf58389cec46 - isActive = true
commited - userId = 1e62b2fd-5170-45a5-a139-bf58389cec46
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = f43adb89-f0fc-4898-b367-c6d6fe2795da - isActive = true
-- upsert - userId = e6ec38c9-24e5-4497-8303-90b994a7c916 - isActive = true
commited - userId = e6ec38c9-24e5-4497-8303-90b994a7c916
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 140da3b6-f213-4627-938a-d3b93f6f101f - isActive = true
-- upsert - userId = 25eadd95-0960-412f-9706-c57c05c1f903 - isActive = true
commited - userId = 25eadd95-0960-412f-9706-c57c05c1f903
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 29d6dc5e-a1bb-4390-948b-d417db53537b - isActive = true
-- upsert - userId = f7405e25-e78c-4d81-b3ed-bace7490ef20 - isActive = true
-- upsert - userId = f43adb89-f0fc-4898-b367-c6d6fe2795da - isActive = true
-- upsert - userId = 889df848-9a2e-4ed9-8126-a45c0c33a101 - isActive = true
commited - userId = 889df848-9a2e-4ed9-8126-a45c0c33a101

req.url - http://chipintozamac.local:9001/gss-api/users/upsert
InternalServerError: err.code: 10
err.message: Transaction was aborted. It was wounded by a higher priority transaction due to conflict on keys in range [[889df848-9a2e-4ed9-8126-a45c0c33a101], [889df848-9a2e-4ed9-8126-a45c0c33a101]), column isActive in table users.
err.status: undefined
err location:
User.upsert\select Users...eMail...
data:
{"userId":"140da3b6-f213-4627-938a-d3b93f6f101f","isActive":true,"created":1384853268917,"timestamp":1491156814361,"modifierId":"system-id","modifierName":"system","eMail":"a.muller@gss.ge","contactName":"ka:ალექსანდრე მიულერი","mobile":"","webBrowserID":"8a0d1345-d688-43eb-8f48-a3239aae23f0","photoID":null,"gssGroupID":"0267345b-a116-4879-9990-81f22d1db9ad","timeZoneID":"Georgian Standard Time","gssClientID":null,"purchasedServicesCasheExpiryDateString":"{\"B09DE416-A2EA-4419-B5F0-70B9CE956007\":{\"e9e5b8c9-bc25-4925-9b86-329928c3583e\":\"28/02/2014\"}}","note":"","environmentsIds":[],"googleName":"","googlePictureUrl":"","googleProfile":"","facebookName":"","facebookPictureUrl":"","facebookProfile":"","signedInBy":"eMail","envLanguage":"ka","registered":false}
apiResponse:

    at Object.getDsError (/Users/Chipintoza/GSS Projects/accounts.gss.ge/server/node_modules/database-utils.js:186:20)
    at error (/Users/Chipintoza/GSS Projects/accounts.gss.ge/server/tables/system/users-table.js:924:64)
    at DestroyableTransform.tx.run.err (/Users/Chipintoza/GSS Projects/accounts.gss.ge/server/tables/system/users-table.js:947:35)
    at emitOne (events.js:101:20)
    at DestroyableTransform.emit (events.js:188:7)
    at /Users/Chipintoza/GSS Projects/accounts.gss.ge/node_modules/through2/through2.js:19:12
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
-- upsert - userId = 29d6dc5e-a1bb-4390-948b-d417db53537b - isActive = true
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 2b264cff-5a31-4481-97ce-d7da401a0db2 - isActive = true
commited - userId = 29d6dc5e-a1bb-4390-948b-d417db53537b
-- upsert - userId = f43adb89-f0fc-4898-b367-c6d6fe2795da - isActive = true
commited - userId = f43adb89-f0fc-4898-b367-c6d6fe2795da
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 094d4afd-2ce9-464e-9fd3-e7e6efbd0401 - isActive = true
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 22a4305a-6670-4de2-8de0-eeed85fe4227 - isActive = true
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = 92545fa4-356e-4219-881c-fddae5be8888 - isActive = true
-- upsert - userId = 2b264cff-5a31-4481-97ce-d7da401a0db2 - isActive = true
commited - userId = 2b264cff-5a31-4481-97ce-d7da401a0db2
http://chipintozamac.local:9001/gss-api/users/upsert
post - userId = bab0f638-55c5-4c8d-b2bc-ff1462dc77cf - isActive = true
-- upsert - userId = 094d4afd-2ce9-464e-9fd3-e7e6efbd0401 - isActive = true
-- upsert - userId = 22a4305a-6670-4de2-8de0-eeed85fe4227 - isActive = true
-- upsert - userId = 92545fa4-356e-4219-881c-fddae5be8888 - isActive = true
-- upsert - userId = f7405e25-e78c-4d81-b3ed-bace7490ef20 - isActive = true

req.url - http://chipintozamac.local:9001/gss-api/users/upsert
InternalServerError: err.code: 10
err.message: Transaction was aborted. It was wounded by a higher priority transaction due to conflict on keys in range [[92545fa4-356e-4219-881c-fddae5be8888], [92545fa4-356e-4219-881c-fddae5be8888]), column isActive in table users.
err.status: undefined
err location:
User.upsert\select Users...eMail...
data:
{"userId":"bab0f638-55c5-4c8d-b2bc-ff1462dc77cf","isActive":true,"created":1384507909720,"timestamp":1491156817894,"modifierId":"system-id","modifierName":"system","eMail":"t.tsirekidze@gss.ge","contactName":"ka:თინათინ ცირეკიძე|ru:Тинатин Цирекидзе","mobile":"995 32 555333350","webBrowserID":"534474e9-6e05-4f03-9dda-441933797cff","photoID":"ba34529c-955f-4fa3-be23-873fcdc6a8fe","gssGroupID":"0267345b-a116-4879-9990-81f22d1db9ad","timeZoneID":"Georgian Standard Time","gssClientID":null,"purchasedServicesCasheExpiryDateString":"{\"40454df0-59ed-4e11-a914-4340bc97cbf6\":{\"e9e5b8c9-bc25-4925-9b86-329928c3583e\":\"28/02/2014\"},\"b97e23bf-07a5-4ad4-b925-2b8003e971cc\":{}}","note":"","environmentsIds":[],"googleName":"","googlePictureUrl":"","googleProfile":"","facebookName":"","facebookPictureUrl":"","facebookProfile":"","signedInBy":"eMail","envLanguage":"ka","registered":false}
apiResponse:

    at Object.getDsError (/Users/Chipintoza/GSS Projects/accounts.gss.ge/server/node_modules/database-utils.js:186:20)
    at error (/Users/Chipintoza/GSS Projects/accounts.gss.ge/server/tables/system/users-table.js:924:64)
    at DestroyableTransform.tx.run.err (/Users/Chipintoza/GSS Projects/accounts.gss.ge/server/tables/system/users-table.js:947:35)
    at emitOne (events.js:101:20)
    at DestroyableTransform.emit (events.js:188:7)
    at /Users/Chipintoza/GSS Projects/accounts.gss.ge/node_modules/through2/through2.js:19:12
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
commited - userId = f7405e25-e78c-4d81-b3ed-bace7490ef20
-- upsert - userId = 22a4305a-6670-4de2-8de0-eeed85fe4227 - isActive = true
-- upsert - userId = 92545fa4-356e-4219-881c-fddae5be8888 - isActive = true
commited - userId = 92545fa4-356e-4219-881c-fddae5be8888
-- upsert - userId = 094d4afd-2ce9-464e-9fd3-e7e6efbd0401 - isActive = true
commited - userId = 094d4afd-2ce9-464e-9fd3-e7e6efbd0401
-- upsert - userId = 22a4305a-6670-4de2-8de0-eeed85fe4227 - isActive = true
commited - userId = 22a4305a-6670-4de2-8de0-eeed85fe4227
```

when selecting -  "userId":"140da3b6-f213-4627-938a-d3b93f6f101f"
error is:
conflict on keys in range [[889df848-9a2e-4ed9-8126-a45c0c33a101], [889df848-9a2e-4ed9-8126-a45c0c33a101]), column isActive in table users

if i remove
tx.run(`SELECT * FROM ${tableName} WHERE eMail = '${user.eMail}' AND isActive = TRUE`,
        (err, rows) => 

everything is ok

Thank you

### LINKED ISSUE #2197: System test coverage for Cloud Spanner
_From @vkedia on January 19, 2017 7:20_

System tests for Cloud Spanner should cover all these scenarios:

1. Basic read,query and writes for all the data types.
2. Read with indexes, read limits, different concurrency options, range reads. 
3. Reading back large amount of data (lot of rows, large rows) to exercise the stream resume, flow control and chunking logic.
4. Parameter binding in query for all the data types including nulls, null arrays, arrays containing null, array of structs, special values for double type (Nan, Infinity,-Infinity).
5. Read/modify/write using transactions.
6. Tests for admin operations such as creating/deleting a database, updating database schema, pagination for list operations.

_Copied from original issue: GoogleCloudPlatform/google-cloud-node-private#20_

### LINKED ISSUE #2273: Spanner - transaction.runStream : fails for large query results : Uncaught TypeError: Cannot read property 'rowType' of null
#### Environment details

  - OS: Mac OSX 10.12.4
  - Node.js version: 6.7.0
  - npm version: 3.10.8
  - @google-cloud/spanner version: 0.4.2

#### Steps to reproduce

  1. require `@google-cloud/spanner`
  2. Run transaction.runStream on a large result set (~10k records)

```
        collectionSelf.connection.runTransaction(function(err, transaction) {
          if (err) {
            var msg = `${name} : could not run transaction : ${JSON.stringify(err)}`;
            log.error(msg);
            log.error(err.stack);
            reject(msg);
          }
          var results = [];
          transaction.runStream(query)
            .on('error', function(err) {
              log.error(`${name} : runStream error : ${JSON.stringify(err)}`);
              reject(err);
            })
            .on('data', function(row) {
              log.silly(`${name} : runStream : ${JSON.stringify(row)}`);
              results.push(row);
            })
            .on('end', function() {
              log.debug(`${name} : runStream end`);
              resolve(results);
            });
        });
```

3. The following error happens

Uncaught TypeError: Cannot read property 'rowType' of null
      at Function.partialResultStream.formatRow_ (node_modules/@google-cloud/spanner/src/partial-result-stream.js:167:28)

it seems that row.metadata is null in the response

https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/spanner/src/partial-result-stream.js#L164

As printing out row gives the following

{"values": [\<very long array of objets\>,"metadata":null,"chunkedValue":false,"resumeToken":[],"stats":null}]}

Are large result sets not supported even with runStream? Thanks

callmehiphop:
Closes #2273 
Closes #2197 
Closes #2176

### DIFF
--- a/packages/spanner/src/database.js
+++ b/packages/spanner/src/database.js
@@ -481,7 +481,7 @@ Database.prototype.getSchema = function(callback) {
  * //-
  * var query = {
  *   sql: 'SELECT * FROM Singers WHERE name = @name AND id = @id',
- *   params: {,
+ *   params: {
  *     id: spanner.int(8),
  *     name: null
  *   },


packages/spanner/src/codec.js
```diff
@@ -188,3 +188,89 @@ function encode(value) {
...
+ * Database.getType_(NaN);
+ * // 'float64'
+ */
+function getType(field) {
+  if (is.bool(field)) {
+    return 'bool';
+  }
+
+  var isSpecialNumber = is.infinite(field) ||
+    (typeof field === 'number' && isNaN(field));
```
stephenplusplus: Would `is.number` work here?

packages/spanner/src/database.js
```diff
@@ -577,6 +593,25 @@ Database.prototype.run = function(query, options, callback) {
  *   .on('end', function() {});
  *
  * //-
+ * // If you need to enforce a specific param type, a types map can be provided.
+ * // This is typically useful if your param value can be null.
```
stephenplusplus: If this is the only reason we need the types map (?), then I think it would be easier on the user to just give us a `null` value for the property that can be null.

packages/spanner/src/database.js
```diff
@@ -611,9 +653,44 @@ Database.prototype.runStream = function(query, options) {
...
   }
 
+  if (reqOpts.types) {
+    var types = {};
+
+    for (var prop in reqOpts.types) {
+      var type = reqOpts.types[prop]
+      var child;
+
+      if (is.object(type)) {
```
stephenplusplus: How does it happen that `type` is an object?

packages/spanner/src/database.js
```diff
@@ -598,11 +633,18 @@ Database.prototype.runStream = function(query, options) {
...
+  var fields = {};
+
   if (reqOpts.params) {
-    var fields = {};
+    reqOpts.types = reqOpts.types || {};
 
     for (var prop in reqOpts.params) {
       var field = reqOpts.params[prop];
+
+      if (!reqOpts.types[prop]) {
```
stephenplusplus: I think we need handling here if a user provides an invalid type, we should throw.

packages/spanner/src/table.js
```diff
@@ -347,10 +349,10 @@ Table.prototype.insert = function(keyVals, callback) {
  *     be yielded. If using a composite key, provide an array within this array.
  *     See the example below.
  * @param {string=} query.index - The name of an index on the table.
+ * @param {number} query.limit - The number of rows to return.
```
stephenplusplus: Let's also move the callback JSDocs below the `options`. They're a little out of sort.

packages/spanner/src/database.js
callmehiphop: Well, this was my solution for encountering Arrays since you need to specify that it is both an array and has a single child type.

packages/spanner/src/database.js
callmehiphop: When that happened I had been setting the types to 'unspecified' and assumed that the upstream would handle it, but we can throw an exception too

packages/spanner/src/database.js
callmehiphop: I know it seems odd, but in my system tests for querying types, if you don't specify a type for certain values the tests will fail because it assumes the wrong type. I think for the most part our users won't need to do this though

packages/spanner/src/database.js
stephenplusplus: Yeah, I think we want to be specific and throw in this case.

packages/spanner/src/database.js
stephenplusplus: Ah, that makes sense. Would you mind leaving a short code comment about it?

packages/spanner/src/codec.js
callmehiphop: Actually yes! I thought it did `NaN` sniffing, but it does not.

packages/spanner/src/database.js
stephenplusplus: Is it just this one case, where the user provides `null` and we have no way to know what the value's non-null type is? Or are there others as well?

This being the solution for the null thing makes sense to me. The only change I would suggest is, if that *is* the only case, we should update the example to show the value passed in as `name` being `null`. Otherwise, at a glance, it seems like we force explicit listing of parameter types, when we actually get it right most of the time.

packages/spanner/system-test/spanner.js
```diff
@@ -1002,141 +2515,356 @@ var spanner = new Spanner(env);
...
+              Key STRING(MAX) NOT NULL,
+              StringValue STRING(MAX)
+            ) PRIMARY KEY (Key)
+          */}));
+        })
+        .then(onPromiseOperationComplete)
+        .then(function() {
+          var promise = Promise.resolve();
+
+          for (var i = 0; i < 5; i++) {
```
stephenplusplus: Could you add some code comments explaining this?

packages/spanner/system-test/spanner.js
```diff
@@ -1002,141 +2515,356 @@ var spanner = new Spanner(env);
...
-          }, execAfterOperationComplete(next));
-        },
-        session.create.bind(session)
-      ], done);
-    });
+    before(function() {
+      return database.create()
+        .then(onPromiseOperationComplete)
+        .then(function() {
+          return table.create(multiline.stripIndent(function() {/*
```
stephenplusplus: We can just use backticks now:

```js
return table.create(`
  CREATE TABLE TxnTable (
  ...
`)
```

packages/spanner/src/transaction.js
```diff
@@ -283,11 +314,28 @@ Transaction.prototype.commit = function(callback) {
...
+
   var reqOpts = extend({
     session: this.session.formattedName_
   }, config.reqOpts);
 
-  config.method(reqOpts, callback);
+  var gaxOptions = reqOpts.gaxOptions;
+  delete reqOpts.gaxOptions;
+
+  config.method(reqOpts, gaxOptions, function(err, resp) {
```
callmehiphop: @stephenplusplus so I refactored the way we handle aborted errors from transactions - however, I'm not 100% how I should handle this for streams.

My understanding is that reads can trigger an abort error, so my thought would be to create a new through stream and pipe the gax stream to it, handle errors in the same way as below and if we get an aborted error to destroy the stream?

My overall concern here is that abort can probably occur in between requests, meaning the user could get a handful of results before we have to retry the whole transaction.

stephenplusplus: We might have to assign some type of reference ID to items we've already pushed to the user so we avoid pushing again. Or if the API has a way of starting from a specific token, so it won't respond with those items in the retries. And if that's not the right way to handle it, I think we are limited in solutions, because the point of the stream is to keep memory low and flush response items to the user as quickly as possible. If the request can be invalidated at an arbitrary time before the response is complete, we might need some advice from @vkedia on how this should be handled.

callmehiphop: @vkedia we need a little guidance here on if we should be discarding results from a stream that came in right before an aborted error.

callmehiphop: @stephenplusplus I've been thinking about this quite a bit this weekend and I think our best course of action would probably be to treat streams like we do everything else and fail it silently and retry the entire transaction.

At first I thought this was not such a great idea since it could error out mid-stream and our users might have to write code to undo any non-transaction things they might have done, but that pretty much applies to anything done in `runTransaction` at any time.

Because of this, I think we should create a public method that allows the user to get a transaction (`getTransaction`, `beginTransaction`, etc.) that does NOT do all the retry on abort stuff - that way if a user requires a higher level of control they can do whatever it is they want. This all has the added benefit of letting a user use promises.

stephenplusplus: I don't think letting the same stream mix both invalid and valid response items is an acceptable solution. I think we should be as loud about the failure as possible, and destroy the stream, so the user can decide to retry on their own.

@vkedia we could still use your guidance on when a retry should be performed, as well as how you see that working with a streaming request.

vkedia: Sorry for the late response. I was busy last week at I/O.
As @callmehiphop said, a stream failing with aborted should be treated like everything else and we should just retry the whole transaction. @stephenplusplus I am not sure what do you mean by "mixing valid and invalid responses". 
This does not seem any different from the scenario where user makes two read requests in the transaction and one of them succeed completely and the second one fails with aborted error. Even in that case if they have made some non transactional changes using the results of the previous stream, they can run into the scenario where they replay those changes when the first stream is retried. We can add the public method that allows users to do there own retry when someone asks for it or if users report issues using the automatic retry method.

packages/spanner/src/database.js
callmehiphop: No, when I wrote the tests for querying the various param types, I essentially copied the tests from the Java repo, in doing so certain queries will assume a type (in my case int64) and would throw an error if the param was not that same type. For whatever reason specifying the type made everything happy.

stephenplusplus: For those cases, we have the custom type methods like `spanner.int(8)`. If a user provided the values that was, as opposed to e.g. `8`, would that solve those cases?

callmehiphop: Well, if you were to supply a string, the upstream API is still expecting an INT64 and throws an error, so I don't think that's the correct solution here.

stephenplusplus: But if they provided us the value in it's correct custom type, we would be able to write out the correct configuration for them.

callmehiphop: And that's exactly what we do!

packages/spanner/src/database.js
stephenplusplus: I think we could still use a comment about this.

packages/spanner/src/table.js
stephenplusplus: We don't actually have two separate options parameters, right? It's just `query`?

packages/spanner/src/database.js
stephenplusplus: Can you show the cases where that doesn't work?

packages/spanner/src/table.js
callmehiphop: No, we have two. We can change this if you like though. Essentially the query object is specific to the query, options (which is optional) is defining single use transaction options (min timestamps, etc.)

packages/spanner/src/database.js
callmehiphop: Not sure I follow

packages/spanner/src/table.js
stephenplusplus: Ah, sorry for missing that. A quick thing we need to do is rename that parameter to match (`keyVals` => `query`). We can go with `query` as a name or `config`.

Beyond that, I think we should instead use `query.transactionOptions`, which future-proofs us when we add support for `query.gaxOptions`. Should we separate-issue that or do you want to get it done while we're in here?

packages/spanner/src/database.js
stephenplusplus: Would you mind changing the example to show the null case, and recommend passing values in their native-type format, e.g. `spanner.int` instead of just the number?

packages/spanner/src/transaction.js
```diff
@@ -347,14 +348,42 @@ Transaction.prototype.request = function(config, callback) {
...
+
   var reqOpts = extend({
     session: this.session.formattedName_
   }, config.reqOpts);
 
   var gaxOptions = reqOpts.gaxOptions;
   delete reqOpts.gaxOptions;
 
-  return config.method(reqOpts, gaxOptions);
+  // return config.method(reqOpts, gaxOptions);
```
stephenplusplus: Comment about a comment!

packages/spanner/src/transaction.js
```diff
@@ -347,14 +348,42 @@ Transaction.prototype.request = function(config, callback) {
...
+  if (!is.fn(self.runFn_)) {
+    return requestStream;
+  }
+
+  var userStream = through.obj();
+
+  requestStream.on('error', function(err) {
+    if (err.code !== ABORTED) {
+      userStream.emit('error', err);
+      return;
```
stephenplusplus: I believe the stream will still need to be ended, so you can do:

```js
userStream.destroy(err);
return;
```

callmehiphop: Done!

packages/spanner/src/database.js
callmehiphop: Ok, I've updated the example.. let me know if it is in line with what you were thinking

packages/spanner/src/database.js
```diff
@@ -480,11 +480,13 @@ Database.prototype.getSchema = function(callback) {
  * // This is typically useful if your param value can be null.
  * //-
  * var query = {
- *   sql: 'SELECT * FROM Singers WHERE name = @name',
- *   params: {
- *     name: 'Eddie Wilson'
+ *   sql: 'SELECT * FROM Singers WHERE name = @name AND id = @id',
+ *   params: {,
```
stephenplusplus: Rogue comma.

callmehiphop: Doh, nice find. Fixed!

PR #2314, opened 2017-05-17
error-reporting: Fix bug with errors not published

DominicKramer:
This commit fixes a bug where if
`errors.event().setMessage(string)` was used to build an error to
report, the error wouldn't appear in the error reporting console.

### DIFF
--- a/packages/error-reporting/README.md
+++ b/packages/error-reporting/README.md
@@ -160,7 +160,7 @@ errors.report('My error message');
 
 The stack trace associated with an error can be viewed in the error reporting console.
 * If the `errors.report` method is given an `ErrorMessage` object built using the `errors.event` method, the stack trace at the point where the error event was constructed will be used.
-* If the `errors.report` method is given an `Error` object, the stack trace where the error originated will be used.
+* If the `errors.report` method is given an `Error` object, the stack trace where the error was instantiated will be used.
 * If the `errors.report` method is given a string, the stack trace at the point where `errors.report` is invoked will be used.
 
 ### Using Express


packages/error-reporting/src/interfaces/manual.js
```diff
@@ -74,6 +74,10 @@ function handlerSetup(client, config) {
     }
 
     if (err instanceof ErrorMessage) {
+      var fauxError = new Error(err.message);
```
matthewloring: Are these error message objects constructed by the user (could they be queued before publishing)? We want to make sure that the stack reflects the accurate stack where the error was captured if possible.

packages/error-reporting/src/interfaces/manual.js
```diff
@@ -74,6 +74,10 @@ function handlerSetup(client, config) {
...
+      var fullStack = fauxError.stack.split('\n');
+      var cleanedStack = fullStack.slice(0, 1).concat(fullStack.slice(2)).join('\n');
```
matthewloring: I think `fullStack.splice(1, 1)` may make it more clear what this does. Could you add a comment describing why only the second frame needs to be removed?

DominicKramer: This matches what is done elsewhere in the code, but I'll add a comment for clarity.

matthewloring: If it is documented elsewhere that is fine as well.

packages/error-reporting/src/interfaces/manual.js
DominicKramer: A user does construct error message objects and, with this implementation, the user would see the stack trace at the point where they run `errors.report(errorMessage)`, not at the point where the error message object was created.

However, I think this is the correct approach.  That is, I think it would make more sense to get a trace back to the point where an error was reported.  Also, I expect the `errors.event()` method will probably be used to create messages with additional information rather than queueing messages.

If a customer has a use case where queuing is needed, we can address it at that point since we will then have a more clearly defined goal on what supporting queueing will need to look like.

packages/error-reporting/src/interfaces/manual.js
DominicKramer: I've added a comment.

packages/error-reporting/src/interfaces/manual.js
matthewloring: After discussion offline it sounds like there may be an efficient way to get creation site stack traces? This behavior will mirror that of native JS errors.

DominicKramer:
@matthewloring This is ready for another look.

packages/error-reporting/src/interfaces/message-builder.js
```diff
@@ -33,9 +33,26 @@ function handlerSetup(config) {
    * @returns {ErrorMessage} - returns a new instance of the ErrorMessage class
    */
   function newMessage() {
-    return new ErrorMessage().setServiceContext(
+    // The API expects a reported error to contain a stack trace.
+    // However, users do not need to provide a stack trace for stack
```
matthewloring: Should this be "for ErrorMessage objects using the message builder"?

packages/error-reporting/src/interfaces/manual.js
```diff
@@ -74,6 +76,22 @@ function handlerSetup(client, config) {
...
+      // append the stack trace of the point where the error was constructed.
+      // See the `message-builder.js` file for more details.
+      if (err._autoGeneratedStackTrace) {
+        err.setMessage(err.message + '\n' + err._autoGeneratedStackTrace);
+        // Delete the property so that if the ErrorMessage is reported a second
+        // time, the stack trace is not appended a second time.  Also, the API
+        // will not accept the ErrorMessage if it has additional properties.
+        delete err._autoGeneratedStackTrace;
+      }
+      else {
```
matthewloring: Move `else` up a line.

packages/error-reporting/src/interfaces/manual.js
```diff
@@ -74,6 +76,22 @@ function handlerSetup(client, config) {
     }
 
     if (err instanceof ErrorMessage) {
+      // The API expects the error to contain a stack trace.  Thus we
+      // append the stack trace of the point where the error was constructed.
+      // See the `message-builder.js` file for more details.
+      if (err._autoGeneratedStackTrace) {
```
matthewloring: Is it documented publicly that the ErrorMessage constructor will handle adding stack traces and that user messages should not include them?

packages/error-reporting/src/interfaces/message-builder.js
DominicKramer: Good catch.  Fixed.

packages/error-reporting/src/interfaces/manual.js
DominicKramer: Done

packages/error-reporting/src/interfaces/manual.js
DominicKramer: I've updated the `README`.  Can you take a quick look?

packages/error-reporting/README.md
```diff
@@ -145,19 +145,24 @@ var errors = require('@google-cloud/error-reporting')({
...
 errors.report(new Error('My error message'), () => console.log('done!'));
 
-// One can even just use a string
+// or one can even just use a string.
 errors.report('My error message');
 ```
 
+The stack trace associated with an error can be viewed in the error reporting console.
+* If the `errors.report` method is given an `ErrorMessage` object built using the `errors.event` method, the stack trace at the point where the error event was constructed will be used.
+* If the `errors.report` method is given an `Error` object, the stack trace where the error originated will be used.
```
matthewloring: Maybe "...where the error was instantiated..." would be more precise?

DominicKramer: Updated.

DominicKramer:
@stephenplusplus This is ready to land since the broken tests appear to be unrelated to the `error-reporting` library.

Here is the end of the test failure:
```

module.js:327
    throw err;
    ^

Error: Cannot find module '/home/ubuntu/google-cloud-node/packages/video-intelligence/package.json'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at new Module (/home/ubuntu/google-cloud-node/scripts/helpers.js:63:17)
    at Module (/home/ubuntu/google-cloud-node/scripts/helpers.js:57:12)
    at Array.map (native)
    at Function.Module.getUpdated (/home/ubuntu/google-cloud-node/scripts/helpers.js:142:24)
    at Object.<anonymous> (/home/ubuntu/google-cloud-node/scripts/build.js:50:22)
    at Module._compile (module.js:409:26)

. scripts/circle/test.sh returned exit code 1
```

stephenplusplus:
Thanks!

PR #2315, opened 2017-05-17
error-reporting: Address unresponsive Hapi apps

DominicKramer:
The Hapi plugin was too restrictive when determining if it should
continue a reply, which could cause some Hapi apps to become
unresponsive.

### DIFF
--- a/packages/error-reporting/test/unit/interfaces/hapi.js
+++ b/packages/error-reporting/test/unit/interfaces/hapi.js
@@ -114,31 +114,31 @@ describe('Hapi interface', function() {
       fakeServer.removeAllListeners();
     });
     it('Should call continue when a boom is emitted if reply is an object',
-    function(done) {
-      plugin.register(fakeServer, null, function() {});
-      fakeServer.emit(EVENT, {response: {isBoom: true}},
-        {
-          continue: function() {
-            // The continue function should be called
-            done();
+      function(done) {
+        plugin.register(fakeServer, null, function() {});
+        fakeServer.emit(EVENT, {response: {isBoom: true}},
+          {
+            continue: function() {
+              // The continue function should be called
+              done();
+            }
           }
-        }
-     );
+       );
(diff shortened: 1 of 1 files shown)


DominicKramer:
I've added a test.  It looks like we thought that the `reply` object given to listeners of the `onPreResponse` event would satisfy `is.object(reply)`.  As a result, our test made a `reply` object using an object literal in the test.

However, I made a Hapi app and found that it is possible that `reply` is a function that has a `reply.continue`.  Thus, I've added a test to cover the case when `reply` is a function.

packages/error-reporting/test/unit/interfaces/hapi.js
```diff
@@ -113,7 +113,8 @@ describe('Hapi interface', function() {
     afterEach(function() {
       fakeServer.removeAllListeners();
     });
-    it('Should call continue when a boom is emitted', function(done) {
+    it('Should call continue when a boom is emitted if reply is an object',
+    function(done) {
```
ofrobots: style nit: this line should be indented

packages/error-reporting/test/unit/interfaces/hapi.js
```diff
@@ -124,6 +125,21 @@ describe('Hapi interface', function() {
         }
      );
     });
+    it('Should call continue when a boom is emitted if reply is a function',
+    function(done) {
```
ofrobots: indentation

packages/error-reporting/test/unit/interfaces/hapi.js
DominicKramer: Fixed

packages/error-reporting/test/unit/interfaces/hapi.js
DominicKramer: Fixed

DominicKramer:

PR #2318, opened 2017-05-17
Video Intelligence (GAPIC only)

lukesneeringer:
This is the video intelligence (GAPIC only) release for Node.js.

### DIFF
--- a/packages/video-intelligence/package.json
+++ b/packages/video-intelligence/package.json
@@ -14,8 +14,7 @@
     "extend": "^3.0.0"
   },
   "devDependencies": {
-    "mocha": "3.2.0",
-    "through2": "2.0.3"
+    "mocha": "3.2.0"
   },
   "license": "Apache-2.0",
   "engines": {


packages/video-intelligence/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "url": "https://github.com/googleapis/googleapis",
+  "name": "@google-cloud/video-intelligence",
+  "version": "0.1.0",
```
stephenplusplus: Start this at `0.0.0`. Our `publish-module` script handles bumping it 👍

packages/video-intelligence/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "devDependencies": {
+    "mocha": "3.2.0",
+    "through2": "2.0.3"
+  },
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=4.0.0"
+  },
+  "scripts": {
+    "test": "mocha"
```
stephenplusplus: Add a "publish-module" script:

```js
"publish-module": "node ../../scripts/publish.js video-intelligence",
```

packages/video-intelligence/src/index.js
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright 2016 Google Inc. All rights reserved.
```
stephenplusplus: 2017

packages/video-intelligence/src/index.js
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ 'use strict';
+
+ var gapic = {
+   v1beta1: require('./v1beta1'),
```
stephenplusplus: It's a best practice to reference the direct file, saving some overhead at runtime-- not sure of the specifics, but I read of this being necessary going forward when more robust module loaders are available. In this case, I think we want `v1beta1/index.js`.

packages/video-intelligence/src/index.js
```diff
@@ -0,0 +1,55 @@
...
+ };
+
+ const VERSION = require('../package.json').version;
+
+
+ /**
+  * Create an imageAnnotatorClient with additional helpers for common
```
stephenplusplus: It doesn't look like we have additional helpers, so I don't think we would want to include this in the docs.

packages/video-intelligence/src/index.js
```diff
@@ -0,0 +1,55 @@
...
+  * tasks.
+  *
+  * @param {Object=} opts - The optional parameters.
```
stephenplusplus: This doesn't mention how authentication is handled-- does it accept `opts.keyFilename`, `opts.credentials`?

packages/video-intelligence/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "url": "https://github.com/googleapis/googleapis",
```
stephenplusplus: This property isn't necessary. `npm` figures it out based on the `repository` property, which we need to add:

```js
"repository": "googlecloudplatform/google-cloud-node",
```

packages/video-intelligence/src/index.js
```diff
@@ -14,41 +14,59 @@
...
+  var client = v1beta1(options).videoIntelligenceServiceClient(options);
   return client;
 }
 
 // The default export should be the latest version.
 // Assign all versions as version properties on the default.
-module.exports = videointelligenceV1beta1;
-module.exports.v1beta1 = videointelligenceV1beta1;
+module.exports = VideoIntelligence;
+module.exports.v1beta1 = v1beta1;
```
stephenplusplus: This is how we usually do it-- we export the raw gapic module as the version number, without a wrapper.

packages/video-intelligence/src/index.js
```diff
@@ -14,41 +14,59 @@
...
-
 /**
- * Create an imageAnnotatorClientfor the appropriate version.
- *
- * @param {Object=} opts - The optional parameters.
- * @param {String=} opts.servicePath
- *   The domain name of the API remote host.
- * @param {number=} opts.port
- *   The port on which to connect to the remote host.
- * @param {grpc.ClientCredentials=} opts.sslCreds
```
stephenplusplus: I removed the custom type references (`grpc.ClientCredentials` and `gax.constructSettings`) for now, because they wouldn't link to anything or make much sense. We can find a better solution when we get GAX docs out... or I can revert them if you'd still like them there.

stephenplusplus:
@lukesneeringer PTAL and feel free to merge. That will be my queue to get out the 0.1.0 release.

PR #2324, opened 2017-05-21
Allow a base header for setEncryptionKey.

### LINKED ISSUE #2322: Storage - file.js. Any way to set Source Encryption Key?
Hi,
In File.js of the Storage package, the function setEncryptionKey() helps set the "x-goog-encryption-key" header. Is there a function to set the "x-goog-copy-source-encryption-key header"? Can a function be added for this please?

lukesneeringer:
@subra-vr Let me know if this gives you what you need.

Fixes #2322.

### DIFF
--- a/packages/storage/src/file.js
+++ b/packages/storage/src/file.js
@@ -1164,7 +1164,7 @@ File.prototype.download = function(options, callback) {
  * });
  */
 File.prototype.setEncryptionKey = function(encryptionKey, options) {
-  options = options || {}
+  options = options || {};
   options.baseHeader = options.baseHeader || 'x-goog-encryption';
   this.encryptionKey = encryptionKey;
 
--- a/packages/storage/test/file.js
+++ b/packages/storage/test/file.js
@@ -2427,10 +2427,9 @@ describe('File', function() {
 
   describe('setEncryptionKey', function() {
     var KEY = crypto.randomBytes(32);
-    var _file;
 
     it('should localize the key', function() {
-      var _file = file.setEncryptionKey(KEY);
+      file.setEncryptionKey(KEY);
       assert.strictEqual(file.encryptionKey, KEY);
     });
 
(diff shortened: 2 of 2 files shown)


subra-vr:
@lukesneeringer Thanks, this will work.

stephenplusplus:
The `encryptionKey` option is accepted in a couple other places:

- The `File` constructor: https://github.com/lukesneeringer/google-cloud-node/blob/185bd501ecfd9b6bbf398c8251ca6852a6f85ab6/packages/storage/src/file.js#L83
- `bucket.file`: https://github.com/lukesneeringer/google-cloud-node/blob/185bd501ecfd9b6bbf398c8251ca6852a6f85ab6/packages/storage/src/bucket.js#L638 (note: there is apparently a typo in these docs-- the property name is not `key`)
- `bucket.upload`: https://github.com/lukesneeringer/google-cloud-node/blob/185bd501ecfd9b6bbf398c8251ca6852a6f85ab6/packages/storage/src/bucket.js#L1059

 Should we accept a `sourceEncryptionKey` option where these are as well?

subra-vr:
@stephenplusplus I guess your query was targeted at @lukesneeringer as he requested the review? But in the interest of moving things along so I can get my code finalised, here goes. 
My original request was specifically for x-goog-copy-source-encryption-* headers which are applicable for rewrite operations. On the other hand, the above four finally end up being consumed in File object constructors, and something generic like options.baseHeader can be confusing there. So my opinion would be that options.baseHeader is enough just in the setEncryptionKey function. Over to @lukesneeringer. Thanks.

subra-vr:
Hi @lukesneeringer @stephenplusplus , this is the first time I have raised an issue on a forum like this, so please bear with me if I ask the obvious. Any help on how and when this will get into the master branch?

conundrum888:
@lukesneeringer Any chance this will be merged soon? Thanks.

conundrum888:
Hi, my project really needs a way for the SDK to set source encryption key. Is the coverage check the reason this pull request is not merged?

Issue #2251, opened 2017-04-24
improved processing of structured logs


travi:
I recently got `@google-cloud/logging-bunyan` working in a [hapi](http://hapijs.com/) app (leveraging [good-bunyan](https://github.com/muzzley/good-bunyan)) after a very helpful conversation with @ofrobots. 

This got us to a much better state of logging than our previous use of [good-console](https://github.com/hapijs/good-console) to simply write to `stdout`. However, the log events that are sent to Stackdriver are still not tokenized nearly as well as the logs that we have for some other (non-node) apps we have running in the standard environment.

It appears that the events sent through `@google-cloud/logging-bunyan` are structured in the `bunyan` [recommended formats](https://github.com/trentm/node-bunyan#recommendedbest-practice-fields), rather than the Stackdriver format. If I understand the docs correctly, Stackdriver would automatically process these better if they were transformed to meet the formats like:
* https://cloud.google.com/logging/docs/reference/v1beta3/rest/v1beta3/RequestLog
* https://cloud.google.com/logging/docs/reference/v1beta3/rest/v1beta3/LogLine

I have not looked very deeply into trying to perform any transformations myself because I wanted to see if the team agreed that it made more sense to be part of the `@google-cloud/logging-bunyan` library directly. It seems like that would be the most helpful place so that each consumer doesn't need to build the transformations themselves (if event possible).

Would it be reasonable to add this type of transformation into the library to support better log tokenization?

ofrobots:
Hi @travi, I think it would be a good idea to have more structured request logs. The challenge is that, unlike the standard environment, there are too many combinations of web frameworks and logging middleware involved, each with their own way of formatting log entries. In the Node.js case, hapi + good seems to have very [specific way](https://github.com/muzzley/good-bunyan/blob/master/lib/formatters.js) of formatting log entries. Restify is [entirely different](http://restify.com/#audit-logging). Users may configure and/or modify these formatters making it hard to predict how the data would be structured.

In other words, it is quite difficult to predict if the `userAgent` is available through `req.headers['user-agent]` (raw http request), `response.source.userAgent` (hapi + good), or is not available at all (restify doesn't provide headers by default). It is very hard (error prone) to build structure out of unstructured input.

One way to deal with this would be:

* If the bunyan logging record contains a `httpRequest` property, we will report it as HttpRequest field of the LogEntry (as opposed to leaving it as part of the unstructured payload).
[...]

PR #2325, opened 2017-05-22
logging-bunyan: HttpRequest support

ofrobots:
If the bunyan provided metadata includes a httpRequest property, format
that as the HttpRequest message. This allows Stackdriver logging to
format request logs nicely.

Fixes: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2251

EDIT: note that this is semver-major.

### DIFF
--- a/packages/logging-bunyan/src/index.js
+++ b/packages/logging-bunyan/src/index.js
@@ -173,7 +173,7 @@ LoggingBunyan.prototype.formatEntry_ = function(record) {
   // If the record contains a httpRequest property, provide it on the entry
   // metadata. This allows Stackdriver to use request log formatting.
   // https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest
-  // Note that the httpRequest field must properly validate as a HttpRequest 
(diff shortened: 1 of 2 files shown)


packages/logging-bunyan/README.md
```diff
@@ -66,7 +66,29 @@ var loggingBunyan = require('@google-cloud/logging-bunyan')({
 // ...you're good to go!
 ```
 
+## Formatting Request Logs
+
+To format your request logs you can provide a `httpRequest` property on the bunyan metadata you provide along with the log message. We will format this as the [`HttpRequest`][http-request-message] message and Stackdriver logging will show this as a request log. Example:
```
stephenplusplus: Instead of `we will format this`, maybe we should say, `we will treat this as` or `we will consider this`... the former might suggest we are going to take potentially invalid input and "correct" it.

ofrobots: Done.

ofrobots:
CI is green now. PTAL.

stephenplusplus:
Cool, thanks!

travi:
thanks so much for this update! is there a release planned soon so i can give it a shot?

PR #2330, opened 2017-05-24
Allow for custom host with pubsub

mkamioner:
When using the emulator, I don't want to be restricted to using the environment variable of  `PUBSUB_EMULATOR_HOST` - Especially when I have certain topics in the cloud and other in an emulator. 

This will allow the programmer to decide what host to use easily

### DIFF
--- a/packages/pubsub/test/index.js
+++ b/packages/pubsub/test/index.js
@@ -1043,6 +1043,8 @@ describe('PubSub', function() {
     it('should set base url to parameter sent', function() {
       var defaultBaseUrl_ = 'defaulturl';
       var testingUrl = 'localhost:8085';
+
+      setHost(defaultBaseUrl_);
       pubsub.defaultBaseUrl_ = defaultBaseUrl_;
 
       pubsub.determineBaseUrl_(testingUrl);


mkamioner:
I signed it!

stephenplusplus:
Thank you for this! We support this in the Datastore API, but haven't brought it over here yet, so your help is appreciated!

To match what's going on in Datastore, could you rename `host` to `apiEndpoint`? You can see (and please add) the JSDocs description here: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/13ec5fef0da92b74e7adf7a2479f8c1636be72d4/packages/datastore/src/index.js#L61

And if you're up for it, we could use some tests as well. You can probably get by with copying what we have in Datastore: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/13ec5fef0da92b74e7adf7a2479f8c1636be72d4/packages/datastore/test/index.js#L363 -- if you'd rather have me do it, no problem, just let me know.

Thanks again!

mkamioner:
@stephenplusplus  - Happily! I will do it now!

Happy to contribute :-)

stephenplusplus:
It's perfect, thank you very much again!

mkamioner:
Honor and a pleasure :-)

PR #2336, opened 2017-05-26
error-reporting: Fix an issue where authorization would fail

DominicKramer:
The `keyFilename` and `credentials` values specified in the
configuration to the error-reporting library were not being
used to communicate through the API.  This would cause
communication through the error-reporting API to fail.


DominicKramer:
A follow-up PR will address adding regression tests.

DominicKramer:

PR #2344, opened 2017-05-31
spanner: make getTransaction public

### LINKED ISSUE #2152: spanner: how should transactional promises look?
*[Originally posted by @callmehiphop](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2118#issuecomment-289919717)*

@vkedia @stephenplusplus @lukesneeringer @jgeewax so our typical callback vs. promise convention is pretty simple - omit the callback and you shall receive a promise. I think this is the first API that really complicates this approach.

Consider the following example

```js
database.runTransaction()
  .then(transaction => {
    return Promise.all([
      transaction,
      transaction.getRows('Singers')
    ]);
  })
  .then([transaction, rows] => {
    // do some kind of mutation based on the rows
    return transaction.commit();
  })
  .then(() => {
    // do something async/unrelated to the transaction
  });
```

My main issue is that in order to properly retry our transaction here, we would need to capture all functions passed to the promise before `transaction.commit()` and retry them in the correct order before allowing any chaining that occurs after to continue. Technically I don't think this is even feasible, but even if it were I think the overall behavior itself is totally unpredictable for our users. Based on this opinion, I think that our traditional approach will not fly here.

That being said, if we were to continue to try and let both callbacks and promises live together within a single method, we need a different way of letting the user tell us that they want promises.

My current suggestion is to let the user return a promise within the callback (like so)

```js
database.runTransaction((err, transaction) => {
  if (err) return Promise.reject(err);

  // do some transactional based stuff
  return transaction.commit();
})
.then(() => {
  // transaction is committed, so some other stuff now
});
```

However after some discussion in my PR, this also may be confusing for the users. So I would like to revisit this specific API since we don't have a precedent for this specific issue.

An idea that we've been kicking around would basically be to not directly support promises in *and* out of this method, but *only* in.. This means the code would resemble the following

```js
database.runTransaction((err, transaction) => {
  if (err) {
    // an error occurred while trying to get/create a session + transaction
  }

  // do reads/writes/etc. here
  
  // now the user has a choice between using a promise or a callback
  // and while this function will be ran multiple times in the event of an aborted error
  // commit will not resolve until either a success or final error occurs
  transaction.commit()
    .then(() => console.log('committed successfully!'))
    .catch((err) => console.error(err));
});
```

At which point if a user doesn't feel that this is the way they want to use Promises, it would be fairly easy to wrap that into something like

```js
function runTransactions() {
  return new Promise((resolve, reject) => {
    database.runTransaction((err, transaction) => {
      if (err) return reject(err);
      // do transaction stuff
      transaction.commit().then(resolve, reject);
    });
  });
}
```

callmehiphop:
Closes #2152

### DIFF
--- a/packages/spanner/src/database.js
+++ b/packages/spanner/src/database.js
@@ -402,9 +402,21 @@ Database.prototype.getSchema = function(callback) {
 /**
  * Get a read/write ready Transaction object.
  *
- * @param {object=} options - Transaction options.
- * @param {boolean} options.readOnly - Specifies if the transaction is read
- *     only.
+ * @param {object=} options - [Transaction options](https://cloud.google.com/spanner/docs/timestamp-bounds).
+ * @param {number} options.timeout - Specify a timeout for the transaction. The
+ *     transaction will be ran in its entirety, however if an abort error is
+ *     returned the transaction will be retried if the timeout has not been met.
+ *     Default: `60000` (milliseconds)
+ * @param {boolean} options.readOnly - Specifies if the transaction is
+ *     read-only. Default: `false`.
+ * @param {number} options.exactStaleness - Executes all reads at the timestamp
+ *     that is `exactStaleness` old.
+ * @param {date} options.readTimestamp - Execute all reads at the given
+ *     timestamp.
+ * @param {boolean} options.returnTimestamp - If `true`, returns the read
+ *     timestamp.
+ * @param {boolean} options.strong - Read at the timestamp where all previously
+ *     committed transactions are visible.
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while getting the
  *     transaction object.
@@ -448,6 +460,7 @@ Database.prototype.getTransaction = function(options, callback) {
 
     transaction.begin(function(err) {
       if (err) {
+        transaction.end();
         callback(err);
         return;
       }
--- a/packages/spanner/test/database.js
+++ b/packages/spanner/test/database.js
@@ -1150,11 +1150,15 @@ describe('Database', function() {
 
       it('should return an error if transaction cannot begin', function(done) {
         var error = new Error('err');
+        var endCalled = false;
 
         var SESSION = {};
         var TRANSACTION = {
           begin: function(callback) {
             callback(error);
+          },
+          end: function() {
+            endCalled = true;
           }
         };
 
@@ -1169,6 +1173,7 @@ describe('Database', function() {
 
         database.getTransaction(OPTIONS, function(err) {
           assert.strictEqual(err, error);
+          assert.strictEqual(endCalled, true);
           done();
         });
       });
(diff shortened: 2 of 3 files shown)


bjwatson:
Thanks @callmehiphop! Adding @vkedia as a reviewer, since he understands this code better than I do.

packages/spanner/src/database.js
```diff
@@ -400,6 +400,64 @@ Database.prototype.getSchema = function(callback) {
 };
 
 /**
+ * Get a read/write ready Transaction object.
+ *
+ * @param {object=} options - Transaction options.
+ * @param {boolean} options.readOnly - Specifies if the transaction is read
+ *     only.
```
vkedia: please document that by default this is false.

packages/spanner/src/database.js
```diff
@@ -400,6 +400,64 @@ Database.prototype.getSchema = function(callback) {
 };
 
 /**
+ * Get a read/write ready Transaction object.
+ *
+ * @param {object=} options - Transaction options.
+ * @param {boolean} options.readOnly - Specifies if the transaction is read
```
vkedia: what about timestamp bound for read only txns?

packages/spanner/src/database.js
```diff
@@ -400,6 +400,64 @@ Database.prototype.getSchema = function(callback) {
...
+    }
+
+    options = extend({}, options);
+    delete options.readOnly;
+
+    var transaction = self.pool_.createTransaction_(session, options);
+
+    transaction.begin(function(err) {
+      if (err) {
+        callback(err);
```
vkedia: dont you need to return the session to the pool?

callmehiphop: That's up to the user, when we call [`createTransaction_`](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/spanner/src/session-pool.js#L401) we override the `end` method which should get called on commit/rollback/etc., releasing the session back to the pool.

vkedia: But in this error case since the transaction is not even passed to the user specified callback, how would they end it?

bjwatson:
@callmehiphop Can you close this out this week?

callmehiphop:
@bjwatson yes! Sorry, I've been side tracked by other things.

bjwatson:
Thanks @callmehiphop!

bjwatson:
@swcloud FYI, we want to get this done this week.

packages/spanner/src/database.js
callmehiphop: Ah, yes, you are correct!

callmehiphop:
@stephenplusplus @vkedia this should be good to go now!

packages/spanner/src/database.js
```diff
@@ -400,6 +400,77 @@ Database.prototype.getSchema = function(callback) {
 };
 
 /**
+ * Get a read/write ready Transaction object.
+ *
+ * @param {object=} options - [Transaction options](https://cloud.google.com/spanner/docs/timestamp-bounds).
```
stephenplusplus: I think we need the `https://` for this to work in the docs.

stephenplusplus: Apparently, mobile UI strips https:// or I'm crazy. Ignore this.

PR #2363, opened 2017-06-06
error-reporting: Fix `report()` silently failing

DominicKramer:
If the `report()` method is called with a number, an associated
error never appears in the error reporting console, and no log
message is printed to the user to indicate why the error has not
appeared.

### DIFF
--- a/packages/error-reporting/src/build-stack-trace.js
+++ b/packages/error-reporting/src/build-stack-trace.js
@@ -16,32 +16,27 @@
 
 'use strict';
 
-var is = require('is');
+const SRC_ROOT = __dirname;
 
 /**
  * Constructs a string representation of the stack trace at the point where
  * this function was invoked.  Note that the stack trace will not include any
- * references to this function itself.
+ * references to frames specific to this error-reporting library itself.
  * @param {String?} message - The message that should appear as the first line
  *   of the stack trace.  This value defaults to the empty string.
- * @param {Number?} numSkip - The number of additional lines to not display
- *   in the stack trace.  That is, the stack trace returned will have the
- *   specified message with the top `numSkip` lines of the stack not displayed.
- *   This value defaults to 0 if it is not defined or is not a number.
  * @returns {String} - A string representation of the stack trace at the point
  *   where this method was invoked.
  */
-function buildStackTrace(message, numSkip) {
+function buildStackTrace(message) {
   var target = {};
   // Build a stack trace without the frames associated with `buildStackTrace`.
   // The stack is located at `target.stack`.
   Error.captureStackTrace(target, buildStackTrace);
-  // Ignore the first line, which contains the error message, because we will
-  // add our own error message later.
-  var fullStack = target.stack.split('\n').slice(1);
   var prefix = message ? message + '\n' : '';
-  var realNumSkip = is.number(numSkip) ? numSkip : 0;
-  return prefix + fullStack.slice(realNumSkip).join('\n');
+  return prefix + target.stack.split('\n').slice(1).filter(function(line) {
+    // Filter out all frames that are specific to the error-reporting library
+    return !line || line.indexOf(SRC_ROOT) === -1;
+  }).join('\n');
 }
 
 module.exports = buildStackTrace;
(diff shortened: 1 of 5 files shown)


packages/error-reporting/package.json
```diff
@@ -4,7 +4,7 @@
   "main": "./src/index.js",
   "repository": "GoogleCloudPlatform/google-cloud-node",
   "scripts": {
-    "test": "nyc --exclude=\"fuzzer.js\" mocha ./test/unit/**/*.js",
+    "test": "nyc --exclude=\"fuzzer.js\" mocha ./test/unit/*.js ./test/unit/**/*.js",
```
stephenplusplus: With the new pattern matching expression, you should be able to remove `./test/unit/*.js`

DominicKramer: On my machine, without the `./test/unit/*.js`, the js files in the `./test/unit` directory are not being tested when `npm test` is run.

packages/error-reporting/src/build-stack-trace.js
```diff
@@ -0,0 +1,45 @@
...
+
+'use strict';
+
+var is = require('is');
+
+/**
+ * Constructs a string representation of the stack trace at the point where
+ * this function was invoked.  Note that the stack trace will not include any
+ * references to this function itself.
+ * @function buildStackTrace
```
DominicKramer: Removed.

packages/error-reporting/src/interfaces/message-builder.js
```diff
@@ -44,9 +45,7 @@ function handlerSetup(config) {
     //       error is used instead of the stack trace where the error is
     //       reported to be consistent with the behavior of reporting a
     //       an error when reporting an actual Node.js Error object.
-    var fauxError = new Error('');
-    var fullStack = fauxError.stack.split('\n');
-    var cleanedStack = fullStack.slice(2).join('\n');
+    var cleanedStack = buildStackTrace('', 1);
```
DominicKramer: I've updated `buildStackTrace` to use `Error.captureStackTrace`.  To remove all error-reporting frames would require restructuring the `src/interfaces/manual.js` file to expose the `reportManualError` function.  This would take more refactoring than I think is needed at this point.

packages/error-reporting/src/populate-error-message.js
```diff
@@ -0,0 +1,179 @@
...
+ * @function populateErrorMessage
+ * @param {Any} ob - the object information to extract from
+ * @param {ErrorMessage} em - an instance of ErrorMessage to marshal object
+ *  information into
+ * @returns {Undefined} - does not return a value
+ */
+function populateErrorMessage(ob, em) {
+  if (ob === null || ob === undefined) {
+    populateFromUnknown(ob, em);
+  } else if (ob instanceof Error) {
```
DominicKramer: Yes.  Updated.

packages/error-reporting/src/populate-error-message.js
```diff
@@ -0,0 +1,179 @@
...
+ * stack trace for submission to the API and check to confirm that the given
+ * value is of type string.
+ * @function populateFromString
+ * @param {String} str - the String indicated as the content of the error
+ * @param {ErrorMessage} errorMessage - the error message instance to marshal
+ *  error information into.
+ * @returns {Undefined} - does not return anything
+ */
+function populateFromString(str, errorMessage) {
+  errorMessage.setMessage(buildStackTrace(str, 3));
```
DominicKramer: It is.  I've consolidated the code.

packages/error-reporting/src/populate-error-message.js
```diff
@@ -0,0 +1,179 @@
...
+ * a stack-trace for submission to the Error API and will attempt to caste the
+ * given number to a string for submission to the Error API.
+ * @function populateFromNumber
+ * @param {Number} num - the number submitted as content for the error message
+ * @param {ErrorMessage} errorMessage - the error messag instance to marshall
+ *  error information into.
+ * @returns {Undefined} - does not return anything
+ */
+function populateFromNumber(num, errorMessage) {
+  var message = isNumber(num) && isFunction(num.toString) ? num.toString() : '';
```
DominicKramer: I've consolidated the code so that this method is not needed.

DominicKramer:

PR #2366, opened 2017-06-07
error-reporting: Elaboration on GKE Scopes in Docs

kjin:
Added an example CLI for getting the right scope in GKE.

### DIFF
--- a/packages/error-reporting/README.md
+++ b/packages/error-reporting/README.md
@@ -40,7 +40,7 @@ runtime configuration object is set to `true`.  See the [Configuration](#configu
 
   In your project, on the command line:
 
-```bash
+```sh
   # Install through npm while saving to the local 'package.json'
   npm install --save @google-cloud/error-reporting
 ```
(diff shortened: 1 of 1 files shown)


packages/error-reporting/README.md
```diff
@@ -96,15 +96,26 @@ If you already have VMs that were created without API access and do not wish to
...
 
-Container Engine nodes need to also be created with the `https://www.googleapis.com/auth/cloud-platform` scope, which is configurable during cluster creation. Alternatively, you can follow the instructions for using a service account under [running elsewhere](#running-elsewhere). It's recommended that you store the service account credentials as [Kubernetes Secret](http://kubernetes.io/v1.1/docs/user-guide/secrets.html).
+As with Compute Engine, Container Engine nodes need to be created with the `https://www.googleapis.com/auth/cloud-platform` scope, which is configurable during cluster creation:
+* If the cluster is being created with the `gcloud` CLI, pass the scope to the command with the `--scopes` command (multiple scopes can be [delimited with a comma](https://cloud.google.com/sdk/gcloud/reference/container/clusters/create)):
+
+  ```sh
+  gcloud container clusters create example-cluster-name --scopes https://www.googleapis.com/auth/cloud-platform
+  ```
+
+* If the cluster is being created through the Cloud Console UI, ensure that the "Cloud Platform" project access is set to "Enabled" (it's disabled by default).
```
DominicKramer: grammar nit: I think "it's" should be capitalized or the parenthetical text could be a separate sentance.

kjin: I disagree -- "it's" should not be capitalized. Even excluding capitalization rules, this is the convention in our top-level README.

PR #2369, opened 2017-06-08
RFC: express request logger

ofrobots:
**Do not land yet.**

It would be nice if we could get AppEngine-standard like request logs for all environments people are using Node.js express servers in. Today we provide [documentation](https://github.com/GoogleCloudPlatform/google-cloud-node/tree/master/packages/logging-winston#formatting-request-logs) on how to properly attach a `HttpRequest` message to a log message for it to get formatted as a request. It would be nice if users wouldn't have to do that manually.

This PR adds support for getting an express middleware.

@stephenplusplus: any thoughts?

Potential topics for discussion
* Normally we would add framework integrations into logging-{winston,bunyan}. In this case, the bulk of the code is independent of winston, bunyan. Any comments with the integration being present in shared g-c/logging code? I would expect that we would want to add hapi, restify and probably koa integrations too down the road. Another option would be to introduce middle tier module that could be shared by logging-{winston,bunyan}, but I am not sure if the complexity is worth it. The code being added to g-c/logging does have utility independent of logging-{winston,bunyan} e.g. for folks that want to use a different logger (say pino).

### DIFF
--- a/packages/logging-winston/package-lock.json
+++ b/packages/logging-winston/package-lock.json
@@ -1,0 +1,1629 @@
+{
+  "name": "@google-cloud/logging-winston",
+  "version": "0.4.1",
+  "lockfileVersion": 1,
+  "dependencies": {
+    "@google-cloud/common": {
+      "version": "https://registry.npmjs.org/@google-cloud/common/-/common-0.13.3.tgz",
+      "integrity": "sha1-1z7j+lEfKf+8xEZniYaFcJ6f7Iw="
+    },
+    "@google-cloud/common-grpc": {
+      "version": "https://registry.npmjs.org/@google-cloud/common-grpc/-/common-grpc-0.3.4.tgz",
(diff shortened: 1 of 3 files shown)


stephenplusplus:
If you're a user of one of those frameworks, you might not be in need of any advanced Logging API controls. You are going to look on npmjs.com for "logging express". If you get the big, `@google-cloud/logging` module in your result, it's up to our docs to quickly point you to the `requestExpress()` method, otherwise you're overwhelmed. I would lean towards having modules that stand alone to serve as helpers for these integrations, i.e. `@google-cloud/logging-express`, `@google-cloud/logging-hapi`, etc.

ofrobots:
I think I agree with standalone modules as being the better path. The hesitation was that the standalone expressLogger module will need to have a `module:logging/log` object injected into it. However, given that logging-{winston,bunyan} are going to the starting point for most framework users, I don't think this detail gets exposed to the users.

Closing.

PR #2384, opened 2017-06-13
model is not whitelisted anymore

merla18:
we went GA back in April. the parameter is no more whitelisted on Google Translation API


stephenplusplus:
Thanks!

PR #2387, opened 2017-06-14
storage/bucket: add label accessor methods

### LINKED ISSUE #2370: Add GCS bucket labels feature

stephenplusplus:
Fixes #2370

#### To Dos

- [x] Docs
- Tests
  - [x] Unit
  - [x] System

This adds three methods:

- `bucket.deleteLabels([labelName, ...], function(err) {})`
- `bucket.getLabels(function(err, labels) {})`
- `bucket.setLabels({ labelName: 'labelValue' }, function(err) {})`

### DIFF
--- a/packages/storage/src/bucket.js
+++ b/packages/storage/src/bucket.js
@@ -646,12 +646,15 @@ Bucket.prototype.deleteFiles = function(query, callback) {
  * //-
  * // Delete a single label.
  * //-
- * bucket.deleteLabels('label', function(err, metadata) {});
+ * bucket.deleteLabels('labelone', function(err, metadata) {});
  *
  * //-
  * // Delete a specific set of labels.
  * //-
- * bucket.deleteLabels(['label', 'labeltwo'], function(err, metadata) {});
+ * bucket.deleteLabels([
+ *   'labelone',
+ *   'labeltwo'
+ * ], function(err, metadata) {});
  *
  * //-
  * // If the callback is omitted, we'll return a Promise.
@@ -1148,7 +1151,7 @@ Bucket.prototype.makePublic = function(options, callback) {
  *
  * @example
  * var labels = {
- *   label: 'labelvalue',
+ *   labelone: 'labelonevalue',
  *   labeltwo: 'labeltwovalue'
  * };
  *


danoscarmike:
@jskeet for vis.

jskeet:
Not sure what you mean by "for via" - I'm not really an appropriate person to review Node code.

stephenplusplus:
Probably "for visibility" -- can you tag someone else?

jskeet:
I'm afraid I don't know anyone offhand to review the Node code here... And I'm only on my phone, so can't really look up folks appropriately.

stephenplusplus:
It's not so much the code that needs review from a GCS member, but the methods we've created to support this feature, and the ways they can be used. We want to be sure we are supporting the intended use cases properly.

Having someone read over the code comment blocks in the PR would be helpful. They will demonstrate the methods and list their possible arguments.

jskeet:
I still don't think I'm the most appropriate person to do this, and I wouldn't be to do it before Monday anyway.

jskeet:
(Note that I'm not part of GCS myself.)

stephenplusplus:
Okay, sorry for getting you looped up in here. @danoscarmike had you listed as the POC for Storage Veneer.

@danoscarmike any suggestions on how to proceed?

danoscarmike:
Hi both, apologies Jon - thought you'd be able to review the approach as
opposed to the code itself - I'll find somebody on GCS team to have a look.

packages/storage/src/bucket.js
```diff
@@ -627,6 +627,72 @@ Bucket.prototype.deleteFiles = function(query, callback) {
...
+ *
+ * //-
+ * // Delete a single label.
+ * //-
+ * bucket.deleteLabels('label', function(err, metadata) {});
+ *
+ * //-
+ * // Delete a specific set of labels.
+ * //-
+ * bucket.deleteLabels(['label', 'labeltwo'], function(err, metadata) {});
```
swcloud: To be consistent with the unit test below, maybe we should use 

bucket.deleteLabels(['labelone', 'labeltwo']

packages/storage/src/bucket.js
```diff
@@ -1032,6 +1135,41 @@ Bucket.prototype.makePublic = function(options, callback) {
...
+ * This makes an underlying call to {module:storage/bucket#setMetadata}, which
+ * is a PATCH request. This means an individual label can be overwritten, but
+ * unmentioned labels will not be touched.
+ *
+ * @param {type} labels - Labels to set on the bucket.
+ * @param {function} callback - The callback function.
+ * @param {?error} callback.err - An error returned while making this request.
+ * @param {object} callback.metadata - The bucket's metadata.
+ *
+ * @example
```
swcloud: To be consistent with the unit test below, maybe we should use

var labels = {
   labelone: 'labelonevalue',
   labeltwo: 'labeltwovalue'
};

packages/storage/src/bucket.js
stephenplusplus: Good catch, will do!

packages/storage/src/bucket.js
stephenplusplus: Done!

PR #2390, opened 2017-06-14
update gapic retry config

swcloud:


swcloud:
@stephenplusplus What's wrong with circleci?

$nvm ls
N/A: version "N/A -> N/A" is not yet installed.

You need to run "nvm install N/A" to install it before using it.
            N/A
default -> 4.2.6 (-> N/A)
node -> stable (-> N/A) (default)
iojs -> N/A (default)

nvm ls returned exit code 3

Action failed: nvm ls

stephenplusplus:
Circle equips Linux builds with `nvm` by default, but when we were testing OS X, I had to add a step in the build to install it. I removed that line, restarted the build, and we're back to green.

PR #2396, opened 2017-06-15
Tweak Cloud Bigtable information in README.md

rameshdharan:
I believe this doc was originally written for the v1 API and so it referred to creating a cluster rather than an instance. Moreover the link actually took you to the "creating a compute instance" page rather than the "creating a Bigtable instance" page; the latter is more appropriate for this specific text since it's referring to the fact that creating the instance auto-enables the Cloud Bigtable API for the project.


stephenplusplus:
Thanks!

PR #2401, opened 2017-06-20
Add "renaming" instructions to docs

ace-n:
Allows users to CTRL/Cmd-F for "rename" to find the move command.


stephenplusplus:
Makes sense, thanks!

PR #2405, opened 2017-06-22
upgrade to google-auto-auth@0.7.1

ofrobots:
Brings in https://github.com/stephenplusplus/google-auto-auth/pull/21.

### DIFF
--- a/packages/resource/package.json
+++ b/packages/resource/package.json
@@ -57,6 +57,7 @@
   "devDependencies": {
     "arrify": "^1.0.0",
     "async": "^2.1.2",
+    "google-auto-auth": "^0.7.1",
     "methmeth": "^1.1.0",
     "mocha": "^3.0.1",
     "proxyquire": "^1.7.10",


packages/resource/package.json
```diff
@@ -57,7 +57,6 @@
   "devDependencies": {
     "arrify": "^1.0.0",
     "async": "^2.1.2",
-    "google-auto-auth": "^0.5.2",
```
stephenplusplus: Still need this! https://github.com/ofrobots/google-cloud-node/blob/aa5f008e1ebf8106877e23503b4ae508d20a86f7/packages/resource/system-test/resource.js#L22

Although, that can probably be re-written in the future at some point.

ofrobots: My bad. Too used to grepping only in the `src/` and `test/`.

ofrobots: Fixed.

ofrobots:
Thanks for merging! Could we have new releases on Monday?

stephenplusplus:
I will shoot for that 🌠

PR #2406, opened 2017-06-23
spanner: fix tests

### LINKED ISSUE #2397: System test failures on spanner
There are 4 failures in spanner

```
  1) Spanner Transactions read only should accept a read timestamp:

      Uncaught AssertionError: 5 === 1
      + expected - actual

      -5
      +1
      
      at system-test/spanner.js:2719:20
      at DestroyableTransform.<anonymous> (src/transaction.js:532:7)
      at endReadableNT (node_modules/through2/node_modules/readable-stream/lib/_stream_readable.js:920:12)

  2) Spanner Transactions read only should accept an exact staleness:

      Uncaught AssertionError: 5 === 2
      + expected - actual

      -5
      +2
      
      at system-test/spanner.js:2758:20
      at DestroyableTransform.<anonymous> (src/transaction.js:532:7)
      at endReadableNT (node_modules/through2/node_modules/readable-stream/lib/_stream_readable.js:920:12)

  3) Spanner Transactions read only should do an exact read with concurrent updates:

      Uncaught AssertionError: 'v4' === 'v3'
      + expected - actual

      -v4
      +v3
      
      at system-test/spanner.js:2850:24
      at DestroyableTransform.<anonymous> (src/transaction.js:532:7)
      at endReadableNT (node_modules/through2/node_modules/readable-stream/lib/_stream_readable.js:920:12)

  4) Spanner Transactions read only should read with staleness & concurrent updates:

      Uncaught AssertionError: 5 === 1
      + expected - actual

      -5
      +1
      
      at system-test/spanner.js:2878:22
      at src/session-pool.js:314:16
```

stephenplusplus:
Fixes #2397 

## Do not merge unless @callmehiphop approves

I re-wrote some tests that matched my expectations of what you intended, as well as what the options meant. Please make sure I didn't get any of this wrong.

### DIFF
--- a/packages/spanner/system-test/spanner.js
+++ b/packages/spanner/system-test/spanner.js
@@ -2716,7 +2716,7 @@ var spanner = new Spanner(env);
           transaction.run('SELECT * FROM TxnTable', function(err, rows) {
             assert.ifError(err);
 
-            assert.strictEqual(rows.length, 1);
+            assert.strictEqual(rows.length, 2);
 
             var row = rows[0].toJSON();
 
@@ -2747,7 +2747,7 @@ var spanner = new Spanner(env);
       it('should accept an exact staleness', function(done) {
         var options = {
           readOnly: true,
-          exactStaleness: 4
+          exactStaleness: Math.ceil((Date.now() - records[2].timestamp) / 1000)
         };
 
         database.runTransaction(options, function(err, transaction) {
@@ -2757,9 +2757,7 @@ var spanner = new Spanner(env);
             assert.ifError(err);
             assert.strictEqual(rows.length, 2);
 
-            rows = rows.map(function(row) {
-              return row.toJSON();
-            });
+            rows = rows.map(exec('toJSON'));
 
             assert.strictEqual(rows[0].Key, 'k0');
             assert.strictEqual(rows[0].StringValue, 'v0');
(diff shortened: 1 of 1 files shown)


packages/spanner/system-test/spanner.js
```diff
@@ -2747,7 +2747,7 @@ var spanner = new Spanner(env);
       it('should accept an exact staleness', function(done) {
         var options = {
           readOnly: true,
-          exactStaleness: 4
+          exactStaleness: Math.ceil((Date.now() - records[2].timestamp) / 1000)
```
stephenplusplus: `4` meant "4 seconds ago". That might work when this test is run in isolation, but not when this describe suite runs all of the tests serially. I used this calculation to get a relative staleness value.

packages/spanner/system-test/spanner.js
```diff
@@ -2835,19 +2833,21 @@ var spanner = new Spanner(env);
...
               assert.ifError(err);
 
               transaction.run(query, function(err, rows_) {
                 assert.ifError(err);
 
-                var row = rows_.pop().toJSON();
-                assert.strictEqual(row.StringValue, 'v3');
+                rows_ = extend(true, {}, rows_);
+
+                assert.deepStrictEqual(rows_, originalRows);
```
stephenplusplus: I believe this assertion is basically "make sure I'm still reading data in an old state". I took out the specific value expectation, "v3", hopefully making it more clear that all rows should be the same as they were before the update.

packages/spanner/system-test/spanner.js
```diff
@@ -2716,7 +2716,7 @@ var spanner = new Spanner(env);
           transaction.run('SELECT * FROM TxnTable', function(err, rows) {
             assert.ifError(err);
 
-            assert.strictEqual(rows.length, 1);
+            assert.strictEqual(rows.length, 2);
```
stephenplusplus: This just seems like the correct value should be `2`. The transaction is run with a timestamp of the second insertion, so I would assume getting 2 rows back is correct:

- insert row with timestamp: 101
- insert row with timestamp: 102
- start transaction with timestamp: 102
- read all rows
- get 2 rows back

PR #2410, opened 2017-06-26
enable internal load balancing

blueGinko:
This option will enable internal load balancing. The client channel will build connections to all the targets that the name (like bigtable.googleapis.com) resolves to and do load balancing among them.


stephenplusplus:
@lukesneeringer this sounds like a good thing. What do you think?

stephenplusplus:
@lukesneeringer WDYT?

stephenplusplus:
It's probably best we not default anyone into anything. The user actually has control over this for most (and soon, all) of our GRPC APIs. To take Spanner for example, you can set this flag like this:

```js
var Spanner = require('@google-cloud/spanner')

var spanner = new Spanner({
  projectId: '...',
  keyFilename: '...',
  'grpc.lb_policy_name': 'round_robin'
});
```

Thanks for your patience on this topic.

PR #2417, opened 2017-06-28
spanner: Properly handle multi-row mutations with heterogeneous columns

### LINKED ISSUE #2411: spanner: Unable to insert multiple rows in a single table.insert/transaction.insert call when some rows lack a value for nullable columns
In [transaction-request.js](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/ca0d96cb94f7ef4e176edf221f9ee154dcf8c850/packages/spanner/src/transaction-request.js#L687-L695), there is the following (L687-695):

```javascript
mutation[method] = {
  table: table,
  columns: Object.keys(keyVals[0]),
  values: keyVals.map(function(keyVal) {
    return Object.keys(keyVal).map(function(key) {
      return codec.encode(keyVal[key]);
    });
  })
};
```

There are at least two bugs here:

1. When an array is provided to table.insert(), if some objects do not have a given key (which is perfectly valid if the column said key corresponds to is nullable), the array within `values` for this object will not be the same length as the columns array, and indices in `columns` will no longer correspond to the correct value in the `values` array for this row/object. This results in (best case) misleading error messages from Spanner about incorrect data types. Worse still, if it so happens that the value in the `values` array for this row and incorrect column index is the same type as the value for the correct column index, the insert request will succeed, and end up storing data in the wrong column entirely.

2. If a consumer of this library works around the issue above by populating missing nullable keys with `null`, as the code above relies on the order of values returned by Object.keys, the newly added values will be at the end of the array returned by Object.keys (which returns keys in insertion order), and again, the indices between the `columns` array, and the `values` array for affected rows will no longer align, with the same end result as what occurs with issue 1 above.

This could be fixed by iterating over all rows (rather than naively assuming all rows have identical keys to the first row) and collecting all unique keys into a single array which is subsequently sorted lexically, to build the value for `columns`. Likewise, the value for `values` would be constructed by mapping over `columns` (rather than `Object.keys(keyVal)`) and pulling out the corresponding value from `keyVal` to pass to `codec.encode`.

jscinoz:
Prior to this change, attempting to insert multiple rows at once with Table#insert or
Transaction#insert where some rows lack a value for nullable columns would result in data attempting
to be inserted into the incorrect column, due to a flawed assumption in transaction-request.js that
every single row contains exact same set of columns, and that the order of keys returned by
Object.keys would be identical for each and every row.

This fixes #2411.

### DIFF
--- a/packages/spanner/src/transaction-request.js
+++ b/packages/spanner/src/transaction-request.js
@@ -684,12 +684,29 @@ TransactionRequest.prototype.mutate_ = function(method, table, keyVals, cb) {
 
   var mutation = {};
 
+  var columns = keyVals.reduce(function(allKeys, keyVal) {
+    var keys = Object.keys(keyVal);
+    var key;
+
+    for (var i = 0, ii = keys.length; i < ii; ++i) {
+      key = keys[i];
+
+      if (allKeys.indexOf(key) === -1) {
+        allKeys.push(key);
+      }
+    }
+
+    return allKeys;
+  }, []).sort();
+
   mutation[method] = {
     table: table,
-    columns: Object.keys(keyVals[0]),
+    columns: columns,
     values: keyVals.map(function(keyVal) {
-      return Object.keys(keyVal).map(function(key) {
-        return codec.encode(keyVal[key]);
+      return columns.map(function(key) {
+        var value = keyVal[key];
+
+        return codec.encode(value === undefined ? null : value);
       });
     })
   };


stephenplusplus:
Thank you very much! @callmehiphop and I are in the middle of a few other tasks for the week, but we will check this out as soon as possible. 👍

vkedia:
@jscinoz Thanks for sending this. @stephenplusplus @lukesneeringer Can we please prioritise this. Anything that can cause incorrect data to be stored in Spanner makes me really nervous.

stephenplusplus:
I made a few changes, but didn't have permission to push to the PR branch. I've continued this PR in #2426.

PR #2425, opened 2017-06-30
spanner: fix broken system test

### LINKED ISSUE #2397: System test failures on spanner

callmehiphop:
Closes #2397

### DIFF
--- a/packages/spanner/system-test/spanner.js
+++ b/packages/spanner/system-test/spanner.js
@@ -2716,7 +2716,7 @@ var spanner = new Spanner(env);
           transaction.run('SELECT * FROM TxnTable', function(err, rows) {
             assert.ifError(err);
 
-            assert.strictEqual(rows.length, 2);
+            assert.strictEqual(rows.length, 1);
 
             var row = rows[0].toJSON();
 


packages/spanner/system-test/spanner.js
```diff
@@ -2716,7 +2716,7 @@ var spanner = new Spanner(env);
           transaction.run('SELECT * FROM TxnTable', function(err, rows) {
             assert.ifError(err);
 
-            assert.strictEqual(rows.length, 2);
+            assert.strictEqual(rows.length, 1);
```
callmehiphop: @stephenplusplus this was my fault. So when we capture the timestamp from inserts, they're approximate, in that we capture them before an insert happens, meaning when we grab the timestamp from insert `n`, it should return `n - 1` values.

callmehiphop: However I am a little confused where 3 comes from in the failures you are seeing, locally I consistently get 1 returned value.

stephenplusplus: I'll give it a run and see what we get.

stephenplusplus: Still 3 consistently. Here are the rows being returned (`console.log(rows)` added on the line above this change):

```
[ [ { name: 'Key', value: 'k0' },
    { name: 'StringValue', value: 'v0' },
    { name: 'NumberValue', value: null } ],
  [ { name: 'Key', value: 'k1' },
    { name: 'StringValue', value: 'v1' },
    { name: 'NumberValue', value: null } ],
  [ { name: 'Key', value: 'k2' },
    { name: 'StringValue', value: 'v2' },
    { name: 'NumberValue', value: null } ] ]
```

stephenplusplus: Note as well that @vkedia was receiving `5` in his original report from #2397.

callmehiphop: I'm kind of at a loss because mine consistently returns 1 row.

stephenplusplus: Have you tried running all of the Spanner system tests a couple times, not just the one?

swcloud: In my case, with the this fix, the test passed. I run the test three times.

callmehiphop: Yeah, I've tried various combinations of running just the single test, just txn tests and running all tests and it passes every time.

stephenplusplus: Alright. I'll just not run these :)

PR #2426, opened 2017-06-30
spanner: Properly handle multi-row mutations with heterogeneous columns

### LINKED ISSUE #2411: spanner: Unable to insert multiple rows in a single table.insert/transaction.insert call when some rows lack a value for nullable columns

stephenplusplus:
Fixes #2411 

This adds a commit with tests on top of the work done by @jscinoz in #2417.

### DIFF
--- a/packages/spanner/src/transaction-request.js
+++ b/packages/spanner/src/transaction-request.js
@@ -686,7 +686,18 @@ TransactionRequest.prototype.mutate_ = function(method, table, keyVals, cb) {
 
   var columns = uniq(flatten(keyVals.map(Object.keys))).sort();
 
-  var values = keyVals.map(function(keyVal) {
+  var values = keyVals.map(function(keyVal, index) {
+    var keys = Object.keys(keyVal);
+
+    var missingColumns = columns.filter(column => keys.indexOf(column) === -1);
+
+    if (missingColumns.length > 0) {
+      throw new Error([
+        `Row at index ${index} does not contain the correct number of columns.`,
+        `Missing columns: ${JSON.stringify(missingColumns)}`
+      ].join('\n\n'));
+    }
+
     return columns.map(function(column) {
       var value = keyVal[column];
       return codec.encode(value);
--- a/packages/spanner/test/transaction-request.js
+++ b/packages/spanner/test/transaction-request.js
@@ -872,6 +872,31 @@ describe('TransactionRequest', function() {
       assert.strictEqual(returnValue, requestReturnValue);
     });
 
+    it('should throw when rows have incorrect amount of columns', function() {
+      var invalidEntry = { key1: 'val' };
+      var caughtError;
+
+      try {
+        transactionRequest.mutate_(METHOD, TABLE, [
+          invalidEntry,
+          { key1: 'val', key2: 'val' }
+        ], assert.ifError);
+      } catch(e) {
+        caughtError = e;
+      } finally {
+        if (!caughtError) {
+          throw new Error('Expected error was not thrown.');
+        }
+
+        var expectedErrorMessage = [
+          'Row at index 0 does not contain the correct number of columns.',
+          'Missing columns: ["key2"]'
+        ].join('\n\n');
+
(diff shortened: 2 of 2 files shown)


packages/spanner/system-test/spanner.js
```diff
@@ -108,6 +108,22 @@ var spanner = new Spanner(env);
...
+          {
+            Key: generateName('id'),
+            BoolValue: true
+          },
+          {
+            Key: generateName('id'),
+            BoolValue: true,
+            IntValue: 10
+          }
+        ], done);
```
vkedia: These semantics of automatically setting unset keys to null is very confusing in the case of updates.
If in the above code if you had `table.update` instead of `table.insert`, it would be reasonable to expect that to leave IntValue untouched in the first row but that is not what it will do.

stephenplusplus: What's the recommendation? Do I need to change the source code (transaction-request.js) or just the test?

vkedia: The issue is with the source code. I think we should make it mandatory for users to specify all the values explicitly. So if something is null they should explicitly specify it as null.

stephenplusplus: Does that mean simply not accepting this change, and keeping our code the way it is currently?

vkedia: I don't think so. The issue mentioned that even explicitly passing null can cause problem. Also we should throw an error if number of values in each array is not same.

stephenplusplus: Thank you. The code now correct point no. 2 from #2411, but not point no. 1.

vkedia:
What would happen if the two rows have different number of columns? I think you should error out in that case.

packages/spanner/src/transaction-request.js
```diff
@@ -682,16 +684,28 @@ TransactionRequest.prototype.upsert = function(table, keyVals, callback) {
...
-        return codec.encode(keyVal[key]);
-      });
-    })
+  var values = keyVals.map(function(keyVal, index) {
+    var keys = Object.keys(keyVal);
+
+    var missingColumns = columns.filter(column => keys.indexOf(column) === -1);
+
+    if (missingColumns.length > 0) {
+      throw new Error([
```
stephenplusplus: @vkedia error handling added here.

Issue #2400, opened 2017-06-20
Why is "dlp" in the left-nav in lower case?


jgeewax:
<img width="273" alt="screen shot 2017-06-20 at 10 03 57 am" src="https://user-images.githubusercontent.com/112928/27322859-d614956a-559f-11e7-9f79-c53eebc0d81a.png">

This looks totally inconsistent with the rest of the UI....

/cc @omaray

jgeewax:
It's also basically empty of comments on how to use this. Can we either take it away until it has useful content or fix it and add content that helps people to use the API?

jgeewax:
@lukesneeringer : We're showing ugliness on our docs for over a week now... can we get this fixed?

lukesneeringer:
Sorry, I was OOO when you filed this.

@stephenplusplus Does the @name field need to be added to autogen?

stephenplusplus:
Yep, it does.

lukesneeringer:
Okay, I can do that.

stephenplusplus:
@lukesneering can the generated layer add a sample snippet in the `@example` block? That is what is required to resolve:

> It's also basically empty of comments on how to use this. Can we either take it away until it has useful content or fix it and add content that helps people to use the API?

callmehiphop:
This appears to be resolved.

PR #2429, opened 2017-07-01
docs: find module name from @name tag

stephenplusplus:
RE: #2400 (part 1)

This cases the module name properly when the module is not a typical Class constructor (gapic-only APIs like DLP).

### DIFF
--- a/scripts/docs/parser.js
+++ b/scripts/docs/parser.js
@@ -340,21 +340,43 @@ function getMixinMethods(comments) {
 }
 
 function parseFile(fileName, contents, umbrellaMode) {
-  var comments = dox.parseComments(contents).filter(isPublic);
+  var comments = dox.parseComments(contents);
   var constructor = comments.filter(prop('isConstructor'))[0];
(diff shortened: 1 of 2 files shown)


jgeewax:
:-D

scripts/docs/parser.js
```diff
@@ -340,21 +340,43 @@ function getMixinMethods(comments) {
...
+
+      if (isModuleTag) {
+        tags.forEach(tag => {
+          if (tag.type === 'name') {
+            name = tag.string;
+          }
+        });
+      }
+    });
+  }
```
callmehiphop: I think this can be simplified into this

```js
var moduleTags = getTagsByType(comments, 'module');

if (moduleTags.length) {
  name = getTagsByType(comments, 'name').pop().string;
}
```

stephenplusplus: `getTagsByType` seems to work on an object level, and doesn't expect an array (which `comments` is).

PR #2448, opened 2017-07-12
Datastore: fix Queries created in Transaction with unspecified namespace

### LINKED ISSUE #2447: Datastore - Queries created by Transaction don't fall back to Datastore namespace
Queries created in a transaction without an explicitly specified namespace do not fall back to the datastore's constructor-specified namespace.

Because of this, the namespace must always be explicitly specified for ancestor queries created within a transaction. 

This is probably not desired behaviour, since the docs for Transaction#createQuery provide a namespace-less example (`transaction.createQuery('Company');`)

#### Steps to reproduce

```javascript
const ds = require('@google-cloud/datastore')({namespace: 'myNamespace', projectId: 'my-project'});

const ancestorKey = ds.key(['ancestorKind', 1234]);

// This is OK, query namespace matches key namespace:
ds.createQuery('childKind')
  .hasAncestor(ancestorKey)
  .run()
  .then(([results]) => results);

ds.transaction().run()
  .then(([activeTransaction]) => {

// This fails! 
// "Error: The query namespace is  but ancestor namespace is myNamespace"
    activeTransaction.createQuery('childKind')
             .hasAncestor(ancestorKey)
             .run()
             .then(([results]) => results);
  });
```

#### Cause
This occurs because Transaction binds `this` to itself when calling datastore's createQuery:
```javascript
Transaction.prototype.createQuery = function() {
  return this.datastore.createQuery.apply(this, arguments);
};
```
Then, in createQuery:
```javascript
  if (arguments.length < 2) {
    kind = namespace;
    namespace = this.namespace;
  }
```
Unfortunately `this.namespace` is `undefined` 😭  

This could be fixed in the Transaction constructor:
```javascript
function Transaction(datastore) {
  this.datastore = datastore;
// Add this line to match datastore's fallback namespace:
  this.namespace = datastore.namespace;
```

#### Environment details
Probably not too relevant, but:
  - OS: macOS 10.12.5
  - Node.js version: 6.10.2
  - npm version: 5.1.0
  - google-cloud/datastore version: 1.0.2

beaulac:
Fixes #2447 

Previously, queries created by Transaction could not fall back to the datastore's constructor-specified namespace, because the fallback in `createQuery` needs `namespace` to be defined on the bound `this`, which in this case is the Transaction instance. 

This PR sets the datastore's `namespace` in the Transaction constructor, analogously to `projectId`.

### DIFF
--- a/packages/datastore/src/transaction.js
+++ b/packages/datastore/src/transaction.js
@@ -55,6 +55,7 @@ function Transaction(datastore) {
   this.datastore = datastore;
 
   this.projectId = datastore.projectId;
+  this.namespace = datastore.namespace;
 
   this.request = datastore.request.bind(datastore);
 
--- a/packages/datastore/test/transaction.js
+++ b/packages/datastore/test/transaction.js
@@ -63,10 +63,12 @@ describe('Transaction', function() {
   var transaction;
   var TRANSACTION_ID = 'transaction-id';
   var PROJECT_ID = 'project-id';
+  var NAMESPACE = 'a-namespace';
 
   var DATASTORE = {
     request: function() {},
-    projectId: PROJECT_ID
+    projectId: PROJECT_ID,
+    namespace: NAMESPACE
   };
 
   function key(path) {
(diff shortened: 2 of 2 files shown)


stephenplusplus:
Thanks @beaulac!

stephenplusplus:
Released under @google-cloud/datastore @ 1.0.4.

PR #2449, opened 2017-07-12
docs: add param tags for mixin methods

### LINKED ISSUE #2431: Datastore insert/upsert/update methods aren't documented?
@stephenplusplus mentioned this in a comment, so I went looking for some code...

Seems like `insert` is here: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/datastore/src/request.js#L464

Why are these methods not documented ? (I'm looking at https://googlecloudplatform.github.io/google-cloud-node/#/docs/datastore/1.0.0/datastore)

callmehiphop:
Closes #2431

### DIFF
--- a/packages/datastore/src/request.js
+++ b/packages/datastore/src/request.js
@@ -461,8 +461,6 @@ DatastoreRequest.prototype.get = function(keys, options, callback) {
 /**
  * Maps to {module:datastore#save}, forcing the method to be `insert`.
  *
- * @throws {Error} If an unrecognized method is provided.
- *
  * @param {object|object[]} entities - Datastore key object(s).
  * @param {Key} entities.key - Datastore key object.
  * @param {object|object[]} entities.data - Data to save with the provided key.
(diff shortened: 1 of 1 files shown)


stephenplusplus:
*convo carried on here: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2431#issuecomment-314738617*

packages/datastore/src/request.js
```diff
@@ -460,6 +460,18 @@ DatastoreRequest.prototype.get = function(keys, options, callback) {
 
 /**
  * Maps to {module:datastore#save}, forcing the method to be `insert`.
+ *
+ * @throws {Error} If an unrecognized method is provided.
```
stephenplusplus: These methods don't need this line, since these wrappers define the method.

callmehiphop: Ah, right, fixed!

PR #2450, opened 2017-07-12
Increase the deadline for spanner rpcs

vkedia:
Cloud Spanner RPCs such as Commit and Streaming calls can take a long time to run in certain cases. Increased the deadline for those to be 1 hour. Ideally we would not want any deadlines for these calls.
Also marked most of the rpcs as idempotent.

### DIFF
--- a/packages/spanner/src/v1/spanner_client_config.json
+++ b/packages/spanner/src/v1/spanner_client_config.json
@@ -20,7 +20,7 @@
           "rpc_timeout_multiplier": 1.0,
           "max_rpc_timeout_millis": 60000,
           "total_timeout_millis": 600000
-        }
+        },
         "long_running": {
           "initial_retry_delay_millis": 1000,
           "retry_delay_multiplier": 1.3,


packages/spanner/src/v1/spanner_client_config.json
```diff
@@ -18,11 +21,20 @@
           "max_rpc_timeout_millis": 60000,
           "total_timeout_millis": 600000
         }
+        "long_running": {
```
stephenplusplus: I believe the preceding line needs a trailing "," after the close of the "default" object.

vkedia: Done

stephenplusplus:
This is a generated file-- does https://github.com/googleapis/api-client-staging/blob/master/generated/nodejs/spanner/src/v1/spanner_client_config.json need to be updated first?

stephenplusplus:
Also, PHP's file: https://github.com/googleapis/api-client-staging/blob/master/generated/php/google-cloud-spanner-v1/src/Spanner/V1/resources/spanner_client_config.json, and Python's: https://github.com/googleapis/api-client-staging/blob/master/generated/python/gapic-google-cloud-spanner-v1/google/cloud/gapic/spanner/v1/spanner_client_config.json

vkedia:
I see. Ok I will update that. Once I do that, will it automatically get copied here?

stephenplusplus:
I don't believe it's automatic, but @lukesneeringer or @bjwatson should know how to get it copied over here.

lukesneeringer:
It is not automatic, but we can do it. @stephenplusplus, if you know the new values, feel free to just change it to the new values. That would be faster.

Issue #2356, opened 2017-06-05
Spanner - noStatus - TypeError: requestStream.destroy is not a function


Chipintoza:
OS: Mac OSX 10.12.4
Node.js version: 6.9.3
npm version: 5.0.0
version: 0.4.4

**After closing #2176  i run my data export in spanner. When a number of requests were in parallel the following error occurred several times.**

```
noStatus - TypeError: requestStream.destroy is not a function
    at StreamProxy.<anonymous> (/Users/Chipintoza/GSS Projects/accounts.gss.ge/node_modules/@google-cloud/spanner/src/transaction.js:375:19)
    at emitOne (events.js:96:13)
    at StreamProxy.emit (events.js:188:7)
    at ClientReadableStream.<anonymous> (/Users/Chipintoza/GSS Projects/accounts.gss.ge/node_modules/google-gax/lib/streaming.js:130:17)
    at emitOne (events.js:96:13)
    at ClientReadableStream.emit (events.js:188:7)
    at ClientReadableStream._emitStatusIfDone (/Users/Chipintoza/GSS Projects/accounts.gss.ge/node_modules/grpc/src/node/src/client.js:204:12)
    at ClientReadableStream._receiveStatus (/Users/Chipintoza/GSS Projects/accounts.gss.ge/node_modules/grpc/src/node/src/client.js:180:8)
    at /Users/Chipintoza/GSS Projects/accounts.gss.ge/node_modules/grpc/src/node/src/client.js:592:14
``` 

**After certain quantity of data export, the following error kept occurring constantly:**

- req.url - http://accounts.gss.local:9001/gss-api/environments/upsert
InternalServerError: err.code: 8
- err.message: Too many active sessions in database projects/spanner-gss-ge/instances/business-data/databases/system-development, limit is 30000. Increase the node count to allow more sessions.
- err.status: undefined
- err location: Environment.upsert\database.runTransaction((err, tx) => {...
- data:

```json
{
  "environmentId": "5e9d511f-e300-49d8-80ba-21c421f98b03",
  "isActive": true,
  "created": 1325376000000,
  "timestamp": 1496655116694,
  "modifierId": "system-id",
  "modifierName": "system",
  "name": "ka:ჩემი 'GSS გარემო' - paab@mail.ru|en:My 'GSS Environment' - paab@mail.ru|ru:Моя 'GSS Среда' - paab@mail.ru",
[...]

PR #2452, opened 2017-07-12
spanner: promote to beta

stephenplusplus:


vkedia:
I think we should resolve #2356 before merging this.

stephenplusplus:
I'll consider this blocked, unless @lukesneeringer feels otherwise.

PR #2453, opened 2017-07-12
logging-winston: make winston a peerDependency

ofrobots:
Users of this library need to depend on winston directly. There are
scenarios where the user, or other libraries may depend on a semver
incompatible version of winston. Not only does it end up installing an
unnecessary copy of winston; it may also lead to our transport not
being visible in `winston.transports` that the user has visibility to.

### DIFF
--- a/packages/logging-winston/package.json
+++ b/packages/logging-winston/package.json
@@ -32,12 +32,15 @@
     "@google-cloud/logging": "^1.0.0",
     "extend": "^3.0.0",
     "is": "^3.2.0",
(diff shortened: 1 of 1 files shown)


ofrobots:
This is breaking, so this should bump us the logging-winston module to 0.5.0.

stephenplusplus:
Need a release now?

ofrobots:
Hold off for https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2454.

PR #2459, opened 2017-07-15
Speech partial veneer

lukesneeringer:
Supercedes #2421.

### DIFF
--- a/packages/speech/package.json
+++ b/packages/speech/package.json
@@ -1,7 +1,7 @@
 {
   "repository": "GoogleCloudPlatform/google-cloud-node",
   "name": "@google-cloud/speech",
-  "version": "0.10.0",
+  "version": "0.9.4",
   "author": "Google Inc.",
   "description": "Cloud Speech Client Library for Node.js",
   "contributors": [
@@ -62,9 +62,7 @@
     "through2": "^2.0.3"
   },
   "devDependencies": {
-    "intelli-espower-loader": "^1.0.1",
     "mocha": "^3.2.0",
-    "nyc": "^10.3.0",
     "power-assert": "^1.4.2",
     "sinon": "^2.2.0"
   },
--- a/packages/speech/src/helpers.js
+++ b/packages/speech/src/helpers.js
@@ -57,9 +57,7 @@ module.exports = () => {
    *
    * @example
    *
-   * var client = require('@google-cloud/speech')();
-   *
-   * var stream = client.streamingRecognize({
+   * var stream = speech.streamingRecognize({
    *   config: {
    *     encoding: 'LINEAR16',
    *     languageCode: 'en-us',
@@ -104,12 +102,7 @@ module.exports = () => {
           });
         }),
 
-        requestStream,
-
-        // Pass through the results.
-        through.obj(function(obj, _, next) {
-          next(null, obj);
-        })
+        requestStream
       ]);
     });
 


packages/speech/package.json
```diff
@@ -1,6 +1,7 @@
 {
+  "repository": "GoogleCloudPlatform/google-cloud-node",
   "name": "@google-cloud/speech",
-  "version": "0.9.4",
+  "version": "0.10.0",
```
stephenplusplus: Keep the version 0.9.4, it will be bumped to 0.10.0 from the release script

packages/speech/package.json
```diff
@@ -51,36 +50,28 @@
...
   },
   "devDependencies": {
-    "@google-cloud/storage": "*",
-    "async": "^2.0.1",
-    "methmeth": "^1.1.0",
-    "mocha": "^3.0.2",
-    "proxyquire": "^1.7.10",
-    "tmp": "^0.0.31",
-    "uuid": "^3.0.1"
+    "intelli-espower-loader": "^1.0.1",
```
stephenplusplus: I don't think this is used (don't see in it any of the `test` scripts)

packages/speech/package.json
```diff
@@ -51,36 +50,28 @@
...
+    "mocha": "^3.2.0",
+    "nyc": "^10.3.0",
```
stephenplusplus: I don't think we need this, since we're not using it from a `cover` task specifically inside this module.

packages/speech/src/helpers.js
```diff
@@ -0,0 +1,120 @@
...
+        // Format the user's input.
+        through.obj(function(obj, _, next) {
+          next(null, {
+            audioContent: obj
+          });
+        }),
+
+        requestStream,
+
+        // Pass through the results.
```
stephenplusplus: We should be able to remove this part. The following stream:

```js
through.obj(function(obj, _, next) {
  next(null, obj);
})
```

is the same as just:

```js
through.obj();
```

which is just a passthrough stream, which means we might as well save the memory (streams are relatively heavy) and not use it at all.

packages/speech/src/helpers.js
```diff
@@ -0,0 +1,120 @@
...
+   * @returns {Stream}
+   *   An object stream which is both readable and writable. It accepts
+   *   [StreamingRecognizeRequest]{@link StreamingRecognizeRequest}-like
+   *   objects for the write() method, and will emit objects representing
+   *   [StreamingRecognizeResponse]{@link StreamingRecognizeResponse} on the
+   *   'data' event asynchronously.
+   *
+   * @example
+   *
+   * var client = require('@google-cloud/speech')();
```
stephenplusplus: We don't have to do this for each method. We use the Overview / description of the instantiating function to show how to require the module / initialize it.

stephenplusplus: By default, the module is required and available for the snippets to use under the name of the module (`speech`). If you prefer `client`, I will make a change in a configuration file to support that. Otherwise, just rename `client` to `speech`:

```js
    * var stream = speech.streamingRecognize({
    *   config: {
    *     encoding: 'LINEAR16',
    *     languageCode: 'en-us',
    *     sampleRateHertz: 44100,
    *   },
    * }).on('data', function(response) {
    *   // doThingsWith(response);
    * });
    * var request = {};
    * // Write request objects.
    * stream.write(request);
```

packages/speech/src/helpers.js
```diff
@@ -0,0 +1,113 @@
...
+
+/*!
+ * Return a dictionary-like object with helpers to augment the Speech
+ * GAPIC.
+ *
+ * @return {Object} - An object with keys and functions which are placed
+ *   onto the pure GAPIC.
+ */
+module.exports = () => {
+  var methods = {};
```
jmdobry: @lukesneeringer These Speech helpers are missing the file loading helpers that the Vision partial gapic helpers have: https://github.com/lukesneeringer/google-cloud-node/blob/ada694f1ced83e8d46d685ed8d6a7d20f204e494/packages/vision/src/helpers.js#L51

PR #2461, opened 2017-07-17
speech: add a sink to `requestStream` so events are not dropped

stephenplusplus:


packages/speech/src/helpers.js
```diff
@@ -80,29 +79,39 @@ module.exports = () => {
...
     var recognizeStream = streamEvents(pumpify.obj());
 
-    recognizeStream.once('writing', function() {
-      requestStream.on('error', function(err) {
+    // Attach the events to the request stream, but only do so
+    // when the first write (of data) comes in.
+    //
+    // This also means that the sending of the initial request (with the
+    // config) is delayed until we get the first burst of data.
+    recognizeStream.once('writing', () => {
```
stephenplusplus: `() => {` binds the current context with it, which would be this instance of a Speech constructor. We'll have to change the later line "this.setPipeline" to "recognizeStream.setPipeline"

stephenplusplus: I don't know why I wrote all that. I'll just do it.

stephenplusplus: Done-- let's see what the tests think

lukesneeringer: Derp, thank you for catching that.

PR #2380, opened 2017-06-12
pubsub: api redesign

callmehiphop:
[docs demo](http://callmehiphop.github.io/gcloud-node/#/docs/pubsub/0.14.0/pubsub)

#### TODO:

- [x] Unit Tests
- [x] StreamingPull System Tests
- [x] Docs
- [x] README

PR #2464, opened 2017-07-18
Allow message received to not be parsed to JSON

mkamioner:
Hi All!

I came across a case where I need the **original** string of the message that was published to PubSub (and received). In the current implementation, A message like this: `{"trailingZeros":42.000}` would parse into `{"trailingZeros":42}`. Likewise white space is ignored (`{ "iLoveWhiteSpace": true }` = `{"iLoveWhiteSpace":true}`)

This implementation will allow the subscriber to use the `json` option. If it is set to `true` (which is the default), it will continue to try to parse the JSON automatically. However, if it is explicitly set to `false`, the original string will be returned with no parsing.

Tests included :-)

Also - do I get to add my name to contributors if this is merged? ;-)

### DIFF
--- a/packages/pubsub/test/subscription.js
+++ b/packages/pubsub/test/subscription.js
@@ -312,7 +312,6 @@ describe('Subscription', function() {
 
     it('should not decode stringified JSON to object', function() {
       var str = '{ "hi": "there" }';
-      var obj = JSON.parse(str);
       var stringified = new Buffer(str).toString('base64');
       var attributes = {};
       var publishTime = {


stephenplusplus:
I have some good news!

- This is a great PR-- you found a problem, solved it, and even wrote tests. This is something we really appreciate, as we have so many ongoing things that it's hard to get around to fixing all of them as quickly as we'd like.

- This PR (fortunately for your use case, unfortunately for your effort) won't be necessary shortly, as Pub/Sub is going through a redesign, and it's just about ready to ship: #2380 

@callmehiphop -- can you confirm that this issue will be resolved after #2380?

Thanks again for the help, @mkamioner! Hope to see you more often!

mkamioner:
@callmehiphop -- I will be happy to make a PR into 2380 with this design if you'd like.

stephenplusplus:
I believe the Pub/Sub re-write is going to do away with JSON-izing anything at all. Messages will be accepted and received in raw Buffer format only.

callmehiphop:
@stephenplusplus That is correct, the redesign will solve this problem since we are only going to allow sending/receiving buffers. Meaning from that point the user will have to do JSON parsing on their own.

stephenplusplus:
Thanks again for the PR, @mkamioner. Feel free to subscribe over on #2380 to follow the progress.

Issue #2388, opened 2017-06-14
Cloud Spanner Session Pool should rate limit initial session creations.


vkedia:
Cloud Spanner Session pool upon construction would start creating sessions eagerly upto the minimum number of sessions. It will essentially fire all the create requests in parallel. We should instead rate limit the session creation so that it creates at most X number of sessions at a time (X=10 would be a reasonable value).
I realize that this behavior is manged by the underlying pooling library which does not provide a way to rate limit. One option can be to set `autostart=false` so that the pooling library does not eagerly create sessions. Then in the SessionPool we ourselves can call `pool.acquire` X times so that X number of sessions would be eagerly created.
Another option can be to file a feature request against the pooling library to support this use case.

lukesneeringer:
This issue was moved to googleapis/nodejs-spanner#19

PR #2469, opened 2017-07-19
spanner: session leak detection

### LINKED ISSUE #2292: Spanner library should have a way to track leaked sessions.
Due to a bug in user code or client library code, it is possible that a session is checked out from the pool but never released. We should make it easier for users to be able to detect such leaks and debug where in there code the leak occurred. One way can be to record a trace in a list or a map whenever a session is checked out and remove that trace when the session is released back to the pool. User should be able to look at this list and figure out the location in the code where all the unreleased sessions were checked out at. We can also log these stack traces when the pool is closed.

callmehiphop:
Closes #2292

This change will capture traces for Session as they are checked out and delete said traces once they are released. In the event that a user tries to close a Database connection and we find any Session leaks, we'll return a `SessionLeakError` that contains a list of all traces leading back to the leaked Sessions.

#### TODOs

- [ ] Unit Tests
- [ ] Docs

### DIFF
--- a/packages/spanner/src/database.js
+++ b/packages/spanner/src/database.js
@@ -23,6 +23,7 @@
 var arrify = require('arrify');
 var common = require('@google-cloud/common');
 var commonGrpc = require('@google-cloud/common-grpc');
+var createErrorClass = require('create-error-class');
 var extend = require('extend');
 var is = require('is');
 var util = require('util');
@@ -62,6 +63,14 @@ var Table = require('./table.js');
  * @private
  */
 var TransactionRequest = require('./transaction-request.js');
+
+/**
+ *
+ */
+var SessionLeakError = createErrorClass('SessionLeakError', function(leaks) {
+  this.message = leaks.length + ' session leaks found.';
+  this.messages = leaks;
+});
 
 /**
  * Create a Database object to interact with a Cloud Spanner database.
@@ -223,6 +232,8 @@ Database.formatName_ = function(instanceName, name) {
  *
  * @param {function} callback - The callback function.
  * @param {?error} callback.err - An error returned while making this request.
+ * @param {?string[]} callback.err.messages - If a SessionLeakError is returned,
+ *     this will be an array populated by stack traces for each leak detected.
  *
  * @example
  * database.close(function(err) {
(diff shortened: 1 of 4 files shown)


stephenplusplus:
LGTM 👍

callmehiphop:
@vkedia friendly ping :)

packages/spanner/src/database.js
```diff
@@ -233,6 +244,12 @@ Database.formatName_ = function(instanceName, name) {
  */
 Database.prototype.close = function(callback) {
   var self = this;
+  var leaks = this.pool_.getSessionLeaks();
+
+  if (leaks.length) {
+    callback(new SessionLeakError(leaks));
+    return;
```
vkedia: You should still clear the pool and delete all the sessions before returning here.

callmehiphop: I initially went that route, but I don't think the pooling library allows itself to be drained until all sessions have released. In my testing it would just hang indefinitely. 

Between this and #2388 I think we may want to look into using a different pooling library. WDYT?

vkedia: When `database.close` is called can we release all the sessions to the pool ourselves irrespective of whether they are being used or not so that the pooling library can be drained?

I agree it would be a good idea to investigate a different pooling library or get this pooling library's owner to add these features.

callmehiphop: That's possible, although currently we do not keep track of leased sessions, but I don't think that would be too difficult to do.

stephenplusplus: How should we proceed here? Merge, then improve, or Improve, then merge?

callmehiphop: I don't really have the bandwidth at the moment, so I would vote to merge then improve. I think I'd like to swap out pooling libraries at some point which could also play a role in fixing this.

vkedia: So if I understand correctly, if there is a leak (even a single session), then none of the sessions will get deleted? That's quite bad especially since it seem to me that node.js is more prone to session leaks due to its async nature. I would rather we keep track of all sessions ids somewhere and delete them all ourselves if pooling library will not do this.

PR #2473, opened 2017-07-20
storage/bucket: add "bucket.setStorageClass()"

### LINKED ISSUE #2471: [Storage-request] Provide an example of setting the default storage class for a bucket.
Hi,

This is a request to provide an example of setting the default storage class for a Storage bucket. The current method for updating the storageClass isn't clear or visible in the documentation. Although, I'm not sure if it makes sense for the Nodejs community to have this documented. I'm open to interpretation! 

```
var gcs = require('@google-cloud/storage')({projectId: "spec-test-ruby-samples"});

var bucket = gcs.bucket("bucketname");

bucket.setMetadata({storageClass: "STORAGE_CLASS"});
```

stephenplusplus:
Fixes #2471 

This adds a method `bucket.setStorageClass(newStorageClass, [options], [callback])`.

### DIFF
--- a/packages/storage/src/bucket.js
+++ b/packages/storage/src/bucket.js
@@ -1428,12 +1428,13 @@ Bucket.prototype.setMetadata = function(metadata, options, callback) {
 };
 
 /**
- * Set the storage class for this bucket.
+ * Set the default storage class for new files in this bucket.
  *
  * @resource [Storage Classes]{@link https://cloud.google.com/storage/docs/storage-classes}
  *
  * @param {string} storageClass - The new storage class. (`multi_regional`,
- *     `regional`, `nearline`, `coldline`)
+ *     `regional`, `standard`, `nearline`, `coldline`, or
+ *     `durable_reduced_availability`)
  * @param {object=} options - Configuration object.
  * @param {boolean} options.userProject - If this bucket has `requesterPays`
  *     functionality enabled (see {module:storage/bucket#enableRequesterPays}),


packages/storage/src/bucket.js
```diff
@@ -1428,6 +1428,46 @@ Bucket.prototype.setMetadata = function(metadata, options, callback) {
 };
 
 /**
+ * Set the storage class for this bucket.
```
frankyn: Set a buckets default storage class for new files.  https://cloud.google.com/storage/docs/json_api/v1/buckets/update#storageClass

packages/storage/src/bucket.js
```diff
@@ -1428,6 +1428,46 @@ Bucket.prototype.setMetadata = function(metadata, options, callback) {
...
+ *
+ * @resource [Storage Classes]{@link https://cloud.google.com/storage/docs/storage-classes}
+ *
+ * @param {string} storageClass - The new storage class. (`multi_regional`,
```
frankyn: Pulling in from documentation: Values include MULTI_REGIONAL, REGIONAL, STANDARD, NEARLINE, COLDLINE, and DURABLE_REDUCED_AVAILABILITY.

packages/storage/src/bucket.js
stephenplusplus: Done!

packages/storage/src/bucket.js
stephenplusplus: Good catch, thank you!

frankyn:
Thanks! Example and documentation LGTM.

stephenplusplus:
Thanks for the suggestion and review!

PR #2477, opened 2017-07-20
spanner: documentation corrections

### LINKED ISSUE #2180: Fix spanner API documentation
@lukesneeringer @stephenplusplus @callmehiphop 
I was going through the spanner API documentation and I noticed various issues. Some of which I am enumerating below but we need to go thoroughly over the documentation and make sure it is accurate and complete:

- [ ] In [transaction](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.50.0/spanner/transaction?method=Transaction), the link to database/runTransaction does not work.
- [ ] Under parameters for [database/runTransaction](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.50.0/spanner/database?method=runTransaction) we should document that options like exactStaleness are only relevant for readOnly transactions.
- [ ] [database/runTransaction](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.50.0/spanner/database?method=runTransaction) does not mention that it retries the runFunction on ABORT errors.
- [ ] In [transaction](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.50.0/spanner/transaction),  there is a table of Parameters. But it is not clear where those parameters are to be specified. I think that table can be removed from there. There already is a similar table in runTransaction.
- [ ] The table mentioned above, talks about single-use transaction but nowhere is that term defined and it is not clear how to use single-use transaction in the node client.
- [ ] In [transaction/deleteRows](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.50.0/spanner/transaction?method=deleteRows), Parameters talk about a callback but that method does not take a callback. Ditto for replace, update and upsert.
- [ ] Second argument for insert/update/replace/upsert is called keyVals and it says that it is "A map of names to values of data to insert into this table". That seems incorrect. It is actually a list of rows where a row is a map from column name to column value.
- [ ] [database/run](https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.50.0/spanner/database?method=run) and other similar methods have a parameter, returnReadTimestamp which is true returns the timestamp at which the data was read to the user. But it is not clear how it is returned in example or the documentation.

callmehiphop:
Closes #2180

### DIFF
--- a/packages/spanner/src/transaction.js
+++ b/packages/spanner/src/transaction.js
@@ -71,8 +71,8 @@ var ABORTED = 10;
  *     committed transactions are visible.
  */
 /**
- * Use a Transaction object to read and write against inside of your Cloud
- * Spanner database.
+ * Use a Transaction object to read and write against your Cloud Spanner
+ * database.
  *
  * **This object is created and returned from
  * {module:spanner/database#runTransaction}.**


packages/spanner/src/transaction.js
```diff
@@ -81,6 +69,17 @@ var ABORTED = 10;
  *     timestamp.
  * @param {boolean} options.strong - Read at the timestamp where all previously
  *     committed transactions are visible.
+ */
+/**
+ * Use a Transaction object to read and write against inside of your Cloud
```
vkedia: against your Cloud Spanner database

callmehiphop: Whoops! Nice catch

PR #2485, opened 2017-07-24
generate google-cloud gapic docs correctly

### LINKED ISSUE #2462: docs: (google-cloud) gapic versions acting as first class APIs
The nav bar on the docs site:

image

These are all from the packages _inside_ of google-cloud, e.g. `v1` is from `@google-cloud/datastore` (https://googlecloudplatform.github.io/google-cloud-node/#/docs/datastore/1.0.6/v1). We need to find a way to stuff them back inside of their proper parent, so that `v1` is nested under the `Datastore` link.

stephenplusplus:
Fixes #2462

Preview: https://stephenplusplus.github.io/gcloud-node/#/docs/google-cloud/master/

This will nest the generated versions properly under their API labels.

The one cost is that we can't have a detailed overview page with links, such as https://googlecloudplatform.github.io/google-cloud-node/#/docs/dlp/master/v2beta1. For the bundled package, that page will only say "To explore this API, use the links in the navigation bar." The navigation bar and experience from that point will be the same.

### DIFF
--- a/scripts/docs/builder.js
+++ b/scripts/docs/builder.js
@@ -98,7 +98,7 @@ Builder.prototype.build = function() {
   });
 
   var gapicVersions = docs.reduce((grouped, doc) => {
-    var gapicVersion = doc.id.match(/[/_](v[^\/]*)/);
+    var gapicVersion = doc.id.match(/[/_](v\d[^\/]*)/);
 
     if (gapicVersion) {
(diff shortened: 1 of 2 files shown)


stephenplusplus:
Preview added (https://stephenplusplus.github.io/gcloud-node/#/docs/google-cloud/master)

PR #2488, opened 2017-07-24
circle: refactor test script

stephenplusplus:
I'm trying to reconcile differences between how a test script should exit on AppVeyor and Circle. When one is happy (green build), the other one complains (red build). The first step is a small refactor of the test script run through Circle, so I can find focus on the controversial component.

### DIFF
--- a/scripts/circle/test.sh
+++ b/scripts/circle/test.sh
@@ -43,10 +43,3 @@ done
 npm run lint
 
 node ./scripts/build.js
-
-if [ "$1" == "4" ]
-then
-  export COVERALLS_REPO_TOKEN="vKZ7a3PpW0lRBRWC12dPw2EiZE5ml962J"
-  export CIRCLE_ARTIFACTS="$(pwd)/.coverage"
-  npm run coveralls
-fi


stephenplusplus:
Before:

- test.sh is called from Circle
- test.sh runs tests for all Node versions
- test.sh reports to coveralls with code coverage

After:

- test.sh is called from Circle with argument "4"
- test.sh runs tests for called version (4)
- Circle reports to coveralls with code coverage
- test.sh is called from Circle with argument "6"
- test.sh is called from Circle with argument "7"
- test.sh is called from Circle with argument "8"

This has a couple benefits that I see:

- More configuration is moved to the Circle UI. This makes it easier to test settings without these backwards CI-poking PRs.
- Logging output is less likely to be collapsed by the UI, where Circle forces you to download the raw .txt log file. This is because each Node version now gets its own allowance.

Here is what the Circle UI "test" task is configured to run now:

```
git config --global user.name "circle-ci"
git config --global user.email "circle-ci@circleci.com"
. scripts/circle/test.sh 4
export COVERALLS_REPO_TOKEN="vKZ7a3PpW0lRBRWC12dPw2EiZE5ml962J"; npm run coveralls
. scripts/circle/test.sh 6
. scripts/circle/test.sh 7
. scripts/circle/test.sh 8
```

Let's see how this goes.

PR #2491, opened 2017-07-25
readme: remove Dependency CI

stephenplusplus:
It's been 5 months of pending builds: https://dependencyci.com/github/GoogleCloudPlatform/google-cloud-node?page=23

And if you go before that, we always had failed builds because of things we can't control-- dependencies of dependencies of dependencies, etc. Is it cool to remove this?

cc @lukesneeringer


callmehiphop:
👍

PR #2495, opened 2017-07-27
test: GAE, GKE, GCF run on GCE

ofrobots:
Make tests ensure that getDefaultResource prefers GAE, GKE, GCF over
GCE. Those environments would return true for GCE as well.

### DIFF
--- a/packages/logging/test/metadata.js
+++ b/packages/logging/test/metadata.js
@@ -186,7 +186,8 @@ describe('metadata', function() {
 
           metadata.logging.auth.getEnvironment = function(callback) {
             callback(null, {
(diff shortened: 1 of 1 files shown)


stephenplusplus:
So the source code is doing the right thing, this is just confirming? Should we find a better way to differentiate between the environments, or is it proper to say something is both GAE and GCE?

ofrobots:
The source code is doing the right thing, but there ought to be a test for it. I think it is correct behaviour for IS_COMPUTE_ENGINE and IS_APP_ENGINE to both return true for AppEngine.

stephenplusplus:
Cool, looks good!

PR #2497, opened 2017-07-28
datastore: allow nested unindexed properties

### LINKED ISSUE #1916: Can't store entities with embedded entities which have properties > 1500 bytes
Seems that excluding a top level property doesn't exclude an embedded entities properties from being indexed - on top of that, I don't believe there's a way to exclude an embedded entities individual properties, that would solve my problem too.

**Steps to reproduce**

Just use this script (I'm running the datastore emulator)

```js
const Datastore = require('@google-cloud/datastore');
const datastore = new Datastore({ projectId: 'datastore-test' });

datastore.insert({
  key: datastore.key('User'),
  data: [
    {
      name: 'description',
      value: {
        text: Buffer.alloc(1501, 'a').toString(),
      },
      excludeFromIndexes: true
    }
  ]
}, (err, result) => {
  if (err) {
    console.error(err);
  } else {
    console.log(result.data);
  }
});
```  
I'd expect there to be no error - the embedded entity (`{ text: '...' }`) shouldn't be indexed I believe? But it throw an error saying:

`Error: The value of property "text" is longer than 1500 bytes.`

**Environment**

  - OS: OS X Sierra
  - Node.js: v6.7.0
  - npm: 3.10.3
  - @google-cloud/datastore: v0.6.0

stephenplusplus:
Fixes #1916

#### To Dos

- [x] Docs
- [x] System tests
- [x] Unit tests

This will allow excluding embedded properties from indexing, using the solution discussed [here](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1916#issuecomment-318679971). Here is the sample from our system tests:

```js
datastore.save({
  key: datastore.key('Kind'),
  data: {
    longString: longString,
    notMetadata: true,
    metadata: {
      longString: longString,
      otherProperty: 'value',
      longStringArray: [
        {
          longString: longString,
          nestedLongStringArray: [
            {
              longString: longString,
              nestedProperty: true
            },
            {
              longString: longString
            }
          ]
        }
      ]
    }
  },
  excludeFromIndexes: [
    'longString',
    'metadata.longString',
    'metadata.longStringArray[].longString',
    'metadata.longStringArray[].nestedLongStringArray[].longString'
  ]
})
```

### DIFF
--- a/packages/datastore/src/entity.js
+++ b/packages/datastore/src/entity.js
@@ -396,6 +396,14 @@ function entityToEntityProto(entityObject) {
     }, {})
   };
 
+  if (excludeFromIndexes && excludeFromIndexes.length > 0) {
+    excludeFromIndexes.forEach(function(excludePath) {
+      excludePathFromEntity(entityProto, excludePath);
+    });
+  }
+
+  return entityProto;
+
   function excludePathFromEntity(entity, path) {
     var arrayIndex = path.indexOf('[]');
     var entityIndex = path.indexOf('.');
@@ -404,8 +412,10 @@ function entityToEntityProto(entityObject) {
     var hasEntityPath = entityIndex > -1;
 
     if (!hasArrayPath && !hasEntityPath) {
-      // This is the property to exclude!
-      entity.properties[path].excludeFromIndexes = true;
+      if (entity.properties[path]) {
+        // This is the property to exclude!
+        entity.properties[path].excludeFromIndexes = true;
+      }
       return;
     }
 
@@ -424,6 +434,10 @@ function entityToEntityProto(entityObject) {
     var firstPathPart = splitPath.shift();
     var remainderPath = splitPath.join(delimiter).replace(/^(\.|[])/, '');
 
+    if (!entity.properties[firstPathPart]) {
+      return;
+    }
+
     if (firstPathPartIsArray) {
       var array = entity.properties[firstPathPart].arrayValue;
       array.values.forEach(function(arrayValue) {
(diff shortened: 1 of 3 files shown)


packages/datastore/src/entity.js
```diff
@@ -378,14 +384,70 @@ entity.entityFromEntityProto = entityFromEntityProto;
...
+  if (excludeFromIndexes && excludeFromIndexes.length > 0) {
+    excludeFromIndexes.forEach(function(excludePath) {
+      excludePathFromEntity(entityProto, excludePath);
+    });
+  }
+
+  return entityProto;
+
+  function excludePathFromEntity(entity, path) {
+    var arrayIndex = path.indexOf('[]');
```
callmehiphop: I feel like the bulk of this function could be modularized, I immediately thought of `dot-prop` but upon closer inspection that does not support Arrays. Is there a similar module that does? Or do you think we should consider making this a module?

packages/datastore/system-test/datastore.js
```diff
@@ -81,6 +81,75 @@ describe('Datastore', function() {
...
+          'metadata.longStringArray[].nestedLongStringArray[].longString'
+        ]
+      }, function(err) {
+        assert.ifError(err);
+
+        datastore.get(postKey, function(err, entity) {
+          assert.ifError(err);
+
+          assert.deepEqual(entity, data);
+          assert.deepEqual(entity[datastore.KEY], postKey);
```
callmehiphop: Are there any assertions we can do to make sure that the specified items were indeed not indexed?

stephenplusplus: It's actually the fact that it successfully saved that is the test. If the indexes weren't excluded, the API would error because the strings are > 1500 bytes.

packages/datastore/src/entity.js
stephenplusplus: Maybe. Where I would get stuck is the structure of the Entity proto object, which is assembled differently depending on the type of value.

PR #2502, opened 2017-08-02
feat: we can omit project_id from the resource

ofrobots:
The logging service will automatically insert the project id label on
the resource descriptor - it is not necessary for us to include that
part of the resource on each log entry.

### DIFF
--- a/packages/logging/src/metadata.js
+++ b/packages/logging/src/metadata.js
@@ -127,9 +127,7 @@ Metadata.getGlobalDescriptor = function() {
  * @param {function} callback - The callback function.
  */
 Metadata.prototype.getDefaultResource = function(callback) {
-  var self = this;
-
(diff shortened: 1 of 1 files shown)


packages/logging/src/metadata.js
```diff
@@ -143,32 +129,25 @@ Metadata.getGlobalDescriptor = function(projectId) {
 Metadata.prototype.getDefaultResource = function(callback) {
   var self = this;
 
-  this.logging.auth.getProjectId(function(err, projectId) {
-    if (err) {
-      callback(err);
+  self.logging.auth.getEnvironment(function(err, env) {
```
stephenplusplus: We can get rid of `self` and just use `this` now.

ofrobots: done.

stephenplusplus:
> less self-ish

Nice 👍

PR #2506, opened 2017-08-03
added get project api call 

ashleyschuett:
https://cloud.google.com/compute/docs/reference/v1/projects

### DIFF
--- a/packages/compute/test/index.js
+++ b/packages/compute/test/index.js
@@ -2552,7 +2552,7 @@ describe('Compute', function() {
   });
 
   describe('project', function() {
-    it('should return a Region object', function() {
+    it('should return a Project object', function() {
       var project = compute.project();
       assert(project instanceof FakeProject);
       assert.strictEqual(project.calledWith_[0], compute);


ashleyschuett:
I'm not sure what test I'm missing, or if this is and acceptable way to implement compute projects.

stephenplusplus:
Thank you very much for this! I've got a commit coming with some refactors. No worries at all from this point, I'm happy to take it the rest of the way with regards to the tests and coverage. But, I would appreciate a review of the changes I've made, to make sure I'm not overlooking anything. More details coming in a few minutes!

stephenplusplus:
Changes have been committed, please take a look!

I removed `compute.getProjects()` and `compute.getProjectsStream()`, because it seems like there's only one project as the resource that the `/projects` endpoint refers to. Therefore, the user will never be able to list projects. Please correct me if I'm wrong on that.

Otherwise, I added `project.get()` and `project.getMetadata()` to fit in with the rest of our API, and re-aliased `project.name` (the instance variable) to `project.id`, since `name` is a different piece of data than ID.

Thanks again for helping us out! I look forward to your thoughts.

ashleyschuett:
I debated between implementing it as you have it with the `get` and `getMetadata` calls and just doing the `getProject` top call. I like this implementation better. 

Looks good to me! 👍

PR #2512, opened 2017-08-07
Document keySet query parameter

### LINKED ISSUE #2099: Spanner table.read keys vs keySet?
The official documentation (https://cloud.google.com/spanner/docs/reads) talks about "keySet" as a query parameter on `table.read()` being able to use that to say "give me all rows" (`{keySet: {all: true}}`). I can't find this in our documentation though... Why not?

This gives me all my rows, but isn't documented on the method (https://googlecloudplatform.github.io/google-cloud-node/#/docs/google-cloud/0.49.0/spanner/table?method=read)

```js
table.read({columns: ['id'], keySet: {all: true}}).then(function(data) {
  var rows = data[0];
  console.log(rows);
});
```

evaogbe:
Fixes #2099

### DIFF
--- a/packages/spanner/src/table.js
+++ b/packages/spanner/src/table.js
@@ -349,6 +349,8 @@ Table.prototype.insert = function(keyVals, callback) {
  *     be yielded. If using a composite key, provide an array within this array.
  *     See the example below.
  * @param {string=} query.index - The name of an index on the table.
+ * @param {object=} query.keySet - Defines a collection of keys and/or key
+ *     ranges to read.
  * @param {number} query.limit - The number of rows to return.
  * @param {object=} options - [Transaction options](https://cloud.google.com/spanner/docs/timestamp-bounds).
  * @param {number} options.exactStaleness - Executes all reads at the timestamp


callmehiphop:
Rather than making the users write out the entire `keySet` object we've always supported the `keys` and `ranges` options (although I'm noticing ranges is not documented). These act as convenience options for `keySet`. Are there use cases for wanting to pass in a raw `keySet` object?

/cc @stephenplusplus @vkedia

stephenplusplus:
My choice would be to allow "shorthand" options, but always allow "raw" options. `createReadStream` is called underneath, which accepts the raw `query` object: https://github.com/eoogbe/google-cloud-node/blob/8bc3030d97444be2b0e8bb0e89ebd494a85b08bf/packages/spanner/src/table.js#L113 -- it would be nice to allow the user to pass that through from `read()` as well.

Note that I'm not very well-versed in the conventions of Spanner and our API. If this was closed out for a reason, we probably just need better docs about it.

eoogbe:
@callmehiphop @stephenplusplus The `keySet` option already works; it's just undocumented. This PR fixes that by adding documentation as requested in #2099. If we want to keep the documentation incomplete for some reason, we should close that issue.

stephenplusplus:
I think it's fine to add this to the docs. We typically try to show the common-case options in the docs, and let the others exist on the cloud.google.com site, which we link to. In this case, we probably just overlooked `keySet`. Thanks for fixing this!

PR #2515, opened 2017-08-08
Storage: add system test for unicode filenames.

### LINKED ISSUE #2168: Add integration test for Unicode filenames.
Add a system test to prove that we do not normalize Unicode resource names. See https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3215

landrito:
Resolves #2168 

This is a port of:
https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/933
https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3215

### DIFF
--- a/packages/storage/system-test/storage.js
+++ b/packages/storage/system-test/storage.js
@@ -560,38 +560,40 @@ describe('storage', function() {
 
     // Normalization form C: a single character for e-acute;
     // URL should end with Cafe%CC%81
-    describe('normalization form C', function() {
-      it('should not perform normalization', function(done) {
-        var name = 'Caf\u00e9';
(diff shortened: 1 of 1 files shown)


stephenplusplus:
I pushed a small refactor. PTAL-- most noteworthy:

- Instead of describe blocks with one test each, I just took the tests out into the general "unicode validation" describe
- When you use promises with Mocha, you can just return it to the function without calling done. Your `then` blocks with asserts will still run

landrito:
Thanks for the refactor @stephenplusplus!

> When you use promises with Mocha, you can just return it to the function without calling done. Your then blocks with asserts will still run

Good to know!

LGTM

PR #2517, opened 2017-08-09
common/spanner: introduce `debug` module and nested transaction warning

### LINKED ISSUE #2361: Cloud Spanner Client should prevent nested transactions
Cloud Spanner does not support [nested transactions](https://en.wikipedia.org/wiki/Nested_transaction). But the client library does allow users to call `Database.runTransaction` inside of the callback that is provided to an outer `Database.runTransaction`. This is misleading since users might believe that this will behave like a nested transaction but in reality these will be two independent transactions. This is confusing and a source of bugs. Specifically the inner transaction might succeed while the outer might ABORT, in which case the callback will be rerun thus rerunning the inner transaction. 
We should prevent this from happening by detecting and raising an error if someone calls `Database.runTransaction` inside the callback provided to `Database.runTransaction`.  Note that this needs to be done before Beta since this is a known breaking change.
Also note that if we come across a use case where we do want to allow calling a nested runTransaction, we should be able to enable that in future without making a breaking change. 

cc @jgeewax @bjwatson

stephenplusplus:
Fixes #2361

### DIFF
--- a/packages/spanner/src/database.js
+++ b/packages/spanner/src/database.js
@@ -787,7 +787,9 @@ Database.prototype.runStream = function(query, options) {
  * atomically at a single logical point in time across columns, rows, and tables
  * in a database.
  *
- * Note that Cloud Spanner does not support nested transactions.
+ * Note that Cloud Spanner does not support nested transactions. If a new
+ * transaction is started inside of the run function, it will be an independent
+ * transaction.
  *
  * The callback you provide to this function will become the "run function". It
  * will be executed with either an error or a {module:spanner/transaction}


packages/spanner/system-test/spanner.js
```diff
@@ -2702,7 +2702,7 @@ var spanner = new Spanner(env);
     });
 
     describe('read only', function() {
-      it('should run a read only transaction', function(done) {
+      it.only('should run a read only transaction', function(done) {
```
callmehiphop: Don't forget about this lil guy

packages/spanner/src/database.js
```diff
@@ -892,6 +892,11 @@ Database.prototype.runTransaction = function(options, runFn) {
 
   options = extend({}, options);
 
+  this.debug([
+    'Stating a transaction. Note that nested transactions are not currently',
```
bjwatson: Stating -> Starting

stephenplusplus:
*bump*

packages/spanner/src/database.js
```diff
@@ -892,6 +892,11 @@ Database.prototype.runTransaction = function(options, runFn) {
 
   options = extend({}, options);
 
+  this.debug([
+    'Starting a transaction. Note that nested transactions are not currently',
```
vkedia: Let us remove "currently" from this. "currently" indicates that they might be supported at some point in time.

vkedia:
Looking at this code, it is not clear to me exactly when would this warning be triggered?

stephenplusplus:
The user opts in to this warning behavior by using the popular `debug` module. When they start their app, it would look like:

```sh
export DEBUG=@google-cloud/spanner; npm start
```

The line itself will be printed to their terminal any time a transaction is started.

vkedia:
Oh so this would be printed every time a transaction is started and not just if we detect nested transactions?

stephenplusplus:
Right, we couldn't come up with a semi-reliable way to detect nested transactions without an API redesign. By always warning, in the worst case, it's excess noise to the user. But, by using the `debug` module, the user is expecting to receive a steady stream of output, so I wouldn't expect this to stand out in a way that's overly obtrusive.

vkedia:
@stephenplusplus I am not sure how much utility this would have on top of documentation. It just seems strange that we are telling them something which they might even know exist. It might infact confuse users and they might worry why they are getting this warning and maybe they are doing something wrong.
I would say let us not merge this and instead just document this (if not already documented).

stephenplusplus:
@vkedia updated the docs to make the same mention. Feel free to fork this PR if you have a better idea for the implementation, or words to use to explain.

packages/spanner/src/database.js
```diff
@@ -787,6 +787,8 @@ Database.prototype.runStream = function(query, options) {
  * atomically at a single logical point in time across columns, rows, and tables
  * in a database.
  *
+ * Note that Cloud Spanner does not support nested transactions.
```
vkedia: I think we should elaborate on what exactly that means in the context of this client. For eg something like.
```
Note that Cloud Spanner does not support nested transactions. If a new transaction is started inside the "run function" it will be an independent transaction."
```

Also it might make sense to then put this comment after the paragraph below since that is where we define what is a "run fucntion"

PR #2520, opened 2017-08-09
Logging: allow for camelcase resource labels

### LINKED ISSUE #2494: Logging resource labels use snake_case rather than camelCase
*Originally reported by @jwgoh in https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2484*

I tried specifying the monitored resource manually. The keys for labels seems to only take snake_case i.e. project_id and NOT camelCase projectId.

```
// Works
resource: {
  type: 'gce_instance',
  labels: {
    project_id: projectId,
    instance_id: instanceId,
    zone: zone,
  },
},

// Does not work. Nothing gets logged
resource: {
  type: 'gce_instance',
  labels: {
    projectId: projectId,
    instanceId: instanceId,
    zone: zone,
  },
},
```

I referred to the documentations [here](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/a64518e80742429dd3b4021a6f74c8235faa0375/packages/logging-winston/src/index.js#L82-L86) and [here](https://cloud.google.com/logging/docs/reference/v2/rest/v2/MonitoredResource) when specifying the key-value.

landrito:
Resolves #2494

### DIFF
--- a/packages/logging/test/log.js
+++ b/packages/logging/test/log.js
@@ -362,6 +362,39 @@ describe('Log', function() {
       log.write(ENTRY, optionsWithResource, done);
     });
 
+    it('should transform camelcase label keys to snake case', function(done) {
+      var CUSTOM_RESOURCE = {
+        labels: {
+          camelCaseKey: 'camel-case-key-val'
+        }
+      };
+      var EXPECTED_RESOURCE = {
+        labels: {
+          camel_case_key: 'camel-case-key-val'
+        }
+      };
+      var optionsWithResource = extend({}, OPTIONS, {
+        resource: CUSTOM_RESOURCE
+      });
+
+      log.logging.request = function(config, callback) {
+        assert.strictEqual(config.client, 'loggingServiceV2Client');
+        assert.strictEqual(config.method, 'writeLogEntries');
+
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Awesome, thanks!

ofrobots:
What is the semver-ness of this change?

stephenplusplus:
- Users providing snake case before can continue to provide snake case
- Users providing camel case before would not see those properties saved
- Users providing camel case after will see those properties saved

If that's all true, I would see this as a fix. The only thing that will change unexpectedly for a user is that our advertised API will work as they thought it would (but didn't notice that it wasn't).

Did I miss anything? What do you think?

ofrobots:
That reasoning is good :+1:, thank you!

However, why do we need to special case `resource.labels` for snake_casing? For example, the of the logging code base provides properties of the form `entryMetadata.httpRequest` and that works fine.

Does some layer (common? gax? grpc?) convert those keys to snake_case automatically, but somehow `resource.labels` were missed? Or is it this a problem in the logging service?

landrito:
The resource labels are a predefined list with a specific casing structure that can be found here:
https://cloud.google.com/logging/docs/api/v2/resource-list

PR #2523, opened 2017-08-10
logging: use correct case for cluster-name attribtue

ofrobots:
Metadata services attributes seem to be in kebab-case.

Example from a GKE instance:

```sh
ofrobots@gke-cloudcats-default-pool-xxxxxxx-xxxx ~ $ curl "http://metadata.google.internal/computeMetadata/v1/inst
ance/attributes/" -H "Metadata-Flavor: Google"
cluster-name
configure-sh
created-by
gci-ensure-gke-docker
gci-update-strategy
instance-template
kube-env
user-data
```


ofrobots:
@stephenplusplus PTAL.

PR #2525, opened 2017-08-11
Common: don't assume missingProjectIdError

### LINKED ISSUE #2069: util.decorateRequest should confirm error is a missing project ID error
A bug in the `google-cloud` package resulted in the following error:

```
Error: Sorry, we cannot connect to Cloud Services without a project ID. You may specify one with an environment variable named "GCLOUD_PROJECT". See https://googlecloudplatform.github.io/google-cloud-node/#/docs/guides/authentication for a detailed guide on creating an authenticated connection.
    at Object.<anonymous> (/Users/gajus/Documents/dev/applaudience/movie-editor/node_modules/google-cloud/node_modules/@google-cloud/common/src/util.js:55:29)
    at Module._compile (module.js:571:32)
    at Module._extensions..js (module.js:580:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/gajus/Documents/dev/applaudience/movie-editor/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/gajus/Documents/dev/applaudience/movie-editor/node_modules/google-cloud/node_modules/@google-cloud/common/src/service-object.js:32:12)

```

The reason I am getting this error is [this code block](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/1939deaa32298b3cbe123686e5c52f480c1d7ee2/packages/common/src/util.js#L351-L358):

```js
        try {
          authenticatedReqOpts = util.decorateRequest(
            authenticatedReqOpts,
            extend({ projectId: authClient.projectId }, config)
          );
        } catch(e) {
          err = util.missingProjectIdError;
        }
      }

```

Here `catch` assumes that any error that occurred in `try` is equivalent to `missingProjectIdError`.

Instead of blindly assuming what is the error, use `instanceof` check and a custom error constructor, e.g. https://github.com/gajus/xfetch/blob/0d8cd012354b0bb8b67162f74b2a9e126d3bdabb/src/attemptRequest.js#L52-L67.

Furthermore, what made it hard to debug is the lack stack trace (raised a separate issue https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2068).

landrito:
Resolves #2069

### DIFF
--- a/packages/common/src/util.js
+++ b/packages/common/src/util.js
@@ -353,7 +353,7 @@ function makeAuthenticatedRequestFactory(config) {
             extend({ projectId: authClient.projectId }, config)
           );
         } catch(e) {
-          err = util.missingProjectIdError;
+          err = e;
         }
       }
 


stephenplusplus:
Well, that was easy.

PR #2526, opened 2017-08-11
Update monitoring client

### LINKED ISSUE #1801: Remove references to auto-generated in Monitoring README
### What
Change [these lines](https://github.com/GoogleCloudPlatform/google-cloud-node/blame/master/packages/monitoring/README.md#L8-L10) in the NodeJS Monitoring README. Do not explicitly state that it is an auto-generated library. Also, do not say that "a handwritten layer is not yet available".

### Why
GAPIC-only libraries will become the norm as the code generators become capable of producing client libraries faster than we can hand-write them. We do not want to lead with calling them "auto-generated", since that can have negative connotations and is not the defining feature of the library design.

Stating that "a handwritten layer is not yet available" suggests that one might eventually be available. We don't want to lead developers to believe that they might eventually have a different surface to work with (apart from major version updates).

### Notes
@callmehiphop I think @landrito originally did this, but I could not assign this ticket to him. Could you or @stephenplusplus give him, @jmuk, and @swcloud the necessary permission to own issues here? Thanks!

FYI @omaray

### LINKED ISSUE #2486: Missing dummy docs JS files
@lukesneeringer Would you mind generating the dummy doc JS files for these APIs?:

- [x] Language
- [x] Monitoring
- [x] Video Intelligence
- [x] Vision

Thanks!

evaogbe:
Closes #2486, closes #1801

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ This client supports the following Google Cloud Platform services at a [Beta](#v
 * [Cloud Spanner](#cloud-spanner-beta) (Beta)
 * [Cloud Vision](#cloud-vision-beta) (Beta)
 * [Google BigQuery](#google-bigquery-beta) (Beta)
+* [Google Stackdriver Monitoring](#google-stackdriver-monitoring-beta) (Beta)
 
 This client supports the following Google Cloud Platform services at an [Alpha](#versioning) quality level:
 
@@ -36,7 +37,6 @@ This client supports the following Google Cloud Platform services at an [Alpha](
 * [Google Stackdriver Debugger](#google-stackdriver-debugger-alpha) (Alpha)
 * [Google Stackdriver Error Reporting](#google-stackdriver-error-reporting-alpha) (Alpha)
 * [Google Stackdriver Trace](#google-stackdriver-trace-alpha) (Alpha)
-* [Google Stackdriver Monitoring](#google-stackdriver-monitoring-alpha) (Alpha)
 
 If you need support for other Google APIs, check out the [Google Node.js API Client library][googleapis].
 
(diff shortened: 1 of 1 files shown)


packages/monitoring/package.json
```diff
@@ -1,40 +1,13 @@
...
-    {
-      "name": "Silvano Luciani",
-      "email": "silvano@google.com"
-    },
-    {
-      "name": "Stephen Sawchuk",
-      "email": "sawchuk@gmail.com"
-    }
-  ],
+  "version": "0.3.0",
```
stephenplusplus: Version shouldn't change here.

packages/monitoring/package.json
```diff
@@ -48,19 +21,21 @@
...
-    "google-proto-files": "^0.12.0"
+    "extend": "^3.0",
+    "google-gax": "^0.13.2",
+    "google-proto-files": "^0.12.0",
+    "lodash.union": "^4.6.0"
   },
   "devDependencies": {
-    "mocha": "^3.1.2"
+    "mocha": "^3.2.0",
+    "through2": "^2.0.3"
```
stephenplusplus: Not used.

packages/monitoring/src/v3/index.js
```diff
@@ -31,7 +32,11 @@ function v3(options) {
   return result;
 }
 
+v3.GAPIC_VERSION = '0.0.5';
 v3.SERVICE_ADDRESS = groupServiceClient.SERVICE_ADDRESS;
-v3.ALL_SCOPES = groupServiceClient.ALL_SCOPES;
+v3.ALL_SCOPES = union(
```
stephenplusplus: The `lodash.union` dependency isn't needed-- these `ALL_SCOPES` values are identical.

packages/monitoring/package.json
```diff
@@ -30,8 +30,7 @@
     "lodash.union": "^4.6.0"
```
stephenplusplus: Thanks for the changes! We can remove it from `package.json` here as well.

packages/monitoring/src/index.js
```diff
@@ -20,35 +20,91 @@
...
+const VERSION = require('../package.json').version;
+
 /**
- * [Stackdriver Monitoring](https://cloud.google.com/monitoring/docs) collects
- * metrics, events, and metadata from Google Cloud Platform, Amazon Web Services
- * (AWS), hosted uptime probes, application instrumentation, and a variety of
- * common application components including Cassandra, Nginx, Apache Web Server,
- * Elasticsearch and many others. Stackdriver ingests that data and generates
- * insights via dashboards, charts, and alerts.
+ * Create an groupServiceClient with additional helpers for common
```
stephenplusplus: Create a*

packages/monitoring/src/index.js
```diff
@@ -20,35 +20,91 @@
...
+    libVersion: VERSION
+  });
+
+  // Create the client with the provided options.
+  var client = gapic.v3(options).groupServiceClient(options);
+  return client;
+}
+
+/**
+ * Create an metricServiceClient with additional helpers for common
```
stephenplusplus: Create a*

packages/monitoring/src/index.js
```diff
@@ -20,35 +20,91 @@
```
stephenplusplus: Can you add:

```js
@name Monitoring
```

to the above block, under `@module monitoring`?

PR #2530, opened 2017-08-14
bigquery: api improvements

### LINKED ISSUE #2511: bigquery: API improvements
#### Querying

- [x] Change high-level query to use jobs.
- [x] Auto-generate Job IDs on client
- [x] Allow for user specified prefix for Job IDs

#### Jobs

- [ ] Media upload support
- [ ] Fields
- [ ] Labels
- [ ] "all" flag for job.list

#### Tables

- [x] Partitioning
- [x] Views
- [x] Labels
- [x] Expiration
- [x] etag Support

#### Datasets

- [ ] ACL support
- [x] Labels
- [x] Default expiration
- [x] etag Support

callmehiphop:
Closes #2511

#### TODOs

- [x] Documentation
- [x] Unit Tests
- [x] System Tests

### DIFF
--- a/packages/bigquery/src/table.js
+++ b/packages/bigquery/src/table.js
@@ -198,7 +198,7 @@ function Table(dataset, id) {
      *   }
      * };
      *
-     * table.setMetadata(metadata, function(err, apiResponse) {})
+     * table.setMetadata(metadata, function(err, apiResponse) {});
      *
      * //-
      * // If the callback is omitted, we'll return a Promise.


callmehiphop:
@tswast if you have some free time, could you look through this PR to make sure all the P0 items look ok?

I also want to mention for CRUD calls on specific properties (labels, timePartitioning, etc.), this all happens through a `setMetadata` method, which is just a PATCH request that the user can supply an object to.

e.g.

```js
var metadata = {
  labels: {
    foo: 'bar'
  }
};

table.setMetadata(metadata, (err, apiResponse) => {
  // ...
});
```

packages/bigquery/src/index.js
```diff
@@ -77,6 +79,94 @@ function BigQuery(options) {
...
+      }
+      case 'BYTES': {
+        value = new Buffer(value, 'base64');
+        break;
+      }
+      case 'FLOAT': {
+        value = parseFloat(value);
+        break;
+      }
+      case 'INTEGER': {
```
tswast: I think you might need two cases here eventually. Standard SQL renamed some of the types, such as `INT64` https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#integer-type

packages/bigquery/src/index.js
```diff
@@ -426,6 +516,89 @@ BigQuery.prototype.createDataset = function(id, options, callback) {
...
+ *     }
+ *   }
+ * };
+ *
+ * bigquery.createJob(options, function(err, job) {
+ *   if (err) {
+ *     // Error handling omitted.
+ *   }
+ *
+ *   job.getQueryResults(function(err, rows) {});
```
tswast: Why is `getQueryResults` being called here? If it was meant to wait, shouldn't it poll?

packages/bigquery/src/index.js
```diff
@@ -779,91 +952,49 @@ BigQuery.prototype.job = function(id) {
...
 
-  if (options.params) {
-    options.useLegacySql = false;
-    options.parameterMode = is.array(options.params) ? 'positional' : 'named';
+  var reqOpts = extend({}, query);
 
-    if (options.parameterMode === 'named') {
-      options.queryParameters = [];
+  if (query.params) {
+    reqOpts.useLegacySql = false;
```
tswast: Seems a little confusing to automatically set the syntax type here only if `query.params` is set. Some of the other languages are defaulting to standard SQL in all cases, unless it's explicitly set to legacy.

packages/bigquery/src/job.js
```diff
@@ -266,14 +267,41 @@ Job.prototype.cancel = function(callback) {
...
+    }
+
+    var rows = [];
+
+    if (resp.schema && resp.rows) {
+      rows = self.bigQuery.mergeSchemaWithRows_(resp.schema, resp.rows);
+    }
+
+    var nextQuery = null;
+    if (resp.jobComplete === false) {
```
tswast: I can't tell what's going on here. Is this polling?

callmehiphop: @tswast yep! We recursively call any rpcs that require pagination until we have received all the data. We generally stop when we do not receive a pageToken, however in this case we also check to see if the job is completed or not.

packages/bigquery/src/index.js
callmehiphop: Internally we recursively call `getQueryResults`, should we be waiting to start this process?

packages/bigquery/src/index.js
callmehiphop: This was legacy code, I've gone ahead and set the default value for `useLegacySql` to false.

packages/bigquery/src/job.js
tswast: This concerns me. The result set for a query or table data listing can be very large (hundreds of megabytes to gigabytes, or even petabytes). This implementation means someone could not use this library to process such large result sets that do not fit memory or have too many pages that they blow the stack.

stephenplusplus: Our pagination methods also allow `autoPaginate: false`, where a user will be able to page through the responses at the rate they wish. There is an example here: https://googlecloudplatform.github.io/google-cloud-node/#/docs/bigquery/0.9.6/bigquery/job?method=getQueryResults

```js
function manualPaginationCallback(err, rows, nextQuery, apiResponse) {
  if (nextQuery) {
    // More results exist.
    job.getQueryResults(nextQuery, manualPaginationCallback);
  }
}

job.getQueryResults({
  autoPaginate: false
}, manualPaginationCallback);
```

packages/bigquery/src/index.js
tswast: `getQueryResults` should only be called for query jobs. I don't see a check here for the the type of job. When I try to call `getQueryResults` on a copy job resource, I get an error response

```
{
 "error": {
  "errors": [
   {
    "domain": "global",
    "reason": "invalid",
    "message": "Job 769108155013:8a432e18-336e-489f-bb90-834327622456 is not a query job",
   }
  ],
  "code": 400,
  "message": "Job 769108155013:8a432e18-336e-489f-bb90-834327622456 is not a query job"
 }
}
```

The more general way to check for job completion is to repeatedly [get the job resource](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/get) (note: does not wait for a timeout like `getQueryResults` does, so a polling interval is needed) and check for the [status.state](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#status.state) `DONE`.

packages/bigquery/src/index.js
```diff
@@ -426,6 +519,89 @@ BigQuery.prototype.createDataset = function(id, options, callback) {
...
+ *   job.getQueryResults(function(err, rows) {});
+ * });
+ *
+ * //-
+ * // If the callback is omitted, we'll return a Promise.
+ * //-
+ * bigquery.createJob().then(function(data) {
+ *   var job = data[0];
+ *
+ *   return job.getQueryResults();
```
tswast: Ditto. Are these aways query jobs? `job.get` should be called in the general case.

packages/bigquery/src/job.js
tswast: Is it possible to pass in `autoPaginate` false when the job is created / query is started? It's a bit dangerous to automatically loop through the results, especially if `allowLargeResults` is set in the query job.

packages/bigquery/src/index.js
```diff
@@ -931,62 +1073,58 @@ BigQuery.prototype.query = function(options, callback) {
...
-    var job = that.job(resp.jobReference.jobId);
-    job.metadata = resp;
+  if (query.jobPrefix) {
+    reqOpts.jobPrefix = query.jobPrefix;
+    delete query.jobPrefix;
+  }
 
-    callback(null, job, resp);
-  });
+  this.createJob(reqOpts, callback);
```
tswast: Since it looks like `createJob` is also trying to wait for results, it seems the docs for this method are out of date.

It makes sense to me to have a callback called when the job is done, but it should not fetch all the results (in case they are very large)

packages/bigquery/src/job.js
stephenplusplus: For a definitive answer, I'll defer to @callmehiphop. The way the API is currently (without this PR), the user can pass `autoPaginate: false` to `bigquery.query(...)`, which would work the same was as the example in my last post.

Manually paginating was actually the default behavior when this library started. Across many APIs, we share the concern that the incoming bytes can be too large. I believe the rationale to default autoPaginate to true was the motto we hear throughout discussions like these, something like: "make the common case easy and the advanced case possible".

callmehiphop: It is, however to re-use the same job the user would need to use the `startQuery` method to cache the job object.

packages/bigquery/src/index.js
callmehiphop: For larger queries we would recommend using `startQuery` (the lower level version) to cache the job object and either manually paginate or use the streaming version.

```js
dataset.startQuery('SELECT * FROM `table`', job => {
  job.getQueryResultsStream()
    .on('error', err => console.error(err))
    .on('data', row => console.log(row))
})
```

packages/bigquery/src/index.js
callmehiphop: I don't think it's possible to reliably sniff for this. If a user knows the job id, they could do something like this

```js
const job = bigquery.job('my-job-id')
const rows = await job.getQueryResults()
```

And we would have no way to detect what kind of job it actually is without an additional request. I think our best bet is to allow the upstream API to do the error handling and improve our documentation on when to call `get` and when to call `getQueryResults`

tswast: What do you mean? This function is called `createJob`. You should be able to check for the type of job from the options object.

If it was `createQueryJob` then by all means call `getQueryResults`. But it's no good to call that for all jobs.

callmehiphop: When creating a job it's possible, but if a user were to simply reference a job id, then it would be impossible. At that point we're handling the same error in a different way. For the sake of consistency I'd prefer to let the upstream API handle it

tswast: Maybe I'm missing something? The line I'm referring to is in a generic createJob function, which should not be calling `getQueryResults` since it only applies to query jobs.

stephenplusplus: This comment chain seems to be in reference to our example code in the documentation comment block. The full code of the example snippet:

```js
 * var options = {
 *   configuration: {
 *     query: {
 *       query: 'SELECT url FROM [publicdata:samples.github_nested] LIMIT 100'
 *     }
 *   }
 * };
 *
 * bigquery.createJob(options, function(err, job) {
 *   if (err) {
 *     // Error handling omitted.
 *   }
 *
 *   job.getQueryResults(function(err, rows) {});
 * });
```

So this example shows how to use `createJob()` to create a query job, and use the returned `job` object's `getQueryResults()` to get the results.

The actual code of the method `createJob()` is here: https://github.com/callmehiphop/gcloud-node/blob/af1fc347f8a283faf13fdae0a1cb7f3cdddb3d93/packages/bigquery/src/index.js#L568

Sorry if I'm misunderstanding.

tswast: 🤦‍♂️ You're right. This example code makes sense. I should have realized it was example code, sorry.

packages/bigquery/src/index.js
tswast: Oops, more example code. Sorry for the noise.

callmehiphop:
@stephenplusplus @tswast I think I'm ready for a final review if you can find some time soon.

packages/bigquery/src/dataset.js
```diff
@@ -178,6 +178,17 @@ function Dataset(bigQuery, id) {
   });
 
   this.bigQuery = bigQuery;
+
+  this.interceptors.push({
+    request: function(reqOpts) {
+      if (reqOpts.method === 'PATCH' && reqOpts.json.etag) {
+        reqOpts.headers = reqOpts.headers || {};
+        reqOpts.headers['If-Match'] = reqOpts.json.etag;
```
stephenplusplus: What methods is this used for? I think localizing this to the methods themselves would be better, so no magic is happening.

stephenplusplus:
Facing a system test error:

```
  1) BigQuery BigQuery/Table should insert rows via stream:
     Value '{{projectId}}' in content does not agree with value 'nth-circlet-705'. This can happen when a value set through a parameter is inconsistent with a value set in the request.
  ApiError
      at Object.parseHttpRespBody (node_modules\@google-cloud\common\src\util.js:192:30)
      at Object.handleResp (node_modules\@google-cloud\common\src\util.js:132:18)
      at Request._callback (node_modules\@google-cloud\common\src\util.js:255:14)
      at Request.self.callback (node_modules\request\request.js:186:22)
      at Request.<anonymous> (node_modules\request\request.js:1163:10)
      at IncomingMessage.<anonymous> (node_modules\request\request.js:1085:12)
      at endReadableNT (_stream_readable.js:1059:12)
      at _combinedTickCallback (internal/process/next_tick.js:138:11)
      at process._tickCallback (internal/process/next_tick.js:180:9)
```

This is a console.log of `bigquery`:

```js
BigQuery {
  makeAuthenticatedRequest:
   { [Function: makeAuthenticatedRequest]
     getCredentials: [Function: bound getCredentials],
     authClient:
      Auth {
        authClientPromise: [Object],
        authClient: [Object],
        config: [Object],
        environment: {},
        projectId: 'nth-circlet-705' } },
  authClient:
   Auth {
     authClientPromise: Promise { [Object] },
     authClient:
      JWT {
        transporter: DefaultTransporter {},
        clientId_: undefined,
        clientSecret_: undefined,
        redirectUri_: undefined,
        opts: {},
        credentials: [Object],
        email: 'stephen-windows@nth-circlet-705.iam.gserviceaccount.com',
        keyFile: undefined,
        key: '-----BEGIN PRIVATE KEY----- ... -----END PRIVATE KEY-----\n',
        scopes: [Array],
        subject: undefined,
        gToken: [Function: GoogleToken],
        projectId: 'nth-circlet-705',
        gtoken: [Object] },
     config:
      { baseUrl: 'https://www.googleapis.com/bigquery/v2',
        scopes: [Array],
        packageJson: [Object] },
     environment: {},
     projectId: 'nth-circlet-705' },
  baseUrl: 'https://www.googleapis.com/bigquery/v2',
  getCredentials: [Function: bound getCredentials],
  globalInterceptors: [],
  interceptors: [],
  packageJson: { ... },
  projectId: '{{projectId}}',
  projectIdRequired: true,
  Promise: [Function: Promise] }
```

packages/bigquery/src/table.js
```diff
@@ -644,6 +579,10 @@ Table.prototype.createWriteStream = function(metadata) {
       metadata: {
         configuration: {
           load: metadata
+        },
+        jobReference: {
+          jobId: jobId,
+          projectId: self.bigQuery.projectId
```
stephenplusplus: I don't believe we test that `projectId` is in the request.

packages/bigquery/src/table.js
```diff
@@ -631,6 +559,13 @@ Table.prototype.createWriteStream = function(metadata) {
     }
   });
 
+  var jobId = uuid.v4();
+
+  if (metadata.jobPrefix) {
+    jobId = metadata.jobPrefix + jobId;
+    delete metadata.jobPrefix;
```
stephenplusplus: I don't think there's a test that checks that this was deleted.

packages/bigquery/src/table.js
```diff
@@ -953,102 +822,40 @@ Table.prototype.getRows = function(options, callback) {
...
-      }
-
-      // If no explicit format was provided, attempt to find a match from
-      // the file's extension. If no match, don't set, and default upstream
-      // to CSV.
-      var format = FORMATS[path.extname(src.name).substr(1).toLowerCase()];
-      if (!metadata.sourceFormat && format) {
-        body.configuration.load.sourceFormat = format;
-      }
+    job.on('error', callback);
```
stephenplusplus: ⛓

packages/bigquery/src/table.js
```diff
@@ -1185,15 +1009,33 @@ Table.prototype.insert = function(rows, options, callback) {
...
 
   this.request({
     method: 'POST',
     uri: '/insertAll',
     json: json
   }, function(err, resp) {
     if (err) {
-      callback(err, resp);
+      if (err.code === 404 && autoCreate) {
+        setTimeout(createTableAndRetry, Math.random() * 60000);
```
stephenplusplus: Why not create right away?

packages/bigquery/src/table.js
```diff
@@ -1064,12 +871,21 @@ Table.prototype.import = function(source, metadata, callback) {
  *
  * @param {object|object[]} rows - The rows to insert into the table.
  * @param {object=} options - Configuration object.
+ * @param {boolean} options.autoCreate - Automatically create the table if it
```
stephenplusplus: Maybe we should leave a note that this can take longer than 2 minutes to complete.

packages/bigquery/src/table.js
```diff
@@ -1267,31 +1124,450 @@ Table.prototype.query = function(query, callback) {
...
+ */
+Table.prototype.startCopy = function(destination, metadata, callback) {
+  if (!(destination instanceof Table)) {
+    throw new Error('Destination must be a Table object.');
+  }
 
-  if (metadata.name) {
-    metadata.friendlyName = metadata.name;
-    delete metadata.name;
+  if (is.fn(metadata)) {
```
stephenplusplus: I don't think there's a test for this being optional.

packages/bigquery/src/table.js
```diff
@@ -1267,31 +1124,450 @@ Table.prototype.query = function(query, callback) {
...
+Table.prototype.startCopyFrom = function(sourceTables, metadata, callback) {
+  sourceTables = arrify(sourceTables);
+
+  sourceTables.forEach(function(sourceTable) {
+    if (!(sourceTable instanceof Table)) {
+      throw new Error('Source must be a Table object.');
+    }
+  });
+
+  if (is.fn(metadata)) {
```
stephenplusplus: I don't think this is explicitly tested as being optional.

packages/bigquery/test/table.js
```diff
@@ -1530,100 +1156,751 @@ describe('BigQuery/Table', function() {
...
+    it('should pass the callback to createJob', function(done) {
+      table.bigQuery.createJob = function(reqOpts, callback) {
+        assert.strictEqual(done, callback);
+        callback(); // the done fn
+      };
+
+      table.startCopyFrom(SOURCE_TABLE, done);
+    });
+
+    it('should pass an error to the callback', function(done) {
```
stephenplusplus: We can drop this.

packages/bigquery/test/table.js
```diff
@@ -1530,100 +1156,751 @@ describe('BigQuery/Table', function() {
...
+      };
+
+      table.bigQuery.job = function(id) {
+        return { id: id };
+      };
+
+      table.bigQuery.createJob = function() {};
+    });
+
+    it('should send the correct API request', function(done) {
```
stephenplusplus: This should probably be called `should call createJob correctly`

callmehiphop:
@dwmclary for whatever reason it won't let me add you as a reviewer, so I hope a friendly ping here will suffice.

packages/bigquery/src/dataset.js
callmehiphop: The way it is documented in the upstream docs, it applies for any method that makes a `PATCH` request.

https://cloud.google.com/bigquery/docs/api-performance#read-patch-write

callmehiphop:
@stephenplusplus in regard to https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2530#issuecomment-331916770 - are you using a specific OS/Node version/etc.? I can't reproduce it on my machine.

packages/bigquery/src/dataset.js
stephenplusplus: Ah okay. Can you leave that link in a code comment?

packages/bigquery/test/dataset.js
```diff
@@ -577,6 +639,48 @@ describe('BigQuery/Dataset', function() {
     });
   });
 
+  describe('startQuery', function() {
+    var FAKE_QUERY = 'SELECT * FROM `table`';
+
+    it('should call through to bigQuery#startQuery', function(done) {
+      ds.bigQuery.startQuery = function() {
```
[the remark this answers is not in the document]
callmehiphop: We do in tests below, this test might just be unnecessary.

stephenplusplus:
Windows 10, Node v8.6.0, npm 5.3.0.

It might depend on how you are auth'ing. From `table.createWriteStream()`:

```js
      request: {
        uri: format('{base}/{projectId}/jobs', {
          base: 'https://www.googleapis.com/upload/bigquery/v2/projects',
          projectId: self.bigQuery.projectId
        })
      }
```

I'm using ADC, so `self.bigQuery.projectId` isn't set. This probably isn't new to this PR, but maybe sneak a fix in?

packages/bigquery/test/dataset.js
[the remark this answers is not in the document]
stephenplusplus: I'd go either way. We usually have a test that's just straight-up "are the arguments the same as what we passed in", and then a bunch more tests that handle the non-linear paths.

packages/bigquery/src/table.js
callmehiphop: https://github.com/callmehiphop/gcloud-node/blob/79bcb4d0b6ea0c8ce2b42f83cd95a99b87cb1951/packages/bigquery/test/table.js#L655

packages/bigquery/src/table.js
callmehiphop: https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1635#issuecomment-327543698

callmehiphop:
@stephenplusplus ah, gotcha. I'm wondering what the most correct solution here is, initially I would think to call `util.replaceProjectIdToken(...)` on all the [options](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/common/src/util.js#L219), but that feels wrong. If you think it's more involved than that, I would ask we open a separate issue so we don't delay the release of these changes.

packages/bigquery/src/table.js
stephenplusplus: Thanks-- we should definitely link to that in a code comment as well.

stephenplusplus:
That makes sense, re: separate effort.

callmehiphop:
@stephenplusplus I think we're still waiting for an approval from @dwmclary, but is there anything else I may have forgotten to tackle in this?

stephenplusplus:
@callmehiphop want to just merge this?

PR #2533, opened 2017-08-15
logging-winston: fix `0` priority log levels.

joshuatoenyes:
This PR correct a bug causing log levels with a `0` priority to throw an `Unknown log level: ...` error.

### DIFF
--- a/packages/logging-winston/src/index.js
+++ b/packages/logging-winston/src/index.js
@@ -148,7 +148,7 @@ LoggingWinston.prototype.log = function(levelName, msg, metadata, callback) {
     metadata = {};
   }
 
-  if (typeof this.levels_[levelName] === 'undefined') {
+  if (this.levels_[levelName] === undefined) {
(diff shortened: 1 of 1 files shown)


JoshuaToenyes:
I signed it!

packages/logging-winston/src/index.js
```diff
@@ -148,7 +148,7 @@ LoggingWinston.prototype.log = function(levelName, msg, metadata, callback) {
     metadata = {};
   }
 
-  if (!this.levels_[levelName]) {
+  if (typeof this.levels_[levelName] === 'undefined') {
```
ofrobots: nit: I think a comparison of the form `this.levels_[levelName] === undefined` would work just as well here. I have pushed a change to this effect. Let me know if you disagree.

JoshuaToenyes:
👍 Thanks for approving so quickly and removing my `typeof` usage!

PR #2543, opened 2017-08-18
readme: link to auth info above samples

### LINKED ISSUE #2534: Update readme to include what keyFilename is
The readme uses keyFilename throughout but gives no link or description of what the "keyFilename" is.

stephenplusplus:
Fixes #2534 

cc @gauntface

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,12 @@ With `google-cloud` it's incredibly easy to get authenticated and start using Go
 If you are running this client on Google Cloud Platform, we handle authentication for you with no configuration. You just need to make sure that when you [set up the GCE instance][gce-how-to], you add the correct scopes for the APIs you want to access.
 
 ``` js
+var storage = require('@google-cloud/storage')();
+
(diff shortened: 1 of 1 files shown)


gauntface:
Could you add an actual example to the authentication section with `keyFilename` in the code snippet?

What I ended up doing was doing a search for `keyFilename` throughout the readme and only saw examples using it. Authentication makes total sense but I didn't see it because I expected to see a well of text along the lines of:

> To use our SDK you need to authenticate using a service account. The key you'll download from the service account panel is then passed in via the `keyFilename` parameter.

stephenplusplus:
I think the Authentication section is pretty thorough already because it has to cover multiple forms of authentication (application default, service account JSON/PEM, credentials object), as well as other steps, like billing & enabling APIs. To add more to it would bloat it out even more.

I'll merge this as-is for now, since this should at least be an improvement. A PR would be awesome though, if you can think of a way to make this even better.

Issue #1630, opened 2016-09-26
Custom Endpoint Support


callmehiphop:
Let's add support for any custom endpoint across the board. We already have support for Datastore, PubSub and (soon) Bigtable. For the most part supporting them has just been some minor configuration changes, so we should be able to support it everywhere.

We should have a setting, e.g. `options.apiEndpoint = 'http://localhost:8888` as well as environment variables.

- [ ] GOOGLE_CLOUD_BIGQUERY_ENDPOINT
- [ ] GOOGLE_CLOUD_BIGTABLE_ENDPOINT
- [ ] GOOGLE_CLOUD_COMPUTE_ENGINE_ENDPOINT // GCE_ENDPOINT?
- [ ] GOOGLE_CLOUD_DATASTORE_ENDPOINT
- [ ] GOOGLE_CLOUD_DNS_ENDPOINT
- [ ] GOOGLE_CLOUD_LANGUAGE_ENDPOINT
- [ ] GOOGLE_CLOUD_LOGGING_ENDPOINT
- [ ] GOOGLE_CLOUD_PREDICTION_ENDPOINT
- [ ] GOOGLE_CLOUD_PUBSUB_ENDPOINT
- [ ] GOOGLE_CLOUD_RESOURCE_MANAGER_ENDPOINT
- [ ] GOOGLE_CLOUD_SPEECH_ENDPOINT
- [ ] GOOGLE_CLOUD_STORAGE_ENDPOINT
- [x] GOOGLE_CLOUD_TRANSLATE_ENDPOINT - #1674
- [ ] GOOGLE_CLOUD_VISION_ENDPOINT

stephenplusplus:
Small title change to make sure it's clear we want to support "http://any-uri" as opposed to only official service emulators.

jmuk:
@danoscarmike might also be interested in the idea like this.

rmanyari:
I'm would be super excited to see this feature merged in the next release (custom endpoints are very very helpful for people who operate behind forward proxies). 

I can open a PR in the next days that puts a function very similar to {Datastore|PubSub}.prototype.determineBaseUrl_ in common.utils and then use it across the different services.

Rodrigo

lukesneeringer:
Are these individual environment variables? That seems lousy. I would much rather have this be a parameter that can be passed to clients.

rmanyari:
PubSub does expose a parameter (vs. other services that use environment variables):

[...]

PR #2195, opened 2017-04-10
datastore: convert to gax

stephenplusplus:
RE: #1859 

This converts the Datastore API to use GAX. This involves non-breaking API changes to support a `gaxOptions` configuration object.

#### Blockers

- [x] Allow grpc custom channel without auth: https://github.com/googleapis/gax-nodejs/pull/129
- [x] How to set a gRPC metadata header `google-cloud-resource-prefix`: https://github.com/googleapis/gax-nodejs/pull/130

#### To Dos

- [ ] Set `options.libName` so tracking is enabled in gax
- [x] Docs
- Convert
  - [x] index.js
  - [x] request.js
  - [x] transaction.js
- Tests
  - Unit
    - [x] index.js
    - [x] request.js
    - [x] transaction.js
  - [x] System

PR #2548, opened 2017-08-21
Feature custom endpoint

rmanyari:
Support for custom endpoints at service instantiation. This first PR covers the following services:

- bigquery
- bigtable
- gce
- datastore
- dns
- language

Custom enpoints can be defined using environment variables: GOOGLE_CLOUD_[SERVICE_NAME]_ENDPOINT or the options.{apiEndpoint,servicePath} options

Issue: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1630

Signed-off-by: Rodrigo Manyari <rodrigo@twosigma.com>

### DIFF
--- a/packages/vision/src/index.js
+++ b/packages/vision/src/index.js
@@ -73,7 +73,7 @@ function visionV1(opts) {
   // Create the image annotator client with the provided options.
   var client = gapic.v1(opts).imageAnnotatorClient(opts);
   if (is.undefined(client.annotateImage)) {
-    Object.assign(client.constructor.prototype, helpers('v1'));
+    Object.assign(client, helpers('v1'));
   }
   return client;
 }


rmanyari:
I signed it!

packages/bigquery/src/index.js
```diff
@@ -57,16 +57,30 @@ var Table = require('./table.js');
...
+
+  var baseInfo = common.util.determineBaseUrl(
+      options,
+      'GOOGLE_CLOUD_BIGQUERY_ENDPOINT',
+      'https://www.googleapis.com');
+
   var config = {
-    baseUrl: 'https://www.googleapis.com/bigquery/v2',
+    baseUrl: baseInfo.apiEndpoint + basePath,
+    customEndpoint: baseInfo.customEndpoint,
```
stephenplusplus: What do you think about having `common.util.determineBaseUrl` called from the `common.Service` constructor? We could pass in config data to Service, like `config.defaultEndpoint` and `config.customEndpointEnvVar` (or something similar), and let the Service constructor handle the rest.

Does that seem possible, or am I overlooking any parts of the codebase where that would be a problem?

stephenplusplus:
Thank you very, very much for this!

packages/bigquery/src/index.js
rmanyari: In the case of services extending common.Service it should work fine. I can think of two edge cases though:

- We cannot use the same baseUrl for all the protoServices, for example bigtable has different endpoints for the actual service vs. the admin service. At the moment this is taken care of in the bigtable constructor.
- For backward compatibility reasons (not sure how important it's for you guys) I kept support for [SERVICE]_EMULATOR_HOST when it was already used. Bringing the baseUrl logic down to the services means that we'll have to support an array of environment variables, it's doable but feels like a hack rather than a clean solution

rmanyari:
I signed it!

packages/bigquery/src/index.js
stephenplusplus: Good catches.

> We cannot use the same baseUrl for all the protoServices, for example bigtable has different endpoints for the actual service vs. the admin service. At the moment this is taken care of in the bigtable constructor.

I'm pretty sure that we can leave out `config.protoServices[{{serviceName}}].baseUrl` if the value is the same as `config.baseUrl` (lines like this: https://github.com/rmanyari/google-cloud-node/blob/feature-custom-endpoint/packages/common-grpc/src/service.js). This still leaves room to override the baseUrl on a per-service config level.

> For backward compatibility reasons (not sure how important it's for you guys) I kept support for [SERVICE]_EMULATOR_HOST when it was already used. Bringing the baseUrl logic down to the services means that we'll have to support an array of environment variables, it's doable but feels like a hack rather than a clean solution

An array, defined in order of priority actually sounds like a nice feature, e.g. Bigtable would become:

```js
var config = {
  defaultEndpoint: 'bigtable.googleapis.com',
  endpointEnvVars: [
    'GOOGLE_CLOUD_BIGTABLE_ENDPOINT',
    'BIGTABLE_EMULATOR_HOST'
  ],
  protosDir: path.resolve(__dirname, '../protos'),
  protoServices: {
    Bigtable: {
      path: 'google/bigtable/v2/bigtable.proto',
      service: 'bigtable.v2'
    },
    BigtableTableAdmin: {
      baseUrl: adminBaseUrl,
      path: 'google/bigtable/admin/v2/bigtable_table_admin.proto',
      service: 'bigtable.admin.v2'
    },
    // ...
  },
  // ...
}


```

rmanyari: Actually, for the protoServices, what if we piggy back on the fact that we can tell if the endpoint is custom or not. If it's custom we override every baseUrl otherwise we leave them as they are. As for passing an array I'm fine with that

Slight tangent, googlebot is still complaining about the CLA, is there any way for you to check that  rodrigo@twosigma.com is part of the allowed email addresses to commit code from Two Sigma?

stephenplusplus: > If it's custom we override every baseUrl otherwise we leave them as they are.

I think I'm following, but to be sure:

- Is `config.protoService[service].baseUrl` specified?
  - Yes: Use it
  - No: Is there a custom endpoint?
    - Yes: Use it
    - No: Use `defaultEndpoint`

stephenplusplus:
googlebot is pretty feature-incomplete. @lukesneeringer can you make sure this is okay with the CLA? (re: (re: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2548#discussion_r134335958))

packages/bigquery/src/index.js
rmanyari: Yeah, that works

lukesneeringer:
Yea, I can check on googlebot while you two get tests working. :-)

lukesneeringer:
Update: @rmanyari, our system still thinks you have not signed the CLA, and in the interest of my not being swarmed by mobs of angry lawyers, this is a prerequisite.

The link is here: https://cla.developers.google.com/

If you are pretty sure you already signed it, the most likely problem is that the email address you used to do so does not match the email address on your commits (`git config user.email`).

If you need more help, let us know and I will try to get you in touch with someone who can help more.

rmanyari:
@lukesneeringer - I looked again at that link, it looks like I'm already covered by a corporate agreement from Two Sigma. I also just signed an individual one using my twosigma.com address. If it's not too much to ask, can you check again if I'm still not covered. Me too I have no interested whatsoever in long email threads with lawyers...

packages/bigquery/src/index.js
rmanyari: Actually, looking at the [current](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/bigtable/src/index.js#L318) implementation of bigtable, if there is a custom endpoint it overrides every baseUrl. We can bring this down to service.js like this:

```
  ...

  var baseInfo = util.determineBaseUrl(
    options,
    config.environmentVariables,
    config.defaultApiEndpoint,
    true
  );

  ...

  Object.keys(protoServices).forEach(function(name) {
    var protoConfig = protoServices[name];
    var service = self.loadProtoFile_(protoConfig, config);

    self.protos[name] = service;

    if (baseInfo.customEndpoint) {
      service.baseUrl = baseInfo.apiEndpoint;
    } else if (!protoConfig.baseUrl) {
      service.baseUrl = baseInfo.apiEndpoint;
    }
  });

...

```

Essentially overriding only if there is a custom endpoint or if for whatever reason, the protoService doesn't already have a baseUrl defined

stephenplusplus: Ah, nice catch.

> Essentially overriding only if there is a custom endpoint or if for whatever reason, the protoService doesn't already have a baseUrl defined

That sounds fine, let's go with that.

rmanyari: Perfect. I'll work this changes this morning.

packages/common-grpc/src/service.js
```diff
@@ -202,11 +221,14 @@ function GrpcService(config, options) {
 
   Object.keys(protoServices).forEach(function(name) {
     var protoConfig = protoServices[name];
-    var service = self.loadProtoFile_(protoConfig, config);
+    var service = GrpcService.prototype.loadProtoFile_
```
rmanyari: @stephenplusplus I had to make this change to make the datastore tests work. Everywhere else util.inherit is used to extend from Service except for datastore were modelo.inherits is used (needed because it does multiple inheritance). Modelo doesn't seem to like the prototype loaded by proxyquire and self.loadProtoFile_ is not available.

It suspect it has something to do with the fact that super_ refers to only the first inherited class, modelo has a workaround [this](https://github.com/kevinconway/Modelo.js). When reviewing the change set and running the integration tests, could you pay close attention to this specific change? All the unit tests pass, this was a tricky change that makes me a little nervous considering how js interprets `this` ...

packages/bigquery/test/index.js
```diff
@@ -143,14 +143,38 @@ describe('BigQuery', function() {
       assert(bq instanceof Service);
 
       var calledWith = bq.calledWith_[0];
-
       var baseUrl = 'https://www.googleapis.com/bigquery/v2';
-      assert.strictEqual(calledWith.baseUrl, baseUrl);
+      assert.strictEqual(bq.baseUrl, baseUrl);
```
stephenplusplus: This change means we are relying on behavior from `Service`, which we are trying to isolate from for the unit tests. Let's check:

- `calledWith.defaultApiEndpoint`
- `calledWith.environmentVariables`
- `calledWith.basePath`

to make sure we had a proper instantiation of Service.

packages/bigquery/test/index.js
```diff
@@ -143,14 +143,38 @@ describe('BigQuery', function() {
...
       assert.deepEqual(calledWith.scopes, [
         'https://www.googleapis.com/auth/bigquery'
       ]);
       assert.deepEqual(calledWith.packageJson, require('../package.json'));
     });
+
+    it('should not be considered custom endpoint if default', function() {
```
stephenplusplus: We don't need this test here, because it's not functionality performed by the code in `bigquery/src/index.js`.

packages/bigquery/test/index.js
```diff
@@ -143,14 +143,38 @@ describe('BigQuery', function() {
...
+      assert.strictEqual(bq.customEndpoint, false);
+    });
+
+    it('should use options.apiEndpoint if defined', function() {
```
stephenplusplus: We don't need this test here, because it's not functionality performed by the code in `bigquery/src/index.js`.

packages/bigquery/test/index.js
```diff
@@ -143,14 +143,38 @@ describe('BigQuery', function() {
...
+      var bq = new BigQuery({
+        projectId: 'project-id',
+        apiEndpoint: apiEndpoint
+      });
+
+      assert.strictEqual(bq.baseUrl, apiEndpoint + '/bigquery/v2');
+      assert.strictEqual(bq.customEndpoint, true);
+    });
+
+    it('should use GOOGLE_CLOUD_BIGQUERY_ENDPOINT if defined', function() {
```
stephenplusplus: We don't need this test here, because it's not functionality performed by the code in `bigquery/src/index.js`.

Adding the test earlier to make sure `calledWith.environmentVariables` is `['GOOGLE_CLOUD_BIGQUERY_ENDPOINT ']` will be all we need to know.

packages/bigtable/src/index.js
```diff
@@ -309,24 +317,19 @@ function Bigtable(options) {
...
+    environmentVariables: [
+      'GOOGLE_CLOUD_BIGTABLE_ENDPOINT',
+      'BIGTABLE_EMULATOR_HOST'
+    ],
+    defaultApiEndpoint: 'bigtable.googleapis.com',
     protosDir: path.resolve(__dirname, '../protos'),
     protoServices: {
       Bigtable: {
-        baseUrl: baseUrl,
+        baseUrl: defaultApiEndpoint,
```
stephenplusplus: Do we still need this? I thought `Service` would default this to the proper API endpoint automatically.

packages/bigtable/test/index.js
```diff
@@ -172,10 +172,19 @@ describe('Bigtable', function() {
       ]);
 
       assert.deepEqual(calledWith.packageJson, require('../package.json'));
+
+      Object.keys(calledWith.protoServices).forEach(function(serviceKey) {
```
stephenplusplus: We can skip this test here, because it's not code that was done in `bigtable/src/index.js`.

packages/bigtable/test/index.js
```diff
@@ -172,10 +172,19 @@ describe('Bigtable', function() {
...
+        var service = calledWith.protoServices[serviceKey];
+        var proto = bigtable.protos[serviceKey];
+
+        if (is.object(service)) {
+          assert.strictEqual(proto.baseUrl, service.baseUrl);
+        }
+      });
     });
 
     it('should work with the emulator', function() {
```
stephenplusplus: This test is no longer needed.

packages/bigtable/test/index.js
```diff
@@ -184,21 +193,45 @@ describe('Bigtable', function() {
...
-          assert.strictEqual(service.baseUrl, endpoint);
+        if (is.object(proto)) {
+          assert.strictEqual(proto.baseUrl, endpoint);
+        }
+      });
+
+      delete process.env.BIGTABLE_EMULATOR_HOST;
+    });
+
+    it('should use GOOGLE_CLOUD_BIGTABLE_ENDPOINT if defined', function() {
```
stephenplusplus: This test isn't needed. We only need to check that the correct environment variables were passed to the Service constructor.

packages/bigtable/test/index.js
```diff
@@ -184,21 +193,45 @@ describe('Bigtable', function() {
...
+        if (is.object(proto)) {
+          assert.strictEqual(proto.baseUrl, endpoint);
         }
       });
 
+      delete process.env.GOOGLE_CLOUD_BIGTABLE_ENDPOINT;
       delete process.env.BIGTABLE_EMULATOR_HOST;
     });
 
     it('should work with a custom apiEndpoint', function() {
```
stephenplusplus: This test is no longer needed.

packages/bigtable/test/index.js
```diff
@@ -207,11 +240,34 @@ describe('Bigtable', function() {
...
 
-        if (is.object(service)) {
-          assert.strictEqual(service.baseUrl, options.apiEndpoint);
+        if (is.object(proto)) {
+          assert.strictEqual(proto.baseUrl, endpoint);
+        }
+      });
+    });
+
+    it('should trim the protocol in custom endpoint', function() {
```
stephenplusplus: This test isn't needed.

packages/common/src/service.js
```diff
@@ -52,9 +62,24 @@ function Service(config, options) {
     email: options.email
   });
 
+  // Some services have not been moved to the
```
stephenplusplus: I think we should finish extending support for the new way before we leave ourselves in an incomplete state. Sorry that this didn't seem like a problem before, but since this concept has evolved to support new configurations, I don't think we should leave it incomplete. If you need a break or don't feel like completing it, no worries at all.

packages/common-grpc/src/service.js
stephenplusplus: Thanks for calling this out. It shouldn't be a big deal. `proxyquire` has its quirks, and this seems like a fair thing for it to get confused about.

packages/common-grpc/src/service.js
```diff
@@ -168,6 +176,17 @@ function GrpcService(config, options) {
     return global.GCLOUD_SANDBOX_ENV;
   }
 
+  if (config.defaultApiEndpoint) {
```
stephenplusplus: Ideally, this should all be done in `Service`, but let me know I'm overlooking something.

packages/common/src/util.js
```diff
@@ -788,3 +789,51 @@ function promisifyAll(Class, options) {
...
+ * @param {object=} options - Service configuration object.
+ * @param {array} environmentVariables - Array of accepted environment
+ *     variables through which a custom api endpoint can be specified.
+ * @param {string=} defaultApiEndpoint - Fallback api endpoint to use.
+ * @param {boolean} trimProtocol - If true, trim the leading trimProtocol.
+ *     Useful for grcp based service endpoints.
+ * @return {object=} base - Object containg the baseUrl and a boolean
+ *     property indicating whether the baseUrl is a customEndpoint or not.
+ * */
+function determineBaseUrl(options, environmentVariables,
```
stephenplusplus: I think this can be moved onto `Service.prototype.determineBaseUrl_`.

packages/common/src/service.js
rmanyari: I have no problem wrapping up the other services. I tend to dislike large PR so I don't want to put other maintainers in that position, however if you think it's necessary I can continue with the ones missing.

packages/common/src/util.js
rmanyari: What about services that rely on gapic? These don't extend from services

packages/common-grpc/src/service.js
rmanyari: Looking at it, I'd stop relying on config.{baseUrl, customEndpoint} in the GrpcService constructor and use this.{baseUrl, customEndpoint} since it's being populated by Service. I'll make those changes if you're ok with them

packages/bigquery/test/index.js
rmanyari: Fair enough, I'll make the changes. It should also make the final merge smaller which is a good thing IMO

packages/common-grpc/src/service.js
stephenplusplus: I think that should be fine, basically leaving as much work as possible to `Service`.

packages/common/src/service.js
stephenplusplus: I don't mind-- we've gotten pretty immune to massive PRs after being such a huge monorepo :)

packages/common/src/util.js
stephenplusplus: Gotcha, this is fine then :)

packages/common/src/service.js
```diff
@@ -52,9 +57,20 @@ function Service(config, options) {
     email: options.email
   });
 
+  if (!config.baseUrl) {
```
rmanyari: @stephenplusplus - This is temporary until I wrap up with the other services. Without this the integration tests break. Sorry it's taking this long...

rmanyari:
I wrapped up the remaining services and grpcServices. What's left are the services that use gapic:

- dlp
- language (partially done)
- logging-*
- monitoring
- spanner (partially done)
- speech
- video-intelligence
- vision

To properly support custom endpoint in those services, gax-nodejs needs to have the notion of insecureCrendentials, the same way it's done in [commonGrpc](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/common-grpc/src/service.js#L174). At the moment it forces the calls to go do the auth [roundtrip](https://github.com/googleapis/gax-nodejs/blob/e5059c7eb652cbff29cbdc12556946a32a5b7e1a/lib/grpc.js#L81) which is inconsistent with how we're doing things everywhere else (why would you want Google auth when going to a custom endpoint?!). I can open a PR in gax-nodejs to get that fixed and then wrap this up. If we really want this done I can override _getCredentials but that's just looking for trouble.

Thoughts?

stephenplusplus:
Thanks for all of your help on this. The issue with gax is actually resolved. It originally popped up in [this PR](https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2195/commits/f77643c4945506d2299fd9c43ab563791bb9edb4#r111204782) where I am converting Datastore to gax.

Basically, we need to provide `options.sslCreds` (which is honored here: https://github.com/googleapis/gax-nodejs/blob/e5059c7eb652cbff29cbdc12556946a32a5b7e1a/lib/grpc.js#L83):

```js
optionsForGax.sslCreds = grpc.credentials.createInsecure();
```

rmanyari:
Fantastic, I'll go ahead and make the changes to the remaining services.

packages/speech/src/index.js
```diff
@@ -55,7 +69,7 @@ function speechV1(options) {
 
   // Create the speech client with the provided options.
   var client = gapic.v1(options).speechClient(options);
-  Object.assign(client.constructor.prototype, helpers());
```
rmanyari: `client.constructor.prototype` was adding the helper functions to the base Object prototype, i.e. every object created after this constructor is called will get the helper functions (odd). Since we return an instance of the client anyway I decided to assign the helper functions to the instance instead of the prototype. It didn't seem to have broken anything, but a second pair of eyes would be really appreciated

rmanyari:
Sorry this took forever... Now it has a bunch of conflicts with upstream, lmk if it looks fine and I'll try to fix the conflicts.

Also, for some reason I'm no longer able to build the docs.

```
HEAD is now at 1aa782a... speech @ 0.9.4 tagged.
/home/rodrigo/dev/google-cloud-node/scripts/docs/builder.js:316
    throw e;
    ^

Error: Unable to generate docs for file: speech/src/index.js. Reason: ENOENT: no such file or directory, open 'packages/speech/src/helpers
```

Not sure why it's looking at commits hashes, can someone explain what's going on here? packages/speech/src/helpers exists locally but it might (?) not be looking for that

packages/vision/src/index.js
```diff
@@ -73,7 +73,7 @@ function visionV1(opts) {
   // Create the image annotator client with the provided options.
   var client = gapic.v1(opts).imageAnnotatorClient(opts);
   if (is.undefined(client.annotateImage)) {
-    Object.assign(client.constructor.prototype, helpers('v1'));
```
rmanyari: Same prototype prob here. In this case it was actually breaking the grpc validation. Not sure how this used to work

packages/speech/src/index.js
stephenplusplus: Wow, that was indeed an oversight. Extending `client` is cool 👍

packages/dlp/src/index.js
```diff
@@ -21,10 +21,12 @@
 
 'use strict';
 
+var common = require('@google-cloud/common');
 var extend = require('extend');
 var gapic = {
   v2beta1: require('./v2beta1')
 };
+var grpc = require('grpc');
```
stephenplusplus: We just need to add this to our dlp's dependencies.

packages/common/src/util.js
```diff
@@ -837,3 +833,42 @@ function determineBaseUrl(options, environmentVariables,
...
+ * @param {object=} options - Service configuration object.
+ * @param {array} environmentVariables - Array of accepted environment
+ *     variables through which a custom api endpoint can be specified.
+ * @param {string=} defaultApiEndpoint - Fallback api endpoint to use.
+ * @param {number} defaultPort - Fallback port to use.
+ * @param {object=} credentials - An instance of grpc.ServerCredentials
+ *     these credentials will be used if the api endpoint is considered
+ *     to be a custom endpoint
+ * */
+function resolveGapicOptions(options, environmentVariables,
```
stephenplusplus: Would you switch this method to accept an object instead of multiple arguments?

packages/dlp/src/index.js
```diff
@@ -37,13 +39,25 @@ const VERSION = require('../package.json').version;
...
+ * otherwise the actual API endpoint will be used.
+ *
  * @param {object=} options - [Configuration object](#/docs).
  * @param {number=} options.port - The port on which to connect to
  *     the remote host.
  * @param {string=} options.servicePath - The domain name of the
  *     API remote host.
  */
 function dlpV2beta1(options) {
+  options = common.util.resolveGapicOptions(
```
stephenplusplus: Editing these files will be a bit of a problem, as they are generated. @landrito -- is it possible to have the generator start spitting out these new lines? This PR adds support for custom endpoints, so a user can talk to an emulator or other arbitrary endpoint.

stephenplusplus:
Thanks again for all of the work on this. It's getting a bit hairy when we're mixing in with files that are generated (speech, for example). I don't maintain the generator, so I've asked @landrito to take a look where those edits have been made. If the generator can take it on, we also wouldn't need the new tests-- it would handle that as well.

packages/dlp/src/index.js
landrito: Hm. I'd rather not pull in @google-cloud/common as a dependency to the generated clients. @lukesneeringer this might be where we push for merging @google-cloud/common and google-gax into a single package like python did.

packages/common/src/util.js
rmanyari: Although generally I'm against functions with many arguments (this one has 5, it's borderline too many) in this case I kind of like the fact that it's consistent with determineBaseUrl in the sense that it's very obvious from the arguments that these two functions are somehow related.

When I read index.js in a "common" service and a "gapic" based service, the custom endpoint bits look similar and it's easy to follow what's going on. If we put everything under an object, not only it's less explicit but I'll have to unmarshall the object anyway to pass the args down to determineUrl

Other than that if you feel like 5 args is too many or maybe have another reason that I'm missing (new lint rules for example?) I can make the changes

rmanyari:
Hey guys, any updates in this? Is there anything I can do to help pushing this forward?

packages/dlp/src/index.js
stephenplusplus: Ping!

rmanyari:
Any luck this gets allocated some time? It would be nice to know if there's still a slim chance that this makes it to master or if I should just forget about it. Thanks!

lukesneeringer:
@stephenplusplus What's the situation on this item?

stephenplusplus:
I'm going to close the PR and let #1630 track the request for the feature. Sorry that we got caught up in the repo split. I thought we could get these changes in, but other priorities got in the way, which greatly complicated this. Thank you for all of your work and patience.

Issue #2302, opened 2017-05-11
error-reporting: Errors not sent when providing an API key only


DominicKramer:
- I ran `export NODE_ENV=production`
- I installed express and cloud-errors.
- I run the following code:

I could not see errors in the Error Reporting UI for this project.
I could successfully send manually errors to this project with this API key (by calling the API from cURL).

```
var express = require('express');
var app = express();
var errorHandler = require('@google/cloud-errors')({
  projectId: "<my project>",
  key: "<my api key>",
  serviceContext: {
      service: 'steren-test-node',
      version: 'test1'
  }
});

app.get(
  '/errorRoute',
  function ( req, res, next ) {
    // You can push in errors manually
    res.send("Error");
    res.end();
    next(new Error("Got traffic on the errorRoute"));
  }
);

app.get(
  '/anotherRoute',
  function ( req, res, next ) {
    // It'll even log potentially unexpected errors
    JSON.parse("{\"malformedJson\": true");
  }
)

// Just use the express plugin
app.use(errorHandler.express);

app.listen(
  3000
  , function ( ) {
    console.log('Server has been started on port 3000');
  }
);
```

This issue has been migrated from the cloud-errors-nodejs issue [#10](https://github.com/GoogleCloudPlatform/cloud-errors-nodejs/issues/10).

ivanseidel:
Same problem here. This is really bad, because it seems it's not even attempting to report errors (watched requests from Charles, and none was to `https://clouderrorreporting.clients6.google.com/v1beta1/projects/<myproj>/services?key=...`.

It seems that it's validating OAuth before sending, but not being able to "check" the token, because there isn't none.

stephenplusplus:
Sorry, handing it over to @ofrobots who oversees the logging-* and error-reporting.

kjin:
@ivanseidel, and @steren who filed the original bug

This is fixed in error-reporting v0.2.2, please re-open this bug if you have any issues with it!

PR #2554, opened 2017-08-22
error-reporting: Skip authentication for error-reporting when API key is provided

kjin:
Addresses #2302

When `key` is provided, the error reporting service instance doesn't go through the OAuth2 process. This means that if a provided API key is invalid, errors will never be sent (there is no fallback); however, an error will be printed in this case.

### DIFF
--- a/packages/error-reporting/README.md
+++ b/packages/error-reporting/README.md
@@ -162,9 +162,9 @@ var errors = require('@google-cloud/error-reporting')({
 
 If a key is provided, the module will not attempt to authenticate using the methods associated with locally-stored credentials as mentioned in the previous section.
 
-We recommend storing the API key in a file rather than hard-coding it into your application's source code.
+We recommend using a file, environment variable, or another mechanism to store the API key rather than hard-coding it into your application's source.
 
-**Note:** The Error Reporting instance will check if the provided API key is invalid shortly after it is instantiated. If the key is invalid, an error-level message will be logged.
+**Note:** The Error Reporting instance will check if the provided API key is invalid shortly after it is instantiated. If the key is invalid, an error-level message will be logged to stdout.
 
 ## Configuration
 


packages/error-reporting/src/google-apis/auth-client.js
```diff
@@ -88,26 +88,33 @@ class RequestHandler extends common.Service {
...
+        if (err) {
+          that._logger.error([
+            'Unable to find credential information on instance. This library',
+            'will be unable to communicate with the Stackdriver API to save',
+            'errors.  Message: ' + err.message
+          ].join(' '));
+        }
+      });
+    } else {
+      that._logger.info('API key provided; skipping OAuth2 token generation.');
```
DominicKramer: Can you add a test to verify that OAuth2 token generation is not done and verify that the user is issued a message when an API key is provided?

packages/error-reporting/src/google-apis/auth-client.js
```diff
@@ -88,26 +88,33 @@ class RequestHandler extends common.Service {
...
-    this.authClient.getToken(function(err, token) {
-      if (err) {
-        that._logger.error([
-          'Unable to find credential information on instance. This library',
-          'will be unable to communicate with the Stackdriver API to save',
-          'errors.  Message: ' + err.message
-        ].join(' '));
-      }
-    });
+    if (tryAuthenticate) {
```
DominicKramer: It might be possible to access the [list](https://cloud.google.com/error-reporting/reference/) of reported errors as a way to notify the user if the key is invalid as soon as possible (as apposed to having the user not know the API key is invalid until an error is actually reported).

packages/error-reporting/src/google-apis/auth-client.js
DominicKramer: Can you add a test that verifies that the OAuth2 token generation is skipped and the user is given an info message if an API key is provided?

packages/error-reporting/src/google-apis/auth-client.js
kjin: Inferring from the docs [here](https://cloud.google.com/error-reporting/docs/iam) I believe only reporting errors can be done without OAuth2. I'll look for a more general way to verify an API key but I don't think it's certain that such a method exists.

packages/error-reporting/src/google-apis/auth-client.js
kjin: Done

packages/error-reporting/src/google-apis/auth-client.js
```diff
@@ -127,7 +119,9 @@ class RequestHandler extends common.Service {
     if (this._config.getShouldReportErrorsToAPI()) {
       this.request({
         uri: 'events:report',
-        qs: RequestHandler.manufactureQueryString(this._config.getKey()),
+        qs: {
```
DominicKramer: This looks like it will use `{ key: null }` as `qs` if `this._config.getKey()` is `null` instead of using `null` for `qs` as was done previously.  I'm concerned that this could be introducing a bug.

packages/error-reporting/src/google-apis/auth-client.js
kjin: If we report with an empty request payload, we get a 400 error with a different error message than reporting with an incorrect API key (which also gives a 400 error). From what I can tell this is our only option, though it's somewhat fragile. What do you think of this?

packages/error-reporting/test/unit/google-apis/auth-client.js
```diff
@@ -41,24 +41,35 @@ function verifyReportedMessage(errToReturn, expectedMessage) {
...
 }
-
 describe('RequestHandler', function() {
+  it('should not request OAuth2 token if key is provided', function() {
+    var config = {
+      ignoreEnvironmentCheck: true,
+      key: 'key'
+    };
+    var message = 'Made OAuth2 Token Request';
+    verifyReportedMessage(config, new Error(message), '');
```
DominicKramer: I think you would also want to verify that a call to `info` was done that had the text `"API key provided; skipping OAuth2 token request."` provided to verify that the user is provided with the correct message.

packages/error-reporting/src/google-apis/auth-client.js
DominicKramer: It would be nice to tell the user that the key is invalid as soon as possible, but if it is not something that is possible, we should at least update the documentation to let the users know it is a known limitation.

packages/error-reporting/src/google-apis/auth-client.js
kjin: Done, nice catch

packages/error-reporting/src/google-apis/auth-client.js
kjin: Made a note in the docs.

kjin:
@ofrobots PTAL. The most recent commit warns if the API key is invalid ahead of time, but depends on the error message in the response body. This works, but if you think it's too fragile I'll revert to the previous commit, which adds in the README a message informing users that the we won't be able to detect invalid API keys ahead of time.

packages/error-reporting/README.md
```diff
@@ -145,6 +147,25 @@ If your application is running outside of Google Cloud Platform, such as locally
...
+  projectId: '{your project ID}',
+  key: '{your api key}'
+});
+```
+
+If a key is provided, the module will not attempt to authenticate using the methods associated with locally-stored credentials as mentioned in the previous section.
+
+We recommend storing the API key in a file rather than hard-coding it into your application's source code.
+
+**Note:** The Error Reporting instance will check if the provided API key is invalid shortly after it is instantiated. If the key is invalid, an error-level message will be logged.
```
ofrobots: nit: logged where?

packages/error-reporting/src/google-apis/auth-client.js
```diff
@@ -88,26 +89,46 @@ class RequestHandler extends common.Service {
...
+          ].join(' '));
+        }
+      });
+    } else {
+      this.request({
+        uri: 'events:report',
+        qs: RequestHandler.manufactureQueryString(this._config.getKey()),
+        method: 'POST',
+        json: {}
+      }, (err, body, response) => {
```
ofrobots: @kjin Should we also check `response.statusCode` to be 400?

@0x4b Does this seem like an appropriate way to check if the API key is valid? Is there a test on the service side that ensures that 400 + 'Message cannot be empty.' is not returned when the key is invalid?

0x4b: There isn't such a test, largely because some auth related error cases are handled by a different part of the stack outside the service's control.

lukesneeringer:
@ofrobots Another question that we should ask is: Is this something we want to support? I was of the understanding that not supporting API keys in the client libraries was an intentional decision. I am definitely hesitant to add it as a one-off in _a single_ library.

ofrobots:
@lukesneeringer The API is not being *added*, but rather fixed in this PR. For whatever historical reasons, error-reporting started with API key support and removing it at this point is going to break existing users. I'm :-1: on that.

kjin:
@ofrobots @DominicKramer I've added a system test to check that the messages supplied from the Stackdriver Errors API are in line with what we expect. I think this is a reasonable way to act upon any future changes in the API - let me know your thoughts.

packages/error-reporting/README.md
```diff
@@ -145,6 +147,25 @@ If your application is running outside of Google Cloud Platform, such as locally
...
+```js
+var errors = require('@google-cloud/error-reporting')({
+  projectId: '{your project ID}',
+  key: '{your api key}'
+});
+```
+
+If a key is provided, the module will not attempt to authenticate using the methods associated with locally-stored credentials as mentioned in the previous section.
+
+We recommend storing the API key in a file rather than hard-coding it into your application's source code.
```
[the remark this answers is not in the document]
kjin: Done

kjin:
@stephenplusplus Could you do a release of the `error-reporting` library? I think this is a semver-patch change. Thanks!

stephenplusplus:
Published `0.2.2`!

Issue #2553, opened 2017-08-22
Vision: imageContext isn't documented


jmdobry:
Customer wondered how to pass a crop hint option to `Vision#cropHints()`. Turns out, you can't. The helper methods only accept an `Image` and a grpc options object, and do not accept `imageContext`.

In order to pass say, the `aspectRatios` option as part of the `cropHints` request, you have to use the `annotateImage` method directly, like so:

```js
const Vision = require('@google-cloud/vision');
const vision = Vision();
vision.annotateImage({
  image: {
    source: {
      filename: '/path/to/image.png'
    }
  },
  features: [{ type: Vision.types.Feature.Type.CROP_HINTS }],
  imageContext: {
    cropHintsParams: {
      aspectRatios: [0.25]
    }
  }
}).then(...);
```

I'm not sure how folks are supposed to know this.

`imageContext` itself isn't even documented on the [annotateImage method](https://googlecloudplatform.github.io/google-cloud-node/#/docs/vision/0.12.0/vision?method=annotateImage)

Also, accessing the enum (e.g. `Vision.types.Feature.Type.CROP_HINTS`) is only documented in the client library's README, and even then, only for FACE_DETECTION.

arpwal:
You might want to update these docs also when you're at it -- 
https://cloud.google.com/vision/docs/detecting-crop-hints#vision-crop-hints-detection-nodejs

Thanks for following up jmdobry@

stephenplusplus:
Thanks for reporting and sorry for the trouble @arpwal.

I believe we have two action items here:

1) Add a link to [AnnotateImageRequest](https://googlecloudplatform.github.io/google-cloud-node/#/docs/vision/0.12.0/vision/v1/data_types?method=AnnotateImageRequest) from `annotateImage`'s JSDocs: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/vision-0.12.0/packages/vision/src/helpers.js#L101
2) Allow a raw AnnotateImageRequest to be sent to the shorthand methods, like `cropHints`:
    ```js
    vision.cropHints({
      image: {...},
      imageContext: {...}
    }, gaxOptions, callback)
    ```
    The internal method would extend the user's input with `features: [features.CROP_HINTS]`. This would be a breaking change. Do we want this behavior, or should we refer users to the `annotateImage` method instead?

jmdobry:
Why does 2) have to be a breaking change?

```js
[...]

PR #2555, opened 2017-08-23
Add arbitrary request options to single-feature helpers.

lukesneeringer:
## :warning: Includes Breaking Change

Requests to all Vision methods must be in the full format of an [`AnnotateImageRequest`](https://googlecloudplatform.github.io/google-cloud-node/#/docs/vision/0.12.0/vision/v1/data_types?method=AnnotateImageRequest).

---

@jmdobry @stephenplusplus This is my pass at addressing #2553. It is definitely intended to be up for discussion.

### DIFF
--- a/packages/vision/src/helpers.js
+++ b/packages/vision/src/helpers.js
@@ -73,27 +73,25 @@ var coerceImage = (image, callback) => {
  * @return {Function} - The function that, when called, will call annotateImage
  *   asking for the single feature annotation.
  */
- var _createSingleFeatureMethod = featureValue => {
-   return function(annotateImageRequest, callOptions) {
-     annotateImageRequest.features = annotateImageRequest.features || [{
-       type: featureValue,
-     }];
-
-     // If the user submitted explicit features that do not line up with
-     // the precise method called, throw an exception.
-     for (let feature of annotateImageRequest.features) {
-       if (feature.type !== featureValue) {
-         throw new Error(
-           'Setting explicit features is not supported on this method. ' +
-           'Use the #annotateImage method instead.'
-         );
-       }
-     }
-
-     // Call the underlying #annotateImage method.
-     return this.annotateImage(annotateImageRequest, callOptions);
-   };
- };
+var _createSingleFeatureMethod = featureValue => {
+  return function(annotateImageRequest, callOptions) {
+    annotateImageRequest.features = annotateImageRequest.features || [{
+      type: featureValue,
+    }];
+    // If the user submitted explicit features that do not line up with
+    // the precise method called, throw an exception.
+    for (let feature of annotateImageRequest.features) {
+      if (feature.type !== featureValue) {
+        throw new Error(
+          'Setting explicit features is not supported on this method. ' +
+          'Use the #annotateImage method instead.'
+        );
+      }
+    }
+    // Call the underlying #annotateImage method.
+    return this.annotateImage(annotateImageRequest, callOptions);
+  };
+};
 
 
 /*!


packages/vision/src/helpers.js
```diff
@@ -74,11 +74,12 @@ var coerceImage = (image, callback) => {
  *   asking for the single feature annotation.
  */
 var _createSingleFeatureMethod = featureValue => {
-  return function(image, options) {
-    return this.annotateImage({
+  return function(image, options = {}, callOptions = undefined) {
```
stephenplusplus: `callOptions` seem to have no effect on the method. I believe this runs into the issue I mentioned here: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2553#issuecomment-324385621 about consecutive, optional arguments. For example, if I call `cropHints({image}, {callOptions})`, my `callOptions` are going to be misunderstood for `options`, and end up extending the raw API request.

Side note, I don't think Node v4 supports default parameters.

lukesneeringer: The former is the way Node works, and there is nothing I can do about that.

I will let CI figure out the default parameters part. If it does not then I will change to the old `options = options || {}` idiom.

stephenplusplus: The way the code is in this PR, the API isn't working as the docs advertise. What we can do is find a different way to handle these concepts. JavaScript idioms have always preferred a configuration object over multiple arguments. We can use this to label the parts of the users intent:

```js
method(shortHandArgument, [options])
```

In this case:

```js
cropHints({ image }, {
  request: { imageContext: {} },
  callOptions: {...}
})
```

Or we can combine the `request` with the `callOptions` in a slightly different way:
```js
cropHints({ image }, {
  imageContext: {},

  // this is a whitelist property that is plucked from an options object and treated as `callOptions`
  gaxOptions: {}
})
```

If you want to keep the concept the way it is in this PR, you would have to remove the `optional` part of the arguments, and require users pass in undefined/null:

```js
cropHints({image}, null, {callOptions})
```

lukesneeringer: > JavaScript idioms have always preferred a configuration object over multiple arguments.

Sure. What is not clear in this case is how to set up that object.

My problem with trying to combine the request object with what I renamed `callOptions` is that this would be a single one-off against the entire rest of the library. A user could reasonably ask, "Why is _this_ one different?"

Honestly, maybe the right alternative is to just drop `callOptions` _completely_ for these methods. It is rarely used, and it would be reasonable to say that if you need it, you have to call `annotateImage`.

stephenplusplus: See https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2553#issuecomment-324385621 where I also said that we should not be treating this as a one-off, and coming up with the right solution that works API-wide. I explained in that post other places where `gaxOptions` is used (and how). Maybe we should move the talk back over there?

packages/vision/src/helpers.js
```diff
@@ -73,14 +73,27 @@ var coerceImage = (image, callback) => {
...
-    }, options);
-  };
-};
+ var _createSingleFeatureMethod = featureValue => {
+   return function(annotateImageRequest, callOptions) {
+     annotateImageRequest.features = annotateImageRequest.features || [];
+
+     // Ensure the feature value indicated by the user's method choice exists on the
+     // features array; if it does not, add it.
+     var found = false;
```
stephenplusplus: I think we should only allow the one `featureValue` to be the feature on the request. So basically, that would mean overriding the user's input, because I think it would be a misuse of our API if they did something like this:

```js
cropHints({
  image: {...},
  features: ['labeldetection', 'crophints']
})
```

packages/vision/src/helpers.js
```diff
@@ -196,6 +210,9 @@ module.exports = apiVersion => {
    *
    *   If the key is `content`, the value should be a Buffer.
    * @param {Object=} options
+   *   Any additional options which should be mapped onto the
```
stephenplusplus: I think we need to remove the `image` documentation above, and perhaps rename `options` to `request`.

Then, the docs for `request` can be:

```js
* @param {Object} request
*   A representation of the request being sent to the Vision API.
*   This is an [AnnotateImageRequest]{@link AnnotateImageRequest}.
* @param {Object} request.image
*   A dictionary-like object representing the image. This should have a
*   single key (`source`, `content`).
*
*   If the key is `source`, the value should be another object containing
*   `imageUri` or `filename` as a key and a string as a value.
*
*   If the key is `content`, the value should be a Buffer.
```

I believe the docs are incorrect in places where `request` is marked as optional. `Object=` notation means optional, and should just be `Object` if it is required.

lukesneeringer: Concur on fixing the documentation. I will do that after we resolve your other comment.

packages/vision/src/helpers.js
lukesneeringer: Okay, so let's talk through the situation where someone calls:

```es6
vision.cropHints({
  image: {...},
  features: [{type: features.FACE_DETECTION}],
});
```

There are four options:

  1. Take the union of the feature designated by the method and the features designated in the `features` array. (This is what I implemented.)
  2. Take the feature designated by the method only, throw away the specified `features` array. (This is what you suggest.)
  3. Take the feature designated by the `features` array, throw away the method designation. (This is what @jmdobry's sample code in #2553 did.)
  4. Throw an exception.

I believe that choices (2) and (3) both violate the principle of least surprise: as a user, I wonder why I did not get something that I asked for. One might argue that nobody would ever do this intentionally, but I believe there might be a case where extant code calling `cropHints` exists, and someone decides they _also_ want faces, and in looking up how to do it in docs, stumbles on the `features` array. It is not hard to imagine this scenario.

Therefore, I assert that the only acceptable options are (1) and (4). If we do (4), basically we throw an exception and direct users to `annotateImage`. The advantage to this is that it is explicit: users will not ever be in a situation where they are charged for N features when they only wanted N - 1. On the other hand, we create more cognitive work for the developer.

(1) seems like the best option to me. It passes principle of least surprise as well as holding to Postel's Law.

stephenplusplus: I think this is solved with documentation, and I still see your example as a misuse of the method. The method should be documented as "the way to get crop hints" and not the way to get arbitrary features, because we already have another method for exactly that. `cropHints()` would not document the features property, because that's the part this method controls. If we have such a loose definition of what this method is for, I would lean back towards scrapping the shorthand methods. "I called cropHints() and I got labels! You had one job!"

> as a user, I wonder why I did not get something that I asked for.

By not documenting the `features` property, the user would be misusing the method, which would address why they're not getting something *they think* they're asking for. (What they're actually asking for is "cropHints()" and they will get that)

> someone decides they also want faces, and in looking up how to do it in docs, stumbles on the features array.

But they would have stumbled upon it on another method's documentation.

lukesneeringer: I already accept your premise that a user that does this is doing something unusual. The question is how to handle it. :-) Do we make a guess at which part of the input to honor, do we honor the entire input, or do we throw up our hands and we "we do not know"?

In this case, I am okay with honoring the whole input (which we can do), and okay with failing. I am not okay with guessing which half of the input to honor.

stephenplusplus: I differentiate between "unusual" and "incorrect". I think the way to handle it is do what we say we're going to do, and return crop hints from cropHints() and nothing else. I don't think any of our API methods should have a quirk where they start working as "extra-powered" if the user accidentally unlocks it. We shouldn't enable that with our time of development, maintenance, or support for it. The code should just do what our docs say it does.

jmdobry: Agreed. Just overwrite the `features` property with crop hints. If the user expects anything other than crop hints, they should call something other than `cropHints`.

lukesneeringer: If it is incorrect, then we should throw an exception. We should not just plow over half the input.

stephenplusplus: I would have preferred @jmdobry's solution initially, but typed and backspaced it because it didn't seem worth the challenge of bringing you down from your position. I wanted to make sure all of the options were presented, so that your perspective would be a contribution to the best decision. I was only strongly opposed to having consecutive, optional arguments that would not work as advertised.

You are the charging force that moves us forward, and as this is a mostly-generated API that you developed, I respect your decision. Let's roll with the throwing thing 👍

lukesneeringer: It is _always_ fine to challenge me. :-)

I think we might be at an impasse though. Let's make sure @jmdobry does not strongly disagree before making a final call.

packages/vision/src/helpers.js
```diff
@@ -73,14 +73,27 @@ var coerceImage = (image, callback) => {
...
+   return function(annotateImageRequest, callOptions) {
+     annotateImageRequest.features = annotateImageRequest.features || [{
+       type: featureValue,
+     }];
+
+     // If the user submitted explicit features that do not line up with
+     // the precise method called, throw an exception.
+     for (let feature of annotateImageRequest.features) {
+       if (feature.type !== featureValue) {
+         throw new Error(
```
jmdobry: I do not believe Node v4 supports `throw` closing the iterator in a for..of loop. Check [node.green](http://node.green/). May just want to use a standard for loop, or `annotateImageRequest.features.forEach`.

lukesneeringer: My original reaction was, "But the unit tests passed," but it occurs to me that `assert.throws` is pretty broad. Happy to change this. :-)

lukesneeringer: It does work in Node 4:

```
lukesneeringer@lukesneeringer0:~/Code/Google/google-cloud-node:issue-2553$ nvm use 4
Now using node v4.8.4 (npm v2.15.11)
lukesneeringer@lukesneeringer0:~/Code/Google/google-cloud-node:issue-2553$ node
> for (let i of [1, 2, 3, 4, 5]) {
... console.log(i)
... if (i == 3) {
..... throw new Error('foo')
..... }
... }
1
2
3
Error: foo
    at repl:5:7
    at REPLServer.defaultEval (repl.js:272:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:441:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:203:10)
    at REPLServer.Interface._line (readline.js:542:8)
    at REPLServer.Interface._ttyWrite (readline.js:819:14)
```

stephenplusplus:
This is stuck on linting errors:

```
> google-cloud-node@ lint C:\projects\google-cloud-node
Expected indentation of 0 characters at packages/vision/src/helpers.js :
    74 | *   asking for the single feature annotation.
    75 | */
    76 | var _createSingleFeatureMethod = featureValue => {
--------^
    77 |   return function(annotateImageRequest, callOptions) {
    78 |     annotateImageRequest.features = annotateImageRequest.features || [{
Expected indentation of 2 characters at packages/vision/src/helpers.js :
    75 | */
    76 | var _createSingleFeatureMethod = featureValue => {
    77 |   return function(annotateImageRequest, callOptions) {
----------^
    78 |     annotateImageRequest.features = annotateImageRequest.features || [{
    79 |       type: featureValue,
Expected indentation of 0 characters at packages/vision/src/helpers.js :
    94 |     return this.annotateImage(annotateImageRequest, callOptions);
    95 |   };
    96 | };
--------^
    97 |
    98 |
3 code style er> google-cloud-node@ lint C:\projects\google-cloud-node
Expected indentation of 0 characters at packages/vision/src/helpers.js :
    74 | *   asking for the single feature annotation.
    75 | */
    76 | var _createSingleFeatureMethod = featureValue => {
--------^
    77 |   return function(annotateImageRequest, callOptions) {
    78 |     annotateImageRequest.features = annotateImageRequest.features || [{
Expected indentation of 2 characters at packages/vision/src/helpers.js :
    75 | */
    76 | var _createSingleFeatureMethod = featureValue => {
    77 |   return function(annotateImageRequest, callOptions) {
----------^
    78 |     annotateImageRequest.features = annotateImageRequest.features || [{
    79 |       type: featureValue,
Expected indentation of 0 characters at packages/vision/src/helpers.js :
    94 |     return this.annotateImage(annotateImageRequest, callOptions);
    95 |   };
    96 | };
--------^
    97 |
    98 |
3 code style er=> google-cloud-node@ lint C:\projects\google-cloud-node
Expected indentation of 0 characters at packages/vision/src/helpers.js :
    74 | *   asking for the single feature annotation.
    75 | */
    76 | var _createSingleFeatureMethod = featureValue => {
--------^
    77 |   return function(annotateImageRequest, callOptions) {
    78 |     annotateImageRequest.features = annotateImageRequest.features || [{
Expected indentation of 2 characters at packages/vision/src/helpers.js :
    75 | */
    76 | var _createSingleFeatureMethod = featureValue => {
    77 |   return function(annotateImageRequest, callOptions) {
----------^
    78 |     annotateImageRequest.features = annotateImageRequest.features || [{
    79 |       type: featureValue,
Expected indentation of 0 characters at packages/vision/src/helpers.js :
    94 |     return this.annotateImage(annotateImageRequest, callOptions);
    95 |   };
    96 | };
--------^
    97 |
    98 |
```

PR #2561, opened 2017-08-25
spanner: properly capture readOnly option

### LINKED ISSUE #2542: [Spanner] The default value (strong: true) hasn't been set when create a read-only transaction which cause slow data update
Hi,

If you follow the steps I listed below, I think you will get a very slow data update operation which will consume larger than 10 seconds(```table().update()```).
I found that if I add the option ```{strong: true}``` to call ```runTransaction```, the update function will run at a reasonable speed.

According to the [docs](https://cloud.google.com/spanner/docs/timestamp-bounds), the default timestamp bound types of the read-only transaction is ```Strong```. 
But I think the Spanner SDK ignore the default option in [database.runTransaction](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/spanner/src/database.js#L872-L884).

#### Environment details

  - OS: macOS 10.11.6
  - Node.js version: 8.1.3
  - npm version: 5.0.3
  - google-cloud-node version: 0.7.1

#### Steps to reproduce
1. require ```require('@google-cloud/spanner')```
2. Connect to Spanner
3. Select an instance and a database without any workload
4. create a table with this schema:
```
`CREATE TABLE bar (
	user_id STRING(64) NOT NULL,
	role STRING(64),
	level INT64
) PRIMARY KEY (user_id)`
```
5. insert sample data:
```
let sampleData = [
	{'user_id': '1', 'role': 'coder','level': 1},
	{'user_id': '2', 'role': 'coder', 'level': 1}
];
```
6. create a read-only transaction
7. use this transaction to run the following sql
8. and then update the record
```
let options = {
	readOnly: true
};
console.time('Step 1 read-only-transaction consume:');
database.runTransaction(options, function (err, transaction) {
	if (err) {
		throw (err);
	}

	let query1 = {
		sql: 'SELECT * FROM bar as t WHERE t.role = @role',
		params: {
			role: 'coder'
		}
	};
	transaction.run(query1, function (err1, dataset) {
		if (err1) {
			throw err1;
		}
		console.timeEnd('Step 1 read-only-transaction consume:');
		transaction.end();

		console.time('Step 2 DB update consume:');
		let dbOperation = [
			{user_id: '1', level: 2},
			{user_id: '2', level: 2}
		];

		database.table('bar').update(dbOperation, function (err2, apiResponse) {
			if (err2) {
				throw err2;
			}
			console.timeEnd('Step 2 DB update consume:');
		});
	});
});
```
9. Get the following result(the step 2 consume large than 10 seconds)
```
Step 1 read-only-transaction consume:: 1562.155ms
Step 2 DB update consume:: 15148.733ms
```

Thanks!

callmehiphop:
Closes #2542

### DIFF
--- a/packages/spanner/src/database.js
+++ b/packages/spanner/src/database.js
@@ -460,9 +460,6 @@ Database.prototype.getTransaction = function(options, callback) {
       return;
     }
 
-    options = extend({}, options);
-    delete options.readOnly;
-
     var transaction = self.pool_.createTransaction_(session, options);
 
     transaction.begin(function(err) {
--- a/packages/spanner/src/transaction-request.js
+++ b/packages/spanner/src/transaction-request.js
@@ -51,7 +51,14 @@ var PartialResultStream = require('./partial-result-stream.js');
  * @param {object=} options - Timestamp options.
  */
 function TransactionRequest(options) {
+  this.readOnly = false;
+
   if (options && !is.empty(options)) {
+    options = extend({}, options);
+
+    this.readOnly = !!options.readOnly;
+    delete options.readOnly;
+
     this.options = TransactionRequest.formatTimestampOptions_(options);
   }
 }
--- a/packages/spanner/test/transaction.js
+++ b/packages/spanner/test/transaction.js
@@ -180,6 +180,7 @@ describe('Transaction', function() {
 
   describe('begin', function() {
     var OPTIONS = {
+      readOnly: true,
       boundOptions: true,
       returnReadTimestamp: true
     };
@@ -193,6 +194,7 @@ describe('Transaction', function() {
     it('should make the correct request', function(done) {
       var transaction = new Transaction(SESSION, OPTIONS);
 
+      transaction.readOnly = true;
       transaction.request = function(config) {
         assert.deepEqual(config.reqOpts, EXPECTED_REQ_OPTS);
         assert.strictEqual(config.method(), util.noop);
@@ -203,6 +205,7 @@ describe('Transaction', function() {
     });
 
     it('should not require options', function(done) {
+      transaction.readOnly = false;
       transaction.request = function(config) {
         assert.deepEqual(config.reqOpts, {
           options: {
--- a/packages/spanner/src/transaction.js
+++ b/packages/spanner/src/transaction.js
@@ -164,7 +164,7 @@ Transaction.prototype.begin = function(callback) {
   var self = this;
   var options;
 
-  if (!this.options) {
+  if (!this.readOnly) {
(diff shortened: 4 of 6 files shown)


callmehiphop:
@vkedia it's not letting me add you as a reviewer for some reason.

packages/spanner/src/database.js
```diff
@@ -460,9 +460,6 @@ Database.prototype.getTransaction = function(options, callback) {
       return;
     }
 
-    options = extend({}, options);
-    delete options.readOnly;
```
vkedia: so this line was the root cause of the issue?

packages/spanner/src/transaction-request.js
```diff
@@ -51,7 +51,14 @@ var PartialResultStream = require('./partial-result-stream.js');
  * @param {object=} options - Timestamp options.
  */
 function TransactionRequest(options) {
+  this.readOnly = false;
+
   if (options && !is.empty(options)) {
+    options = extend({}, options);
+
+    this.readOnly = !!options.readOnly;
+    delete options.readOnly;
```
vkedia: why do you need to delete?

packages/spanner/test/transaction.js
```diff
@@ -180,6 +180,7 @@ describe('Transaction', function() {
 
   describe('begin', function() {
     var OPTIONS = {
+      readOnly: true,
```
vkedia: below in EXPECTED_REQ_OPTS you set readOnly to OPTIONS, so it will look like
```
readOnly: {
   readOnly: true,
   boundOptions: true,
   returnReadTimestamp: true,
}
```

So is that expected?

packages/spanner/src/database.js
callmehiphop: More or less yes, we weren't accounting for read only transactions that did not contain timestamp options.

packages/spanner/src/transaction-request.js
callmehiphop: We send this object when creating the transaction, if grpc finds a key that is not listed within a proto file it will throw an error.

packages/spanner/test/transaction.js
callmehiphop: This test is a little confusing, but essentially yes. We have an abstract class of sorts that will format the timestamp options and in the test we stub that class.

https://github.com/callmehiphop/gcloud-node/blob/84befe838b8923d1a61562c94f251e0afb42cec7/packages/spanner/test/transaction.js#L47

So we're essentially just testing that the value of `transaction.options` is sent along in the request.

PR #2574, opened 2017-08-30
Add trace ID to log metadata when available

draffensperger:
This makes the `logging-bunyan` and `logging-winston` libraries add the `LogEntry.trace` field based on the current trace context of the `@google-cloud/trace-agent` library when available.

Users can override that default `LogEntry.trace` assignment by specifying a `logging.googleapis.com/trace` field in the Winston entry metadata or Bunyan structured log entry. That naming makes logging backward-compatible for anyone who might happen to already use a `trace` field in their logs and it the follows the similar behavior for the Stackdriver logging agent, see https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/91835bb12b728ba88f63c38008e539a047d216a6/lib/fluent/plugin/out_google_cloud.rb#L105.

This will make use of the `getWriterProjectId` function that https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/548 adds to `cloud-trace-nodejs`.


draffensperger:
@matthewloring and @ofrobots could you take a look at this when you get a chance?

The `getCurrentTraceFromAgent` function is duplicated in both `logging-bunyan` and `logging-winston`, but it's pretty short and the alternative would be to move it to the core `logging` library, but that would require multiple PRs and updating dependencies (which I'm open to do if you prefer).

packages/logging-bunyan/src/index.js
```diff
@@ -133,8 +140,6 @@ LoggingBunyan.prototype.formatEntry_ = function(record) {
     );
   }
 
-  record = extend({}, record);
```
matthewloring: Is this record user provided? If so, deleting properties from the original object would probably be considered api breaking.

draffensperger: Yes, that definitely would be. The reason I removed this line here is that I added a call to `extend` in the `write` function (line 222 below). So essentially, it gets copied at ingestion into the buffer instead of on the write of the buffer to the API.

packages/logging-winston/src/index.js
```diff
@@ -126,6 +131,30 @@ winston.transports.StackdriverLogging = LoggingWinston;
 util.inherits(LoggingWinston, winston.Transport);
 
 /**
+ * Gets the current fully qualified trace ID when available from the
+ * @google-cloud/trace-agent library in the LogEntry.trace field format of:
+ * "projects/[PROJECT-ID]/traces/[TRACE-ID]".
+ */
+function getCurrentTraceFromAgent() {
```
ofrobots: @stephenplusplus Instead of duplication between `logging-{winston,bunyan}`, it would be nice to have this in `logging`. WDYT?

stephenplusplus: That's okay with me.

draffensperger: Cool. I just added a pull request for the `logging` library for that: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2597

I think we will need to release a new `logging` npm module version and then update `logging-winston` and `logging-bunyan` to depend on it as part of this change.

packages/logging-bunyan/src/index.js
```diff
@@ -41,6 +41,13 @@ var BUNYAN_TO_STACKDRIVER = {
 };
 
 /**
+ * Key to use in the Bunyan payload to allow users to indicate a trace for the
+ * request, and to store as an intermediate value on the log entry before it
+ * gets written to the Stackdriver logging API.
+ */
+var LOGGING_TRACE_KEY = 'logging.googleapis.com/trace';
```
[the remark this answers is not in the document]
draffensperger: Yeah, that's a good point on exporting this and also that this (and the feature generally) should be documented in the README.

I'll wait until we get the change to the `logging` library done (and new version updated), then I'll come back to this and make the changes.

packages/logging-winston/src/index.js
stephenplusplus: A second thought on this (I'm sorry) is maybe this code belongs in a different layer. `@google-cloud/logging` is a consumer module, as opposed to a utility suite. Bending the rules seems *okay*, although I could see an accident where a future maintainer removes this method because it's not clear why it's there (since the `@google-cloud/logging` codebase never directly calls it). And similarly, if the logging-* modules decide not to use this method anymore, its removal from `@google-cloud/logging` could be forgotten.

Alternatively, I've thought of a couple solutions. More are welcome:

- (A) Put this code in `@google-cloud/common`, i.e. `require('@google-cloud/common').util.getCurrentTraceFromAgent()`.
  - 👍 The code is in a utility library
  - 👎 The code isn't useful for all API modules, just `logging-bunyan` and `logging-winston`
- (B) Duplicate the code, as it is in this PR.
  - 👍 Code relevant to `logging-bunyan` and `logging-winston` stays close to their respective codebases.
  - 👎 Code is duplicated, which could mean a patch in one is forgotten to be placed in the other.
- (C) ?

I'm leaning towards (B), as the duplication is minor. And as is evident by this PR, usually a modification to `logging-winston` means a modification to `logging-bunyan` is simultaneous.

draffensperger: I think for (C), the eventual best solution would probably be to separate out a context propagation library that the trace SDK, logging-{winston,bunyan} can all three depend on that would expose this "get current context id" functionality. But that's a longer term approach not something that will help us directly here in this PR.

I have heard it said that "duplication is better than the wrong abstraction", and the actual function we would be duplicating is pretty small and straightforward.

draffensperger: @ofrobots, are you OK with keeping the getCurrentTraceFromAgent duplicated in both logging-winston and logging-bunyan for now?

ofrobots:
Sgtm

ofrobots:
@stephenplusplus Is this good to land from your POV?

PR #2577, opened 2017-08-31
Spanner auto-gen refresh.

lukesneeringer:
This updates the auto-gen component for Spanner.

@stephenplusplus I do not think proto changes are involved, so this should just be a "merge and launch". We should verify that in the auto-gen surface however.

### DIFF
--- a/packages/spanner/src/v1/index.js
+++ b/packages/spanner/src/v1/index.js
@@ -1,11 +1,11 @@
-/*!
- * Copyright 2016 Google Inc. All Rights Reserved.
+/*
+ * Copyright 2017, Google Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


packages/spanner/src/v1/index.js
```diff
@@ -26,10 +26,9 @@ function v1(options) {
   var gaxGrpc = gax.grpc(options);
   return spannerClient(gaxGrpc);
 }
+
+v1.GAPIC_VERSION = '0.0.5';
 v1.SERVICE_ADDRESS = spannerClient.SERVICE_ADDRESS;
 v1.ALL_SCOPES = spannerClient.ALL_SCOPES;
-v1.admin = {
```
stephenplusplus: This going away breaks our handwritten Spanner constructor:

```js
function Spanner(options) {
  // ...

  this.api = {
    Database: v1.admin.database(options).databaseAdminClient(options),
    Instance: v1.admin.instance(options).instanceAdminClient(options),
    Spanner: v1(options).spannerClient(options)
  };
```

Is there a new way to get to those clients?

lukesneeringer: No, this looks like something that was special for Spanner and autogen does not know that. I will set it back for now. :-)

lukesneeringer:
Re-genning shortly with new stuff.

PR #2581, opened 2017-08-31
storage: handle validation at the same rate as the destination pipe

### LINKED ISSUE #2569: storage - readStream hangs on long running time pipe
#### Environment details
  - OS: Ubuntu 16.04
  - Node.js version: 6.11.2
  - npm version: 3.10.10
  - google-cloud-node version: @google-cloud/storage@1.2.1 

Some more details:
The file size I'm reproducing the issue on is ~2.3 gb.
The issue is happening **only** when using **pipe** and **only** when the download speed is below 5 mb\s.
When the download speed is faster or when using `.on('data', (data) => res.write(data))`, there is no problem.

#### Steps to reproduce

1. require `@google-cloud/storage`
2. use the following route function, where bucketFile is a pure file instance from the storage lib:

```javascript
function DownloadOva (req, res) {
    console.log(new Date() + ' start');

    res.on('finish', function () {
        console.log(new Date() + ' finish');
    });

    let bucketFile = OvaFileHelpers.GetBucketFile('agent_ova/agent.ova');
    bucketFile.createReadStream({validation: false})
        .on('error', (err) => {
            console.log(new Date() + ' error: ' + err.stack);
        })
        .on('response', (streamResponse) => {
            console.log(new Date() + ' streamResponse');
            res.setHeader('Content-Length', streamResponse.headers['content-length']);
            res.setHeader('Content-Type', streamResponse.headers['content-type']);
            res.setHeader('Content-disposition', 'attachment; filename="agent_download.ova"');
        })
        .on('end', () => {
            console.log(new Date() + ' end');
            res.end();

            return true;
        })
        // .on('data', (data) => {
        //     Response.write(data);
        // })
        .pipe(res);
}
```
3. limit the download speed in chrome using the throttling option in the network developer tools to 5000 kb/s
4. call the route

**Expected results**:
all console logs outputs should be written to console.
**Actual results**:
The file is downloaded successfully
the 'end' and 'finish' events are never called (no console log outputs and doesn't get into the breakpoint). 
Can't call the route again until restarting the application

**more details**:
When the download speed is not limited, it works fine (on high speed connection). but some users with low speed connections, faced this issue (that's why I reproduced this issue).
When using the commented `.on('data' ...` and commenting out the `pipe`, it also works fine also on slow download.

It seems that when the `pipe` takes too long (seems like 3-4 minutes), it looses all the event listening functions and gets hangs even though the download finish successfully on the client side.

Thanks!

stephenplusplus:
Fixes #2569 

## Update: 9/1/17

Okay, looks like I had the whole thing wrong. In the process of debugging, I came across a couple of other issues that had to be cleared before I could uncover the problem. Therefore, this PR:

- Stops our sending a keepAlive header for file download requests. This turned out to be the sole fix for #2569 in my testing-- will need confirmation from @tomgrossman
- When receiving a file that is gzip'd, we do the decompression ourselves, as opposed to letting `request` do it. This allows us to:
  - use the compressed bytes for validation
  - pass the decompressed bytes to the user
  - do these steps incrementally at the rate that is asked for by the user's stream
- Prevents the stream we hand to the user from ending before we have validated the data. **TODO: System test for this**

#### To Dos
- [x] Unit tests
- [ ] System tests

cc @tomgrossman -- please take a look!
cc @lukesneeringer  -- I marked this as `do not merge` until we know the right place to merge this (gcn or [probably] gcn-storage)

### DIFF
--- a/packages/storage/src/file.js
+++ b/packages/storage/src/file.js
@@ -522,6 +522,16 @@ File.prototype.createReadStream = function(options) {
       })
       .resume();
 
+    throughStream.on('error', function() {
+      // An error can occur before the request stream has been created
+      // (during authentication).
+      if (requestStream.abort) {
+        requestStream.abort();
+      }
+
+      requestStream.destroy();
+    });
+
     // We listen to the response event from the request stream so that we can...
     //
     //   1) Intercept any data from going to the user if an error occurred.
@@ -546,16 +556,7 @@ File.prototype.createReadStream = function(options) {
       rawResponseStream
         .pipe(shouldRunValidation ? validateStream : through())
         .pipe(isCompressed ? zlib.createGunzip() : through())
-        .pipe(throughStream, { end: false })
-        .on('error', function() {
-          // An error can occur before the request stream has been created
-          // (during authentication).
-          if (requestStream.abort) {
-            requestStream.abort();
-          }
-
-          requestStream.destroy();
-        });
+        .pipe(throughStream, { end: false });
     }
 
     // This is hooked to the `complete` event from the request stream. This is
--- a/packages/storage/system-test/storage.js
+++ b/packages/storage/system-test/storage.js
@@ -1418,9 +1418,14 @@ describe('storage', function() {
     });
 
     it('should allow changing the storage class', function(done) {
+      var bucket = storage.bucket(generateName());
       var file = bucket.file(generateName());
 
       async.series([
+        function(next) {
+          bucket.create(next);
+        },
+
         function(next) {
           bucket.upload(FILES.logo.path, { destination: file }, next);
         },
--- a/packages/storage/test/file.js
+++ b/packages/storage/test/file.js
@@ -70,6 +70,13 @@ fakeRequest.defaults = function(defaultConfiguration) {
   return fakeRequest;
 };
 
+var hashStreamValidationOverride;
+var hashStreamValidation = require('hash-stream-validation');
+function fakeHashStreamValidation() {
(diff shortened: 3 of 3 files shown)


stephenplusplus:
@tomgrossman -- my original PR didn't have the right solution at all. I've updated it with a pretty sure fix from my testing. Can you give this a shot in your environment?

packages/storage/src/file.js
```diff
@@ -491,63 +505,77 @@ File.prototype.createReadStream = function(options) {
...
-          md5: md5
-        });
+      var headers = rawResponseStream.toJSON().headers;
+      var isCompressed = headers['content-encoding'] === 'gzip';
 
-        res.pipe(validateStream).on('data', common.util.noop);
-      }
+      rawResponseStream
+        .pipe(shouldRunValidation ? validateStream : through())
+        .pipe(isCompressed ? zlib.createGunzip() : through())
```
callmehiphop: I think we could optimize this if we were to conditionally pipe here. My understanding is that `through` uses a moderate amount of memory, so ideally we wouldn't use it as a noop. I think `pumpify` might be useful here

stephenplusplus: Good call, will do!

stephenplusplus:
RE: system tests, all I want to cover is that we validate the data before we end the stream the user has. Previously, the stream could end, then be discovered to be invalid.

stephenplusplus:
This is going over to googleapis/nodejs-storage

PR #2596, opened 2017-09-06
DLP changes

lukesneeringer:
@stephenplusplus This probably needs a new `google-proto-files`.

### DIFF
--- a/packages/dlp/src/v2beta1/dlp_service_client.js
+++ b/packages/dlp/src/v2beta1/dlp_service_client.js
@@ -312,13 +312,11 @@ DlpServiceClient.prototype.redactContent = function(request, options, callback) 
  *   identifier for the Operation, and the `count` is a counter used for
  *   tracking the number of files written. <p>The CSV file(s) contain the
  *   following columns regardless of storage type scanned: <li>id <li>info_type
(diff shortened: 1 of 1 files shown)


stephenplusplus:
We also have the dummy doc files that probably need updating, e.g. this line: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/49c4cd41c97c6de71ef1e8396aa10b892d5e255a/packages/dlp/src/v2beta1/doc/doc_dlp.js#L490

RE: google-proto-files, just change the dependency range in packages/dlp/package.json to `^0.13.1` and it will have the changes.

packages/dlp/src/v2beta1/dlp_service_client.js
```diff
@@ -425,7 +423,7 @@ DlpServiceClient.prototype.createInspectOperation = function(request, options, c
  * @param {string} request.name
  *   Identifier of the results set returned as metadata of
  *   the longrunning operation created by a call to CreateInspectOperation.
- *   Should be in the format of `inspect/results/{id}`.
+ *   Should be in the format of `inspect/results/{id}.
```
ace-n: Bug?

PR #2602, opened 2017-09-11
Export trace key in logging-{winston,bunyan}

draffensperger:
This makes the LOGGING_TRACE_KEY constant exported by the logging-{winston,bunyan} modules per conversation here: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2574#discussion_r136679515

### DIFF
--- a/packages/logging-bunyan/test/index.js
+++ b/packages/logging-bunyan/test/index.js
@@ -233,7 +233,7 @@ describe('logging-bunyan', function() {
     });
 
     it('should promote prefixed trace property to metadata', function(done) {
-      var recordWithTrace = extend({}, RECORD);
+      const recordWithTrace = extend({}, RECORD);
       recordWithTrace[LoggingBunyan.LOGGING_TRACE_KEY] = 'trace1';
 
       loggingBunyan.log_.entry = function(entryMetadata, record) {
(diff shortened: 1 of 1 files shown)


packages/logging-bunyan/test/index.js
```diff
@@ -285,9 +284,9 @@ describe('logging-bunyan', function() {
         getWriterProjectId: function() { return 'project1'; }
       };
       const recordWithoutTrace = extend({}, RECORD);
-      const recordWithTrace = extend({
-        'logging.googleapis.com/trace': 'projects/project1/traces/trace1'
-      }, RECORD);
+      var recordWithTrace = extend({}, RECORD);
```
ofrobots: nit: [Use a const unless the variable needs to be reassigned](https://google.github.io/styleguide/jsguide.html#features-use-const-and-let). In other languages `const` might connote immutability, but that's not the case for JS (AFAIK).

draffensperger: Done.

draffensperger:
Is this good to merge? @stephenplusplus, did you want to take a look?

stephenplusplus:
Sorry I didn't catch this one. LGTM!

Issue #2621, opened 2017-09-20
PubSub deadlock with flushQueue / ConnectionPool.isConnected() always false


ctavan:
#### Environment details

  - OS: Debian Jessie
  - Node.js version: 8.5.0
  - google-cloud-node version: 0.14.2

#### Steps to reproduce

I think there is an issue with queue flushing and reconnecting in the PubSub client. Here's a minimal example to reproduce it (locally, with the latest version of PubSub emulator):

```js
const pubsub = require('@google-cloud/pubsub')();

let last = Date.now();

function onError(err) {
  console.log('ON ERROR', err);
}
function onMessage(message) {
  console.log('ON MESSAGE', message.attributes, (Date.now() - last) / 1000);
  last = Date.now();
  message.ack();
}

const topicName = `my-topic-${Math.floor(Math.random() * 10000)}`;
const subscriptionName = `my-subscription-${Math.floor(Math.random() * 10000)}`;

pubsub.createTopic(topicName, () => {
  const topic = pubsub.topic(topicName);
  const publisher = topic.publisher();
  topic.createSubscription(subscriptionName, () => {
    const subscription = pubsub.subscription(subscriptionName);
    subscription.on('error', onError);
    subscription.on('message', onMessage);

    publisher.publish(new Buffer('x'.repeat(4000)), { batch: 'first' }, (err, messageId) => {
      console.log('DONE publishing immediately', err, messageId);
    });
    setTimeout(() => {
      publisher.publish(new Buffer('x'.repeat(4000)), { batch: 'second' }, (err, messageId) => {
        console.log('DONE publishing after timeout', err, messageId);
      });
    }, 2000);
  });
});

process.on('SIGINT', () => {
  const subscription = pubsub.subscription(subscriptionName);
  subscription.removeListener('message', onMessage);
  subscription.removeListener('error', onError);

  process.exit();
});
```

Output:
```
DONE publishing immediately null 229
ON MESSAGE { batch: 'first' } 1.36
DONE publishing after timeout null 230
ON MESSAGE { batch: 'second' } 1.988
ON MESSAGE { batch: 'second' } 10.008
ON MESSAGE { batch: 'second' } 10.006
```

As you can see the second message is re-delivered over and over again.

What I could find out so far is the following:

In
[...]

Issue #2619, opened 2017-09-19
Error: Received RST_STREAM with error code 2


lafama:
#### Environment details

  - OS: darwin x64
  - Node.js version:8.5.0
  - npm version:5.4.2
  - google-cloud/pubsub version:0.14.1

#### Steps to reproduce

  1. require `google-cloud/pubsub`
  2. 
`subscription=pubsub.topic('awesomeTopic).subscription('awesomeTopicSubcription',{reuseExisting: true})`
   3. 
```
subscription.on('error',function(error){
    logger.error('subscription',error);
  })
```
 4. This (3) will log the following 
```
{"message":"subscription",
"stack":"Error: Received RST_STREAM with error code 2\n    at ClientDuplexStream.<anonymous> 
(/app/node_modules/@google-cloud/pubsub/src/connection-pool.js:218:21)\n    
at emitOne (events.js:115:13)\n    at ClientDuplexStream.emit (events.js:210:7)\n    
at ClientDuplexStream._emitStatusIfDone (/app/node_modules/grpc/src/node/src/client.js:260:10)\n    
at ClientDuplexStream._receiveStatus (/app/node_modules/grpc/src/node/src/client.js:233:8)\n    
at /app/node_modules/grpc/src/node/src/client.js:757:12","code":13,"timestamp":"2017-09-19T04:57:56.510Z",
"level":"error"}
```

callmehiphop:
@lafama thanks for reporting. I believe this already being investigated, but this message is almost always received when the PubSub api closes a stream. However we usually retry on that specific error, so I'm intrigued as to why you're seeing it.

I'm unable to reproduce with the steps you've provided, so if you could provide us with a gist/repo/etc. that reproduces the error you're seeing, that would help tremendously.

lafama:
What would make PubSub api closes a stream? My worker process hangs after this error and i have to keep restarting it. The challenge reproducing this is that it occurs infrequently and at random time in production. It also only stated when upgrade to 0.14.0 from 0.13.0

callmehiphop:
The streaming API is new, hence why this only occurs in 0.14.x. Prior to this release we manually polled the servers.

Usually the PubSub API closes the stream after about 30 minutes. However, the only time we would emit an error like this would be if the client was unable to re-establish a connection within 5 minutes of the previous connections closing. So it might be that the server closes the connection and won't allow for any new connections. @jganetsk for thoughts here.

[...]

Issue #2598, opened 2017-09-07
subscription.on message wasn't received of messagActually unrelated failures, but we should bump this anyway.es after 10 mins or more.


QTGate:
- OS: Debian
  - Node.js version:  8.4.0
  - npm version:   5.4.0
  - google-cloud-node version:  0.56.0
  - @google-cloud/pubsub version: 0.14.0

#### Steps to reproduce

```js
const pubsub = require ( '@google-cloud/pubsub' )( connectOption )

const subscription = pubsub.topic ( topicName ).subscription ( subScription )

subscription.on ('message', message => {
    message.ack()
    ....
})

subscription.once('error', err => {
    ....
})
```

I received all un-received messages when I restart it.
I checked the connectionPool object when it have not received message:
connectionPool.isPaused [false]
connectionPool.isOpen [true]
connectionPool.connections [5]

Thank you.

callmehiphop:
Thanks for reporting! We've had multiple reports of similar behavior and we're currently investigating this issue. In the meantime you might want to use pubsub v0.13.x as a temporary workaround.

QTGate:
Thank you.
I did a temp code to use the pubsub v0.14.0 like this:
```js
const pubsub = require ( '@google-cloud/pubsub' )( connectOption )

const listenSub = () => {
    let subscription = pubsub.topic ( topicName ).subscription ( subScription )
    subscription.on ( 'message', message => {
        message.ack()
        ....
   })

    subscription.once('error', err => {
        ....
    })

    //        each 10 mins restart
    const tt = setTimeout (() => {
	subscription.close().then(() => {
		listenPubSubMaster = null
		return listenSub ()
    }, 600000 )
}
listenSub()
```

caseyduquettesc:
We see a similar behavior, although it takes a few hours. Eventually the subscription stops processing new messages and we have to restart the subscription. This only started after upgrading to 0.14.

mscottx88:
I too am seeing the same problem.  We set up a forever running process on GKE and it seems to go unresponsive after some time.  I came to the same conclusion of putting a restarter.  

Let's hope this gets resolved soon!

callmehiphop:
This is definitely a result of moving towards the new streaming API, I'm working on resolving this issue, but in the mean time I suggest downgrading to 0.13. I'll update this issue once it's been resolved.

pongad:
[...]

Issue #2572, opened 2017-08-27
PubSub locks up Node process at 100% CPU


ehacke:
#### Environment details

  - OS: Linux Mint 18.2 (dev) 
  - Node.js version: 8.3.0
  - npm version: 5.3.0
  - @google-cloud/pubsub version: 0.13.2 and 0.14

#### Steps to reproduce

  1.Try to read from PubSub subscription (even if it's completely empty)
  2. Node process goes to 100% of CPU and eventually is killed by K8S
  3. Apologize for not having any good reproduction steps

So we've been having intermittent issues with reading from subscriptions attached to one specific topic using Node. The same code will run fine for weeks, and then suddenly every Node process attached to that subscription is pegged to 100% CPU until the process is killed externally. The crazy part is that even after the subscription is completely drained, this continues to happen. There is nothing to read, and yet it's completely locked up. It will continue to lock up every process attached to it for days, then suddenly stop.

I completely agree that this sounds like an implementation issue on my side, but hear me out.

Important details:.
- It does not seem to matter if there is anything in the subscription. An empty subscription also locks up the reading process.
- Running a flamegraph analysis (attached) shows that 97.18% of the time, some function called Node::MakeCallback is on the top of the stack. In my mind this points to the GPRC library being involved in the problem.
- The same application code functions fine using Kafka as the queue, with the only difference being the very simple read and write operations.
- Other Node applications, written significantly differently, also attached to subscriptions on that topic will also fail in the same way, but at different times. Which suggests to me that it's not an issue with the content of the subscriptions (it can even be empty), but the operation of reading from the subscription itself.
- We have many many other Node processes, using the same reading and writing code, connected to different topics, that never experience this issue.
- If it helps, another unique detail is that the source of this topic is our only Dataflow process

Is there anything else I can do to try to trace this problem? 

I accept that I haven't given enough info for this be resolvable on your side, but I'm at a loss on how to debug this further.

![makecallback](https://user-images.githubusercontent.com/4016786/29754334-d3f5508a-8b50-11e7-8135-222d6f2618cb.png)

ehacke:
Things I tried that did not solve the issue:
- Recreating the Dataflow process that writes to the topic
- Recreating each individual subscription that is attached to the topic
- Recreating the processes that read from the topic
- Shutting down the source entirely and allowing the subscription to drain
- Shutting down every process involved simultaneously, waiting, and then bringing them all back up
[...]

PR #2627, opened 2017-09-25
pubsub: refactor subscription connection

callmehiphop:
Relates to #2621, #2619, #2598, #2572

### DIFF
--- a/packages/pubsub/src/connection-pool.js
+++ b/packages/pubsub/src/connection-pool.js
@@ -35,6 +35,10 @@ var v1 = require('./v1');
 var CHANNEL_READY_EVENT = 'channel.ready';
 var CHANNEL_ERROR_EVENT = 'channel.error';
 
+// if we can't establish a connection within 5 minutes, we need to back off
+// and emit an error to the user.
+var MAX_TIMEOUT = 300000;
+
 // codes to retry streams
 var RETRY_CODES = [
   0, // ok
@@ -342,7 +346,7 @@ ConnectionPool.prototype.getAndEmitChannelState = function() {
       elapsedTimeWithoutConnection = now - self.noConnectionsTime;
     }
 
-    deadline = now + (300000 - elapsedTimeWithoutConnection);
+    deadline = now + (MAX_TIMEOUT - elapsedTimeWithoutConnection);
 
     client.waitForReady(deadline, function(err) {
       self.isGettingChannelState = false;
@@ -392,7 +396,7 @@ ConnectionPool.prototype.getClient = function(callback) {
     }
 
     self.client = new Subscriber(address, credentials, {
-      'grpc.keepalive_time_ms': 300000,
+      'grpc.keepalive_time_ms': MAX_TIMEOUT,
       'grpc.max_receive_message_length': 20000001,
       'grpc.primary_user_agent': common.util.getUserAgentFromPackageJson(PKG)
     });
@@ -539,7 +543,7 @@ ConnectionPool.prototype.shouldReconnect = function(status) {
   }
 
   var exceededRetryLimit = this.noConnectionsTime &&
-    Date.now() - this.noConnectionsTime > 300000;
+    Date.now() - this.noConnectionsTime > MAX_TIMEOUT;
 
   if (exceededRetryLimit) {
     return false;
--- a/packages/pubsub/test/connection-pool.js
+++ b/packages/pubsub/test/connection-pool.js
@@ -757,15 +757,11 @@ describe('ConnectionPool', function() {
     var fakeChannelState;
     var dateNow;
     var fakeTimestamp;
+    var fakeChannel = {};
 
     var fakeClient = {
       getChannel: function() {
-        return {
-          getConnectivityState: function(shouldConnect) {
-            assert.strictEqual(shouldConnect, false);
-            return fakeChannelState;
-          }
-        };
+        return fakeChannel;
       }
     };
 
@@ -774,6 +770,10 @@ describe('ConnectionPool', function() {
     });
 
     beforeEach(function() {
+      fakeChannel.getConnectivityState = function() {
+        return fakeChannelState;
+      };
+
       fakeChannelState = 0;
       fakeClient.waitForReady = fakeUtil.noop;
 
@@ -828,6 +828,11 @@ describe('ConnectionPool', function() {
 
     it('should emit the ready event if the channel is ready', function(done) {
       fakeChannelState = channelReadyState;
+
+      fakeChannel.getConnectivityState = function(shouldConnect) {
+        assert.strictEqual(shouldConnect, false);
+        return fakeChannelState;
+      };
 
       pool.on(channelReadyEvent, function() {
         assert.strictEqual(pool.isGettingChannelState, false);
--- a/packages/pubsub/test/subscription.js
+++ b/packages/pubsub/test/subscription.js
@@ -778,6 +778,26 @@ describe('Subscription', function() {
       subscription.flushQueues_();
     });
 
+    it('should cancel any pending flushes', function() {
+      var fakeHandle = 'abc';
+      var cleared = false;
+
+      var _clearTimeout = global.clearTimeout;
+
+      global.clearTimeout = function(handle) {
+        assert.strictEqual(handle, fakeHandle);
+        cleared = true;
+      };
+
+      subscription.flushTimeoutHandle_ = fakeHandle;
+      subscription.flushQueues_();
+
+      assert.strictEqual(subscription.flushTimeoutHandle_, null);
+      assert.strictEqual(cleared, true);
+
+      global.clearTimeout = _clearTimeout;
+    });
+
     describe('with connection pool', function() {
       var fakeConnection;
 


packages/pubsub/src/connection-pool.js
```diff
@@ -32,6 +32,9 @@ var uuid = require('uuid');
 var PKG = require('../package.json');
 var v1 = require('./v1');
 
+var CHANNEL_READY = 'channel.ready';
```
stephenplusplus: Can we long-form name these, i.e. `CHANNEL_READY_EVENT_NAME`

packages/pubsub/src/connection-pool.js
```diff
@@ -177,6 +190,23 @@ ConnectionPool.prototype.createConnection = function() {
     var connection = client.streamingPull(self.metadata_);
     var errorImmediateHandle;
 
+    if (self.isPaused) {
+      connection.pause();
+    }
+
+    self.once(CHANNEL_ERROR, function(err) {
```
stephenplusplus: `err` isn't handled here

packages/pubsub/src/connection-pool.js
```diff
@@ -459,4 +477,65 @@ ConnectionPool.prototype.shouldReconnect = function(status) {
...
+    var state = channel.getConnectivityState(false);
+
+    if (state === READY) {
+      self.removeAllListeners(CHANNEL_ERROR);
+      self.emit(CHANNEL_READY);
+      return;
+    }
+
+    var elapsedTimeWithoutConnection = 0;
+    var deadline, timeout;
```
stephenplusplus: I think we're a multi-line var decl style kinda repo.

packages/pubsub/src/connection-pool.js
```diff
@@ -459,4 +477,65 @@ ConnectionPool.prototype.shouldReconnect = function(status) {
...
+    }
+  });
+};
+
+/**
+ *
+ */
+ConnectionPool.prototype.waitForReady = function() {
+  var self = this;
+  var READY = 2;
```
stephenplusplus: `READY_STATE_ID` or something.

packages/pubsub/src/connection-pool.js
```diff
@@ -459,4 +477,65 @@ ConnectionPool.prototype.shouldReconnect = function(status) {
   return true;
 };
 
+/**
+ *
+ */
+ConnectionPool.prototype.watchForReady = function() {
```
stephenplusplus: The pattern of `watching` then `waiting` "forReady" is a bit confusing. Can these two methods be merged into one, where it's just "waitForReady()"?

packages/pubsub/src/connection-pool.js
```diff
@@ -459,4 +477,65 @@ ConnectionPool.prototype.shouldReconnect = function(status) {
...
+  var self = this;
+
+  this.on('newListener', function(eventName) {
+    if (eventName == CHANNEL_READY && !self.waiting) {
+      clearImmediate(self.waitHandle);
```
stephenplusplus: I don't follow the clear & then setImmediate pattern here. What case does this handle?

packages/pubsub/src/connection-pool.js
```diff
@@ -142,16 +146,25 @@ ConnectionPool.prototype.acquire = function(id, callback) {
...
+  }
 
+  this.connections.clear();
+  this.queue.forEach(clearTimeout);
   this.queue.length = 0;
+
+  this.isOpen = false;
+  this.waiting = false;
+
+  this.removeAllListeners('newListener')
```
stephenplusplus: Are we the only ones who access `this`? Just thinking this could be bad if users might also have assigned a listener to `newListener` in their own apps for their own reasons.

packages/pubsub/src/connection-pool.js
```diff
@@ -177,6 +190,23 @@ ConnectionPool.prototype.createConnection = function() {
...
+    if (self.isPaused) {
+      connection.pause();
+    }
+
+    self.once(CHANNEL_ERROR, function(err) {
+      connection.cancel();
+    });
+
+    self.once(CHANNEL_READY, function() {
+      connection.isConnected = true;
```
stephenplusplus: This seems to be setting a property on a gRPC connection object. Is it supposed to be used this way? (Without knowing more, it seems like it's setting an internal property). I see this was part of the code before, but just wondering.

packages/pubsub/src/connection-pool.js
callmehiphop: Sure!

packages/pubsub/src/connection-pool.js
callmehiphop: We should be, yes, the pool isn't even documented.

packages/pubsub/src/connection-pool.js
callmehiphop: It was a mistake to add that argument, we shouldn't be handling the error there.

packages/pubsub/src/connection-pool.js
callmehiphop: It's not setting an internal property, but I can store it in a different way if you prefer.

packages/pubsub/src/connection-pool.js
callmehiphop: Sure! Done.

packages/pubsub/src/connection-pool.js
callmehiphop: I've since changed it, but this can fire off roughly 5 times at almost the same exact time, but it only should be called once.

packages/pubsub/src/connection-pool.js
callmehiphop: Sure, done.

packages/pubsub/src/connection-pool.js
callmehiphop: Yeah, we're not even using `timeout`, I just hadn't gotten around to cleaning up the code yet.

callmehiphop:
@stephenplusplus I think this is ready for another look through.

packages/pubsub/test/connection-pool.js
```diff
@@ -347,6 +392,66 @@ describe('ConnectionPool', function() {
...
+
+          pool.emit(channelErrorEvent);
+          assert.strictEqual(pool.listenerCount(channelReadyEvent), 0);
+        });
+
+        it('should cancel the connection', function() {
+          pool.createConnection();
+          pool.emit(channelErrorEvent);
+
+          assert.strictEqual(fakeConnection.canceled, true);
```
stephenplusplus: `canceled` being set to `true` looks to be done somewhere else, not in the `createConnection` body of code. Can you just stub the `connection.cancel` method to check that it was called?

packages/pubsub/test/connection-pool.js
```diff
@@ -674,15 +750,162 @@ describe('ConnectionPool', function() {
...
+
+      pool.on('error', function(err) {
+        assert.strictEqual(err, fakeError);
+        errorEmitted = true;
+      });
+
+      pool.getClient = function(callback) {
+        callback(fakeError);
+
+        assert.strictEqual(pool.isGettingChannelState, false);
```
stephenplusplus: I'm somewhat confused why this batch of asserts isn't a victim of the event queue. `callback` sends out async events, which apparently are being executed synchronously before these asserts 😕 . Just to be safe, can we do a `setImmediate` wrapper around these 3 asserts and done()? Unless I'm missing something about why this works.

packages/pubsub/test/connection-pool.js
```diff
@@ -674,15 +750,162 @@ describe('ConnectionPool', function() {
...
+    var channelReadyState = 2;
+    var fakeChannelState;
+    var dateNow;
+    var fakeTimestamp;
+
+    var fakeClient = {
+      getChannel: function() {
+        return {
+          getConnectivityState: function(shouldConnect) {
+            assert.strictEqual(shouldConnect, false);
```
stephenplusplus: `asserts` should be kept within the `it`s, otherwise they're the first to go during someone's refactor. Could you keep this mock here, so that it still doesn't have to be written for each test, but do the assert in an explicit test?

packages/pubsub/src/connection-pool.js
```diff
@@ -278,6 +305,59 @@ ConnectionPool.prototype.createMessage = function(connectionId, resp) {
...
+
+    var elapsedTimeWithoutConnection = 0;
+    var now = Date.now();
+    var deadline;
+
+    if (self.noConnectionsTime) {
+      elapsedTimeWithoutConnection = now - self.noConnectionsTime;
+    }
+
+    deadline = now + (300000 - elapsedTimeWithoutConnection);
```
stephenplusplus: Can we alias `300000` to a constant so that we know what/why it is that value?

packages/pubsub/src/subscription.js
```diff
@@ -520,6 +523,9 @@ Subscription.prototype.exists = function(callback) {
 Subscription.prototype.flushQueues_ = function() {
   var self = this;
 
+  clearTimeout(this.flushTimeoutHandle_);
```
stephenplusplus: I don't think this and the next line is tested.

packages/pubsub/test/connection-pool.js
callmehiphop: Actually that's exactly what is happening here. https://github.com/callmehiphop/gcloud-node/blob/728575b113a3715e4f83554c4859c4ce7aed898f/packages/pubsub/test/connection-pool.js#L65

packages/pubsub/test/connection-pool.js
callmehiphop: Looking at https://github.com/nodejs/node/blob/master/lib/events.js#L207
and https://github.com/nodejs/node/blob/master/lib/events.js#L103 it seems as though they are not processed asynchronously.

stephenplusplus: Fascinating! Never would have expected that.

packages/pubsub/package.json
```diff
@@ -59,6 +59,7 @@
     "google-auto-auth": "^0.7.1",
     "google-gax": "^0.13.0",
     "google-proto-files": "^0.12.0",
+    "grpc": "^1.6.0",
```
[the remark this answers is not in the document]
callmehiphop: Sure! If it's ok I'll just bump it here and release it first.

packages/pubsub/test/connection-pool.js
callmehiphop: Fair enough!

packages/pubsub/src/connection-pool.js
callmehiphop: Sure

packages/pubsub/src/subscription.js
callmehiphop: Ah! Yeah, good catch.

packages/pubsub/package.json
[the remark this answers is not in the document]
stephenplusplus: Sure sure, thanks!

callmehiphop:
@stephenplusplus this should be good to go btw!

ApeNox:
Could this be merged ? seems it fixes the issue with acked messages.

PR #2629, opened 2017-09-26
ci: fix lint command to support relocated APIs

stephenplusplus:
Fixes #2629

I can't reproduce the error locally, so this will be trial and error with CI on this PR.

### DIFF
--- a/scripts/helpers.js
+++ b/scripts/helpers.js
@@ -127,15 +127,20 @@ Module.getUpdated = function() {
 
   var files = output.stdout.toString();
 
+  console.log(files);
+
   var modules = files
     .trim()
     .split('\n')
     .filter(function(file) {
       return /^packages\/.+\.js/.test(file);
     })
-    .filter(file => test('-e', file)) // File exists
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Well, these tests passed, which I'm not sure means anything. I'll try an edit of a random file from a random package.

scripts/helpers.js
```diff
@@ -60,6 +60,11 @@ function Module(name) {
   this.name = name;
   this.directory = path.join(ROOT_DIR, 'packages', name);
 
+  if (!test('-e', this.directory)) {
```
callmehiphop: I think this is the wrong spot, if I had a guess I would say you want to filter out deleted modules [here](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/scripts/helpers.js#L135).

stephenplusplus: Thanks for that!

stephenplusplus:
@callmehiphop PTAL!

PR #2633, opened 2017-09-28
Log to console if client succeeded to connect & start streaming or not

ellenaua:


packages/speech/src/v1/speech_client.js
```diff
@@ -106,8 +106,26 @@ function SpeechClient(gaxGrpc, grpcClients, opts) {
...
+
+          // log to console if gRPC client succeeded to connect in 3 seconds or not
+          const CONNECT_TIMEOUT = 3 * 1000;
+          speechStub.waitForReady(Date.now() + CONNECT_TIMEOUT, (err) => {
+            if (err) {
+              console.log(`${methodName}: Failed to connect in ${CONNECT_TIMEOUT}msec`);
+            } else {
+              console.log(`${methodName}: Connected in ${CONNECT_TIMEOUT}msec`);
+            }
+          });
```
ellenaua: I just log it to console to show that it works.
Ideally, it should: 
- get CONNECT_TIMEOUT from options (options are 1st argument, args[0])
- throw an error if connection failed within CONNECT_TIMEOUT
- or pass an error to some callback (callback can be passed in options also)

Issue #2639, opened 2017-10-01
PubSub: Unable to create a pull subscription to a topic where acknowledged messages are retained


braxtonb:
#### Environment details

  - OS: macOS Sierra version 10.12.2
  - Node.js version: v8.2.1
  - npm version: 4.6.1
  - google-cloud-node version: 0.14.2

#### Steps to reproduce

  1. require `google-cloud`
  2. initialize pubsubClient
  3. create subscription to a topic with config option `retainAckedMessages` set to true (see code below)
  4. push 5 messages to a topic
  5. pull and acknowledge the 5 messages from the topic successfully the first time
  6. in a second attempt to pull the 5 messages again, 0 are retrieved
  7. get the metadata for the subscription created to the topic in step 3, and `retainAckedMessages` is set to false (and `messageRetentionDuration` is set to null even though it should default to 7 days)

Below is the code I use to create a subscription
```js
const PubSub = require('@google-cloud/pubsub');

const Keys = require('./json/pubsub-editor-keys');

const projectId = 'my-project-id';
let pubsubClient;
const subscriptionName = 'my-subscription-name';
const topicName = 'my-topic';

const initializePubSubClient = (client_email, private_key) => {
  const options = {
    projectId,
    credentials: {
      client_email,
      private_key
    }
  };
  pubsubClient = PubSub(options);
}

const createSubscription = (topic, name) => {
  const options = {
    retainAckedMessages: true
  };

  console.log('createSubscription called');

  pubsubClient
    .createSubscription(topic, name, options)
    .then((data) => {
      const subscription = data[0];
      const apiResponse = data[1];
      console.log('subscription', subscription);
      console.log('apiResponse', apiResponse);
    })
    .catch((err) => {
      console.log('error', err);
    });
}

initializePubSubClient(Keys.client_email, Keys.private_key);
createSubscription(topicName, subscriptionName);

```

- I would like to create a pull subscription to a topic where acknowledged messages are retained for a set amount of time. 
- I have tried accomplishing this by setting retainAckedMessages to true and leaving messageRetentionDuration as the default 7 days; however, I have not had success in initializing a subscription with those option values. 
- Regardless of the options I use to initialize the subscription, acknowledged messages are removed from the subscribed topic's stored messages.
- Related to #2240

callmehiphop:
[...]

PR #2643, opened 2017-10-02
pubsub: fix setMetadata method

callmehiphop:
Relates to #2639


packages/pubsub/src/subscription.js
```diff
@@ -1042,9 +1071,16 @@ Subscription.prototype.setMetadata = function(metadata, gaxOpts, callback) {
     gaxOpts = {};
   }
 
+  var body = Subscription.formatMetadata_(metadata);
```
stephenplusplus: Can we rename this `subscription`? We usually use `body` to mean "JSON payload" of a request.

callmehiphop: Done!

PR #2644, opened 2017-10-02
Increased timeout to match timeout in config file

vkedia:


packages/spanner/src/transaction.js
```diff
@@ -98,7 +98,7 @@ function Transaction(session, options) {
   this.retries_ = 0;
   this.runFn_ = null;
 
-  this.timeout_ = 60000;
+  this.timeout_ = 3600000;
```
callmehiphop: Could we just require the config file so we don't have to mange it in multiple places?

stephenplusplus: That makes sense to me.

vkedia: I agree. Currently if config file is present does this value get overridden by the config file value somewhere?

stephenplusplus:
@callmehiphop do we still need this?

callmehiphop:
@stephenplusplus yep!

stephenplusplus:
Could you create an issue / send the PR on the new repo?

stephenplusplus:
Survived by https://github.com/googleapis/nodejs-spanner/pull/75

PR #2663, opened 2017-10-10
Update logging-{winston,bunyan} versions

draffensperger:
@stephenplusplus and @ofrobots what would you think about doing a new NPM release of the logging-{winston,bunyan} libraries now that they have the ability to add trace IDs when available (plus a few other changes)?


stephenplusplus:
Just need the sign-off from @ofrobots. I'll release manually if so.

ofrobots:
+1 on releasing.

stephenplusplus:
Done!

PR #2666, opened 2017-10-11
bigquery: add simple benchmarks

pongad:
Benchmarks make various queries and measure
1. how long until we get the first row
2. how long we need to go through all rows

### DIFF
--- a/packages/bigquery/benchmark/bench.js
+++ b/packages/bigquery/benchmark/bench.js
@@ -19,14 +19,16 @@
 const async = require('async');
 const fs = require('fs');
 const BigQuery = require('../src/index.js');
+const env = require('../../../system-test/env.js');
 
 if (process.argv.length < 3) {
-  throw `need query file; usage: \
-'${process.argv[0]} ${process.argv[1]} <queries.json>'`;
+  throw new Error(`need query file; ` +
+    `usage: '${process.argv[0]} ${process.argv[1]} <queries.json>'`);
 }
+
 var queryJson = fs.readFileSync(process.argv[2]);
 var queries = JSON.parse(queryJson);
-var client = new BigQuery();
+var client = new BigQuery(env);
 
 var doQuery = function(queryTxt, callback) {
   var startMilli = new Date().getTime();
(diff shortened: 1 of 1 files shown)


packages/bigquery/benchmark/bench.js
```diff
@@ -0,0 +1,68 @@
...
+    numRows,
+    numCols,
+    timeFirstByteMilli/1000,
+    timeTotalMilli/1000));
+  });
+}
+
+for (queryTxt of queries) {
+  doQuery(queryTxt);
+}
```
pongad: Is it possible to run queries serially? We're pulling A LOT of data and I think running queries in parallel might significantly skew results due to network and CPU contention.

callmehiphop: Yeah, you'll probably want to add a callback function to `doQuery` and call it on the `end` event. Then you can use the `async` module (it's already listed as a dev dependency) to do something like

```js
async.eachSeries(queries, doQuery, err => {});
```

pongad: Done. Thank you for teaching me Node :)

pongad:
@callmehiphop PTAL

packages/bigquery/benchmark/bench.js
```diff
@@ -0,0 +1,69 @@
+/*!
+ * Copyright 2014 Google Inc. All Rights Reserved.
```
callmehiphop: I think we'll want to update the year on this

packages/bigquery/benchmark/bench.js
```diff
@@ -0,0 +1,69 @@
...
+ * limitations under the License.
+ */
+
+const async = require('async');
+const fs = require('fs');
+const util = require('util');
+const BigQuery = require('../src/index.js');
+
+if (process.argv.length < 3) {
+  throw util.format("need query file; usage: '%s %s <queries.json>'", process.argv[0], process.argv[1]);
```
callmehiphop: We could just use an es6 template string here

```js
throw `need query file; usage: '${process.argv[0]} ${process.argv[1]} <queries.json>'`;
```

packages/bigquery/benchmark/bench.js
```diff
@@ -0,0 +1,69 @@
...
+      this.end();
+      callback(util.format("query %j: wrong number of columns, want %d got %d",
+          queryTxt,
+          numCols,
+          Object.keys(row).length));
+    }
+    numRows++;
+  })
+  .on('end', function() {
+    timeTotalMilli = new Date().getTime() - startMilli;
```
callmehiphop: This variable doesn't appear to be defined anywhere

packages/bigquery/benchmark/bench.js
pongad: I keep missing this. Done.

packages/bigquery/benchmark/bench.js
pongad: Done

packages/bigquery/benchmark/bench.js
pongad: Done

pongad:
@callmehiphop Thank you for all the pointers, PTAL. The long strings look a little weird to me, I'm not sure if this is the node-idiomatic way to do it.

As to the CI, eventually I'd like to make this benchmark run in all languages on schedule and auto-publish result to dashboard. It's somewhat resource-intensive so we'll have to run on dedicated machine, etc. I don't know how to do all this yet, so I think we can make it manually-runnable now and defer automation for a little later. WDYT?

pongad:
@callmehiphop ping

callmehiphop:
@pongad you are right, the strings looked weird. I added a small commit that tried to clean them up and I also ran `prettier` against the code.

PR #2673, opened 2017-10-13
spanner: use retry info

### LINKED ISSUE #2646: Cloud Spanner client should use RetryInfo in gRPC trailer to figure out retry delays
Cloud spanner client library automatically retries transactions that failed with ABORTED error. Currently it computes the retry delay in code using exponential backoff. But Cloud Spanner sends a RetryInfo trailer whenever it aborts a transaction that tells after how much the transaction should be retried.
Client library should be changed to extract the retry delay from this trailer and use that to sleep.

https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L41

callmehiphop:
Closes #2646

### DIFF
--- a/packages/spanner/src/transaction.js
+++ b/packages/spanner/src/transaction.js
@@ -22,7 +22,9 @@
 
 var common = require('@google-cloud/common');
 var extend = require('extend');
+var grpc = require('grpc');
 var is = require('is');
+var path = require('path');
 var through = require('through2');
 var util = require('util');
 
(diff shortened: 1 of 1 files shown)


callmehiphop:
@vkedia I can't add you as a reviewer but PTAL.

stephenplusplus:
@callmehiphop can you port this over to https://github.com/googleapis/nodejs-spanner?

lukesneeringer:
@callmehiphop We should be doing a Spanner release tomorrow -- would love to get this in, but it does need to be ported to the dissected repo. Sorry. :-/

(It is not just a matter of applying the patch; the code has changed significantly.)

PR #2691, opened 2017-10-23
chore: update deps: google-{gax,proto-files}

ofrobots:


ofrobots:
Removed the change from logging, monitoring and spanner as those code-bases are already migrated.

PR #2708, opened 2017-10-30
Support BeginTransaction TransactionOptions.

stephenplusplus:
# Why <kbd>do not merge</kbd>?
https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2708#discussion_r147817198

---

Change Overview and Requirements: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4278

This PR:

- Updates the Datastore proto files
- Allows setting the `BeginTransaction` rpc's `TransactionOptions`.

Examples:

#### Global-Level Setting
```js
var transaction = datastore.transaction({ readOnly: true });
transaction.run(function(err) {
  transaction.get(key, function() { /*...*/ });
});
```
```js
var transaction = datastore.transaction({ id: 'previous-id' });
transaction.run(function(err) {
  transaction.save({ key, data });
  transaction.commit(function(err) {});
});
```

#### Method-Level Setting
```js
var transaction = datastore.transaction();
transaction.run({ readOnly: true }, function(err) {
  transaction.get(key, function() { /*...*/ });
});
```
```js
var transaction = datastore.transaction();
transaction.run({ transactionId: 'previous-id' }, function(err) {
  transaction.save({ key, data });
  transaction.commit(function(err) {});
});
```

#### Raw Configuration (method level only)
```js
var options = {
  transactionOptions: {
    readWrite: {
      previousTransaction: 'previous-id'
    }
  }
};

var transaction = datastore.run(options, function(err) {
  transaction.save({ key, data });
  transaction.commit(function(err) {});
});
```

### DIFF
--- a/packages/datastore/system-test/datastore.js
+++ b/packages/datastore/system-test/datastore.js
@@ -927,5 +927,35 @@ describe('Datastore', function() {
         });
       });
     });
+
+    it('should read in a readOnly transaction', function(done) {
+      var transaction = datastore.transaction({ readOnly: true });
+      var key = datastore.key(['Company', 'Google']);
+
+      transaction.run(function(err) {
+        assert.ifError(err);
+        transaction.get(key, done);
+      });
+    });
+
+    it('should not write in a readOnly transaction', function(done) {
+      var transaction = datastore.transaction({ readOnly: true });
+      var key = datastore.key(['Company', 'Google']);
+
+      transaction.run(function(err) {
+        assert.ifError(err);
+
+        transaction.get(key, function(err) {
+          assert.ifError(err);
+
+          transaction.save({ key: key, data: {} });
+
+          transaction.commit(function(err) {
+            assert(err instanceof Error);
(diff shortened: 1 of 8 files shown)


packages/datastore/system-test/datastore.js
```diff
@@ -927,5 +927,35 @@ describe('Datastore', function() {
...
+      var transaction = datastore.transaction({ readOnly: true });
+      var key = datastore.key(['Company', 'Google']);
+
+      transaction.run(function(err) {
+        assert.ifError(err);
+        transaction.get(key, done);
+      });
+    });
+
+    it('should not write in a readOnly transaction', function(done) {
```
stephenplusplus: Note that this test fails, which is mentioned in https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4278 under "Integration Tests":

> - A read-only transaction fails if it attempts to write. This test currently fails against the production endpoint.

PR #2739, opened 2017-11-09
Generate insertId if not provided.

### LINKED ISSUE #1041: bigquery table.insert fails for identical rows

stephenplusplus:
Fixes #1041

### DIFF
--- a/packages/bigquery/src/table.js
+++ b/packages/bigquery/src/table.js
@@ -1007,6 +1007,7 @@ Table.prototype.insert = function(rows, options, callback) {
   if (!options.raw) {
     json.rows = arrify(rows).map(function(row) {
       return {
+        insertId: uuid.v4(),
         json: Table.encodeValue_(row)
       };
     });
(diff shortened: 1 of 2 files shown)


callmehiphop:
@tswast based on some dialog within the attached issue, it seems like a uuid per row might not be the way to go. Would you mind taking a peak at this?

tswast:
UUID-per-row is the way to go (if the user does not provide one).

After some discussion with the BigQuery team, we determined that a row ID should always be included in the request. If not present, create a unique ID for the row through a suitably random source.

The reason behind this is so that the API request can safely be retried by the client library.

callmehiphop:
Awesome, thanks for the feedback!

PR #2742, opened 2017-11-11
Default namespace in authentication docs

HugoReeves:
The authentication config object can also contain a namespace key. The namespace key defines the default namespace. Added as a third option under the explanation of the config object.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -16,11 +16,14 @@ A `config` object is not required if you are in an environment which supports [A
 If this doesn't describe your environment, the `config` object expects the following properties:
 
 1. One of the following:
-  1. `credentials` object containing `client_email` and `private_key` properties.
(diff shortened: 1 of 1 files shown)


HugoReeves:
I signed it

stephenplusplus:
Thanks for this PR, @HugoReeves! We actually don't want to document this here, because the file you edited is shown across all languages in our docs. The current docs mention the `namespace` configuration option on its overview page here: https://googlecloudplatform.github.io/google-cloud-node/#/docs/datastore/1.1.0/datastore.

Datastore is going to relocate to a new repository very soon, and its documentation will be hosted on cloud.google.com (not on GitHub pages), with a different layout and navigation structure. The new documentation will keep all of the options together on one page, which should make this less confusing.

Thanks again for taking the time to help out!

Issue #2699, opened 2017-10-27
Latest pubsub version breaks dependencies with grpc (doesn't use common-grpc package)


lostpebble:
#### Environment details

  - OS: Linux
  - Node.js version: 8.7.0
  - npm version: 5.4.2
  - google-cloud-node version:

#### Steps to reproduce

Try use the latest versions of `@google-cloud/datastore` and `@google-cloud/pubsub` together.

Because the latest version of `@google-cloud/pubsub` includes grpc as a direct dependency instead of using the package which seems to have been created for linking grpc together between libraries (`@google-cloud/common-grpc`), the versions of GRPC are clashing and causing problems such as #2667 

To get my project working I had to revert to `@google-cloud/pubsub:0.14.2`

It appears to have been changed during this commit: #2627

ThomWright:
I can reproduce with the following in my `package.json`:

```
"@google-cloud/pubsub": "^0.14.5",
"grpc": "^1.6.6",
```

Fixing `@google-cloud/pubsub:0.14.2` worked for me too.

watermanio:
I think I'm seeing something related using PubSub on GCF with `@google-cloud/pubsub:0.14.6`:

```javascript
var Buffer = require("safe-buffer").Buffer;
var pubsub = require('@google-cloud/pubsub')();

exports.publish = function publish(event, callback) {
    var topic = pubsub.topic('outbound-topic');
    var publisher = topic.publisher();
    var message = new Buffer('New message!');
    publisher.publish(message, function(err, data) {
        callback(err, data);
    });
};
```

Throws the following error:

```
info: Function crashed
info: TypeError: message.encode(...).finish is not a function
    at BundleDescriptor.getByteLength [as byteLengthFunction] (./node_modules/google-gax/lib/grpc.js:262:36)
    at ./node_modules/google-gax/lib/bundling.js:338:38
    at Array.forEach (native)
    at BundleExecutor.schedule (./node_modules/google-gax/lib/bundling.js:337:16)
    at ./node_modules/google-gax/lib/bundling.js:474:20
    at Canceller.call (./node_modules/google-gax/lib/api_callable.js:110:19)
    at Bundleable.call (./node_modules/google-gax/lib/bundling.js:473:12)
    at ./node_modules/google-gax/lib/api_callable.js:356:17
```

stephenplusplus:
cc @jsididris who reported this in #2717.

callmehiphop:
[...]

PR #2744, opened 2017-11-14
pubsub: convert iam to use gax

callmehiphop:
Relates to #2699

### DIFF
--- a/packages/common-grpc/package.json
+++ b/packages/common-grpc/package.json
@@ -42,7 +42,7 @@
     "dot-prop": "^2.4.0",
     "duplexify": "^3.5.1",
     "extend": "^3.0.0",
-    "grpc": "^1.6.0",
+    "grpc": "^1.7.1",
     "is": "^3.2.0",
     "modelo": "^4.2.0",
     "retry-request": "^3.0.0",


packages/pubsub/src/iam.js
```diff
@@ -68,32 +63,16 @@ var util = require('util');
...
-
-  commonGrpc.Service.call(this, config, pubsub.options);
 }
 
-util.inherits(IAM, commonGrpc.Service);
-
 /**
  * Get the IAM policy
  *
+ * @param {object=} gaxOpts - Request configuration options, outlined
```
stephenplusplus: Can you call these `gaxOptions`?

packages/pubsub/system-test/pubsub.js
```diff
@@ -513,7 +513,6 @@ describe('pubsub', function() {
         assert.ifError(err);
 
         assert.deepEqual(policy.bindings, []);
-        assert.strictEqual(policy.etag, 'ACAB');
```
stephenplusplus: What happens now?

callmehiphop: It returns a Buffer, I tried decoding using the generated Policy object but that returns a ByteBuffer from protobufjs.

Issue #2745, opened 2017-11-15
Job.prototype.poll_ does not callback with err when job failed.


ziplokk1:
Description: 
Job.prototype.poll_ does not callback with err when job failed.
This is causing issues with Job.promise since the promise will always resolve instead of reject

Problem snippet:

```javascript
/**
 * Poll for a status update. Execute the callback:
 *
 *   - callback(err): Job failed
 *   - callback(): Job incomplete
 *   - callback(null, metadata): Job complete
 *
 * @private
 *
 * @param {function} callback
 */
Job.prototype.poll_ = function(callback) {
  this.getMetadata(function(err, metadata, apiResponse) {
    if (!err && apiResponse.status && apiResponse.status.errors) {
      // ******************************************
      // Here is where the issue lies.
      // For some reason err is undefined which is 
      // causing the following if statement condition to 
      // be false, thus not calling the callback with the error.
      // 
      // The solution is just putting `new` before common.util.ApiError.
      // err = new common.util.ApiError(apiResponse.status);
      err = common.util.ApiError(apiResponse.status);
    }
    if (err) {
      callback(err);
      return;
    }

    if (metadata.status.state !== 'DONE') {
      callback();
      return;
    }

    callback(null, metadata);
  });
};
```

Environment Details:
  - OS: agnostic
  - Node.js version: v6.11.2
  - npm version: 3.10.10
  - @google-cloud/bigquery version: 0.10.0

Steps to reproduce:

```javascript
const bigquery = require('@google-cloud/bigquery')();
// Query a table that doesn't exist to cause the job to fail.
bigquery.startQuery({query: "SELECT * FROM dataset.abc_this_table_doesnt_exist;"})
    .then(([job, apiResponse]) => {
        job.on('error', function (err) {
            // Never called
            console.error(err);
        });
        job.on('complete', function (meta) {
            console.log('success');
            console.log(meta.status.errors);
        });
    });
```

stephenplusplus:
Interesting! `bigquery` is on a small code-freeze as it is relocated to https://github.com/googleapis/nodejs-bigquery. This issue will be moved over to that repo when the relocation is complete, and we will have to re-implement the solution then. Thanks for finding this and fixing it!

stephenplusplus:
This issue was moved to googleapis/nodejs-bigquery#14

PR #2746, opened 2017-11-15
Fix Job.prototype.poll_

ziplokk1:
#2745 

Error was not being instantiated and thus never triggering the errback.


ziplokk1:
I signed it!

stephenplusplus:
Closing as explained in: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2745#issuecomment-344752615

PR #2751, opened 2017-11-20
Excise common.

alexander-fenster:
Common package has been moved to [googleapis/nodejs-common](https://github.com/googleapis/nodejs-common) and the latest version is successfully published from that new location. Removing it from here.


stephenplusplus:
This is a big moment. This is some ancient code that has been through a lot with us. You will be missed, common.

PR #2752, opened 2017-11-21
Make ackDeadline editable

mkamioner:
Love the new changes, but I miss the ability to specify my `ackDeadline` -- Sometimes I have processes with long running jobs and I want to be able to change it in once place

### DIFF
--- a/packages/pubsub/src/subscription.js
+++ b/packages/pubsub/src/subscription.js
@@ -92,6 +92,9 @@ var Snapshot = require('./snapshot.js');
  *     messages. Default: Infinity.
  * @param {number} options.maxConnections - Use this to limit the number of
  *     connections to be used when sending and receiving messages. Default: 5.
+ * @param {number} options.ackDeadline - The maximum time after receiving a
+ *     message that you must ack a message before it is redelivered, in
+ *     milliseconds. Default: 10000.
  *
  * @example
  * //-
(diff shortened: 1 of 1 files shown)


callmehiphop:
@mkamioner thanks for the PR! PubSub is in a code freeze at the moment, but I'll try and make sure to squeeze this in once that is over. 😃

callmehiphop:
@mkamioner pubsub has moved to https://github.com/googleapis/nodejs-pubsub, so unfortunately we'll have to close this PR. 😞 

Could you explain to me the end goal of setting a default ack deadline? Are you hoping to opt out of auto-leasing?

worudso:
I can give you example when manual ack deadline setting is necessary. I have some kinds of messages which spawn hundreds of messages when it completes successfully. And those spawned messages can produce hundres of messages again recursively. And the first 'root' message can take a long time to be finished, longer than 10 seconds at least. I agree that current implementation(automatic control of ack deadline) has many benefits. But in the above mentioned case, before ack deadline conveges to ideal value, there happens more than thousands duplicated messages, because first 'root' message can be processed multiple times redundantly.

stephenplusplus:
Ping @callmehiphop -- is this something we still want over on nodejs-pubsub?

callmehiphop:
@stephenplusplus yes, although we need to make a few changes to make the configurable deadline stick.

stephenplusplus:
Opened an issue to track the request here: https://github.com/googleapis/nodejs-pubsub/issues/6

PR #2767, opened 2017-12-04
Redirect to new API reference docs.

jmdobry:
We'll remove the pubsub check later.


callmehiphop:
@jmdobry @lukesneeringer I had opened a PR against the docs site to add redirects a bit differently.

https://github.com/GoogleCloudPlatform/google-cloud-common/pull/239

Do we want to allow users to view older versions of the docs since I assume they weren't formatted correctly to be on the cloud site? Or do we not care about old versions anymore??

jmdobry:
I'll leave it up to @lukesneeringer.

PR #2771, opened 2017-12-12
Update retry-request to 3.3.1

kolodny:
This is to enable not needing to call setTimeout when making a followup request. See https://github.com/googleapis/nodejs-bigtable/pull/17#discussion_r156229018 for more details


stephenplusplus:
Thanks, I was just doing this :)

Issue #2510, opened 2017-08-06
[datastore] new excludeFromIndexes syntax should allow for a catch-all on object properties


lostpebble:
Playing around with this today (after seeing this issue #1916 has been resolved) and I'm noticing that you have to define each and every embedded object property that you would like unindexed, instead of being able to just define a single "catch all" `excludeFromIndexes` option.

This works (actually it doesn't really, see edit below...):

```
{
  "key": { ... },
  "excludeFromIndexes": [
    "testEmbeddedObject.otherStringThing",
    "testEmbeddedObject.stringThing",
    "testLongString"
  ],
  "data": {
    "testLongString": "really long string here",
    "testEmbeddedObject": {
        "stringThing": "really long string here",
        "otherStringThing": "really long string here"
     }
  }
}
```

But this does not:

```
{
  "key": { ... },
  "excludeFromIndexes": [
    "testEmbeddedObject",
    "testLongString"
  ],
  "data": {
    "testLongString": "really long string here",
    "testEmbeddedObject": {
        "stringThing": "really long string here",
        "otherStringThing": "really long string here"
     }
  }
}
```

In the second example I'm still getting an error for `stringThing` and `otherStringThing` being longer than 1500 bytes. Is there no way to define that it catches all the properties in an embedded object?

Maybe something like this, if you'd like the catch all to have a more intentional syntax:

```
 "excludeFromIndexes": [
    "testEmbeddedObject.*",
    "testLongString"
  ],
```

Though, I would hope that for this syntax of `propertyName.*`, which seems to say catch all properties of the __object__ at propertyName, would also catch things that are not embedded objects but also simply a long string at that propertyName. I would just like a way to define that data stored at a certain property of an entity should not be indexed _at all_ - be it a string, embedded object or whatever.

EDIT:

Upon thinking about it more, why does the syntax in my second example not work? I think putting a `*` wildcard shouldn't be necessary actually. You've deliberately indicated you do not want that property indexed and that should mean the _entire_ property, be it an embedded object and all it's properties or just a long string. If you'd still like to index some properties of an embedded object, then you'd define those which you want unindexed and leave out the ones you want indexed.

This is confusing because upon looking at my entities in the datastore console it appears that my first example is actually wrong. I should have included the base `testEmbeddedObject` in my exclusion array too, so it would look like this now:

```
"excludeFromIndexes": [
    "testEmbeddedObject",
    "testEmbeddedObject.otherStringThing",
[...]

PR #2773, opened 2017-12-12
datastore: Support excludeFromIndexes for primitives in arrays.

### LINKED ISSUE #2615: Datastore: field types of array are always showing indexed regardless of excludeFromIndexes
Saving array's shows it is indexed (not all the actual items in the array, just the field), even when excludeFromIndexes: true.

I am not sure if this is normal behavior and the whole array is not actually stored in the index.  I had asked this originally in gstore-node issues and he said to ask here :)

## You can see the "circles" field shows indexed
image

## Inside the "circles" it is showing all array items are excluded from index
image

#### Environment details

  - OS: Windows 10
  - Node.js version: 6.11.2
  - npm version: 3.10.10
  - google-cloud-node version: 
 Replecated on google-cloud: 0.56.0
and @google-cloud/datastore: 1.1.0
I originally was using gstore-node and saw this problem, and now I am using just raw datastore and still have it.

#### Steps to reproduce
Seed database with a field of array

## My test function
```js
createEntity([
  {
    name: 'kind',
    value: 'Pages',
    excludeFromIndexes: true,
  },
  {
    name: '_id',
    value: '6',
  },
  {
    name: 'title',
    value: 'my first title',
  },
  {
    name: 'dateCreated',
    value: new Date(),
  },
  {
    name: 'string',
    value: 'A long string goes here',
    excludeFromIndexes: true,
  },
  {
    name: 'circles',
    value: ['1', '2', '3', '4'],
    excludeFromIndexes: true,
  },
]);
```
The actual save function
```js
export default async function createEntity(entity) {
  let response = null;
  let kind = null;
  let dsKey = null;

  try {
    entity.map((entityFeilds) => {
      if (entityFeilds.name === '_id') {
        dsKey = entityFeilds.value;
        if (dsKey === '') {
          dsKey = null;
        }
        return dsKey;
      }
      if (entityFeilds.name === 'kind') {
        kind = entityFeilds.value;
        return kind;
      }
      return null;
    });
    const key = datastoreClient.key([kind, dsKey]);
    const data = entity;
    const newEntity = {
      key,
      data,
    };

    response = await datastoreClient.save(newEntity);
  } catch (err) {
    console.error('save err', err);
  }
  return response;
}
```

beaulac:
Resolves #2615 .

This slightly modifies the index exclusion syntax discussed in #1916, which was merged in #2497.
 
Primitives in an array can now be excluded from indexes like this:
```javascript
{
  key: key,
  data: {
      array: [
          'a string', // Excluded
          {name: 'string in entity'} // Not excluded
      ]
  },
  excludeFromIndexes: ['array[]']
}
```

Previously, `excludeFromIndexes: ['array[]']` would have thrown a server-side error because it was interpreted as excluding the arrayValue itself from indexes, which is disallowed by the Datastore API.

This PR changes this behaviour: **now, an `excludeFromIndexes` path ending with `[]` means that non-entity values in the array should be excluded**. 

To exclude properties of entities in arrays, the existing nested-property syntax (e.g. `array[].name`) is still used. 

For example, to exclude both from indexes:
```javascript
{
  key: key,
  data: {
      array: [
          'a string', // Excluded
          {name: 'string in entity'} // 'name' property is also excluded
      ]
  },
  excludeFromIndexes: ['array[]', 'array[].name']
}
```

**Nested arrays remain unsupported**. The following will still break, as it currently does:
```javascript
{
  key: key,
  data: {
      array: [
          [1] // 'excludeFromEntities' will be set on the arrayValue, which breaks. 
      ]
  },
  excludeFromIndexes: ['array[]']
}
```

I wanted to change as little as possible of how entity properties are excluded, since there is no clear conclusion on how #2510 ("catch-all" index-exclusion) will be addressed. As such, I didn't want to redo the core implementation, but I realize that [`excludePathFromEntity`](https://github.com/beaulac/google-cloud-node/blob/b29308b63f0f1c25c65bcf1f86c199babef528aa/packages/datastore/src/entity.js#L455) is becoming rather convoluted. If you'd prefer, I'd be happy to extract it into something that's more manageable.

### DIFF
--- a/packages/datastore/system-test/datastore.js
+++ b/packages/datastore/system-test/datastore.js
@@ -89,6 +89,9 @@ describe('Datastore', function() {
       var data = {
         longString: longString,
         notMetadata: true,
+        longStringArray: [
+          longString,
+        ],
         metadata: {
           longString: longString,
           otherProperty: 'value',
@@ -130,6 +133,7 @@ describe('Datastore', function() {
         data: data,
         excludeFromIndexes: [
           'longString',
+          'longStringArray[]',
           'metadata.obj.longString',
           'metadata.obj.longStringArray[].longString',
           'metadata.obj.longStringArray[].nestedLongStringArray[].longString',


stephenplusplus:
This looks good to me. I added the scenario into the system tests (the tests that run against the actual API).

Only problem now is that we're blocked for a short time while we are extracting the Datastore module into its own repo (soon to be https://github.com/googleapis/nodejs-datastore). After that, I'll port the PR over, and get it released ASAP.

Thank you for helping out!

lostpebble:
Ah thank you @beaulac ! This has been a much wanted resolution.

stephenplusplus:
@beaulac The other repo is now open: https://github.com/googleapis/nodejs-datastore -- I plan to port your work this week, but if you're willing, feel free to beat me to it! :)

beaulac:
[Merged in new repository.](https://github.com/googleapis/nodejs-datastore/pull/18)

PR #2785, opened 2018-01-08
chore: license checking in posttest

jinwoo:
This can't be checked in until a license issue with
`@google-cloud/common-grpc` is fixed. It has a transitive dependency,
`jsonify`, which has "Public Domain" as its license. But Public Domain
is not a valid license for Google's open source projects.

Also the 'log-driver' package is whitelisted.

Additonal minor changes:
- add the `version` field: this is a required field
- fix the `private` field: this is a boolean field. "true" --> true.
- package-lock.json is committed.

### DIFF
--- a/js-green-licenses.json
+++ b/js-green-licenses.json
@@ -1,0 +1,5 @@
+{
+  "packageWhitelist": [
+    "log-driver"
+  ]
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7557 @@
+{
+  "name": "google-cloud-node",
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "JSV": {
+      "version": "4.0.2",
+      "resolved": "https://registry.npmjs.org/JSV/-/JSV-4.0.2.tgz",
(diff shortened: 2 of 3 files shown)


jinwoo:
The license checking fails due to `jsonify`. The issue is tracked by https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2786.

jinwoo:
Huh? All CI tests passed, which is unexpected. They should've failed in the `posttest` phase.

Looking at the test setup, it looks like none of the tests are actually run. Test scripts just run lint and the build script. Is this intended?

jinwoo:
@stephenplusplus Can you check the CI setup? None of the tests seem to run on CI.

stephenplusplus:
Hey @jinwoo, we just branched common-grpc out into its own repo: https://github.com/googleapis/nodejs-common-grpc. Can you add the license checking changes there? I'm not sure what the fate of this repo will be, but that was the last bit of code we had here.

jinwoo:
Got it. Branching out common-grpc seems like the way to go. Thanks.

PR #2797, opened 2018-01-24
Replace req.url with req.originalUrl in Express extractor

koistya:


stephenplusplus:
Sorry for the redirect, but this code now lives on https://github.com/googleapis/nodejs-error-reporting.

PR #2831, opened 2018-06-21
Deprecate google-cloud package

crwilcox:
The package hasn't been updated in months and users are running into install issues because of it. 

1) Update package.json to have a preinstall script that fails on latest version and informs user to install API specific packages
2) publish as 0.58.0
3) mark all other versions as deprecated with same message so pinned users get warnings about the old package.

Resolves #2892


packages/google-cloud/package.json
```diff
@@ -133,7 +115,8 @@
   },
   "scripts": {
     "publish-module": "node ../../scripts/publish.js google-cloud",
-    "test": "mocha test/*.js"
+    "test": "mocha test/*.js",
```
JustinBeckwith: Do we need to keep the test?

scripts/preinstall.js
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+'use strict';
+
+console.error('The google-cloud package has been deprecated. For a list of ' + 
+            'Google Cloud Platform API specific packages please visit ' + 
+            'https://cloud.google.com/nodejs/docs/reference/libraries');
+            
+process.exit(-1);
```
JustinBeckwith: nit, add a new line

scripts/preinstall.js
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+console.error('The google-cloud package has been deprecated. For a list of ' +
```
JustinBeckwith: After the first sentence: 

> We strongly recommend installing individual API packages, such as `@google-cloud/storage`.

JustinBeckwith:
FYI @GoogleCloudPlatform/node-team @ofrobots

packages/google-cloud/package.json
crwilcox: I think I can delete the tests. I just realized if I keep them I need to also keep the dependencies. It likely makes sense to remove all of it.

PR #2934, opened 2019-06-10
feat: add command for generating README.md

### LINKED ISSUE #2818: Link out to each github repo hosting the specific client lib
This repo when mentioning "Spanner Nodejs" should have a link to where the code lives as well as a link back to the google cloud webpage.

bcoe:
Add tool for generating README so that it's easier to keep up-to-date, and so we can use this as the landing page from cloud.google.com when linking to all Node.js libraries.

fixes #2818

### DIFF
--- a/bin/generate-readme.js
+++ b/bin/generate-readme.js
@@ -70,7 +70,9 @@ async function collectRepoMetadata (repos) {
 function generateReadme (repoMetadata) {
   const template = readFileSync('./bin/README.mustache', 'utf8');
   const libraries = [];
-  // partition libraries by release type.
+
+  // filter libraries to only contain those with Google Cloud api_id,
+  // standardizing naming along the way.
   Object.keys(repoMetadata).forEach(key => {
     const metadata = repoMetadata[key];
     if (metadata.api_id) {


bin/generate-readme.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// grab files using GitHub API.
+async function getContent (owner, repo, path) {
+  const headers = {}
+  if (argv.token) headers.authorization = `token ${argv.token}`
+  const resp = await fetch(`${apiUrl}/repos/${owner}/${repo}/contents/${path}`, {
+    headers: headers
+  })
+    .then(res => {
```
JustinBeckwith: Mixing promises and async functions feels weird.  Can we just use `await` here?

bin/generate-readme.js
```diff
@@ -0,0 +1,107 @@
...
+      if (res.status !== 200) {
```
JustinBeckwith: If you used `gaxios` it would just work this way 😛

bin/generate-readme.js
```diff
@@ -0,0 +1,107 @@
...
+        const err = Error(`unexpected status = ${res.status}`)
+        err.status = res.status
+        throw err
+      }
+      return res.json()
+    })
+  return JSON.parse(Buffer.from(resp.content, 'base64').toString('utf8'))
```
JustinBeckwith: Can we use 2 lines here to make it more readable?

package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "google-cloud-node",
+  "version": "1.0.0",
```
JustinBeckwith: Could we lose the `version`, and add a `private: true` in here?

bcoe:
doesn't automate things yet, but at least makes generating a new README.md a single command.

package.json
```diff
@@ -0,0 +1,39 @@
...
+  "description": "Google Cloud client libraries for Node.js",
+  "main": "bin/generate-readme.js",
```
JustinBeckwith: This is going to confuse people.  I don't expect anyone to install this.  I'd much rather just have an `npm run` command

package.json
```diff
@@ -0,0 +1,39 @@
...
+  "bin": "bin/generate-readme.js",
+  "directories": {
+    "doc": "docs"
+  },
+  "scripts": {
+    "fix": "standard --fix",
```
JustinBeckwith: Eugh. I see the appeal here, but it is inconsistent with literally every other module we have.  I'll let it pass if you use `semistandard` instead.

bin/generate-readme.js
bcoe: wasn't sure when to reach for teeny-request or gaxios ... so reached for `node-fetch` :stuck_out_tongue:

PR #3000, opened 2020-04-24
build: fix synthfile

JustinBeckwith:


synth.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 """This script is used to synthesize generated parts of this library."""
 
+import subprocess
 
-node.install()
-node.fix()
+subprocess.run(['npm', 'install'])
```
alexander-fenster, line 18: What didn't work?

alexander-fenster: Oh, I see. The proper fix is `import synthtool.languages.node as node`.

PR #3007, opened 2020-06-15
docs: generate libraries.json to provide index

JustinBeckwith:


JustinBeckwith:
Exactly!  I am going to write a little nightly synth script that reads this file, and uses it to identify GAPIC/Veneers that match based on API name.  If it matches up, I add the warning we chatted about to the README.

PR #3008, opened 2020-06-16
docs: add "support_documentation" to libraries.json

stephenplusplus:
RE: https://github.com/googleapis/synthtool/pull/582

This will add a best-effort link to each product's "Getting Support" page. If it can't be found, it will re-use the "product_documentation" link.


bcoe:
I like this idea, let's try it.

Issue #3023, opened 2020-07-27
Why are we deleting these libraries from libraries.json?


sofisl:
PR #3022 seems to be deleting some important libraries from the json file. It may be an issue with this query: https://github.com/googleapis/google-cloud-node/blob/master/bin/generate-readme.js#L126.

We should investigate to see what is going on with that document generation.

JustinBeckwith:
So I have a suspect here, and it is me.  Looking at this code:
https://github.com/googleapis/google-cloud-node/blob/master/bin/generate-readme.js#L127

That's what is responsible for regenerating this file, and it's hooked into synthtool.  We should run that code, and dump a log of all found libraries. Now when I wrote that code, what I *didn't* know is that it's possible for the search API to not be entirely exhaustive: 
https://docs.github.com/en/rest/reference/search#timeouts-and-incomplete-results

If it turns out that the `incomplete_results` property is getting set to `true`, we need to re-work this code to:
- fetch all repositories in the organization, using pagination ([example](https://docs.github.com/en/rest/reference/repos#list-organization-repositories))
- perform a filter on the combined results, removing archived, non-TypeScript/non-JavaScript repos, archived repos, etc

JustinBeckwith:
I think you fixed this :)

PR #3024, opened 2020-07-31
fix: add incomplete results view

sofisl:
This PR addresses #3023 
I ran the code and it generated all the libraries correctly - I don't know why PR #3022 is showing up wrong. However, I think it's still probably a good idea to add a 'just-in-case' clause to the library generation in case there are incomplete results.

### DIFF
--- a/bin/generate-readme.js
+++ b/bin/generate-readme.js
@@ -130,7 +130,7 @@ async function getRepos () {
   const repos = [];
   while (url) {
     const res = await github.request({ url: url.href });
-    repos.push(...res.data.filter(r => (r.language === 'TypeScript' || r.language === 'JavaScript') && r.archived === false).map(r => r.full_name));
+    repos.push(...res.data.filter(r => (r.language === 'TypeScript' || r.language === 'JavaScript') && r.archived === false && r.private === false).map(r => r.full_name));
     url = null;
     if (res.headers['link']) {
       const link = parseLinkHeader(res.headers['link']);


bin/generate-readme.js
```diff
@@ -132,6 +132,30 @@ async function getRepos () {
   while (url) {
     const res = await github.request({ url: url.href });
     repos.push(...res.data.items.map(r => r.full_name));
+    if (res.incompleteResults === true) {
```
JustinBeckwith, line 135: So, I think I did a bad job of explaining what's up here.  What we've learned recently is that the Search API can never be trusted to return complete results :/  As a result, it's a bad fit for cases where complete results are a requirement.  

I think we specifically need to query this endpoint, using pagination, to build a full list of available repositories:
https://docs.github.com/en/rest/reference/repos#list-organization-repositories

It's also possible I'm misreading the "incomplete results" docs

bin/generate-readme.js
```diff
@@ -124,14 +124,13 @@ async function generateReadme (repoMetadata) {
...
-  let url = new URL('/search/repositories', baseUrl);
-  url.searchParams.set('q', q);
+  let url = new URL('/orgs/googleapis/repos', baseUrl);
+  url.searchParams.set('type', 'public');
   url.searchParams.set('per_page', 100);
   const repos = [];
   while (url) {
     const res = await github.request({ url: url.href });
-    repos.push(...res.data.items.map(r => r.full_name));
+    repos.push(...res.data.filter(r => (r.language === 'TypeScript' || r.language === 'JavaScript') && r.archived === false).map(r => r.full_name));
```
JustinBeckwith, line 133: I think we want to limit this to `r.private === false` as well.  I'm a little pleasantly surprised `r.language` was right, given the acrobatics I had to go through in the repo-automation-tools repository.

PR #3087, opened 2021-04-20
docs: de-emphasize downloading service account keys

sethvargo:

### DIFF
--- a/bin/README.mustache
+++ b/bin/README.mustache
@@ -39,6 +39,8 @@ If all of the above are impossible, you can
 [download a service account JSON file](https://cloud.google.com/iam/docs/creating-managing-service-account-keys), but this may be
 forbidden by your organization's policies.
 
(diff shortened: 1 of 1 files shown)


sofisl:
@sethvargo, this seems like a pretty substantial change in direction. Is there guidance coming from elsewhere to make this change? Otherwise, perhaps we might want to talk it over with other Node owners (apologies if I missed a convo/message about this).

sethvargo:
@sofisl in general, downloading service account JSON keys should be the **last** choice. Check the yoshi chat.

JustinBeckwith:
We did have a discussion on this today in the team chat, and decided we needed to address this in a more holistic way.  This may turn out to be the answer (I hope it is!) but we have to address some of the key issues first, and then make sure we're tracking all the places this would have to be changed.  This is but one of 100 places we would need to update.

tmatsuo:
Let's close this for now.
(We can always reopen)

PR #3117, opened 2021-07-08
build: migrate to ESM

bcoe:
_Accidentally landed a few ESM deps, because we didn't have any unit tests for generation. Given this._

1. I've migrated the library to work with ESM (_this is a good low risk repo to play with._).
2. I've added a smoke test so that this doesn't happen again as easily.

### DIFF
--- a/bin/generate-readme.mjs
+++ b/bin/generate-readme.mjs
@@ -22,15 +22,14 @@ import parseLinkHeader from 'parse-link-header';
 const token = process.env.GITHUB_TOKEN;
 const smokeTest = !!process.env.SMOKE_TEST;
 
-if (!token) {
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,26 @@
...
+name: ci
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [14]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
```
JustinBeckwith, line 15: I think there's a setup-node v2 right?

JustinBeckwith (approved):

PR #3174, opened 2022-05-19
build: create new templating engine to bootstrap Node mono-repos

sofisl:
Should call this a feat, but not sure if it will release a new version.

### DIFF
--- a/containers/node-bootstrap-container/tests/helper-test.js
+++ b/containers/node-bootstrap-container/tests/helper-test.js
@@ -50,5 +50,11 @@ describe('helper functions', () => {
     
         await createGhIssue.createGHIssue(octokit);
         scope.done();
+
+        delete process.env.MONO_REPO_NAME;
+        delete process.env.API_ID;
+        delete process.env.PROJECT_ID;
+        delete process.env.BUILD_ID;
+        delete process.env.GITHUB_TOKEN;
       });
 });
--- a/containers/node-bootstrap-container/package.json
+++ b/containers/node-bootstrap-container/package.json
@@ -17,8 +17,7 @@
   },
   "homepage": "https://github.com/googleapis/google-cloud-node#readme",
   "dependencies": {
-    "@octokit/rest": "^18.12.0",
-    "octokit": "^1.7.2"
+    "@octokit/rest": "^18.12.0"
   },
   "devDependencies": {
     "@octokit/rest": "^18.12.0",


bcoe (approved):

sofisl:
Hi @bcoe! Would you mind re-reviewing this PR? I added the container and more tests. Thanks!

.github/workflows/ci.yaml
```diff
@@ -24,3 +24,23 @@ jobs:
...
+  unit-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [14]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
```
bcoe, line 36: I'm glad you got initial testing in place 🎉 

Would like to work with @chingor13 to make our test runner smart about selectively running against only folders that have changed, as follow up work.

containers/node-bootstrap-container/entrypoint.sh
```diff
@@ -0,0 +1,28 @@
...
+WORKSPACE_DIR="/workspace"
+cd  $WORKSPACE_DIR/$MONO_REPO_NAME/packages/gapic-node-templating
+npm ci
+npm run compile
+npm link .
+
+cd $WORKSPACE_DIR/$MONO_REPO_NAME/containers/node-bootstrap-container
+FOLDER_NAME=$(node create-folder-name.js $API_ID) || node create-gh-issue.js
+mkdir -p "$WORKSPACE_DIR/$MONO_REPO_NAME/packages/$FOLDER_NAME" || node create-gh-issue.js
+cd  $WORKSPACE_DIR/$MONO_REPO_NAME
```
bcoe, line 27: Did you ever figure out why this wasn't liking absolute paths?

containers/node-bootstrap-container/tests/helper-test.js
```diff
@@ -0,0 +1,54 @@
...
+    it('should return the API path without the version, with dashes', async () => {
+      assert.deepStrictEqual(cp.execSync(
+        `node ${createFolderNamePath} 'google.cloud.kms.v1'`).toString('utf-8'),
+        'google-cloud-kms\n'
+      );
+    });
+
+  
+      it('should open an issue on a given repo', async () => {
+       process.env.MONO_REPO_NAME = 'google-cloud-node';
```
bcoe, line 38: it would be good to cleanup environment variable after the test run, so that the behavior doesn't carry over to the next test, I believe you could do something like:

`sinon.stub(process.env, 'FOO').value('bar');`

bcoe (approved):

PR #3204, opened 2022-06-28
build: add deploy script for node-bootstrap-container

sofisl:


chingor13 (approved):

bcoe (approved):
Looks reasonable to me, I assume we'll only run the build process when the `owlbot-bootstrap-prod` folder changes?

sofisl:
@bcoe, I haven't created the trigger yet but I will specify it should run when we change the `node-bootstrap-container` folder

PR #3206, opened 2022-06-30
build: add owlbot.yaml path as env variable

sofisl:
See: https://github.com/googleapis/repo-automation-bots/pull/4033

### DIFF
--- a/containers/node-bootstrap-container/add-to-well-known-path.js
+++ b/containers/node-bootstrap-container/add-to-well-known-path.js
@@ -1,3 +1,17 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 const fs = require('fs');
 function getApiPathWithDashes(args) {
     const [workdir, folderName] = args


containers/node-bootstrap-container/entrypoint.sh
```diff
@@ -25,4 +25,5 @@ cd $WORKSPACE_DIR/$MONO_REPO_NAME/containers/node-bootstrap-container
 FOLDER_NAME=$(node create-folder-name.js $API_ID) || node create-gh-issue.js
 mkdir -p "$WORKSPACE_DIR/$MONO_REPO_NAME/packages/$FOLDER_NAME" || node create-gh-issue.js
 cd  $WORKSPACE_DIR/$MONO_REPO_NAME
+_OWLBOT_YAML_PATH="packages/$FOLDER_NAME"
```
chingor13, line 28:
```suggestion
export _OWLBOT_YAML_PATH="packages/$FOLDER_NAME"
```
You probably need to export this env otherwise it's only a local variable and not passed to following commands

chingor13 (changes requested):

containers/node-bootstrap-container/entrypoint.sh
```diff
@@ -25,4 +25,5 @@ cd $WORKSPACE_DIR/$MONO_REPO_NAME/containers/node-bootstrap-container
 FOLDER_NAME=$(node create-folder-name.js $API_ID) || node create-gh-issue.js
 mkdir -p "$WORKSPACE_DIR/$MONO_REPO_NAME/packages/$FOLDER_NAME" || node create-gh-issue.js
 cd  $WORKSPACE_DIR/$MONO_REPO_NAME
+export _OWLBOT_YAML_PATH="packages/$FOLDER_NAME/.OwlBot.yaml"
```
chingor13, line 28: Where is this env var being used? Should it be an argument to the gapic-node-templating CLI instead?

containers/node-bootstrap-container/entrypoint.sh
```diff
@@ -24,5 +24,6 @@ npm link .
 cd $WORKSPACE_DIR/$MONO_REPO_NAME/containers/node-bootstrap-container
 FOLDER_NAME=$(node create-folder-name.js $API_ID) || node create-gh-issue.js
 mkdir -p "$WORKSPACE_DIR/$MONO_REPO_NAME/packages/$FOLDER_NAME" || node create-gh-issue.js
+echo "packages/$FOLDER_NAME/.OwlBot.yaml" > $WORKSPACE_DIR/owlbotYamlPath.md
```
chingor13, line 27: nit: I wouldn't make it a markdown extension - it's just a text file so probably `.txt` or perhaps even exclude an extension

chingor13: We should also consider a single metadata file with a defined schema: https://github.com/googleapis/repo-automation-bots/pull/4033#pullrequestreview-1025409428

sofisl: done!

chingor13 (approved):

PR #3236, opened 2022-07-22
fix: path in .OwlBot.yaml path for new directory structure, and feat: add default_version to .repo-metadata.json

sofisl:
Also cleaning up GH token that was not being used.


containers/node-bootstrap-container/create-gh-issue.js
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 const {Octokit} = require('@octokit/rest');
-const octokit = new Octokit({auth: process.env.GITHUB_TOKEN});
```
bcoe, line 16: Is this intentionally removed?

bcoe (approved):

PR #3254, opened 2022-08-05
build: add samples test triggers

chingor13:
Similar to #3198

We still cannot create these triggers via terraform because the internal provider does not yet support the `    commentControl: COMMENTS_ENABLED_FOR_EXTERNAL_CONTRIBUTORS_ONLY` option

### DIFF
--- a/packages/typeless-sample-bot/package.json
+++ b/packages/typeless-sample-bot/package.json
@@ -18,6 +18,7 @@
     "test": "c8 mocha build/test --recursive",
     "snapshots-update": "SNAPSHOT_UPDATE=1 npm test",
     "pretest": "npm run compile",
+    "samples-test": "echo no system tests yet 🙀",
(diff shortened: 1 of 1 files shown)


ci/export/system-continuous-node14.yaml
```diff
@@ -11,5 +11,5 @@ name: system-continuous-node14
 serviceAccount: projects/long-door-651/serviceAccounts/kokoro-system-test@long-door-651.iam.gserviceaccount.com
 substitutions:
   _BUILD_TYPE: continuous
-  _TEST_TYPE: system
   _LOGS_BUCKET: cloud-node-public-logs
+  _TEST_TYPE: system
```
sofisl, line 15: should this be samples?

sofisl (approved):

bcoe (approved):

PR #3269, opened 2022-08-11
feat: add better template for samples quickstart test

sofisl:
This feature, alongside copying the quickstart sample, will make it so that we won't have to do any tweaking to samples tests to make it run.


bcoe (approved):
This is awesome 😄

PR #3288, opened 2022-08-22
build: add script to import split repo code

chingor13:
Example usage:

`Usage: $0 <source-repo> <target-repo> <source-path> <target-path> [comma,separated,folders,to,ignore] [files,to,keep] [branch-name]`

```bash
./bin/migrate-history.sh \
  googleapis/nodejs-asset \
  googleapis/google-cloud-node \
  "" \
  packages/google-cloud-asset \
  .kokoro,.github \
  .github/.OwlBot.yaml
```

creates: https://github.com/googleapis/google-cloud-node/pull/3287

Also, if you set the `UPDATE_SCRIPT` env var, we will run that script before opening the pull request in case you need to do any post-processing.


sofisl:
@chingor13, I think it would be great if we could import these repos with:

1. A change to the .Owlbot.yaml file to the monorepo style,
2. A change to the package.json to get the samples and system tests to pass correctly.

chingor13:
@sofisl I added the ability to specify an extra update script via an env variable. This keeps the script generic - it can be used to migrate git history from any repo to another (remapping directories as well).

sofisl (approved):

PR #3298, opened 2022-08-26
build: add post-process script for migrating split repo

chingor13:
Adds a post-process script for cleaning up the owlbot and release-please config files.

Also adds a wrapper script for the generic `migrate-git-history` script that is used specifically for migrating the nodejs split repos into google-cloud-node.

Example commit: https://github.com/googleapis/google-cloud-node/pull/3287/commits/840c9497ec2cfad23055cbe4b1e6ed4649bf9fa1


bin/split-repo-post-process.sh
```diff
@@ -0,0 +1,46 @@
...
+
+echo "Moving ${PACKAGE_PATH}/.github/.OwlBot.yaml"
+mv "${PACKAGE_PATH}/.github/.OwlBot.yaml" "${PACKAGE_PATH}/.OwlBot.yaml"
+
+echo "Fixing format of ${PACKAGE_PATH}/.OwlBot.yaml"
+# remove `docker:` line
+sed -i "/docker:/d" "${PACKAGE_PATH}/.OwlBot.yaml"
+# remove `image:` line
+sed -i "/image:/d" "${PACKAGE_PATH}/.OwlBot.yaml"
+sed -i "s/dest: \/owl-bot-staging/dest: \/owl-bot-staging\/${PACKAGE_NAME}/" "${PACKAGE_PATH}/.OwlBot.yaml"
```
sofisl, line 42: From looking at the example commit, this should only have one regex grouping: 

`/owl-bot-staging/google-cloud-asset/$1` instead of `/owl-bot-staging/google-cloud-asset/$1/$2`

Also, the source should also be changed to have one regex grouping, from this:

`/google/cloud/asset/(.*)/.*-nodejs/(.*)` to this `/google/cloud/asset/(.*)/.*-nodejs`

sofisl (changes requested):

sofisl:
Also @chingor13 , would you mind creating a tracking bug for all the things we'll need to change to make the samples & systems tests pass for migrated libraries, if we're not doing it in this script?

bin/split-repo-post-process.sh
chingor13: This should be fine -- it's copying each subfolder

sofisl: I'm a bit concerned because the post processor does some extra processing to bundle the versions correctly, and I'm sure the other way works (since it's how we have it currently, which would provide consistency).

chingor13: I updated it to try to match the .OwlBot.yaml style we currently have in this repo.

chingor13:
> if we're not doing it in this script?

The script is now updating the samples/system test scripts and those tests should now pass.

sofisl (approved):

PR #3334, opened 2022-09-09
build: enable sentence-case plugin for release notes

chingor13:
This is used for standardizing release notes.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -19,6 +19,6 @@
     {
       "type": "sentence-case"
     }
-  ]
+  ],
   "release-type": "node"
 }


sofisl (approved):

release-please-config.json
```diff
@@ -15,5 +15,10 @@
     "packages/typeless-sample-bot": {},
     "packages/google-cloud-bigquery-analyticshub": {}
   },
+  "plugins": [
+    {
+      "type": "sentence-case"
+    }
+  ]
```
bcoe, line 22:
```suggestion
  ],
```

bcoe:
@chingor13 one thing I noticed in recent week's release notes is that the plugin isn't working for go, with this configuration:

https://github.com/googleapis/google-cloud-go/blob/main/spanner/release-please-config.json#L11

☝️ should the plugins work as a `string`, or should they be an object with the key `type`?

chingor13:
> should the plugins work as a `string`, or should they be an object with the key `type`?

Either way should work. Originally, plugins were not configurable, so they could be loaded with just their name. Then we needed the ability to configure plugins, so we have an option where you provide a JSON object where the `type` field is the name of the plugin.

PR #3367, opened 2022-09-17
fix: preserve default values in x-goog-request-params header for artifact registry

sofisl:
Had to run owlbot separately for artifact registry, since it was being migrated during this update.

@alexander-fenster, do these changes look right?


sofisl:
Fixes https://github.com/googleapis/google-cloud-node/pull/3366

PR #3377, opened 2022-09-28
build: add update split repo script to update the README and delete the rest of the repo

sofisl:
According to agreement with DEE org, since the split repo will stay un-archived during the migration, we'll need to delete everything but the samples, while updating the README.md. We'll also need to add or replace the owlbot.py entirely to ignore the README.


bin/update-split-repo.sh
```diff
@@ -0,0 +1,85 @@
...
+  exit 1
+fi
+
+# repo name (e.g. nodejs-asset)
+SPLIT_REPO=$1
+# destination directory (e.g. google-cloud-asset)
+ARTIFACT_NAME=$2
+
+rm -rf /tmp/${SPLIT_REPO}
+git clone git@github.com:googleapis/${SPLIT_REPO}.git /tmp/${SPLIT_REPO}
```
chingor13, line 30: Whenever doing string interpolation in bash, be sure to quote the values so you cannot inject commands.
```suggestion
git clone "git@github.com:googleapis/${SPLIT_REPO}.git" "/tmp/${SPLIT_REPO}"
```

bin/update-split-repo.sh
```diff
@@ -0,0 +1,85 @@
...
+git checkout -b 'removeOwlBotAndUpdateREADME'
+
+shopt -s extglob
+
+# Delete everything first that's not the samples, README.md, and git folder
+rm -rf !("samples"|"README.md"|".git") 
+rm -rf /tmp/${SPLIT_REPO}/.[!.git]*
+rm -rf /tmp/${SPLIT_REPO}/.github
+
+IGNORE_README_TXT="# Copyright 2021 Google LLC
```
chingor13, line 43: 2022?

bin/migrate-split-repo.sh
```diff
@@ -41,3 +41,6 @@ ${SCRIPT_DIR}/migrate-git-history.sh \
   "${PACKAGE_PATH}" \
```
chingor13, line 41: It looks like this file's mode is changing to not be executable?

chingor13 (changes requested):

bin/migrate-split-repo.sh
sofisl: Fixed it, thanks!

sofisl:
@chingor13 thanks for the review, PTAL

bin/update-split-repo.sh
```diff
@@ -0,0 +1,85 @@
...
+  echo "Usage: $0 <split-repo-name> <target-path>"
+  exit 1
+fi
+
+# repo name (e.g. nodejs-asset)
+SPLIT_REPO=$1
+# destination directory (e.g. google-cloud-asset)
+ARTIFACT_NAME=$2
+
+rm -rf /tmp/${SPLIT_REPO}
```
chingor13, line 29: All of these string interpolations need quotes. I didn't mark all of them because there were so many.

chingor13 (changes requested):

chingor13 (approved):

PR #3398, opened 2022-10-11
build: update what gets deleted in Kokoro file during migration, to make sure samples tests continue to pass

sofisl:


bin/delete-everything-split-repo.sh
```diff
@@ -36,9 +36,18 @@ git checkout -b 'deleteMostOfLibrary'
 shopt -s extglob
 
 # Delete everything first that's not the samples, README.md, .kokoro folder, and git folder
-rm -rf !("samples"|"README.md"|".git"|".kokoro") 
-rm -rf "/tmp/${SPLIT_REPO}/.[!.git|!.kokoro]*"
+rm -rf !("samples"|"README.md"|".git")
```
alexander-fenster, line 39: Does this syntax actually work? The way I would do it is

```sh
ls | grep -v 'samples\|README.md\|.git' | xargs rm -f
```

bin/delete-everything-split-repo.sh
```diff
@@ -57,13 +66,79 @@ IGNORE_README_TXT="# Copyright 2022 Google LLC
...
 node.owlbot_main(templates_excludes=[
-README.md'
+'README.md'
 ])"
 
 OWLBOT_PY="/tmp/${SPLIT_REPO}/owlbot.py"
 
 echo "$IGNORE_README_TXT" >> "$OWLBOT_PY"
 
+SAMPLES_CFG_TXT="#!/bin/bash
```
alexander-fenster, line 76: Nit: here-document syntax is useful for this kind of things, you won't need to quote all quotes and dollar signs if you just do

```
SAMPLES_CFG_TXT=<<'EOD'
# Copyright 2018 Google LLC
..... (unquoted content here)
EOD
```

bin/delete-everything-split-repo.sh
sofisl: Wouldn't this say delete samples, README, and .git? The syntax I opted for should delete everything but those. I think it does work! See: https://github.com/googleapis/nodejs-language/pull/701

sofisl: I'm still interested in bash magic though, if you have another way please share :)

alexander-fenster: I mean, I have never seen this syntax, and it did not work for me when I tried, that's why I asked :)

PR #3461, opened 2022-10-27
fix: typeless bot updates for a fully functioning packaged version

feywind:
This updates the typeless sample bot to fix some issues discovered when trying to integrate with OwlBot.


bcoe (approved):

packages/typeless-sample-bot/src/bin/typeless-sample-bot.ts
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env node
```
bcoe, line 1: Adding a shebang and `bin` the main differences?

feywind: Yeah, and there are some weirdnesses around module naming/paths and ESM, so I went ahead and separated it out with the right name.

PR #3463, opened 2022-10-28
fix: update typeless bot build to compile before pack

feywind:
It looks like I was missing a "prepare" line in package.json, so the package was missing the actual code.


feywind:
I tested against Node 14 and 15, and it looks fine, so I moved the engines requirement down to 14 as well.

bcoe (approved):

PR #3465, opened 2022-10-31
fix: typeless bot packaging wasn't including source/build files

feywind:
This fixes the build.


feywind:
Apparently this ended up being merged last week >_>

PR #3474, opened 2022-11-02
build: fix readme generation, do not fail if product documentation does not exist

### LINKED ISSUE #3387: libraries.json missing repo metadata update
The `libraries.json` file is missing metadata for `@google-cloud/iam`, so it may not be updating to reflect all client library additions/updates.

sofisl:
Fixes #3387

### DIFF
--- a/bin/generate-readme.mjs
+++ b/bin/generate-readme.mjs
@@ -152,7 +152,7 @@ async function processMetadata (repoMetadata) {
         supportDocsUrl = 'https://cloud.google.com/stackdriver/docs/getting-support';
       }
 
-      if (!supportDocsUrl.match(/https/)) {
+      if (!supportDocsUrl.match(/^https/)) {
         supportDocsUrl = `https://${supportDocsUrl}`
       }
       


bin/generate-readme.mjs
```diff
@@ -152,13 +152,23 @@ async function processMetadata (repoMetadata) {
...
-        validateStatus: () => true
-      });
-      const remoteUrlExists = res.status !== 404;
+      try {
+        res = await request({
+          url: supportDocsUrl,
+          method: 'HEAD',
+          validateStatus: () => true
+        });
+      } catch (err) {
```
bcoe, line 168: I would look at the error `code` and only set `remoteUrlExists = false` if the error returned is a 404 (indicating that it doesn't exist).

This way we don't mess up the README.md if there happen to be spurious 500s.

bcoe (changes requested):

bin/generate-readme.mjs
```diff
@@ -152,13 +152,25 @@ async function processMetadata (repoMetadata) {
         supportDocsUrl = 'https://cloud.google.com/stackdriver/docs/getting-support';
       }
 
+      if (!supportDocsUrl.match(/https/)) {
```
alexander-fenster, line 155: Let's do `/^https:/` to avoid possible matches in the middle of the string.

alexander-fenster (approved):

PR #3677, opened 2022-11-22
build: grab latest version of cloudrad

sofisl:


.kokoro/release/docs-devsite-single.sh
```diff
@@ -21,4 +21,4 @@ set -eo pipefail
 npm install
 
 # publish docs to devsite
-npx @google-cloud/cloud-rad@0.2.5
+npx @google-cloud/cloud-rad
```
bcoe, line 24:
```suggestion
npx @google-cloud/cloud-rad@0.2.18
```

bcoe (changes requested):
I think it would still be good to pin to a version of `@google-cloud/cloud-rad`, otherwise if there are major changes to cloud-rad in the future our builds will break.

.kokoro/release/docs-devsite-single.sh
dansaadati: Just published latest `npx @google-cloud/cloud-rad@0.2.18`

bcoe (approved):

PR #3739, opened 2022-12-13
docs: exclude sample files from doc generation

dansaadati:
This excludes samples under a package from having docs generated.


sofisl:
merged into #3737

PR #3818, opened 2022-12-22
build: add presubmit unit tests node 16 and 18

### LINKED ISSUE #3496: need to test on all LTS versions
We want to run unit tests on all current LTS versions of Node.js - that is, 14, 16, 18.  12 is optional but nice to have because we claim we still support it.

sofisl:
Fixes #3496

I created these triggers separately since I didn't want to run samples, unit, lint and system for all the versions, just presubmit unit tests for 16 and 18.

I also had to increase the timeout of the other tests since they've been timing out recently when touching all files, which this PR does since it touches the CI.

Note: some tests need to be refactored since they actually fail Node 18.

### DIFF
--- a/packages/google-cloud-dataplex/.jsdoc.js
+++ b/packages/google-cloud-dataplex/.jsdoc.js
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2023 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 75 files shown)


bcoe (approved):

packages/gapic-node-templating/test/get-bootstrap-template-vars.test.ts
```diff
@@ -108,22 +104,17 @@ describe('get bootstrap template vars', () => {
...
     it('should get the distribution name', async () => {
-      const fileRequest = nock('https://api.github.com')
-        .get(
-          '/repos/googleapis/googleapis/contents/google%2Fcloud%2Fkms%2Fv1%2FBUILD.bazel'
-        )
-        .reply(200, {
-          name: 'BUILD.bazel',
-          content: 'content',
-        });
+      const getContentStub = sinon
```
bcoe, line 107: One other non blocking nit, what was the reason for switching from nock to sinon here?

PR #3839, opened 2023-01-10
chore: minor updates to migration script

sofisl:


bcoe (approved):
LGTM.

nit: would help me review if the minor updates were summarized.

PR #3912, opened 2023-01-28
build: owlbot audit 8

sofisl:


bcoe:
speech was the cause of this issue, I've fixed its OwlBot.py.

PR #3976, opened 2023-02-09
build: minify protos.json in the build directory

alexander-fenster:
Slightly reduces the size of the unpacked files (from 58MB to 55MB). We do it for some other packages (e.g. Firestore) and we need to start doing it for all packages.


sofisl (approved):

sofisl:
cool

PR #4189, opened 2023-04-19
feat!: move @google-cloud/translate to ESM and Node 14

sofisl:
A few notes on this move:

1. A special `post-compile.js` file was required to modify compiled commonJS code after transpiling from Typescript, since some of the code won't compile if the `module` type is set in the package.json. This post-compile script is a hacky proof-of-concept, but captures some important steps, for example:
    - Replacing instances of `import.meta` with `__dirname`
    - Using proxyquire instead of esmock
    - Renaming .js files to .cjs (in order for them to be interpreted as common JS)

If the idea of this script looks OK, I can clean it up.
  
2. Translate seems to be an outlier in that it uses the `protos.js` file in its tests, which should not be a load-bearing file. Protos.js uses [UMD](https://stackoverflow.com/questions/60365052/why-is-functionglobal-factory-used-in-so-many-js-libraries) modules. I created a `protos-hack.js` file that outlines what this file would need to look like to be used in ESM. Perhaps we can change what this file would look like as its generated in gax (@alexander-fenster to weigh in).

3. Currently, we're only running unit tests on ESM files, and we're still using CJS to run sample & system tests. If we wanted to run system-tests on ESM, we could do so by providing options [here](https://github.com/google/pack-n-play/blob/705e581f30b839135ee5c9526fd379ca0bd61e09/src/pack-n-test.ts#L91) and [here](https://github.com/google/pack-n-play/blob/705e581f30b839135ee5c9526fd379ca0bd61e09/src/pack-n-test.ts#L122) to install it in an ESM package.

~~4. Since a .cjs file cannot just import an ESM file, we do need to copy over the protos files into each subdirectory (and rename them to .cjs in that directory), since we use it in our [`client.ts` file](https://github.com/googleapis/gapic-generator-typescript/blob/327e234bc2bcc54c3e6a672e59ab5242f61f3047/templates/typescript_gapic/src/%24version/%24service_client.ts.njk#L44). Perhaps @alexander-fenster can confirm this. The concern is that this would double our current library size (instead of just 1/3 as we'd hoped).~~ --> nevermind, this is OK for all "regular" gapics. For translate, I think we'd still need #2 since they're using it in tests. Or we could refactor the tests.

5. Subsequently, some gax stuff expects there to be a build/protos file, instead of build/cjs/protos and build/esm/protos. This isn't a big deal since we can just add a directory argument (for example, to minifyProtosJson), but it might be worth considering changing the default behavior.

6. Some templates need to be changed in synthtool from js --> cjs.

7. Node 12 test will fail because this moves to Node 14. Once I see tests passing, I'll skip this module from being tested on Node 12.

### DIFF
--- a/packages/google-cloud-translate/src/v3/translation_service_client.ts
+++ b/packages/google-cloud-translate/src/v3/translation_service_client.ts
@@ -29,7 +29,7 @@ import type {
 } from 'google-gax';
 import * as gax from 'google-gax';
 import {Transform} from 'stream';
-import * as protos from '../../protos/protos.js';
+import type * as protos from '../../protos/protos';
 import fs from 'fs';
 import path from 'path';
 import {fileURLToPath} from 'url';
--- a/packages/google-cloud-translate/src/v3beta1/translation_service_client.ts
+++ b/packages/google-cloud-translate/src/v3beta1/translation_service_client.ts
@@ -29,7 +29,7 @@ import type {
   GaxCall,
 } from 'google-gax';
 import {Transform} from 'stream';
-import * as protos from '../../protos/protos.js';
+import type * as protos from '../../protos/protos';
 import path from 'path';
 import * as translation_service_client_config from './translation_service_client_config.json';
 import fs from 'fs';


packages/google-cloud-translate/package.json
```diff
@@ -79,8 +94,7 @@
     "pack-n-play": "^1.0.0-2",
     "proxyquire": "^2.0.1",
```
ddelgrosso1, line 95: Can this be removed as well?

packages/google-cloud-translate/tsconfig.json
```diff
@@ -2,11 +2,15 @@
...
+    "outDir": "build/cjs",
     "resolveJsonModule": true,
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "sourceMap": false,
+    "module": "CommonJS",
     "lib": [
       "es2018",
-      "dom"
+      "DOM"
```
bcoe, line 13: curious what we actually get from this plugin, I would think that DOM would be mostly for web code.

packages/google-cloud-translate/post-compile.js
ddelgrosso1, line 1: If we decide that using a post compile step is the best path forward it would probably make sense to extract this to some kind of utility repo so that it is reusable.

packages/google-cloud-translate/package.json
sofisl: I think we need it to run tests for CJS (since it doesn't support esmock)

packages/google-cloud-translate/package.json
```diff
@@ -4,21 +4,33 @@
   "version": "7.2.0",
   "license": "Apache-2.0",
   "author": "Google Inc.",
+  "type": "module",
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=14.18.0"
```
ddelgrosso1, line 9: Was there an issue with 14.0.0?

packages/google-cloud-translate/tsconfig.json
sofisl: We get the definition for window here:
https://github.com/googleapis/google-cloud-node/blob/b9cd90cc4e8eaf732fde676198d8a694e7df200b/packages/google-cloud-translate/src/v3/translation_service_client.ts#L123

packages/google-cloud-translate/package.json
sofisl: good call, just using it while I was testing.

packages/google-cloud-translate/package.json
```diff
@@ -37,9 +49,13 @@
     "predocs": "npm run compile",
     "lint": "gts check",
     "samples-test": "npm run compile && cd samples/ && npm link ../ && npm i && npm test",
-    "system-test": "npm run compile && c8 mocha build/system-test",
-    "test": "c8 mocha build/test",
-    "compile": "tsc -p . && cp -r protos build/ && minifyProtoJson",
+    "system-test": "c8 mocha build/cjs/system-test",
+    "test:cjs": "c8 mocha build/cjs/test",
+    "test:esm": "c8 mocha --loader=esmock build/esm/test",
```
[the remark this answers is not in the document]
bcoe: What I think would be best would be to extend pack-n-play to test the ESM build of the package, at this point we could run unit tests against just ESM or CJS.

packages/google-cloud-translate/post-compile.js
```diff
@@ -0,0 +1,69 @@
...
+// ** This file is automatically generated by gapic-generator-typescript. **
+// ** https://github.com/googleapis/gapic-generator-typescript **
+// ** All changes to this file may be overwritten. **
+
+import * as fs from 'fs';
+import * as path from 'path';
+const fsp = fs.promises;
+
+const regexMetaInput =
+  /let dirToUse = '';\ntry {\n {4}dirToUse = __dirname;\n\}\ncatch \(e\) \{\n {4}\/\/ eslint-disable-next-line @typescript-eslint\/ban-ts-comment\n {4}\/\/ @ts-ignore\n {4}dirToUse = import\.meta\.url;\n\}\nconst filename = \(0, url_1\.fileURLToPath\)\(dirToUse\);\nconst dirname = path_1\.default\.dirname\(filename\);/i;
```
bcoe, line 24: This regex seems pretty fragile, I'm betting when we update to a new typescript version at some point it will stop working silently. I'll leave a comment further down in code review.

packages/google-cloud-translate/src/v2/index.ts
```diff
@@ -14,19 +14,62 @@
...
   DecorateRequestOptions,
   BodyResponseCallback,
-} from '@google-cloud/common/build/src/util';
+} from '@google-cloud/common/build/src/util.js';
+import htmlTags from 'html-tags';
+import path from 'path';
+import {fileURLToPath} from 'url';
+let dirToUse = '';
+try {
+  dirToUse = __dirname;
```
bcoe, line 29: I would pull this logic into a separate utility library, perhaps in the generator or gax so that we don't have to add this in every generated library.

I think if we're creative with how this helper is implemented, we should be able to avoid the `regexMetaInput` replacement.

What I did in yargs was to have a different helper file for ESM vs., CJS, during the build step you can then use the appropriate helper. Such that the generated code only has __dirname for CJS and import.meta.url for ESM.

packages/google-cloud-translate/tsconfig.json
bcoe: SGTM!

bcoe (changes requested):
First pass of review. The main thing that jumped out at me is that I think we should pull the path helpers into a module. For the prototype, this could be in the packages/google-cloud-translate module.

I think if you do this refactor, you might be able to eliminate some of the regex post processing you're doing, also it will allow you to delete a lot of code.

bcoe:
Thanks for doing this Sofia 🥳  left a first round of comments.

packages/google-cloud-translate/package.json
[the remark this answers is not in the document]
sofisl: Yep! That's what I suggest in #3 in the PR body!

packages/google-cloud-translate/src/v2/index.ts
sofisl: I agree that the regex is fragile; it's meant to be a working prototype, but I do think regex is ultimately the way to go for the __dirname --> import.meta.url problem. Say, for example, we created a helper file that exports a default function called `dirnameOrImportMeta()`, that lives in gax, for example, that looks like this, and is imported in each file that asks for the __dirname:
```
export default function dirnameOrImportMeta() {
    let dirToUse = '';
    try {
      dirToUse = __dirname;
    } catch (e) {
      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
      // @ts-ignore
      const filename = fileURLToPath(import.meta.url);
      dirToUse = path.dirname(filename);
    }
  
    return dirToUse;
  }
```

This option is fragile in that __dirname would actually be pointing to where this particular file lives, not the file that is calling this function. This could lead to wrong paths. Even if we passed the context of the currently executing directory, such as `process.cwd()`, this also wouldn't necessarily be called from the file that is executing, since we wouldn't know what context that particular file is being run in. 

Another option is to just use this logic in each file that needs it:

```
    let dirToUse = '';
    try {
      dirToUse = __dirname;
    } catch (e) {
      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
      // @ts-ignore
      const filename = fileURLToPath(import.meta.url);
      dirToUse = path.dirname(filename);
    }
  
    return dirToUse;
```
However, CJS tests will complain during runtime:

```
TypeError [ERR_INVALID_URL]: Invalid URL
    at __node_internal_captureLargerStackTrace (node:internal/errors:484:5)
    at new NodeError (node:internal/errors:393:5)
    at URL.onParseError (node:internal/url:565:9)
    at new URL (node:internal/url:645:5)
    at fileURLToPath (node:internal/url:1483:12)
```
Therefore, we'd still need to do some regex replacing to remove the filename part in build/cjs (which is what we're doing in this prototype, hence the fragility).

Another option, we could do something like:

```
const dirname = CHANGE_THIS
```
And change it to be what each file needs in the post-compiling step. However, this option wouldn't allow the file to be runnable (say, for example, from Deno), since it would need to be compiled to be executable.

Lastly, there are libraries that do this transition for you - like `[tsup](https://www.npmjs.com/package/tsup)`, according to this [article](https://antfu.me/posts/publish-esm-and-cjs). The downsides of this alternative are that 1) we'd probably need other custom post-compiling done, like esmock --> proxyquire, and 2) we'd have another dependency that we'd have to maintain and probably does not overlap perfectly with what we need it to do.

PR #4205, opened 2023-04-24
build: make tests run faster

sofisl:
Partially addresses #4140 

- [x] Previous builds are cancelled when new builds are started on the same branch
- [x] MinifyProtos is moved to postpack for publishing, removed from running before test execution; sample & system tests do not compile before running, since the `prepare` script will do that as well
- [x] Switch to using github actions for unit tests and `pnpm` instead of `npm`
- [x] Check returns immediately if it fails

### DIFF
--- a/ci/deletecloudbuild.sh
+++ b/ci/deletecloudbuild.sh
@@ -12,12 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# A script for exporting Cloud Build build triggers.
-
-# `-e` enables the script to automatically fail when a command fails
-# `-o pipefail` sets the exit code to the rightmost comment to exit
-# with a non-zero
-
 now=$(date -u -d "-15 min" +"%Y-%m-%dT%H:%M:%SZ")
 gcloud builds list --format="value(id)" --filter="substitutions.REF_NAME="$REF_NAME" AND status="WORKING" AND create_time<"$now"" | while IFS= read -r line ; do
     echo "gcloud builds cancel "$line""


make-change-to-package.cjs
```diff
@@ -0,0 +1,28 @@
+const fs = require('fs');
+const path = require('path');
+const cp = require('child_process');
+const dirsToSkip = ['gapic-node-templating', 'typeless-sample-bot'];
```
bcoe, line 4: what purpose does this script serve?

packages/google-maps-routing/package.json
```diff
@@ -57,6 +58,7 @@
     "jsdoc-fresh": "^2.0.0",
     "jsdoc-region-tag": "^2.0.0",
     "linkinator": "^4.0.0",
+    "long": "^5.2.3",
```
bcoe, line 61: I don't think we should need to add this `long` dependency, as it's pulled in by google-gax. I'm curious why this was needed.

packages/google-cloud-beyondcorp-clientconnectorservices/samples/test/quickstart.js
sofisl, line 1: commenting these tests out per b/278007126

make-change-to-package.cjs
sofisl: I will delete it! Shouldn't be in here.

packages/google-maps-routing/package.json
sofisl: The way pnpm installs dependencies doesn't grab this dependency. It's really only for [this line](https://github.com/googleapis/google-cloud-node/blob/8db807ae80a695bb1ad4fa290bc9a50ce29a8365/packages/google-api-servicemanagement/protos/protos.d.ts#L16). I spoke with @alexander-fenster about this, and he recommended this option, over skipping type checking in the tsconfig.

bcoe (approved):
ship it!

PR #4238, opened 2023-05-04
build(generativeai): add long dependency as dev dependency for pnpm

sofisl:


bcoe (approved):

bcoe (approved):
We should update our  library generation logic to include this additional dependency going forward.

PR #4272, opened 2023-05-11
fix: do not exit with 1 on empty folder

### LINKED ISSUE #4073: [typeless-sample-bot] Exit code 1 when no TypeScript files found
When the `typeless-sample-bot` doesn't find any TypeScript files, it exits with code 1. That will cause builds to fail if `check=True` when running via the synthtool `shell.run()` command.  I do think it's worth flagging as a warning, but I think that only errors in the process should lead to an exit code 1 outcome.

bcoe:
Fixes #4073

### DIFF
--- a/packages/typeless-sample-bot/test/index.ts
+++ b/packages/typeless-sample-bot/test/index.ts
@@ -206,12 +206,23 @@ describe('command line option', () => {
       await assert.rejects(stat(antiTarget));
     } finally {
       targets.forEach(t => {
-        rm(path.join(fixturePath, t)).catch(() => {});
+        rm(t).catch(() => {});
       });
     }
   });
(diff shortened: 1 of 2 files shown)


packages/typeless-sample-bot/test/index.ts
```diff
@@ -206,12 +206,23 @@ describe('command line option', () => {
       await assert.rejects(stat(antiTarget));
     } finally {
       targets.forEach(t => {
-        rm(path.join(fixturePath, t)).catch(() => {});
+        rm(t).catch(() => {});
```
bcoe, line 209: There was a small bug with cleanup logic.

feywind (approved):

PR #4770, opened 2023-10-27
docs: Remove region tags

camiekim:
Per b/296916003


ddelgrosso1:
Hi @camiekim per the buganizer issue you linked should these be replaced with `workflows_client_library` as opposed to just removing them?

camiekim:
Hi @ddelgrosso1, there is already an existing Node.js sample with the correct region tag (now) that is being used here: https://cloud.google.com/workflows/docs/samples/workflows-client-library?hl=en#workflows_client_library-nodejs_javascript

Thanks for checking though!

PR #4983, opened 2024-01-24
docs(nodejs): Link broken on NPM

FossPrime:
Docs Link wasn't displaying on NPM. The parser may have just given up for performance reasons, or may not support the commonmark extension.

image

Fixes #0 🦕


sofisl:
Thanks for your fix! These docs are autogenerated, so we need to fix this upstream: https://github.com/googleapis/google-cloud-node/issues/5045 (in another library).

PR #5419, opened 2024-05-31
build: rename close-invalid-link.js to close-invalid-link.cjs

sofisl:

### DIFF
--- a/.github/scripts/close-invalid-link.cjs
+++ b/.github/scripts/close-invalid-link.cjs
@@ -35,7 +35,7 @@ module.exports = async ({ github, context }) => {
 
     try {
         const link = issue.data.body.match(/(https?:\/\/github.com\/.*)/)[0];
-        const isValidLink = (await fetch(link)).statusText === 'OK');
+        const isValidLink = (await fetch(link)).ok;
         if (!isValidLink) {
            await closeIssue();
         }


.github/scripts/close-invalid-link.cjs
```diff
@@ -35,11 +35,11 @@ module.exports = async ({ github, context }) => {
 
     try {
         const link = issue.data.body.match(/(https?:\/\/github.com\/.*)/)[0];
-        const isValidLink = (await fetch(link)).status === 200;
+        const isValidLink = (await fetch(link)).status >= 200 && (await fetch(link)).status < 300;
```
danielbankhead, line 38: Looks good, perhaps move the fetch call into a variable so that way it’s called once. You can also use `.ok` as well for status 200-299

sofisl: Thanks!

.github/scripts/close-invalid-link.cjs
```diff
@@ -35,11 +35,11 @@ module.exports = async ({ github, context }) => {
 
     try {
         const link = issue.data.body.match(/(https?:\/\/github.com\/.*)/)[0];
-        const isValidLink = (await fetch(link)).status === 200;
+        const isValidLink = (await fetch(link)).statusText === 'OK');
```
danielbankhead, line 38: Actually we’d want the `.ok` property of the response; the status text ‘OK’ is only for 200

sofisl: Thank you! TIL

danielbankhead (approved):

PR #5440, opened 2024-06-06
chore: remove requiring fetch

sofisl:
We should just magically get it... https://github.com/danmichaelo/github-script/blob/da8818015eb5c3d01f8ad6db6bd8e3a2b460c71c/README.md


danielbankhead (approved):
Yep, Node 18+ includes `fetch` globally

PR #5460, opened 2024-06-12
chore: fix issue closing script

sofisl:
1. I think the script is failing because we aren't including an issue number
```
  status: 404,
  response: {
    url: 'https://api.github.com/repos/googleapis/google-cloud-node/issues/',
    status: 404,
``` 

(this should have an issue number at the end)

2. Forgot to pass parameters
3. we should make sure the issue is a bug, we don't want to look for links where we aren't expecting them.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -107,7 +107,6 @@
   "packages/google-cloud-recommender": "6.4.0",
   "packages/google-cloud-redis": "4.3.0",
   "packages/google-cloud-resourcemanager": "5.3.0",
-  "packages/google-cloud-resourcesettings": "4.0.0",
   "packages/google-cloud-retail": "3.4.0",
   "packages/google-cloud-run": "1.3.0",
   "packages/google-cloud-scheduler": "4.3.0",
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -108,7 +108,6 @@
     "packages/google-cloud-recommender": {},
     "packages/google-cloud-redis": {},
     "packages/google-cloud-resourcemanager": {},
-    "packages/google-cloud-resourcesettings": {},
     "packages/google-cloud-retail": {},
     "packages/google-cloud-run": {},
     "packages/google-cloud-scheduler": {},


.github/scripts/close-invalid-link.cjs
```diff
@@ -23,21 +23,25 @@ async function closeIssue() {
...
-    try {
-        const link = issue.data.body.match(/(https?:\/\/github.com\/.*)/)[0];
-        const isValidLink = (await fetch(link)).ok;
-        if (!isValidLink) {
-           await closeIssue();
+    const isBugTemplate = issue.data.body.includes("Link to the code that reproduces this issue");
+
+    if (isBugTemplate) {
+        try {
+            const link = issue.data.body.match(/(https?:\/\/github.com\/.*)/)[0];
```
danielbankhead, line 38: A bit out of scope for this PR, but would this capture references with `#` (e.g. issue `#12321`)?

danielbankhead: What if there are multiple links? Should they all be valid, at least one, or just the first?

If not just the first, perhaps `String#matchAll` would be nice:
```ts
let foundValid = false;

for (const link of issue.data.body.matchAll(/https?:\/\/\w*/g)) {
  try {
    const url = new URL(link);
    const validHostname = url.hostname.endsWith('github.com');

    if (!validHostname) continue;
    if ((await fetch(url)).ok) {
      foundValid = true;
      break;
    }
  } catch {
    continue;
  }
}
```

sofisl: I think just one should be fine! Perhaps people want to include other links that are somewhat relevant but not faithful reproductions of the issue

sofisl: Also, as to your first point: yes, technically it would match `issues`, `pulls`, etc. However, I'm thinking there's one of two cases: either 1) they accidentally linked to pulls, etc., of a real reproduction repo, in which case we can just navigate to it, or 2) they linked to a non-reproduction repo's pulls, issues, etc. If 1, we should still review. If 2, then there's nothing stopping them from just submitting a github repo that isn't actually a reproduction repo (i.e., submitting something like google-cloud-node). Hopefully this is the minority of cases, and we can just close those manually.

danielbankhead: Nice, sg  👌🏽

danielbankhead (approved):

PR #5467, opened 2024-06-13
chore: update close-invalid-link.js

sofisl:
[This action](https://github.com/googleapis/google-cloud-node/actions/runs/9490647589) passed, because there was a github link... in the template of the bug. This updates the logic so we're specifically looking where it should be inputted.

### DIFF
--- a/.github/scripts/close-invalid-link.cjs
+++ b/.github/scripts/close-invalid-link.cjs
@@ -35,7 +35,7 @@ module.exports = async ({ github, context }) => {
 
     if (isBugTemplate) {
         try {
-            const link = issue.data.body.match(/(https?:\/\/github.com\/.*)/)[0];
+            const link = issue.data.body.split("\n")[18].match(/(https?:\/\/g?i?s?t\.?github.com\/.*)/);
             const isValidLink = (await fetch(link)).ok;
             if (!isValidLink) {
             await closeIssue(owner, repo, number);


.github/scripts/close-invalid-link.cjs
```diff
@@ -35,7 +35,7 @@ module.exports = async ({ github, context }) => {
 
     if (isBugTemplate) {
         try {
-            const link = issue.data.body.match(/(https?:\/\/github.com\/.*)/)[0];
+            const link = issue.data.body.split("\n")[18].match(/(https?:\/\/g?i?s?t\.?github.com\/.*)/);
```
danielbankhead, line 38: This seems sensitive to subtle changes; e.g. a minor change in the bug template could break the `18` positioning and would be difficult to debug. Perhaps a `.matchAll` with URL parsing would work more consistently for the long-term: https://github.com/googleapis/google-cloud-node/pull/5460/files#r1637082948

sofisl: True - not ideal. However your solution would fail (pass?) for the same reason, i.e., it would catch github links in the template. I think, ultimately, we need to check the input value itself. Ideally github would provide a JSON-like object for their [yaml templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms), especially since they support `required` fields.

sofisl: Ah, seems like a feature request: https://github.com/orgs/community/discussions/63402#discussioncomment-6697333

danielbankhead (approved):

PR #5640, opened 2024-08-27
fix(functions): fix invalid region tag

iennae:
making sure this sample includes a valid region tag; it could be that this sample needs to be archived but in case it's in use within the repo for testing, making sure it's associated with the right region tag. Additionally, the region tag may need to change in the near future and this will make it more discoverable that it needs to be updated. Also recognize that functions_ is what's in the configuration, but functions is in broader use and cloudfunctions is the appropriate one.


iennae:
I made a very small change, and I'm very unclear why owlbot made a bunch of changes to other files that I didn't change.

sofisl (approved):

PR #5696, opened 2024-09-20
fix: remove extra protos file

sofisl:
For some reason we're publishing an extra protos file. Adding this in to make sure we don't re-add protos once we've run compile. FYI that we'll also need to make this change in the generator (will do). Publishing as a fix since this will reduce its package size from 17MB to 9MB.

### DIFF
--- a/packages/google-cloud-tasks/protos/protos.json
+++ b/packages/google-cloud-tasks/protos/protos.json
@@ -1,4 +1,7 @@
 {
+  "options": {
+    "syntax": "proto3"
+  },
   "nested": {
     "google": {
       "nested": {


packages/google-cloud-tasks/package.json
```diff
@@ -75,7 +75,7 @@
     "compile:esm": "tsc -p ./tsconfig.esm.json && cp -r esm/src/json-helper.cjs build/esm/src/json-helper.cjs",
     "babel": "babel esm --out-dir build/cjs --ignore \"esm/**/*.d.ts\" --extensions \".ts\" --out-file-extension .cjs --copy-files",
     "compile:cjs": "tsc -p ./tsconfig.json && npm run babel",
-    "compile": "npm run compile:esm && npm run compile:cjs && cp -r protos build/protos"
+    "compile": "npm run compile:esm && npm run compile:cjs && rm -rf build/protos && cp -r protos build/protos"
```
danielbankhead, line 78: To enable non-POSIX compatibility (e.g. Windows) we could make this an inline node script rather than `rm -rf`:
```suggestion
    "compile": "npm run compile:esm && npm run compile:cjs && node -e \"fs.rm('build/protos', {recursive: true})\" && cp -r protos build/protos"
```

sofisl: Since technically we don't support Windows (we don't have windows tests), and *technically speaking* you could compile this library without having Node, I think it's better to keep as is.

danielbankhead: With `npm` as a prerequisite means there is a version of `node` available in the `$PATH`. Also we do support Windows; while we don't have any OS-specific tests customers are certainly using Windows. Here are a few examples where this can bite customers:
- https://github.com/actions-on-google/smart-home-dashboard/issues/43

Either way, this would be best practice, but no one's filed an issue yet.

danielbankhead (approved):

PR #5705, opened 2024-09-23
chore: update invalid link regex

sofisl:


.github/scripts/close-invalid-link.cjs
```diff
@@ -41,7 +41,7 @@ module.exports = async ({ github, context }) => {
 
     if (isBugTemplate) {
         try {
-            const link = issue.data.body.split("\n")[18].match(/(https?:\/\/g?i?s?t\.?github.com\/.*)/);
+            const link = issue.data.body.split("\n")[18].match(/(https?:\/\/g?i?s?t?\.?github.com\/.*)/);
```
danielbankhead, line 44: [optional] The `gist` could be its own matching group:
```suggestion
            const link = issue.data.body.split("\n")[18].match(/(https?:\/\/(gist\.)?github.com\/.*)/);
```

danielbankhead (approved):

Issue #5838, opened 2024-11-22
Need to fix issue templates for bugs


sofisl:
We are trying to use github's use of yaml validation [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms). However, they do not provide validation for their  issue templates and so our solution has been hacky at best (we are trying to confirm that a proper URL exists but right now it's just been too fragile). Instead of doing another quick fix, I need to create a more durable solution that doesn't frustrate people, including writing tests and making sure it's able to withstand various scenarios.

sofisl:
see #5826

PR #5905, opened 2024-12-26
build: improve logic for issue template closing and add tests

sofisl:
This PR was meant to only be to add  better logic and testing for our issue automation since it was looking a little fragile, see #5838. I also will follow up for our split libraries, see https://github.com/googleapis/synthtool/pull/2052

However, it also includes some changes that were incorrectly added to the library, namely #5914, #5381, #5922. The more permanent fixes will come in a new version of the generator, and cl/715614266.

The only relevant changes to review are in .github folder

### DIFF
--- a/.github/scripts/package.json
+++ b/.github/scripts/package.json
@@ -12,7 +12,6 @@
     },
     "devDependencies": {
         "@octokit/rest": "^21.0.2",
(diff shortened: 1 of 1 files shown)


danieljbruce (approved):
Looks fine to me, but I don't know too much about this library.

PR #6461, opened 2025-07-02
docs(googlemaps): add terms of service section to readme

usefulthink:
I'm currently creating PRs for all @googlemaps packages on npm to include a Paragraph about the terms of service and especially the changes ToS for the EEA market.

CC @anglarett


sofisl (approved):

sofisl:
ACtually, these are all autogenerated, so you'll need to add them in a .readme-partials.yaml file (or else they'll get overwritten). They're not overwritten right now because the post-processor didn't run on this PR. See Example: https://github.com/googleapis/google-cloud-node/blob/main/packages/google-cloud-compute/.readme-partials.yaml

usefulthink:
Thanks a lot, I'll look into this!

usefulthink:
@sofisl do you think we could add a conditional section (e.g. "Terms of Service") to the README template in synthtool ([I think this is the one](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/node_library/README.md?plain=1)), somewhere towards the bottom? I have the feeling that the `introduction` partial would be too far up the top of the document.

We could then add the Google Maps ToS to a `.readme-partials.yaml` under the key `tos` or `terms-of-service`. But I'm not sure if something like that would be useful for projects outside of Google Maps.

usefulthink:
@sofisl would you mind having another look? This should be a solution that is fine for the maps team, and should work without having to change anything in the `synthtool` templates.

sofisl (approved):

usefulthink:
What is the process for getting this merged? Is there anything left that I have to do?

leahecole (approved):

PR #6699, opened 2025-09-09
fix: [gapic-node-processing] remove hardlinks

sofisl:
see b/443125466
gemini seems to think the hardlinks error has to do with the node_modules and with the LICENSE.md.


leahecole:
Weird! I was worried about the license file then realized it has a regular license file, so this should be fine to remove.

PR #6701, opened 2025-09-09
chore: update all libraries .OwlBot.yaml and owlbot.py file to librarian.js

sofisl:
The corresponding PR to cl/804624707

This PR does the following:

1. Updates the .Owlbot.yaml file to copy over the files from the correct location
2. Any  owlbot.py files are removed if they do nothing (these are now captured in `templates_excludes`) or rewritten to not depend on synthtool.
3. Adds `[//]: # "partials.introduction"` to all READMEs. This is because currently, we need a hacky way of determining whether the library is the "new" process or the "old" process in synthtool. I've created this hacky determination since that is what the new templates in the generator have. So, adding this line will force the post-processor to run the new code and makes sure the post-processor runs correctly for all repos.


sofisl:
@leahecole this is good to merge!

PR #6784, opened 2025-10-03
chore: refactor gapic node processing

sofisl:
This is a pre-PR for actual changes that I need. I'm refactoring the library so I can easily add a feature later on. For the time being, this merely consolidates all the library information into a library object.

Essentially we're moving a lot of target-library-specific information from the generate-index.ts command to a top-level library object so that we can use the information more globally.

The motivation for this PR is https://github.com/googleapis/google-cloud-node/pull/6777. The test is failing because the system-test is testing all the clients in v1 - but not all those clients exist in v1beta1. Since the default version is still v1beta1, we need to test those clients, as opposed to the clients in v1. (separately, we should probably also upgrade this library). Basically this exposes the need to test the same clients in the default version, which means the clients and default version exposed globally vs. per function, hence the new library object.

I've tried to make as few changes as possible here so as to not break anything too much. I'll create a separate follow up PR putting more methods into the actual library object, for now we can just pass in the library object as is since it preserves functionality the best.


packages/gapic-node-processing/test/library.test.ts
```diff
@@ -0,0 +1,187 @@
...
+      assert.strictEqual(libraryConfig.sourcePath, FAKE_DIRECTORY);
+      assert.strictEqual(libraryConfig.destinationPath, FAKE_DEST_DIRECTORY);
+    });
+
+    it('should set whether isEsm and the appropriate srcPath', () => {
+      const libraryConfig = new LibraryConfig(
+        FAKE_DIRECTORY,
+        FAKE_DEST_DIRECTORY,
+        undefined,
+        true,
```
leahecole: maintainability nit: can these be passed as named parameters rather than positional ones so it's easier to know which value is which parameter?

packages/gapic-node-processing/test/generate-index.test.ts
```diff
@@ -136,91 +101,37 @@ describe('generate index.ts', () => {
...
+      ),
       'utf8',
     );
     // Confirm all versions were generated
     assert.match(contents, /index.js/);
     // Confirm default version is exported
     assert.match(contents, /..\/..\/protos\/protos.js/);
   });
-
-  it('should guess the default version if not provided', async () => {
```
leahecole, line 159: Why is this test removed?

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+      throw new Error(
+        `Default version ${this.defaultVersion} not found in available versions: ${this.versions.join(
+          ', ',
+        )}`,
+      );
+    }
+    return defaultVersionAndClients;
+  }
+
+  // In case a default version isn't provided, this function should
```
leahecole: nit - regular comment before more thorough docstring

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+/export\s*{\s*(\w+Client)\s*}/g;
+*/
+const CLIENT_EXTRACTION_REGEX = /export\s*{\s*(\w+Client)\s*}/g;
+
+interface VersionSpec {
+  version: string; // Stores the "best" version found so far
+  major: number;
+  precedence: number;
+  preReleaseQualifier: number; // For e.g., beta1 vs beta2
+}
```
leahecole, line 53: add a docstring for this please!

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+  public destinationPath: string;
+  public defaultVersion?: string = '';
+  public defaultVersionAndClients: VersionsAndClients = {
+    version: '',
+    clients: [],
+  };
+
+  /**
+   * @param isEsm Whether the library supports ES Modules.
+   * @param releaseLevel The current stability level of the library.
```
leahecole: I don't think this is a parameter anymore?

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+   * A getter to provide a list of all available versions.
+   */
+  public async getVersions(): Promise<string[]> {
+    return this.clientsAndVersions.map(x => x.version);
+  }
+
+  /**
+   * A getter to provide a list of clients and their corresponding versions.
+   */
+  public async getClientsAndVersions() {
```
leahecole, line 121: Why is there a `getClientsAndVersions` that's separate from `getVersions`? What's different about these versions?

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+    const allVersionedLibraries = await getAllTopLevelDirectories(
+      this.sourcePath,
+    );
+    for (const directory of allVersionedLibraries) {
+      const versions = await getAllTopLevelDirectories(
+        path.join(this.sourcePath, directory, this.srcPath),
+      );
+      // even though this looks nested, it ends up being o(1) since
+      // we only have one directory per versioned library (the single
+      // version of the library)
```
leahecole, line 142: nit - is it technically o(n) where n is the number of versions? 

Either way, are there circumstances where this could change and we would need to be more concerned about this being nested?

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+        );
+        if (await fs.stat(indexFile)) {
+          const fileContent = await fs.readFile(indexFile, 'utf8');
+          const clientsRegexMatch = Array.from(
+            fileContent.matchAll(CLIENT_EXTRACTION_REGEX),
+          );
+          clientsAndVersions.push({
+            version,
+            clients: Array.from(
+              new Set(clientsRegexMatch.map((x: any) => x[1])),
```
leahecole, line 161: what does this line do?

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+      '': 3, // Stable (no suffix) is highest precedence
+      beta: 2,
+      alpha: 1,
+    };
+
+    let currentVersionSpec: VersionSpec = {
+      version: versions[0], // Stores the "best" version found so far
+      major: -1,
+      precedence: -1,
+      preReleaseQualifier: -1, // For e.g., beta1 vs beta2
```
leahecole, line 217: why are these initialized at -1s?

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    // Define the precedence of pre-release types
+    const precedence = {
+      '': 3, // Stable (no suffix) is highest precedence
+      beta: 2,
+      alpha: 1,
+    };
+
+    let currentVersionSpec: VersionSpec = {
+      version: versions[0], // Stores the "best" version found so far
```
leahecole: is versions always guaranteed to be ordered? i.e. can we really assume 0 will be the "best" version always?

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+      major: -1,
+      precedence: -1,
+      preReleaseQualifier: -1, // For e.g., beta1 vs beta2
+    };
+
+    for (const version of versions) {
+      const match = version.match(/^v(\d+)(alpha|beta(\d*))?$/);
+
+      if (match) {
```
leahecole, line 223: should we be erroring if there isn't a match?

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+        const majorVersion = parseInt(match[1], 10);
+        const preReleaseType = match[2]
```
leahecole: this chained ternary operator hurts my brain a little bit - could this instead be an easier to read helper function? Advantage of that would mean that we would be checking if the string in `match[2]` is 'alpha' in addition to if it's 'beta' for a little bit of extra safety - otherwise that suffix could be anything

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    return currentVersionSpec.version;
+  }
+}
+
+// Comparison Logic:
+// 1. Higher Precedence (Stable > Beta > Alpha)
+// 2. Higher Major Version
+// 3. Within same pre-release type, higher qualifier (e.g., beta2 > beta1)
+// 4. If everything else is equal, the current one is just as good
```
leahecole: nit - should this be a docstring?

packages/gapic-node-processing/test/fixtures/combined-library/google-cloud-tasks/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@google-cloud/tasks",
   "version": "0.1.0",
-  "private": true,
```
leahecole, line 4: is removing this an issue? or was this in the fixtures in error?

packages/gapic-node-processing/src/library.ts
sofisl: The getVersions is just calling a .map function on the result of getClientsAndVersions. It's just so it's easier to access instead of calling .map for every instance of this.clientsAndVersions

packages/gapic-node-processing/src/library.ts
sofisl: Nope, since we're calling this function before libraries are combined, meaning each library is a single version - it will look like test/fixtures/combined-library/google-cloud-speech-nodejs! And if we have combined libraries already, this function would fail because it's not the correct directory structure. I can add an error about it not being in expected format if we can't access the following path.

packages/gapic-node-processing/src/library.ts
sofisl: This line just makes sure we don't have any duplicates in the regex matching (with Set), and then creates an array from that set which is what the clients type is.

packages/gapic-node-processing/src/library.ts
sofisl: this is just a starting point! It gets updated as we traverse the versions.

packages/gapic-node-processing/src/library.ts
sofisl: This is just for comparison - it's guaranteed to be ovewritten by the first comparison. We could just put the first values, but then we'd have to do the regex parsing beforehand, so I think -1 is just an easy way to guarantee comparison.

packages/gapic-node-processing/test/generate-index.test.ts
sofisl: The functionality was moved to library.ts so it was moved to the library.test.ts file.  The corresponding test is: should correctly initialize versions and clients if no default version is provided.

packages/gapic-node-processing/src/library.ts
sofisl: this is just a starting point! It gets updated as we traverse the versions.

packages/gapic-node-processing/test/fixtures/combined-library/google-cloud-tasks/package.json
sofisl: this was in the fixtures, I think it was an old attempt at making sure these files don't try to get published. It's ok to be removed!

sofisl:
@leahecole this is ready for review, PTAL!

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,314 @@
...
+          this.srcPath,
+          version,
+          INDEX_PATH,
+        );
+        if (await fs.stat(indexFile)) {
+          const fileContent = await fs.readFile(indexFile, 'utf8');
+          const clientsRegexMatch = Array.from(
+            fileContent.matchAll(CLIENT_EXTRACTION_REGEX),
+          );
+          clientsAndVersions.push({
```
leahecole, line 158:
```suggestion
          // ensures we don't have any duplicates in the regex matching
          // creates an array from the set which is what the client type is
          clientsAndVersions.push({
```

packages/gapic-node-processing/src/library.ts
leahecole: made a suggestion to add this as a comment

packages/gapic-node-processing/src/library.ts
```diff
@@ -0,0 +1,314 @@
...
+
+    // Define the precedence of pre-release types
+    const precedence = {
+      '': 3, // Stable (no suffix) is highest precedence
+      beta: 2,
+      alpha: 1,
+    };
+
+    let currentVersionSpec: VersionSpec = {
+      version: versions[0], // Stores the "best" version found so far
```
leahecole:
```suggestion
    // use the "best" version found so far, version[0] as a starting point
    // initialize other values to -1, guaranteed to be overwritten at first comparison
    // all will be updated as we traverse the versions
    let currentVersionSpec: VersionSpec = {
      version: versions[0],
```

packages/gapic-node-processing/src/library.ts
leahecole: suggested moving and adjusting the comment

packages/gapic-node-processing/src/library.ts
leahecole: added more of a comment about this!

leahecole (approved):

PR #6790, opened 2025-10-07
fix!: remove deprecated workflows client

sofisl:
see #6789

This is a separate PR since this is a breaking change. We are removing the workflows client from both the v1beta and v1 surfaces. Please use the ExecutionsClient instead.

### DIFF
--- a/packages/google-cloud-workflows-executions/src/index.ts
+++ b/packages/google-cloud-workflows-executions/src/index.ts
@@ -21,10 +21,8 @@ import * as v1beta from './v1beta';
 
 const ExecutionsClient = v1.ExecutionsClient;
 type ExecutionsClient = v1.ExecutionsClient;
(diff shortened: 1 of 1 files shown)


sofisl:
Merge #6789 first

BenJackGill:
This is a very unexpected breaking change

And your advice to "Please use the ExecutionsClient instead" is nonsensical.

ExecutionsClient handles executing workflows (run, cancel, list executions). It has nothing to do with creating or deploying workflow definitions. These are two completely different REST API services:

- workflows.googleapis.com — CRUD on workflow definitions (create, update, delete, get, list)
- workflowexecutions.googleapis.com — manage workflow executions (create execution, cancel, list)

You removed the client for the first one and told us to use the client for the second one, which doesn't cover the same operations at all.

Now there is no way for us to programatically create Workflows, we can only execute Workflows that have already been created.

PR #6795, opened 2025-10-09
fix: make sure we are only generating default version system tests

sofisl:
Fixes #6793 

follow up to #6784

### DIFF
--- a/packages/gapic-node-processing/test/combine-libraries.test.ts
+++ b/packages/gapic-node-processing/test/combine-libraries.test.ts
@@ -222,6 +222,30 @@ describe('combine libraries', () => {
       ),
       /SpeechClient/,
     );
+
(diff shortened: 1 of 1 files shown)


packages/gapic-node-processing/test/combine-libraries.test.ts
```diff
@@ -190,6 +196,58 @@ describe('combine libraries', () => {
...
+          TEST_FIXTURES_PATH,
+          LIB_POST_COMBINATION,
+          'system-test/fixtures/sample/src/index.js',
+        ),
+        'utf-8',
+      ),
+      /SpeechClient/,
+    );
+
+        assert.match(
```
feywind, line 226: I think the spacing here is weird.

PR #6872, opened 2025-11-07
fix: doc linter issues

feywind:
There was a release process failure a few weeks back that seems to have resulted in the capacityplanner library not getting a docs release. Any small change should kick it off again for this library.

(no manual perms to push it!)


quirogas (approved):
LGTM!

PR #6950, opened 2025-12-16
fix: migrate to managed generated samples 

iennae:
replaced by https://github.com/GoogleCloudSamples/client-library-samples/blob/main/bigquerydatatransfer/v1/nodejs/google-cloud-bigquery-data-transfer/data-transfer-service-client-data-sources-list.js

As I'm not sure if this is used for testing in this repo, leaving the code as is. 

The tracker is incorrectly flagging that it will remove a sample browser page. It will not as there are other languages and the conflict is present.


feywind (approved):
You'd know better than I on the samples!

PR #6964, opened 2026-01-07
docs: [storage-control] Fix API reference link in README

danielduhh:
API reference link in README is broken


quirogas:
@danielduhh thank you for submitting this PR. This change must take place at the generator level; otherwise, the fix will not be persistent. I'll followup on this fix separately.

danielduhh:
sg, thank you!

PR #6992, opened 2026-01-14
build: change migration script to support folder path for packages

miguelvelezsa:
Add support to define folder path in target package.

As part of monorepo consolidation we need to migrate other kind of packages to this repo, therefore we are updating the migration scripts to support folder definition before packages.

### DIFF
--- a/ci/run_conditional_tests.sh
+++ b/ci/run_conditional_tests.sh
@@ -73,6 +73,7 @@ fi
 subdirs=(
     containers
     packages
+    handwritten
     .github/scripts
 )
 


bin/delete-everything-split-repo.sh
```diff
@@ -23,7 +23,7 @@ fi
 
 # repo name (e.g. nodejs-asset)
 SPLIT_REPO=$1
-# destination directory (e.g. google-cloud-asset)
+# destination directory (e.g. packages/google-cloud-asset)
```
sofisl, line 26: this should be handwritten (we're going to create a new directory).

bin/delete-everything-split-repo.sh
```diff
@@ -154,7 +154,7 @@ chmod u+x "/tmp/${SPLIT_REPO}/.kokoro/system-test.sh"
 README_MD="/tmp/${SPLIT_REPO}/README.md"
 
 README_CONTENT=$(cat $README_MD)
-echo -e "**_THIS REPOSITORY IS DEPRECATED. ALL OF ITS CONTENT AND HISTORY HAS BEEN MOVED TO [GOOGLE-CLOUD-NODE](https://github.com/googleapis/google-cloud-node/tree/main/packages/${ARTIFACT_NAME})_**\n\n$README_CONTENT" > "$README_MD"
+echo -e "**_THIS REPOSITORY IS DEPRECATED. ALL OF ITS CONTENT AND HISTORY HAS BEEN MOVED TO [GOOGLE-CLOUD-NODE](https://github.com/googleapis/google-cloud-node/tree/main/${ARTIFACT_NAME})_**\n\n$README_CONTENT" > "$README_MD"
```
sofisl, line 157: should have handwritten in the path

sofisl (commented):
I think in addition to these changes, I'm noticing we're going to have to do a few more things:

- In the post-processor, we're going to need to update the logic that it uses to scan packages to include scanning handwritten/. If we want, we can create a new subdirectory in packages/ called handwritten, but ideally I'd like to keep these directories as separate as possible given that they have different codeowners and different requirements, etc.
- we also would need to change the scanning the CI does for handwritten/, unless, again, we put handwritten under packages/, which then might need it's own subdirectory changes

bin/delete-everything-split-repo.sh
miguelvelezsa: It is an example :) now it will accept the folder structure as parameter, for example ./bin/migrate-split-repo.sh nodejs-logging-bunyan handwritten/nodejs-logging-bunyan.

bin/delete-everything-split-repo.sh
miguelvelezsa: Same as above, the artifact has the folder.

miguelvelezsa:
Created changes for synthtool https://github.com/googleapis/synthtool/pull/2147

sofisl (approved):

PR #7023, opened 2026-01-21
docs(storage control): retain jsdocs for overloads and only generate docs for uber stor…

ddelgrosso1:
…age control client


ddelgrosso1:
> LGTM but I think should be docs or build vs. fix?

Yea, now that I think about it I think docs should be sufficient.

PR #7061, opened 2026-01-29
chore: add diff check in migration post processing for kokoro repos

GautamSharda:
Insert diff check logic to migrated repos to determine if kokoro tests should be run.

Some evidence that the diff check logic is added in the right place and works is that adding it to [nodejs-logging-winston/trampoline_v2.sh](https://github.com/googleapis/nodejs-logging-winston/blob/main/.kokoro/trampoline_v2.sh) and diff checking [./github/](https://github.com/googleapis/nodejs-logging-winston/tree/main/.github) results in:

(1) The tests are skipped / trivially passed when a PR does not make a change in ./github/: ["no op" case
](https://github.com/googleapis/nodejs-logging-winston/pull/921)

(2) The tests run when a PR makes changes in ./github/: ["run tests" case
](https://github.com/googleapis/nodejs-logging-winston/pull/922)

In our case, we diff check the package being migrated i.e., handwritten/{package_name} instead. Also, in our case, we have to account for firestore, which uses [trampoline.sh](https://github.com/googleapis/nodejs-firestore/blob/main/.kokoro/trampoline.sh) (v1 / legacy) instead of trampoline_v2.

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/{{metadata['repo']['name']}}/issues) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes b/478965516 🦕

### DIFF
--- a/packages/google-cloud-retail/protos/google/cloud/retail/v2beta/common.proto
+++ b/packages/google-cloud-retail/protos/google/cloud/retail/v2beta/common.proto
@@ -943,3 +943,15 @@ message PinControlMetadata {
   // keyed by pin position.
   map<int64, ProductPins> dropped_pins = 2;
 }
+
+// A list of string values.
+message StringList {
+  // String values.
+  repeated string values = 1;
+}
+
+// A message with a list of double values.
+message DoubleList {
+  // The list of double values.
+  repeated double values = 1;
+}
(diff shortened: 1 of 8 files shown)


sofisl (approved):

bin/split-repo-post-process.sh
```diff
@@ -127,6 +127,57 @@ if [[ -d "${KOKORO_DIR}" ]]; then
...
+    echo "Checking for changes in \${RELATIVE_PKG_PATH}..."
+    
+    # Determine the diff range based on the CI system/event
+    # Safe default: HEAD~1..HEAD
+    DIFF_RANGE="HEAD~1..HEAD"
+
+    if git diff --quiet "\${DIFF_RANGE}" -- "\${RELATIVE_PKG_PATH}"; then
+        echo "No changes detected in \${RELATIVE_PKG_PATH}. Skipping tests."
+        exit 0
+    else
```
miguelvelezsa, line 161: dont need the else here

miguelvelezsa (approved):

PR #7070, opened 2026-01-30
chore: update migration script

sofisl:
1) adds to codeowners file
2) keeps and modifies .kokoro so it works
3) merges in any additional gh action workflows
4) does *not* delete the rest of the library (in the archival state)

### DIFF
--- a/bin/split-repo-post-process.sh
+++ b/bin/split-repo-post-process.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # Copyright 2022 Google LLC
 #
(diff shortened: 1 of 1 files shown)


bin/split-repo-post-process.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 # Copyright 2022 Google LLC
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
+# Licengsed under the Apache License, Version 2.0 (the "License");
```
miguelvelezsa: typo

PR #7082, opened 2026-02-02
migrate code from googleapis/nodejs-logging-winston

GautamSharda:
migration b/475335325


GautamSharda:
Will make new migration PR with updated script.

PR #7147, opened 2026-02-05
feat: Add high precision TIMESTAMP values for queries

danieljbruce:
## Description ##

This PR makes it so that all requests to the /queries endpoint support returning high precision results for the TIMESTAMP type. This ensures that users will always get the highest precision possible for TIMESTAMP typed values requested in queries provided they request getting a higher precision timestamp by specifying a type of TIMESTAMP(12):

```
const query = {
    query: 'SELECT ? as ts',
    params: [bigquery.timestamp('2023-01-01T12:00:00.123456789123Z')],
    types: ['TIMESTAMP(12)']
};
const [rows] = await bigquery.query(query, options);
```

https://github.com/googleapis/nodejs-bigquery/pull/1596 adds similar support for the /data REST endpoint which may be useful for reference. And https://github.com/googleapis/java-bigquery/pull/4010 contains a similar PR to this one, but for the Bigquery Java codebase.

## Impact

Allows users to opt in to fetching high precision timestamps for calls to the /queries endpoint.

## Testing

System tests were added to make sure no matter what formatOptions they provide, the values sent back by the server will be delivered to the user without losing any precision when requested.

## Additional Information

Notice that the PR adds timestampPrecision: 12 only when the user explicitly uses the TIMESTAMP(12) types. This is because if we automatically opt users into timestampPrecision: 12 then their code will suddenly fail for certain queries that were working before. Therefore, we are going to make timestamp precision an opt-in property.

### DIFF
--- a/handwritten/bigquery/test/bigquery.ts
+++ b/handwritten/bigquery/test/bigquery.ts
@@ -3458,7 +3458,7 @@ describe('BigQuery', () => {
           },
         },
         {
-          name: 'TOF: INT64, UI64: false (error)',
+          name: 'TOF: INT64, UI64: false',
           opts: {
             ['formatOptions.timestampOutputFormat']: 'INT64',
             ['formatOptions.useInt64Timestamp']: false,
@@ -3469,7 +3469,7 @@ describe('BigQuery', () => {
           },
         },
         {
-          name: 'TOF: ISO8601_STRING, UI64: true (error)',
+          name: 'TOF: ISO8601_STRING, UI64: true',
           opts: {
             ['formatOptions.timestampOutputFormat']: 'ISO8601_STRING',
             ['formatOptions.useInt64Timestamp']: true,
@@ -3526,7 +3526,7 @@ describe('BigQuery', () => {
           },
         },
         {
-          name: 'TOF: FLOAT64, UI64: omitted (error)',
+          name: 'TOF: FLOAT64, UI64: omitted',
           opts: {
             ['formatOptions.timestampOutputFormat']: 'FLOAT64',
           },


handwritten/bigquery/src/job.ts
```diff
@@ -595,10 +595,21 @@ class Job extends Operation {
         let rows: any = [];
 
         if (resp.schema && resp.rows) {
-          rows = BigQuery.mergeSchemaWithRows_(resp.schema, resp.rows, {
-            wrapIntegers,
-            parseJSON,
-          });
+          try {
+            /*
+            Without this try/catch block, calls to getRows will hang indefinitely if
```
alvarowolfx: nit: getRows = jobsQuery endpoint ?

handwritten/bigquery/system-test/high-precision-query-tests.ts
```diff
@@ -0,0 +1,234 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as assert from 'assert';
+import {describe, it, before} from 'mocha';
+import {BigQuery} from '../src';
+
+describe('High Precision Query System Tests', () => {
```
alvarowolfx: why do we need a separated file ? The previous PR we already deviated a bit from the pattern of having system tests in the https://github.com/googleapis/nodejs-bigquery/blob/main/system-test/bigquery.ts and added https://github.com/googleapis/nodejs-bigquery/blob/main/system-test/timestamp_output_format.ts, can we bundle all timestamp related tests into the same file ?

And gonna bring this again that we don't need the amount of combinations and system tests to check if the support for picosecond resolution is working. More tests doesn't necessarily means more coverage. With this PR and the previous one we are adding 28 query executions, where previously we had a total of around 50. Not saying that the coverage was perfect before, but adding too many system/integration test for a time format change seems bit too much, I'd lean more on unit tests. System tests makes CI slower, and I'm not seeing enough justification for the amount of combination and tests being added here.

We already got bitten by too much client side validation or tests trying to assert internal details of the service in the past that can cause problems. In this case, is not a SDK code, but is test that can fail in the future if things changes in the service.

See internal b/460198628

handwritten/bigquery/system-test/high-precision-query-tests.ts
```diff
@@ -0,0 +1,234 @@
...
+  let bigquery: BigQuery;
+  const expectedTsValueMicroseconds = '2023-01-01T12:00:00.123456000Z';
```
alvarowolfx: I think this was suppose to be Nanosecond. We missed that the previous PR

handwritten/bigquery/system-test/high-precision-query-tests.ts
```diff
@@ -0,0 +1,234 @@
...
+  const expectedTsValueNanoseconds = '2023-01-01T12:00:00.123456789123Z';
```
alvarowolfx: and this picosecond. Same applies to the previous https://github.com/googleapis/nodejs-bigquery/pull/1596/changes#diff-e82cb9428a68678cc305e12e59f2e21761622b9fcfcf45d4caab5e80657e2b9a test file

handwritten/bigquery/system-test/high-precision-query-tests.ts
```diff
@@ -0,0 +1,234 @@
...
+      expectedTsValue: expectedTsValueMicroseconds,
+    },
+    {
+      name: 'TOF: TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED, UI64: false (default ISO8601_STRING)',
+      timestampOutputFormat: 'TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED',
+      useInt64Timestamp: false,
+      expectedTsValue: expectedTsValueMicroseconds,
+    },
+    {
+      name: 'TOF: FLOAT64, UI64: true (error)',
```
alvarowolfx: the service side is going to reject it and is not a reasonable combination that customers would use. Not sure if we need to assert this. If the service side changes behavior (and can happen) we are gonna have a flaky test. I'd add just one error combination here.

handwritten/bigquery/test/bigquery.ts
alvarowolfx, line 3586: we can write similar tests to this one to assert the logic for default dataFormatOptions value. The existing tests assert only not passing any format from the customer and expecting ISO8601_STRING, but we can test passing any timestampOutputFormat or useInt64Timestamp and assert that this overrides the default value. This is faster than the system test and covers the change that is to use ISO8601_STRING as the default, depending on the format option provided by the customer

handwritten/bigquery/src/job.ts
danieljbruce: Yes! This has been fixed now. This was a typo.

handwritten/bigquery/system-test/high-precision-query-tests.ts
danieljbruce: > why do we need a separated file ?

I realize having a separate file deviates from the pattern of putting every test in `system-test/bigquery.ts`, but I really believe that having a separate file is a better pattern. One of the biggest drawbacks with putting every test in one file is that you lose encapsulation. For instance, the before/after hooks in the bigquery.ts test file don't have anything to do with the new tests we are adding. So if we add the new tests to this file then future maintainers have to think about how everything else in this file including the before/after hooks influence these tests and we don't want that.

> we don't need the amount of combinations and system tests to check if the support for picosecond resolution is working

Ok. I needed the tests myself to understand how the client was behaving, but maybe now we can comment some of them out to improve the CI performance. I would prefer to comment the tests out rather than delete them though because the commented tests could be useful later when we want to investigate the product's behaviour. Which ones do you want to comment out?

handwritten/bigquery/system-test/high-precision-query-tests.ts
danieljbruce: Yes. Good catch. I changed this in both places.

handwritten/bigquery/system-test/high-precision-query-tests.ts
danieljbruce: Yes. Good catch. I changed this in both places.

handwritten/bigquery/system-test/high-precision-query-tests.ts
danieljbruce: I commented this test out. Let me know if you want more tests commented out.

handwritten/bigquery/test/bigquery.ts
danieljbruce: Now I understand your preference for unit tests. I'll get to this one next.

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2666,41 +2754,23 @@ export class BigQueryRange {
...
+    /*
+    We moved the internals of fromSchemaValue_ to a method outside of this
+    class, but fromSchemaValue_ calls fromStringValue_ and this fromStringValue_
+    method is private so we needed a fromStringValue_ method outside of this
+    class so it could be accessed by fromSchemaValue_ function.
+     */
+    return fromStringValue_(value);
   }
 
   static fromSchemaValue_(value: string, elementType: string): BigQueryRange {
```
alvarowolfx: this is an internal only method (because it ends with `_`), so we don't need to worry about keeping the public interface.

handwritten/bigquery/test/bigquery.ts
```diff
@@ -3397,12 +3397,141 @@ describe('BigQuery', () => {
...
         jobCreationMode: 'JOB_CREATION_REQUIRED',
         formatOptions: {
-          useInt64Timestamp: true,
+          timestampOutputFormat: 'ISO8601_STRING',
         },
       };
       assert.deepStrictEqual(req, expectedReq);
     });
 
+    describe('format options', () => {
```
alvarowolfx: nit: "timestamp format options"

handwritten/bigquery/system-test/high-precision-query-tests.ts
alvarowolfx: > I realize having a separate file deviates from the pattern of putting every test in system-test/bigquery.ts, but I really believe that having a separate file is a better pattern. One of the biggest drawbacks with putting every test in one file is that you lose encapsulation. For instance, the before/after hooks in the bigquery.ts test file don't have anything to do with the new tests we are adding. So if we add the new tests to this file then future maintainers have to think about how everything else in this file including the before/after hooks influence these tests and we don't want that.

 I still don't understand why we can't bundle all timestamp related tests into the same file. I can understand the reasoning about the `system-test/bigquery.ts` currently being bloated ( and we should plan to refactor it), but those timestamp formats can definitely live together. Am I missing something ? 

> Ok. I needed the tests myself to understand how the client was behaving, but maybe now we can comment some of them out to improve the CI performance. I would prefer to comment the tests out rather than delete them though because the commented tests could be useful later when we want to investigate the product's behaviour. Which ones do you want to comment out?

Was a nice catch to add the nested query test, but I don't think is a good approach to keep tests commented out here. The tests added at `test/bigquery.ts` are already excising all the combinations. For the integration tests we can exercise a few scenarios like:
* Not passing parameters  -> should default to picosecond
* Passing useInt64 -> should override default and return nanoseconds
* Passing timestampOutputFormat: INT64 -> should override default and return nanoseconds
* Passing two incompatible params (useInt64: true and timestampOutputFormat: Float64) -> error

handwritten/bigquery/test/bigquery.ts
```diff
@@ -3397,12 +3397,141 @@ describe('BigQuery', () => {
...
+          timestampOutputFormat: undefined,
+          useInt64Timestamp: true,
+        },
+        {
+          name: 'TOF: omitted, UI64: false (default ISO8601_STRING)',
+          timestampOutputFormat: undefined,
+          useInt64Timestamp: false,
+        },
+        {
+          name: 'TOF: TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED, UI64: omitted (default INT64)',
```
alvarowolfx: both omitted it should default to ISO8601_STRING right ?

handwritten/bigquery/test/bigquery.ts
```diff
@@ -3397,12 +3397,141 @@ describe('BigQuery', () => {
...
+          timestampOutputFormat: 'ISO8601_STRING',
+          useInt64Timestamp: true,
+        },
+        {
+          name: 'TOF: ISO8601_STRING, UI64: false',
+          timestampOutputFormat: 'ISO8601_STRING',
+          useInt64Timestamp: false,
+        },
+        {
+          name: 'TOF: omitted, UI64: omitted (default INT64)',
```
alvarowolfx: both omitted it should default to ISO8601_STRING right ?

handwritten/bigquery/test/bigquery.ts
```diff
@@ -3397,12 +3397,141 @@ describe('BigQuery', () => {
...
+          timestampOutputFormat: 'FLOAT64',
+          useInt64Timestamp: undefined,
+        },
+        {
+          name: 'TOF: INT64, UI64: omitted',
+          timestampOutputFormat: 'INT64',
+          useInt64Timestamp: undefined,
+        },
+        {
+          name: 'TOF: ISO8601_STRING, UI64: omitted (error)',
```
alvarowolfx: why this should be an error ?

handwritten/bigquery/test/bigquery.ts
```diff
@@ -3397,12 +3397,141 @@ describe('BigQuery', () => {
...
+          timestampOutputFormat: 'ISO8601_STRING',
         },
       };
       assert.deepStrictEqual(req, expectedReq);
     });
 
+    describe('format options', () => {
+      const testCases = [
+        {
+          name: 'TOF: TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED, UI64: true',
```
alvarowolfx, line 3415: testsCases can reuse the QueryOptions types and should also have the expectations. Building the expectation based on output feels weird to me. We can see for example that some test cases are described as errors, but we are not capturing any errors. And I would argue that we should not be even throwing error on the client side, as showed with b/460198628.

```suggestion
describe('timestamp format options', () => {
      const testCases: {
        name: string,
        opts: QueryOptions,
        expected?: Pick<QueryOptions, 'formatOptions.timestampOutputFormat' | 'formatOptions.useInt64Timestamp'>,
        bail?: boolean,
      }[] = [
          {
            name: 'TOF: TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED, UI64: true',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED',
              ['formatOptions.useInt64Timestamp']: true,
            },
            expected: {
              ['formatOptions.timestampOutputFormat']: 'TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED',
              ['formatOptions.useInt64Timestamp']: true,
            },
          },
          {
            name: 'TOF: TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED, UI64: false (default ISO8601_STRING)',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED',
              ['formatOptions.useInt64Timestamp']: false,
            },
            expected: {
              ['formatOptions.timestampOutputFormat']: 'ISO8601_STRING',
            },
          },
          {
            name: 'TOF: FLOAT64, UI64: false',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'FLOAT64',
              ['formatOptions.useInt64Timestamp']: false,
            },
            expected: {
              ['formatOptions.timestampOutputFormat']: 'FLOAT64',
            },
          },
          {
            name: 'TOF: INT64, UI64: true',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'INT64',
              ['formatOptions.useInt64Timestamp']: true,
            },
            expected: {
              ['formatOptions.timestampOutputFormat']: 'INT64',
            },
          },
          {
            name: 'TOF: INT64, UI64: false (error)',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'INT64',
              ['formatOptions.useInt64Timestamp']: false,
            },
            bail: true,
          },
          {
            name: 'TOF: ISO8601_STRING, UI64: true (error)',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'ISO8601_STRING',
              ['formatOptions.useInt64Timestamp']: true,
            },
            bail: true,
          },
          {
            name: 'TOF: ISO8601_STRING, UI64: false',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'ISO8601_STRING',
              ['formatOptions.useInt64Timestamp']: false,
            },
            expected: {
              ['formatOptions.timestampOutputFormat']: 'ISO8601_STRING',
            },
          },
          {
            name: 'TOF: omitted, UI64: omitted (default ISO8601_STRING)',
            opts: {},
            expected: {
              ['formatOptions.timestampOutputFormat']: 'ISO8601_STRING',
            },
          },
          {
            name: 'TOF: omitted, UI64: true',
            opts: {
              ['formatOptions.useInt64Timestamp']: true,
            },
            expected: {
              ['formatOptions.useInt64Timestamp']: true,
            },
          },
          {
            name: 'TOF: omitted, UI64: false (default ISO8601_STRING)',
            opts: {
              ['formatOptions.useInt64Timestamp']: false,
            },
            expected: {
              ['formatOptions.timestampOutputFormat']: 'ISO8601_STRING',
            },
          },
          {
            name: 'TOF: TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED, UI64: omitted (default ISO8601_STRING)',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED',
            },
            expected: {
              ['formatOptions.timestampOutputFormat']: 'ISO8601_STRING',
            },
          },
          {
            name: 'TOF: FLOAT64, UI64: omitted (error)',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'FLOAT64',
            },
            bail: true,
          },
          {
            name: 'TOF: INT64, UI64: omitted',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'INT64',
            },
            expected: {
              ['formatOptions.timestampOutputFormat']: 'INT64',
            },
          },
          {
            name: 'TOF: ISO8601_STRING, UI64: omitted (error)',
            opts: {
              ['formatOptions.timestampOutputFormat']: 'ISO8601_STRING',
            },
            bail: true,
          },
        ];

      testCases.forEach(testCase => {
        it(`should handle ${testCase.name}`, () => {

          // TODO: should buildQueryRequest_ throw an error on a bad combination ?
          // I don't think so, as showed with b/460198628, we should avoid that kind
          // of client side validation.
          const req = bq.buildQueryRequest_(QUERY_STRING, testCase.opts);

          const expectedReq = {
            query: QUERY_STRING,
            useLegacySql: false,
            requestId: req.requestId,
            jobCreationMode: 'JOB_CREATION_OPTIONAL',
            formatOptions: testCase.expected,
          };
          assert.deepStrictEqual(req, expectedReq);
        });
      });
    });
```

handwritten/bigquery/src/bigquery.ts
danieljbruce: It sounds like you would like to move the code back to where it was and avoid creating a new method so I have reverted the change.

handwritten/bigquery/system-test/high-precision-query-tests.ts
danieljbruce: > but those timestamp formats can definitely live together. Am I missing something ?

I put the tests in the file like you wanted. I think this is just a discussion about pros and cons. Not bloating bigquery.ts has pros, putting the tests with the other similar test has pros. We can likely move on at this point.

danieljbruce: >For the integration tests we can exercise a few scenarios like

I added three of the scenarios, but I'm not sure what you mean by `Passing timestampOutputFormat: INT64`.

handwritten/bigquery/test/bigquery.ts
danieljbruce: This has been updated now.

handwritten/bigquery/test/bigquery.ts
danieljbruce: Yes. This test case has actually been replaced by your most recent commit.

handwritten/bigquery/test/bigquery.ts
danieljbruce: Yes. This combination results in ISO8601_STRING.

handwritten/bigquery/test/bigquery.ts
danieljbruce: This was addressed in the previous review.

handwritten/bigquery/test/bigquery.ts
danieljbruce: It shouldn't be, but these tests were failing anyway. We need to get the CI set up for the monorepo to catch stuff like this automatically.

In any case, these tests have now been updated. Keep in mind that the exact same logic we used for the /data endpoint is now being applied to the /queries endpoint so if you see any problems with these tests then the same correction should be applied to the /data test as well.

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2666,6 +2701,11 @@ export class BigQueryRange {
   }
 
   private static fromStringValue_(value: string): [start: string, end: string] {
+    /*
```
alvarowolfx: nit: use jsdoc style comments if you want to document the method

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2684,14 +2724,27 @@ export class BigQueryRange {
...
-  static fromSchemaValue_(value: string, elementType: string): BigQueryRange {
+  static fromSchemaValue_(
+    value: string,
+    elementType: string,
+    listParams?:
+      | bigquery.tabledata.IListParams
+      | bigquery.jobs.IGetQueryResultsParams,
+  ): BigQueryRange {
+    /*
+    This method is only used by convertSchemaFieldValue and only when range
```
alvarowolfx: nit: same, use jsdoc style comments if you want to document the method

handwritten/bigquery/system-test/bigquery.ts
```diff
@@ -1498,6 +1503,159 @@ describe('BigQuery', () => {
               },
             );
           });
+          describe.only('High Precision Query System Tests', () => {
```
alvarowolfx: I think this was a unwanted `.only` here

handwritten/bigquery/test/bigquery.ts
```diff
@@ -3397,12 +3398,198 @@ describe('BigQuery', () => {
...
+          expected: {
+            timestampOutputFormat: 'ISO8601_STRING',
+          },
+          bail: true,
+        },
+      ];
+
+      testCases.forEach(testCase => {
+        it(`should handle ${testCase.name}`, () => {
+          // TODO: should buildQueryRequest_ throw an error on a bad combination ?
```
alvarowolfx: you didn't address this TODO here, we are not asserting the error cases and we I think we should not do that per reasons that I raised before.

danieljbruce: That's right. The errors only occur when a bad combination is sent to the server, but these are only unit tests on buildQueryRequest_. It turns out that none of the tests expect an error after the most recent set of changes.

handwritten/bigquery/system-test/bigquery.ts
danieljbruce: Yes. The `.only` has been removed.

handwritten/bigquery/src/bigquery.ts
danieljbruce: Yes. Updated.

handwritten/bigquery/src/bigquery.ts
danieljbruce: Yup. This has been updated.

handwritten/bigquery/test/bigquery.ts
```diff
@@ -3397,12 +3398,191 @@ describe('BigQuery', () => {
...
       assert.deepStrictEqual(req, expectedReq);
     });
 
+    describe('timestamp format options', () => {
+      const testCases: {
+        name: string;
+        opts: QueryOptions;
+        expected?: any;
+        bail?: boolean;
+      }[] = [
```
alvarowolfx, line 3413: nit:  what are we asserting ? 

* Not passing anything and expecting a default makes sense to me
* Passing X and expecting X makes sense too. But then passing Y and expecting Y, passing Z and expecting Z starts to get redundant

danieljbruce: I created https://b.corp.google.com/issues/483681188 to address this in a separate PR.

danieljbruce: https://github.com/googleapis/google-cloud-node/pull/7303/changes should address this issue

PR #7189, opened 2026-02-06
test: Use node 20 instead

danieljbruce:
Adding a test to pass the system-presubmit-node18 check.


quirogas:
@danieljbruce I'm assuming this PR no longer needs to be opened; please feel free to re-open it if I'm mistaken.

PR #7271, opened 2026-02-19
feat: expand workflows to receive both workflows and executions

sofisl:
based on this poorly worded albeit correct user report: https://github.com/googleapis/google-cloud-node/pull/6790#issuecomment-3869204945


packages/google-cloud-workflows-executions/librarian.js
miguelvelezsa, line 1: is this intended?

miguelvelezsa (approved):

PR #7305, opened 2026-02-25
fix(bin): harden README generator link handling

akgitrepos:
## Summary
- hardens support-documentation fallback logic in `bin/generate-readme.mjs` so 404 responses correctly fall back to `product_documentation`
- guards pagination parsing in `getRepos()` when `parse-link-header` returns `null` for malformed/unexpected Link headers
- adds focused unit tests for both edge cases to make regressions easy to catch

## Problem
The generator currently uses `validateStatus: () => true` for support docs HEAD checks. In that mode, HTTP 404 does not throw, so the fallback branch inside `catch` is skipped and broken support links may be retained.

`getRepos()` also assumes `parseLinkHeader(...)` always returns an object. If it returns `null`, `link.next` access can throw and stop README generation.

## Root Cause
1. **404 fallback mismatch**
   - request config intentionally disables throw-on-404
   - fallback logic relied primarily on exception path for 404
2. **Link parser nullability**
   - missing null guard before dereferencing `next`

### 1) 404 fallback correctness
- in `processMetadata()`, after HEAD request, explicitly check `res.status === 404` and mark `remoteUrlExists=false`
- preserve existing behavior for thrown 404s
- rethrow non-404 request errors (avoid swallowing unexpected failures)

### 2) Link-header null safety
- in `getRepos()`, guard with `if (link && link.next && link.next.url)` before reading next page URL

### 3) Testability improvements (no behavior change for CLI usage)
- export `processMetadata` and `getRepos` so they can be unit-tested directly
- allow dependency injection for request/parser/client in tests
- gate CLI execution with an `isCliExecution` check so importing the module does not auto-run `main()`
- keep token warning only for actual CLI invocation

### 4) New tests
Added `bin/generate-readme.test.mjs`:
- `processMetadata` falls back to `product_documentation` when support URL HEAD responds 404
- `getRepos` handles `parseLinkHeader` returning `null` without throwing and still returns filtered repos

## Validation
- `node --test bin/generate-readme.test.mjs`
- `node --check bin/generate-readme.mjs`
- `node bin/generate-readme.mjs --use-cache`

## Risk Assessment
Low.
- changes are isolated to `bin/generate-readme.mjs`
- default call paths are preserved (`downloadRepoMetadata -> getRepos()`, `main -> processMetadata()`)
- tests cover the newly hardened edge cases

### DIFF
--- a/bin/generate-readme.mjs
+++ b/bin/generate-readme.mjs
@@ -18,14 +18,17 @@ import { request, Gaxios } from 'gaxios';
 import figures from 'figures';
 import { readFileSync, writeFileSync } from 'fs';
 import parseLinkHeader from 'parse-link-header';
+import { pathToFileURL } from 'url';
 
 const token = process.env.GITHUB_TOKEN;
 const smokeTest = !!process.env.SMOKE_TEST;
+const isCliExecution =
+  !!process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href;
 
 const headers = {};
 if (token) {
   headers.authorization = `token ${token}`;
-} else {
+} else if (isCliExecution) {
   console.warn('Please provide GITHUB_TOKEN env var for increased quota.');
 }
 const baseUrl = 'https://api.github.com';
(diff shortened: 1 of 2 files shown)


akgitrepos:
@pearigee Would appreciate a review on this too. Thank you!

danieljbruce:
We'll need to formally triage this issue before merging PRs for it. Could you please open an issue for this? If we verify it on our end we can come back to this PR.

Thanks!

akgitrepos:
@danieljbruce Here's the issue that I have opened corresponding to this fix [Issue 7354](https://github.com/googleapis/google-cloud-node/issues/7354)

PR #7313, opened 2026-02-28
chore: update owlbot.pys

sofisl:
I noticed I didn't make the correct changes to the owlbot.py, and with Gautam we noticed the owlbot.py's have been silently failing. This goes back and fixes it.

tracking bug: https://github.com/googleapis/google-cloud-node/issues/7314


codyoss:
@sofisl What needs to still be done with this PR?

sofisl:
> @sofisl What needs to still be done with this PR?

Looks like the post-processor is still not running correctly (but at least it failed this time!). And the smell test is that we shouldn't have the top-level files changed but they were, so those also need to be reverted.

PR #7378, opened 2026-03-04
chore: nodejs spanner migration

sofisl:


.github/workflows/system-tests-against-emulator.yaml
```diff
@@ -22,7 +26,9 @@ jobs:
           node-version: 22
       - run: node --version
       - run: npm install
+        working-directory: handwritten/spanner
```
codyoss, line 29: nit: this is fine for this case, but in the future consider:

```
defaults:
  run:
    working-directory: handwritten/spanner
```

This way if more steps get added in the future we don't need to worry about adding `working-directory` in multiple places.

sofisl: good call, thank you! will do that in the future, for now just want to get this thing through :P

PR #7532, opened 2026-03-06
chore: test bun as a replacement for pnpm.

pearigee:
This is a benchmarking PR, please ignore.

The goal is to test how test performance varies when packages are installed with `bun` instead of `pnpm`!


pearigee:
Closing this for now to remove the distraction.

PR #7546, opened 2026-03-10
feat(storage): Implement robust path validation and structured skip reporting

thiyaguk09:
- Adds protection against path traversal (../) using normalized path resolution.
- Prevents Windows-style drive letter injection while allowing GCS timestamps.
- Implements directory jail logic to ensure absolute-style paths are relative to destination.
- Preserves backward compatibility by returning an augmented DownloadResponse array.
- Automates recursive directory creation for validated nested files.
- Adds a comprehensive 13-scenario test suite for edge-case parity.

### DIFF
--- a/handwritten/storage/package.json
+++ b/handwritten/storage/package.json
@@ -73,7 +73,7 @@
     "samples-test": "npm link && cd samples/ && npm link ../ && npm test && cd ../",
     "system-test:esm": "mocha build/esm/system-test --timeout 600000 --exit",
     "system-test": "mocha build/cjs/system-test --timeout 600000 --exit",
-    "test": "NODE_OPTIONS='--no-deprecation' c8 mocha build/cjs/test"
+    "test": "cross-env NODE_OPTIONS='--no-deprecation' c8 mocha build/cjs/test"
   },
   "dependencies": {
     "@google-cloud/paginator": "^5.0.0",
@@ -130,7 +130,8 @@
     "path-to-regexp": "6.3.0",
     "tmp": "^0.2.0",
     "typescript": "^5.1.6",
-    "yargs": "^17.3.1"
+    "yargs": "^17.3.1",
+    "cross-env": "^7.0.3"
   },
   "homepage": "https://github.com/googleapis/google-cloud-node/tree/main/handwritten/storage"
 }


handwritten/storage/src/transfer-manager.ts
```diff
@@ -571,8 +573,13 @@ export class TransferManager {
       options.concurrencyLimit || DEFAULT_PARALLEL_DOWNLOAD_LIMIT,
     );
     const promises: Promise<DownloadResponse>[] = [];
+    const skippedFiles: DownloadResponseWithStatus[] = [];
     let files: File[] = [];
 
+    const baseDestination = path.resolve(
+      options.prefix || options.passthroughOptions?.destination || '.'
```
krishnamd-jkp: Should we prioritize `options.passthroughOptions?.destination ` first before ` options.prefix`?

handwritten/storage/src/transfer-manager.ts
```diff
@@ -617,20 +651,33 @@ export class TransferManager {
...
-
-          return file.download(passThroughOptionsCopy);
-        }),
+        })
       );
     }
-
-    return Promise.all(promises);
+    const results = await Promise.all(promises);
+    return [...skippedFiles, ...results];
```
krishnamd-jkp: Maintain the input order. Results should be placed in the array at the same index as their corresponding input file. Otherwise this could potentially be a breaking change.

handwritten/storage/src/transfer-manager.ts
```diff
@@ -598,16 +605,43 @@ export class TransferManager {
...
+
+      let dest: string;
+      if (options.stripPrefix) {
+        dest = normalizedGcsName.replace(regex, '');
+      } else {
+        dest = path.join(
           options.prefix || '',
           passThroughOptionsCopy.destination || '',
-          file.name,
+          normalizedGcsName
```
krishnamd-jkp: Why are we combining prefix and destination here? , previously in line 580, you prioritised prefix over baseDestinations.

I think you need to only use baseDestination everywhere and combine it with the object name and do path traversal validations.

handwritten/storage/src/transfer-manager.ts
```diff
@@ -598,16 +605,43 @@ export class TransferManager {
...
         );
       }
-      if (options.stripPrefix) {
-        passThroughOptionsCopy.destination = file.name.replace(regex, '');
+
+      const resolvedPath = path.resolve(baseDestination, dest);
```
krishnamd-jkp, line 653: If dest is an absolute path, path.Resolve will return dest as is. But that is not the behavior we want. If object name is `/etc/passwd` and if the user given `options.passthroughOptions?.destination` is `/usr/downloads`, the final path should be `/usr/downloads/etc/passwd`. Can you check if this is happening in this scenario?

handwritten/storage/src/transfer-manager.ts
thiyaguk09: Switched to a pre-allocated array (`new Array(files.length)`) and internal index assignment (`finalResults[i]`). This ensures the output array strictly matches the input order, even with skips.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: Great catch. We addressed this by 'de-rooting' the object name before resolution. By using `.replace(/^\/+/, '')` on the GCS object name, we force it to be treated as a relative path.

> In your example, `/etc/passwd` becomes `etc/passwd`, and `path.resolve('/usr/downloads', 'etc/passwd')` correctly results in `/usr/downloads/etc/passwd` rather than escaping to the root.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: You're correct—combining them inside the loop was redundant and risky. I've refactored this to use `baseDestination` as the single source of truth, calculated once at the start.
> Now, we only resolve the 'de-rooted' object name against that `baseDestination`. This ensures we don't double-prefix paths and allows for a clean path traversal validation by checking if the final `resolvedPath` still resides within the `baseDestination` jail.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: Yes, agreed. I’ve updated the logic to prioritize `options.passthroughOptions.destination` as the primary local root. This ensures that a specific user-provided destination takes precedence over the broader batch `prefix`, which aligns better with the standard behavior of the individual `file.download()` method.

handwritten/storage/src/transfer-manager.ts
```diff
@@ -570,9 +572,13 @@ export class TransferManager {
     const limit = pLimit(
       options.concurrencyLimit || DEFAULT_PARALLEL_DOWNLOAD_LIMIT,
     );
-    const promises: Promise<DownloadResponse>[] = [];
+    const promises: Promise<void>[] = [];
     let files: File[] = [];
 
+    const baseDestination = path.resolve(
+      options.passthroughOptions?.destination || options.prefix || '.'
```
krishnamd-jkp: I don't think we should be using prefix when downloading to a directory. Prefix is just used to identify objects in the bucket. And moreover, if stripPrefix is used, we shouldn't be using prefix at all when downloading.

handwritten/storage/src/transfer-manager.ts
```diff
@@ -592,45 +598,95 @@ export class TransferManager {
...
+      const file = files[i];
+
+      const hasIllegalDrive = /^[a-zA-Z]:/.test(file.name);
+      if (hasIllegalDrive) {
+        const skippedResult = [Buffer.alloc(0)] as DownloadResponseWithStatus;
+        skippedResult.skipped = true;
+        skippedResult.reason = SkipReason.ILLEGAL_CHARACTER;
+        skippedResult.fileName = file.name;
+        finalResults[i] = skippedResult;
+        continue;
```
krishnamd-jkp, line 644: Do we need to explicitly test this or does the OS report an error anyways when such characters are requested to be downloaded?

handwritten/storage/src/transfer-manager.ts
```diff
@@ -592,45 +598,95 @@ export class TransferManager {
...
       }
-      if (options.stripPrefix) {
-        passThroughOptionsCopy.destination = file.name.replace(regex, '');
+
+      const normalizedGcsName = file.name
+        .replace(/\\/g, '/')
```
krishnamd-jkp: We do not want to change the file name by replacing '\' with '/' and treating it as linux directories. We should download without modifying the separators. Can you remove this?

handwritten/storage/src/transfer-manager.ts
```diff
@@ -592,45 +598,95 @@ export class TransferManager {
...
+        .replace(/^\/+/, '');
```
krishnamd-jkp: This can have a side effect is the above comment of replacing '\' with '/' is removed. Can you check if "\a\b\c.txt" will still be downloaded inside the baseDirectory after removing the replace operation? If not, can you find another way to resolve this?

handwritten/storage/src/transfer-manager.ts
```diff
@@ -592,45 +598,95 @@ export class TransferManager {
...
+
+      const dest = options.stripPrefix
```
krishnamd-jkp: If the above replace condition still exists this stripPrefix might also not work.

handwritten/storage/src/transfer-manager.ts
```diff
@@ -592,45 +598,95 @@ export class TransferManager {
...
+              dirResp.skipped = false;
+              dirResp.fileName = file.name;
+              dirResp.localPath = destination;
+              finalResults[i] = dirResp;
+              return;
+            }
+
+            await fsp.mkdir(path.dirname(destination), {recursive: true});
+
+            const resp = (await file.download(
```
krishnamd-jkp, line 703: Will this change make the download synchronous?

thiyaguk09: The downloads remain asynchronous and concurrent. We use `await` inside the task to ensure the local folder exists before that specific file starts downloading, but `p-limit` continues to manage multiple of these tasks in parallel up to the `concurrencyLimit`.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: You're right—the previous replacement was breaking the prefix match. Since we no longer modify internal separators, `stripPrefix` will now correctly match the original GCS object name provided by the user.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: Agreed. I’ve updated the logic to prioritize `passthroughOptions.destination` and default to the current directory (`.`) if omitted. This ensures `prefix` is only used for object discovery, not for defining the local download path.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: While the OS would eventually throw an error, explicitly checking for drive letters (e.g., `C:`) allows us to skip the file gracefully and return a clear `ILLEGAL_CHARACTER` status in the results array rather than a generic system exception.

If we let the OS handle it, the error would fall into our `catch` block as a generic failure. By testing it upfront, we can return a specific `SkipReason.ILLEGAL_CHARACTER` instead of a vague download error.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: I've removed the global backslash-to-slash replacement to preserve the original filename structure as requested. To prevent `\a\b\c.txt` from escaping the base directory on Windows, I've implemented 'manual de-rooting' using `.replace(/^[\\/]+/, '')`. This strips only leading separators, ensuring `path.resolve` always anchors the file inside our base directory.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: Agreed. I have removed the global replacement of `\` with `/`. This ensures original separators are preserved and the local file structure matches the GCS object name exactly.

handwritten/storage/src/transfer-manager.ts
```diff
@@ -592,45 +598,96 @@ export class TransferManager {
...
+        skippedResult.skipped = true;
+        skippedResult.reason = SkipReason.ILLEGAL_CHARACTER;
+        skippedResult.fileName = file.name;
+        finalResults[i] = skippedResult;
+        continue;
       }
-      if (options.stripPrefix) {
-        passThroughOptionsCopy.destination = file.name.replace(regex, '');
+
+      const normalizedGcsName = file.name.replace(/^[\\/]+/, '');
```
krishnamd-jkp: The stripPrefix operation should happen before this operation, right?

handwritten/storage/src/transfer-manager.ts
```diff
@@ -592,45 +598,96 @@ export class TransferManager {
...
+
+      const dest = options.stripPrefix
+        ? normalizedGcsName.replace(regex, '')
+        : normalizedGcsName;
+
+      const resolvedPath = path.resolve(baseDestination, dest);
+      const relativeFromBase = path.relative(baseDestination, resolvedPath);
+
+      const isOutside =
+        path.isAbsolute(relativeFromBase) ||
```
krishnamd-jkp, line 657: AbsolutePath can still be inside the baseDirectory, right?

handwritten/storage/src/transfer-manager.ts
thiyaguk09: Actually, no — By de-rooting first, we ensure the string starts exactly where the user expects, making the `stripPrefix` regex much more reliable.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: True, but `path.relative` only returns an absolute path if the two paths are on different drives (on Windows) or cannot be related. In those cases, the file is definitely `'outside'` our intended local jail. Including this check serves as a final safety catch for cross-drive traversals.

handwritten/storage/src/transfer-manager.ts
krishnamd-jkp: If object name is "/a/b/c.txt" and if user provides a stripPrefix: "/a/b", if you first remove the leading "/", the normalizedGcsName becomes "a/b/c.txt" after which the stripPrefix will not take any effect. Am I missing something here?

thiyaguk09: Actually it is a valid point, and I agree with your assessment. However, shifting this to the beginning may introduce additional challenges. I will explore further options and follow up with you.

That was a great observation.

thiyaguk09: Actually it's working as expected. I have updated the order of operations as suggested. The `stripPrefix` now runs against the raw GCS name first to ensure any leading slashes in the user's prefix match correctly. I then de-root the resulting string to ensure the final path is relative and stays within the destination 'jail' during `path.resolve`.

handwritten/storage/src/transfer-manager.ts
```diff
@@ -592,45 +598,96 @@ export class TransferManager {
...
+        continue;
       }
-      if (
-        options.skipIfExists &&
-        existsSync(passThroughOptionsCopy.destination || '')
-      ) {
+
+      if (options.skipIfExists && existsSync(resolvedPath)) {
         continue;
       }
```
krishnamd-jkp, line 678: Can you assign an empty result here as well with a skip reason? If you continue, you'll be leaving a hole in the final result array.

handwritten/storage/src/transfer-manager.ts
krishnamd-jkp, line 596: Can you add more notes here about the change in behavior. Previously we were returning only files which actually get downloaed. Previously the input request files size and returned results need not match but now the return result is exactly equal to input request files size. Also communicate the traversal vulnerability changes using a few comments here.

handwritten/storage/src/file.ts
```diff
@@ -396,6 +396,22 @@ export interface CopyCallback {
 
 export type DownloadResponse = [Buffer];
 
+export type DownloadResponseWithStatus = [Buffer] & {
```
krishnamd-jkp, line 399: QQ. This doesn't cause a breaking change to the existing customers because the return type is being changed? The existing users can still continue using this utility as is?

thiyaguk09: It is not a breaking change. The return type remains an array at its core, so existing code using index-based access or destructuring `(e.g., [buffer])` will continue to work. We are simply using a TypeScript intersection to attach additive metadata like `skipped` and `reason` to that array object.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: I have updated the logic to assign a skipped result with `SkipReason.ALREADY_EXISTS` to that index before continuing. This ensures the `finalResults` array remains fully populated and perfectly aligned with the input file array.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: I have updated the JSDoc.

v-pratap (approved):

PR #7548, opened 2026-03-10
feat(storage): implement Object Contexts with advanced filtering and validation

thiyaguk09:
- Add support for Object Contexts metadata in `File` and `Bucket` operations.
- Update `FileMetadata`, `CopyOptions`, and `CombineOptions` types to allow
  null values for context key deletion (PATCH semantics).
- Refactor `validateContexts` to accept a `contexts` object directly for 
  better consistency and simpler call patterns in `save`, `copy`, and `combine`.
- Implement server-side list filtering support via the `filter` query parameter 
  in `getFiles`, supporting NOT logic and existence wildcards.
- Ensure metadata inheritance and explicit overrides work correctly during 
  `File.copy` and `Bucket.combine`.
- Add comprehensive unit and system tests covering CRUD, server-side 
  operations, and complex filtering scenarios.


handwritten/storage/src/file.ts
```diff
@@ -4143,6 +4171,13 @@ class File extends ServiceObject<File, FileMetadata> {
     const options =
       typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
 
+    try {
```
Dhriti07: Can we have a helper for this logic since it seems to be repeated in many places.

thiyaguk09: Sure, I will look into it.

thiyaguk09: Added a shared validation helper to DRY logic.

PR #7589, opened 2026-03-11
feat(firestore): Add logical expressions

milaGGL:
Add new logical expressions `nor` and `switchOn`.

Ported from https://github.com/firebase/firebase-js-sdk/pull/9702


handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -8608,6 +8608,70 @@ export function or(
...
+ * @beta
+ * Creates an expression that performs a logical 'NOR' operation on multiple filter conditions.
+ *
+ * @example
+ * ```typescript
+ * // Check if neither the 'age' field is greater than 18 nor the 'city' field is "London"
+ * const condition = nor(
+ *   greaterThan("age", 18),
+ *   equal("city", "London")
+ * );
```
dlarocque, line 8621: In my opinion it's better to not have variable assignment in the examples. It's more concise, and it's rare to actually assign expressions to variables like that when using pipelines. I saw that the docs for `or` does this `condition = ...`- but I don't think we should follow that example. Since it's already there for `or`, feel free to just ignore this comment and keep it as is.

```suggestion
 * nor(
 *   greaterThan("age", 18),
 *   equal("city", "London")
 * );
```

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -8608,6 +8608,70 @@ export function or(
   return new FunctionExpression('or', [first, second, ...more]).asBoolean();
 }
 
+/**
+ * @beta
+ * Creates an expression that performs a logical 'NOR' operation on multiple filter conditions.
```
dlarocque, line 8613: Why are these called "filter conditions"?

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -8608,6 +8608,70 @@ export function or(
...
+ * If an odd number of arguments is provided, the final argument serves as a default fallback result.
+ * If no default is provided and no condition evaluates to true, it throws an error.
+ *
+ * @example
+ * ```typescript
+ * // Return "Active" if field "status" is 1, "Pending" if field "status" is 2,
+ * // and default to "Unknown" if none of the conditions are true.
+ * switchOn(
+ *   equal(field("status"), 1), "Active",
+ *   equal(field("status"), 2), "Pending",
```
dlarocque: `switchOn` doesn't accept strings, these must be expressions.
```suggestion
 *   equal(field("status"), 1), constant("Active"),
 *   equal(field("status"), 2), constant("Pending"),
```

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -8608,6 +8608,70 @@ export function or(
...
+  first: BooleanExpression,
+  second: BooleanExpression,
+  ...more: BooleanExpression[]
+): BooleanExpression {
+  return new FunctionExpression('nor', [first, second, ...more]).asBoolean();
+}
+
+/**
+ * @beta
+ * Creates an expression that evaluates to the result corresponding to the first true condition.
```
dlarocque, line 8639: Do you know if the result types must all be the same? 

Can we do something like
```
switchOn(
  equal(field("status"), 1), constant("Active") // string
  equal(field("status"), 2), constant(0) // int
)
```

handwritten/firestore/dev/src/pipelines/expression.ts
milaGGL: hmmm, why it is not changing the code when i select "apply suggestion"???

handwritten/firestore/dev/src/pipelines/expression.ts
milaGGL: `and` , `or` called them filter conditions, so `nor` followed the similar description. And, it do make sense as the conditions are used to filter fields.

handwritten/firestore/dev/src/pipelines/expression.ts
milaGGL: `or`,`and`,`xor` all used `const condition = ...`. Lets just keep it the same for `nor` for consistency.

handwritten/firestore/dev/src/pipelines/expression.ts
milaGGL: good question. yes we can, each switch cases are independent from each other, so the type is not required to be the same. So we can do sth like 
```
switchOn(
  equal(field("number"), "one"), constant(1),
  equal(field("number"), "two"), constant(2),
  constant("not a number")
)
```
Do you want me to add a test case for it?

dlarocque: no that's not necessary was just curious

PR #7604, opened 2026-03-13
fix(storage): handle early createReadStream close before piping

michaellatman:
## Summary
- guard `File#createReadStream()` when the returned stream is destroyed before the internal response pipeline is attached
- catch `ERR_STREAM_UNABLE_TO_PIPE` during pipeline setup and clean up the raw response stream/agent
- add a regression test that fails without the fix and passes with it

## Testing
- npm test -- --grep "ERR_STREAM_UNABLE_TO_PIPE"
- npm test -- --grep "createReadStream"

Fixes #7603


pearigee:
@googleapis/gcs-team Any comment on this PR? Otherwise, this PR will autoclose. Please take a look.

PR #7606, opened 2026-03-13
chore: google cloud node core migration

sofisl:
All of the actual changes are in this commit: https://github.com/googleapis/google-cloud-node/pull/7606/changes/5c6f7c9da0e2a67e0cf503dced5231cf845844c1

- Added separate workflows for generator tests
- Added in a whole separate set of google cloud build triggers for "core" and renamed the current triggers to "packages-"
- replaced the kaniko build step in our cloudbuild.yaml files with `cloud-builders/docker`. This was how google-cloud-node-core had it since the latter is currently maintained and a bit more lightweight
- updated the Dockerfile to actually use gcloud cli so that we can run google-cloud-node-core tests, and pnpm (since this test runner uses pnpm)
- updated the export/import trigger script for the new naming conventions (and deployed the triggers, they are on this PR right now)
- updated logic in run_conditional_tests to 1) only run core tests on core triggers, 2) updated ignore logic (it was a bit too permissive and skipping all core tests during runs), 3) added some logic to skip attempting to run tests if there wasn't a package.json, 4) exit early if there's one failure, 5) added logic to run interdependent tests (from the google-cloud-node-core library) for core packages
- cleaned up files in core/ (since it was its own monorepo, deduplicating)
- removed owlbot.py files. these libraries are all handwritten and I believe we will be moving a lot of these templates to the top-level of the library anyways, so they're not necessary.


quirogas:
https://github.com/googleapis/google-cloud-node-core/pull/928

fixes CI error

PR #7815, opened 2026-03-18
fix(spanner): override gcp-metadata to resolve punycode deprecation

alkatrivedi:
## Description
This PR adds overrides for `gcp-metadata` to version `^7.0.1` to resolve Node.js deprecation warnings.
### Issue
When running on **Node.js 22**, deep sub-dependencies (transitively imported via `gcp-metadata`) trigger the following warning:
`DeprecationWarning: The punycode module is deprecated. Please use a userland alternative instead.`
### Solution
Forcing `@google-cloud/spanner` to resolve `gcp-metadata` to `^7.0.1` ensures that updated versions of internal libraries (e.g., `gaxios`) are fetched, correctly bypassing the deprecated `punycode` modules.
### Changes
Added package manager overrides in [package.json](cci:7://file:///Users/gargsurbhi/Work/Github/google-cloud-node/handwritten/spanner/package.json:0:0-0:0):
- `"overrides"` for npm / yarn workspaces compatibility.
- `"pnpm.overrides"` for pnpm resolving.

BEGIN_COMMIT_OVERRIDE
fix(spanner): override gcp-metadata to resolve punycode deprecation](https://github.com/googleapis/google-cloud-node/pull/7815#top)
END_COMMIT_OVERRIDE


surbhigarg92:
> Just to be sure: This change does not change the minimum version of Node or any other major components that are required to use this library, right?

No. `gcp-metadata` package needs `"node": ">=18"` which we are already on.

PR #7816, opened 2026-03-18
feat(firestore): Defer pipeline user data validation to `execute()`

dlarocque:
To support standalone stage initializers (such as `subcollection()`), that can be constructed without an existing pipeline, we need to decouple stage initialization and user data validation. Standalone functions do not have access to `Firestore` or `databaseId` context when called, so validation must be deferred until the pipeline is executed with `execute(pipeline)`.

Implementation Details:
1. Removed user data parsing in stage methods (`select`, `where`, `addFields`, ...).
2. Added a `_validateUserData()` traversal to `Pipeline` class and all stages.
3. Modified `execute()` to trigger `pipeline._validateUserData()` before executing the pipeline.

**Breaking Change:**
Errors regarding invalid user data in stages will now be thrown when `execute(pipeline)` is called, rather than during the initialization of the stage (`pipeline.collection().addFields(/* invalid data */)`).
This aligns with the behaviour on both iOS, Android, and Java SDKs.
```typescript
// Previously: error threw here
const ppl = firestore.pipeline().collection('books').select(field, undefined); // undefined is an unsupported field value

// Now: error throws here
await execute(ppl);
```

Ported from: https://github.com/firebase/firebase-js-sdk/pull/9715

I also updated the ESLint config rules to allow args to be unused if their identifier starts with an underscore (`_name`). Previously the rule was that they can only be underscores (`_`), which causes issues if there's more than one unused argument. I ran into this issue when implementing the `_validateUserData(name: string, ignoreUndefinedProperties: boolean)` methods.

### DIFF
--- a/handwritten/firestore/dev/src/pipelines/pipelines.ts
+++ b/handwritten/firestore/dev/src/pipelines/pipelines.ts
@@ -1590,7 +1590,9 @@ export class Pipeline implements firestore.Pipelines.Pipeline {
     }
 
     // Validates user data in the entire pipeline
-    this._validateUserData('execute');
+    this._validateUserData(
+      this.db._settings.ignoreUndefinedProperties ?? false,
+    );
 
     const util = new ExecutionUtil(this.db, this.db._serializer!);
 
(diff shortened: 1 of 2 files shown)


handwritten/firestore/types/firestore.d.ts
```diff
@@ -11588,6 +11588,10 @@ declare namespace FirebaseFirestore {
      * ```
      */
     export class Pipeline {
+      /**
+       * @internal
+       */
+      _validateUserData(name: string, ignoreUndefinedProperties: boolean): void;
```
MarkDuckworth: It's better to put this internal method on the implementation of Pipeline and not the type definition. This doesn't have to block as we can move it later.

dlarocque: Do we want to avoid putting this on the type, because we're worried users would use it?

I'm not sure how to implement this.
The types for the stage classes reference this `Pipeline` type, so if we remove `_validateUserData` from `Pipeline`, when we try to access `Pipeline._validateUserData`, TS will throw a compiler error for referencing a property that does not exist, even if it is defined on the `Pipeline` instance at runtime.

```
Property '_validateUserData' does not exist on type 'Pipeline'.
```

I want to avoid just silencing the problem by casting to `as any`.

dlarocque: Yeah I think the only way is to cast to `as any`. I think the ideal solution would be to have both private and public dts files like in the JS SDK.

dlarocque: Done

PR #7817, opened 2026-03-18
fix(pubsub): remove messages from leasing on nackWithResponse

feywind:
Re-opening a fix for this in the new repo.

Fixes https://github.com/googleapis/google-cloud-node/issues/7625 🦕

### DIFF
--- a/handwritten/pubsub/test/subscriber.ts
+++ b/handwritten/pubsub/test/subscriber.ts
@@ -924,7 +924,6 @@ describe('Subscriber', () => {
       await subscriber.nack(message);
 
       const [msg, deadline] = stub.lastCall.args;
-
       assert.strictEqual(msg, message);
       assert.strictEqual(deadline, 0);
     });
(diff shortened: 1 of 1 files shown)


handwritten/pubsub/test/subscriber.ts
```diff
@@ -929,6 +929,17 @@ describe('Subscriber', () => {
       assert.strictEqual(deadline, 0);
     });
 
+    it('should modAck the message with a 0 deadline (nackWithResponse)', async () => {
+      const stub = sandbox.stub(subscriber, 'modAckWithResponse');
+
```
pearigee, line 933: This is a super nit of me, but Google's unit test best practices (see go/unit-testing-practices?polyglot=typescript#structure) recommend that unit tests be structured into clear arrange, act, and assert  steps with whitespace.

I think that would make your test look like this:
```
      const stub = sandbox.stub(subscriber, 'modAckWithResponse');
      
      await subscriber.nackWithResponse(message);

      const [msg, deadline] = stub.lastCall.args;
      assert.strictEqual(msg, message);
      assert.strictEqual(deadline, 0);
```

feywind: Oh man... Google best practices folks, meet googleapis 😹

It does seem like a good idea, but fixing up existing tests will be an uphill battle, I think.

PR #7819, opened 2026-03-18
feat(pubsub): add support for streaming pull keepalives from the server

feywind:
Adds the keepalive support.

Fixes https://github.com/googleapis/google-cloud-node/issues/7818 🦕


handwritten/pubsub/src/message-stream.ts
```diff
@@ -347,6 +373,8 @@ export class MessageStream extends PassThrough {
       maxOutstandingBytes: this._subscriber.useLegacyFlowControl
         ? 0
         : this._subscriber.maxBytes,
+      clientId: 'node-pubsub',
```
hongalex: I think clientID should be a UUID that differentiates this specific client instantiation from others. See [Go](https://github.com/googleapis/google-cloud-go/blob/790bab51084e22725bdb2daaa2870d862dc453c6/pubsub/v2/subscriber.go#L72)

handwritten/pubsub/src/message-stream.ts
```diff
@@ -264,10 +270,30 @@ export class MessageStream extends PassThrough {
     // Mark this stream as alive again. (reset backoff)
     const tracker = this._streams[index];
     this._retrier.reset(tracker);
+    this._resetPingTimer(index);
 
     this.emit('data', data);
   }
 
+  private _resetPingTimer(index: number): void {
```
hongalex: I haven't thought about the exact practical differences between what's written here and what's in the [Java client](https://github.com/googleapis/java-pubsub/pull/2605/changes#diff-4fd0fb821e0087d9d1baeed04ae60c2002aaa15cabda1b33a9037a3246e7f96aR137), but I think we should probably just mimic the behavior of the latter.

feywind: This is updated now. Take a look?

feywind:
We're not proceeding with this for now.

handwritten/pubsub/src/message-stream.ts
```diff
@@ -263,11 +272,51 @@ export class MessageStream extends PassThrough {
...
+    const now = Date.now();
+    const lastPingTime = tracker.lastPingTime ?? -1;
+    const lastResponseTime = tracker.lastResponseTime ?? 0;
+    if (lastPingTime <= lastResponseTime) {
+      return;
+    }
+
+    const elapsedSincePing = now - lastPingTime;
+
+    if (elapsedSincePing > 15000) {
```
pearigee: Should this be a constant or configurable item?

handwritten/pubsub/src/message-stream.ts
```diff
@@ -263,11 +272,51 @@ export class MessageStream extends PassThrough {
...
+    }
+  }
+
+  private _setAliveTimer(index: number): void {
+    const tracker = this._streams[index];
+    this._clearAliveTimer(tracker);
+
+    tracker.aliveTimer = setInterval(() => {
+      this._checkAliveTimer(index);
+    }, 10000);
```
pearigee: Should this also be a constant?

pearigee (approved):

handwritten/pubsub/src/message-stream.ts
```diff
@@ -263,11 +272,51 @@ export class MessageStream extends PassThrough {
...
+  }
```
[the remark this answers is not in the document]
feywind: I agree with this conceptually, though it doesn't match the Java impl after that change. I am okay with this, it should do the same thing.

handwritten/pubsub/src/message-stream.ts
feywind: It's fixed, but I like the idea of making it a constant.

handwritten/pubsub/src/message-stream.ts
feywind: It looks like Gemini's suggestion got rid of it anyway.

PR #7826, opened 2026-03-19
feat(firestore): Add timestamp expressions

yvonnep165:
Add new timestamp expressions timestamp_diff and timestamp_extract and integration tests for the expressions

The expression timestamp_trunc was already ported in the repo with integration tests. Amend the example usage documentation for standalone methods.

Ported from https://github.com/firebase/firebase-js-sdk/pull/9728


handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -2429,6 +2429,111 @@ export abstract class Expression
...
+      | 'minute'
+      | 'hour'
+      | 'day'
+      | Expression
+  ): FunctionExpression {
+    return new FunctionExpression(
+      'timestamp_diff',
+      [this, fieldOrExpression(start), valueToDefaultExpr(unit)],
+    );
+  }
```
dlarocque, line 2449: Can we move the `union` string union values into a type?

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -2429,6 +2429,111 @@ export abstract class Expression
...
+   *
+   * @param part - The part to extract from the timestamp (e.g., "year", "month", "day").
+   * @param timezone - The timezone to use for extraction. Valid values are from
+   * the TZ database (e.g., "America/Los_Angeles") or in the format "Etc/GMT-1."
+   * @returns A new `Expression` representing the extracted part as an integer.
+   */
+  timestampExtract(
+    part: firestore.Pipelines.TimePart,
+    timezone?: string | Expression
+  ): FunctionExpression;
```
dlarocque, line 2470: I think we should add documentation explaining how this expression behaves when `timestamp` is not passed.

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -2429,6 +2429,111 @@ export abstract class Expression
...
+   */
+  timestampExtract(
+    part: Expression,
+    timezone?: string | Expression
+  ): FunctionExpression;
+  timestampExtract(
+    part: firestore.Pipelines.TimePart | Expression,
+    timezone?: string | Expression
+  ): FunctionExpression {
+    const internalPart = isString(part) ? part.toLowerCase() : part;
```
dlarocque, line 2496: Can we remove this line, since `TimePart` should already be lower case?

handwritten/firestore/dev/src/pipelines/expression.ts
yvonnep165: Since existing timestampAdd and timestampSub expressions use the same unit value, will do refactoring in a separate PR after this one.

handwritten/firestore/dev/src/pipelines/expression.ts
yvonnep165: `TimePart` includes all the values from `TimeGranularity` and at this moment `TimeGranularity` still have value that's not lower case like `isoWeek`, which I plan to fix when refactoring the `unit` variables as mentioned above. We can remove it after I change all the values to lower case?

handwritten/firestore/dev/src/pipelines/expression.ts
yvonnep165: Add "Defaults to "UTC" if not specified." in the timezone parameter discription

handwritten/firestore/api-report/firestore.api.md
```diff
@@ -1207,6 +1207,10 @@ abstract class Expression implements firestore.Pipelines.Expression, HasUserData
     sum(): AggregateFunction;
     timestampAdd(unit: Expression, amount: Expression): FunctionExpression;
     timestampAdd(unit: 'microsecond' | 'millisecond' | 'second' | 'minute' | 'hour' | 'day', amount: number): FunctionExpression;
+    timestampDiff(start: Expression, unit: Expression): FunctionExpression;
+    timestampDiff(start: string | Expression, unit: 'microsecond' | 'millisecond' | 'second' | 'minute' | 'hour' | 'day'): FunctionExpression;
```
milaGGL, line 1211: I assume this is going to be replaced by the new types u r refactoring.

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -2396,6 +2396,111 @@ export abstract class Expression
...
+   * @beta
+   * Creates an expression that calculates the difference between this timestamp and another timestamp.
+   *
+   * @example
+   * ```typescript
+   * // Calculate the difference in days between 'endTime' and 'startTime' fields.
+   * field("endTime").timestampDiff("startTime", "day");
+   * ```
+   *
+   * @param start - The field name of the starting timestamp.
```
milaGGL: the param description doesn't match the type "string | Expression".

handwritten/firestore/api-report/firestore.api.md
yvonnep165: yes it will be changed by that

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -9508,6 +9613,237 @@ export function timestampTruncate(
...
+): FunctionExpression;
+
+/**
+ * @beta
+ * Creates an expression that calculates the difference between two timestamps.
+ *
+ * @example
+ * ```typescript
+ * // Calculate the difference in days between 'endTime' field and a starting timestamp expression.
+ * timestampDiff('endTime', field('startTime'), 'day')
```
milaGGL, line 9651: putting the "end" first, "start" the second, this param order feels bit weird. But, since backend proto ordered the params like this + examples are documented, should be fine.

handwritten/firestore/dev/src/pipelines/expression.ts
yvonnep165: Oh good catch! Changed!

PR #7839, opened 2026-03-20
feat(firestore): Add support for subqueries

dlarocque:
Adds support for Pipeline subqueries.

Ported from: https://github.com/firebase/firebase-android-sdk/pull/7736

API Proposal: [go/firestore-pipelines-subquery](http://goto.google.com/firestore-pipelines-subquery)

### DIFF
--- a/handwritten/firestore/dev/system-test/pipeline.ts
+++ b/handwritten/firestore/dev/system-test/pipeline.ts
@@ -6432,39 +6432,8 @@ describe.skipClassic('Pipeline class', () => {
           .addFields(reviewsSub.toArrayExpression().as('reviews'))
           .select('title', 'reviews');
 
-        // TODO(dlarocque): Remove these target backend conditionals once the 'get_field' rename has rolled out to prod.
-        const host = firestore._settings.host;
-        const isProd = host === 'firestore.googleapis.com';
-        const isNightly = host === 'test-firestore.sandbox.googleapis.com';
-
-        if (isProd) {
-          // The execution of this pipeline is expected to result in a network error
-          // from the backend until the breaking change to rename the 'field' expression to
-          // 'get_field' has rolled out to prod.
-          try {
-            const results = await ppl.execute();
-
-            // If this is reached, the execution of the pipeline didn't throw an error, so the
-            // breaking change must have rolled out to prod. We can assert the newly expected behaviour.
-            console.warn(
-              "The 'get_field' expression rename has rolled out to the prod backend. Remove the target backend conditionals in this test.",
-            );
-            expectResults(results, {title: '1984', reviews: ['Alice']});
-          } catch (err: unknown) {
-            const error: Error = err as Error;
-            expect(error.message).to.equals(
-              "Request failed with error: The function 'get_field' does not exist, did you mean 'field'?",
-            );
-          }
-        } else if (isNightly) {
-          const results = await ppl.execute();
-          expectResults(results, {title: '1984', reviews: ['Alice']});
-        } else {
-          expect(false).to.equal(
-            true,
-            `This test is only expected to run against firestore.googleapis.com or test-firestore.sandbox.googleapis.com, but it instead ran against ${host}`,
-          );
-        }
+        const results = await ppl.execute();
+        expectResults(results, {title: '1984', reviews: ['Alice']});
       });
     });
 
@@ -6541,39 +6510,8 @@ describe.skipClassic('Pipeline class', () => {
           .addFields(reviewsSub.toArrayExpression().as('reviews'))
           .select('title', 'reviews');
 
-        // TODO(dlarocque): Remove these target backend conditionals once the 'get_field' rename has rolled out to prod.
-        const host = firestore._settings.host;
-        const isProd = host === 'firestore.googleapis.com';
-        const isNightly = host === 'test-firestore.sandbox.googleapis.com';
-
-        if (isProd) {
-          // The execution of this pipeline is expected to result in a network error
-          // from the backend until the breaking change to rename the 'field' expression to
-          // 'get_field' has rolled out to prod.
-          try {
-            const results = await ppl.execute();
-
-            // If this is reached, the execution of the pipeline didn't throw an error, so the
-            // breaking change must have rolled out to prod. We can assert the newly expected behaviour.
-            console.warn(
-              "The 'get_field' expression rename has rolled out to the prod backend. Remove the target backend conditionals in this test.",
-            );
-            expectResults(results, {title: '1984', reviews: []});
-          } catch (err: unknown) {
-            const error: Error = err as Error;
-            expect(error.message).to.equals(
-              "Request failed with error: The function 'get_field' does not exist, did you mean 'field'?",
-            );
-          }
-        } else if (isNightly) {
-          const results = await ppl.execute();
-          expectResults(results, {title: '1984', reviews: []});
-        } else {
-          expect(false).to.equal(
-            true,
-            `This test is only expected to run against firestore.googleapis.com or test-firestore.sandbox.googleapis.com, but it instead ran against ${host}`,
-          );
-        }
+        const results = await ppl.execute();
+        expectResults(results, {title: '1984', reviews: []});
       });
     });
 


handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -10237,6 +10255,190 @@ export function isType(
...
+ */
+export function getField(fieldName: string, key: string): Expression;
+/**
+ * @public
+ * Creates an expression that returns the value of a field from the document with the given field name.
+ *
+ * @example
+ * ```typescript
+ * // Get the value of the "city" field in the "address" document.
+ * getField("address", variable("addressField"))
```
cherylEnkidu, line 10309: I would prefer not introduce the idea of `variable` in the `getField` code sample. It is not intuitive what is the concept of variable here.

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -10237,6 +10255,190 @@ export function isType(
...
+ */
+export function getField(expression: Expression, key: string): Expression;
+/**
+ * @public
+ * Creates an expression that returns the value of a field from a document that results from the evaluation of the expression.
+ *
+ * @example
+ * ```typescript
+ * // Get the value of the key resulting from the "addressField" variable in the "address" document.
+ * getField(field("address"), variable("addressField"))
```
cherylEnkidu: I would prefer not introduce the idea of `variable` in the `getField` code sample. It is not intuitive what is the concept of variable here.

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -10789,6 +10807,190 @@ export function isType(
   return fieldOrExpression(fieldNameOrExpression).isType(type);
 }
 
+/**
+ * @public
+ * Creates an expression that returns the value of a field from a document that results from the evaluation of the expression.
```
MarkDuckworth: consider something like

> Creates an expression that gets a field from a map (object).

Firestore documentation uses "map" to refer to the data type, not document/sub-document.

MarkDuckworth: https://firebase.google.com/docs/firestore/manage-data/data-types

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -10789,6 +10807,190 @@ export function isType(
...
+ *
+ * @example
+ * ```typescript
+ * // Get the value of the "city" field in the "address" document.
+ * getField(field("address"), "city")
+ * ```
+ *
+ * @param expression The expression representing the document.
```
MarkDuckworth: "an expression evaluating to the map from which the field will be extracted"

handwritten/firestore/dev/src/pipelines/expression.ts
```diff
@@ -10789,6 +10807,190 @@ export function isType(
...
+ * @public
+ * Creates an expression that represents the current document being processed.
+ *
+ * @example
+ * ```typescript
+ * // Define the current document as a variable "doc"
+ * firestore.pipeline().collection("books")
+ *     .define(currentDocument().as("doc"))
+ *     // Access a field from the defined document variable
+ *     .select(variable("doc").mapGet("title"));
```
MarkDuckworth: should we use `getField('title')` instead of mapGet?

MarkDuckworth: I forget, what is happening to `mapGet`?

handwritten/firestore/dev/src/pipelines/pipelines.ts
```diff
@@ -513,6 +519,216 @@ export class Pipeline implements firestore.Pipelines.Pipeline {
     return this._addStage(new RemoveFields(innerOptions));
   }
 
+  /**
+   * @public
+   * Defines one or more variables in the pipeline's scope. `define` is used to bind a value to a
```
MarkDuckworth: I still think it would be valuable to state that it binds the variable to the pipeline's current document

For example considering the following 

```
db.pipeline().collection("foo")
  .define(currentTimestamp().as("executionTime"))
  .addField(variable("executionTime").as("executionTime"))
```

If the customer reads this documentation as "define binds a variable to the pipeline", they they may assume that `.define(currentTimestamp().as("executionTime"))` is executed once for the pipeline.

But the reality is that `.define(currentTimestamp().as("executionTime"))` is executed for each document and binds a different value of executionTime for each document.

handwritten/firestore/dev/src/pipelines/pipelines.ts
```diff
@@ -513,6 +519,216 @@ export class Pipeline implements firestore.Pipelines.Pipeline {
...
+      ...options,
+      variables: convertedExpressions,
+    };
+
+    return this._addStage(new Define(internalOptions));
+  }
+
+  /**
+   * @public
+   * Converts this Pipeline into an expression that evaluates to an array of results.
```
MarkDuckworth: Consider "array of map (objects)" for consistent data types.

MarkDuckworth: "Converts this Pipeline into an expression that evaluates to an array of maps (objects), where each result document of the pipeline is represented as a map in the returned array."

handwritten/firestore/dev/src/pipelines/pipelines.ts
```diff
@@ -513,6 +519,216 @@ export class Pipeline implements firestore.Pipelines.Pipeline {
...
+   * </ul>
+   *
+   * @example
+   * ```typescript
+   * // Get a list of reviewers for each book
+   * db.pipeline().collection("books")
+   *     .define(field("id").as("book_id"))
+   *     .addFields(
+   *         db.pipeline().collection("reviews")
+   *             .where(field("book_id").equal(variable("book_id")))
```
MarkDuckworth: might clarify things to choose a field name that does not match the variable name.

```
  ...
  .define(field("id").as("current_book_id"))
  ...
      .where(field("book_id").equal(variable("current_book_id")))
```

handwritten/firestore/dev/src/pipelines/pipelines.ts
```diff
@@ -513,6 +519,216 @@ export class Pipeline implements firestore.Pipelines.Pipeline {
...
+   *  <li>If the item has a single field, its value is unwrapped and returned directly.</li>
+   *  <li>If the item has multiple fields, they are returned as an object.</li>
+   * </ul>
+   *
+   * @example
+   * ```typescript
+   * // Calculate average rating for a restaurant
+   * db.pipeline().collection("restaurants").addFields(
+   *   db.pipeline().collection("reviews")
+   *     .where(field("restaurant_id").equal(variable("rid")))
```
MarkDuckworth: where is rid defined?

handwritten/firestore/dev/src/pipelines/pipelines.ts
```diff
@@ -513,6 +519,216 @@ export class Pipeline implements firestore.Pipelines.Pipeline {
...
+   *   "average_rating": 4.5
+   * }
+   * ```
+   *
+   * Multiple Fields:
+   * ```typescript
+   * // Calculate average rating AND count for a restaurant
+   * db.pipeline().collection("restaurants").addFields(
+   *   db.pipeline().collection("reviews")
+   *     .where(field("restaurant_id").equal(variable("rid")))
```
MarkDuckworth: same

handwritten/firestore/types/firestore.d.ts
MarkDuckworth, line 1: Any changes you decide to make in the ref doc above would also need to be changed here

MarkDuckworth (approved):
LGTM with suggestions for ref doc cleanup

handwritten/firestore/dev/src/pipelines/expression.ts
dlarocque: Thanks! done

handwritten/firestore/dev/src/pipelines/expression.ts
dlarocque: Yes, good idea

handwritten/firestore/dev/src/pipelines/pipelines.ts
dlarocque: I believe the variable is binded to the current document only if the variable was created with `currentDocument()`.

When I run this pipeline which doesn't use `currentDocument`:

```typescript
        const results = await firestore
          .pipeline()
          .collection(randomCol.path)
          .define(currentTimestamp().as('executionTime'))
          .addFields(variable('executionTime').as('docExecTime'))
          .execute();
```

the result has 13 results, and each `docExecTime` is the same, 
`docExecTime":{"timestampValue":{"seconds":"1775580758","nanos":166665000}`

MarkDuckworth: maybe the currentTimestamp example is a bad example. Random would have been better, but I don't think we have a random expression yet. Regardless, the expression passed to define is evaluated for each input document and a new variable `executionTime` is bound to the document.

Let's revisit this later and not let this block. If a documentation change is made, it probably needs to be made for all SDKs.

dlarocque: Oh you're right- if I use `rand()` i get different results for each doc.

dlarocque: Looking more closely at this, I think there were more inaccuracies.
I updated the docs to:

```typescript
  /**
   * Binds one or more expressions to variable names within the pipeline's scope.
   *
   * The `define` stage establishes a variable environment for the pipeline. It assigns
   * the provided expressions to specific aliases. These variables remain in scope for all
   * subsequent stages (and any nested subqueries), where they can be referenced using the
   * `variable()` function.
   *
   * This is primarily used to improve query ergonomics by preventing the duplication of
   * complex expression trees, or to explicitly pass state from an outer pipeline into an
   * inner subquery.
   *
   * @example
   * ```typescript
   * // Bind a mathematical expression to a variable to cleanly reference it multiple times.
   * db.pipeline().collection("products")
   *   .define(
   *     field("price").multiply(0.8).as("discountedPrice")
   *   )
   *   .where(variable("discountedPrice").lessThan(50))
   *   .select("name", variable("discountedPrice"));
   * ```
   *
   * @param aliasedExpression - The first expression to bind to a variable.
   * @param additionalExpressions - Optional additional expressions to bind to a variable.
   * @returns A new Pipeline object with this stage appended to the stage list.
   */
```

Changes
- instead of defining "values", say "expressions". SDK itself doesn't define values, it's just building the AST
- removed phrase "intermediate calculations". I'm not sure if the backend actually stores variables as "intermediate calculations". It may just inline the expressions instead.
- more precisely saying it "improves ergonomics by preventing duplication of complex expression trees" instead of the slightly vague "improving readability and maintainability"

handwritten/firestore/dev/src/pipelines/pipelines.ts
```diff
@@ -513,6 +519,216 @@ export class Pipeline implements firestore.Pipelines.Pipeline {
     return this._addStage(new RemoveFields(innerOptions));
   }
 
+  /**
+   * Defines one or more variables in the current document's scope. `define` is used to bind a value to a
+   * variable for internal reuse within the pipeline body (accessed via the `variable()` function).
```
MarkDuckworth: Overall a very nice improvement to the docs IMO. 

I want to call out that this definition uses the term "document's scope", and the one below uses "pipeline's scope". Probably should make these consistent.

handwritten/firestore/dev/src/pipelines/pipelines.ts
```diff
@@ -513,6 +519,216 @@ export class Pipeline implements firestore.Pipelines.Pipeline {
...
+   * @param additionalExpressions - Optional additional expressions to bind to a variable.
+   * @returns A new Pipeline object with this stage appended to the stage list.
+   */
+  define(
+    aliasedExpression: firestore.Pipelines.AliasedExpression,
+    ...additionalExpressions: firestore.Pipelines.AliasedExpression[]
+  ): Pipeline;
+  /**
+   * Defines one or more variables in the pipeline's scope. `define` is used to bind a value to a
+   * variable for internal reuse within the pipeline body (accessed via the `variable()` function).
```
MarkDuckworth: This definition uses the term "pipeline's scope", and the one below uses "document's scope". Probably should make these consistent.

handwritten/firestore/dev/src/pipelines/pipelines.ts
```diff
@@ -513,6 +519,216 @@ export class Pipeline implements firestore.Pipelines.Pipeline {
...
+   * This stage is useful for declaring reusable values or intermediate calculations that can be
+   * referenced multiple times in later parts of the pipeline.
+   *
+   * @example
+   * ```typescript
+   * db.pipeline().collection("products")
+   *   .define(
+   *     field("price").multiply(0.9).as("discountedPrice"),
+   *     field("stock").add(10).as("newStock")
+   *   )
```
MarkDuckworth: This sample may need to be updated to use the options object

handwritten/firestore/types/firestore.d.ts
```diff
@@ -12246,6 +12376,186 @@ declare namespace FirebaseFirestore {
...
+       *
+       * This stage is useful for declaring reusable values or intermediate calculations that can be
+       * referenced multiple times in later parts of the pipeline.
+       *
+       * @example
+       * ```typescript
+       * db.pipeline().collection("products")
+       *   .define(
+       *     field("price").multiply(0.9).as("discountedPrice"),
+       *     field("stock").add(10).as("newStock")
```
MarkDuckworth: Should show use of the options object.

PR #7848, opened 2026-03-23
feat(bigquery): allow the user to ask for skipping parsing rows when quering

rafal-hawrylak:
**Summary**
This PR introduces a new optional boolean flag, skipParsing, to the BigQuery.query, Table.getRows, and Job.getQueryResults methods. 

When skipParsing is set to true, the library bypasses the internal mergeSchemaWithRows_ logic. Instead of returning idiomatic JavaScript objects (where BigQuery types are mapped to JS types), the rows result will contain the raw BigQuery API wire format (e.g., an array of { f: [{ v: '...' }] } objects).

**Motivation**
Providing access to raw rows offers two primary benefits:

1. Performance: For very large result sets, the overhead of merging the schema with rows and performing type conversions can be significant. Bypassing this step allows for faster processing when the consumer handles their own parsing.
2. Flexibility: Users who require the specific BigQuery JSON structure for downstream systems or custom serialization can now access it directly without the library's opinionated mapping.

### DIFF
--- a/handwritten/bigquery/src/bigquery.ts
+++ b/handwritten/bigquery/src/bigquery.ts
@@ -128,6 +128,7 @@ export type Query = JobRequest<bigquery.IJobConfigurationQuery> & {
   pageToken?: string;
   wrapIntegers?: boolean | IntegerTypeCastOptions;
   parseJSON?: boolean;
+  skipParsing?: boolean;
   // Overrides default job creation mode set on the client.
   jobCreationMode?: JobCreationMode;
 };
@@ -2217,6 +2218,7 @@ export class BigQuery extends Service {
         ? {
             wrapIntegers: query.wrapIntegers,
             parseJSON: query.parseJSON,
+            skipParsing: query.skipParsing,
           }
         : {};
     const callback =
@@ -2268,12 +2270,16 @@ export class BigQuery extends Service {
               'formatOptions.useInt64Timestamp':
                 queryReq.formatOptions?.useInt64Timestamp,
             };
-            rows = BigQuery.mergeSchemaWithRows_(res.schema, res.rows, {
-              wrapIntegers: options.wrapIntegers || false,
-              parseJSON: options.parseJSON,
-              listParams,
-            });
-            delete res.rows;
+            if (options.skipParsing) {
+              rows = res.rows;
+            } else {
+              rows = BigQuery.mergeSchemaWithRows_(res.schema, res.rows, {
+                wrapIntegers: options.wrapIntegers || false,
+                parseJSON: options.parseJSON,
+                listParams,
+              });
+              delete res.rows;
+            }
           } catch (e) {
             (callback as SimpleQueryRowsCallback)(e as Error, null, job);
             return;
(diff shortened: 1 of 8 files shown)


handwritten/bigquery/src/bigquery.ts
```diff
@@ -2268,12 +2270,18 @@ export class BigQuery extends Service {
...
+            if (options.skipParsing) {
+              rows = res.rows;
+            } else {
+              rows = BigQuery.mergeSchemaWithRows_(res.schema, res.rows, {
+                wrapIntegers: options.wrapIntegers || false,
+                parseJSON: options.parseJSON,
+                listParams,
+              });
+            }
+            if (!options.returnRawRows && !options.skipParsing) {
```
pearigee: I don't see returnRawRows as a property in the options interface of this file. Also, I don't see this property being used at all in bigquery.ts other than here?

Do we intended to expose returnRawRows in the public interface? If not, do we need it at all?

handwritten/bigquery/src/job.ts
```diff
@@ -51,6 +51,8 @@ export type QueryResultsOptions = {
   job?: Job;
   wrapIntegers?: boolean | IntegerTypeCastOptions;
   parseJSON?: boolean;
+  returnRawRows?: boolean;
```
pearigee: Do we need returnRawRows? Is it really just to avoid the delete block? Does this need to be separate from skipParsing, or could that be enough for your usecase?

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2268,12 +2270,18 @@ export class BigQuery extends Service {
...
+              rows = res.rows;
+            } else {
+              rows = BigQuery.mergeSchemaWithRows_(res.schema, res.rows, {
+                wrapIntegers: options.wrapIntegers || false,
+                parseJSON: options.parseJSON,
+                listParams,
+              });
+            }
+            if (!options.skipParsing) {
+              delete res.rows;
```
pearigee, line 2281: (Optional) Should this just be moved into the else block above? It seems as though the actions of the else block and the row deletion should always occur together.

```
else {
   rows = BiqQuery.mergeWithSchemaRows_(....)
   delete res.rows;
}
```

handwritten/bigquery/src/job.ts
```diff
@@ -602,10 +603,12 @@ class Job extends Operation {
...
-            rows = BigQuery.mergeSchemaWithRows_(resp.schema, resp.rows, {
-              wrapIntegers,
-              parseJSON,
-            });
+            rows = options.skipParsing
+              ? resp.rows
+              : BigQuery.mergeSchemaWithRows_(resp.schema, resp.rows, {
+                  wrapIntegers,
+                  parseJSON,
+                });
```
pearigee: (optional) Should we do this as an `if` statement for consistency with the other files? If we do, should we move the delete block up as well? As far as I can tell, there is no behavior that requires the deletion to occur below.

pearigee (approved):

PR #7905, opened 2026-03-27
fix(bigquery): allow timestamp format overrides in createQueryStream

rafal-hawrylak:
## Description

This PR fixes a bug in createQueryStream where formatOptions were being stripped from the query object before reaching the internal request builder. This caused the library to always use the default ISO8601_STRING format, which is rejected by some BigQuery backends.

By extracting these options in queryAsStream_ and passing them through to the request builder, users can now successfully opt-out of the new default behavior when necessary.

## Impact

- createQueryStream now correctly handles formatOptions.timestampOutputFormat and formatOptions.useInt64Timestamp if provided in the options object.
- Bypasses 'timestamp_output_format is not supported yet' errors in restricted environments.

## Testing

- Updated unit tests in test/bigquery.ts to verify the 'Gatekeeper' fix.
- Added a specific test case to verify that formatOptions are correctly passed through.
- Verified that all existing BigQuery tests pass.

### DIFF
--- a/handwritten/bigquery/src/bigquery.ts
+++ b/handwritten/bigquery/src/bigquery.ts
@@ -2458,7 +2458,15 @@ export class BigQuery extends Service {
       return;
     }
 
-    const {location, maxResults, pageToken, wrapIntegers, parseJSON} = query;
+    const {
+      location,
+      maxResults,
+      pageToken,
+      wrapIntegers,
+      parseJSON,
+      'formatOptions.timestampOutputFormat': timestampOutputFormat,
+      'formatOptions.useInt64Timestamp': useInt64Timestamp,
+    } = query as any;
 
     const opts = {
       location,
@@ -2466,6 +2474,8 @@ export class BigQuery extends Service {
       pageToken,
       wrapIntegers,
       parseJSON,
+      'formatOptions.timestampOutputFormat': timestampOutputFormat,
+      'formatOptions.useInt64Timestamp': useInt64Timestamp,
       autoPaginate: false,
     };
 
@@ -2474,6 +2484,8 @@ export class BigQuery extends Service {
     delete query.pageToken;
     delete query.wrapIntegers;
     delete query.parseJSON;
+    delete (query as any)['formatOptions.timestampOutputFormat'];
+    delete (query as any)['formatOptions.useInt64Timestamp'];
 
     this.query(query, opts, callback);
   }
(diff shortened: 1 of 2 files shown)


handwritten/bigquery/src/bigquery.ts
```diff
@@ -2474,6 +2484,8 @@ export class BigQuery extends Service {
     delete query.pageToken;
     delete query.wrapIntegers;
     delete query.parseJSON;
+    delete (query as any)['formatOptions.timestampOutputFormat'];
+    delete (query as any)['formatOptions.useInt64Timestamp'];
```
danieljbruce, line 2488: We can't just delete `timestampOutputFormat` and `useInt64Timestamp` properties from being sent to the backend. Some of our users are likely relying on this feature.

The core issue is that this code was developed on the old repository with a project that had allowlisted a feature that we don't plan to make GA until Q3. So we has to pin the version as shown in https://github.com/googleapis/google-cloud-node/issues/7286. We have a few options:

1. Make source code changes so that requests don't set `timestampOutputFormat` by default and then this won't be an issue for users. The tests still won't pass for a project that isn't allowlisted, but it won't affect users.

2. We can stage a PR to reverse the changes between 8.1.1 and 8.2.0 so that you can add newer changes and do a release which will unblock future work. Then we can revert this PR which will introduce the changes again when the feature is GA.

Option #2 is probably cleaner. Let me know what you think.

rafal-hawrylak: > We can't just delete timestampOutputFormat and useInt64Timestamp properties from being sent to the backend. Some of our users are likely relying on this feature.

Both timestampOutputFormat and useInt64Timestamp are copied to ops and sent to backend via ops, just as other options like location or parseJSON. As far as I understand it's just the query struct itself is stripped from those additional options.

> Option #2 is probably cleaner. Let me know what you think.

We do not rely on or expect time formatting changes so for us the most important factor is whichever is the fastest to be released.

danieljbruce: > Both timestampOutputFormat and useInt64Timestamp are copied to ops and sent to backend via ops

I see. But if a user calls the query method directly and we release the latest version of Bigquery then we are going to see https://github.com/googleapis/google-cloud-node/issues/7286 come back again. I am working on this on https://github.com/googleapis/google-cloud-node/compare/main...revert-bq-changes. I'll try to get a fix out today

danieljbruce:
I actually also see failures when I run the system tests for a non-allowlisted timestamp precision project. How did the system tests perform for your project?

image

danieljbruce:
https://github.com/googleapis/google-cloud-node/pull/7946 was merged so I think we can close this as it is obselete.

PR #7907, opened 2026-03-27
feat(firestore): add the parent expression

yvonnep165:
Add support for the parent expression

Ported from https://github.com/firebase/firebase-js-sdk/pull/9773


yvonnep165:
The `@beta` tags have not been removed from the codebase yet for all expressions and removing it now for the `parent` expression triggers errors. So I will merge with `@beta` tags.

PR #7909, opened 2026-03-27
chore: restore pubsub samples until we can convert them to system tests

### LINKED ISSUE #7908: [pubsub] Restore samples
These will be moving elsewhere, but for the moment, they are still doubling as system tests for some things. They need to be restored for now and then migrated into system tests.

feywind:
Fixes #7908 🦕


feywind:
This is out of date now, and will be replaced by converting the samples+tests into system tests.

PR #7917, opened 2026-03-30
refactor(storage): Migrate conformance tests to StorageTransport and enhance retry logic

thiyaguk09:


handwritten/storage/conformance-test/libraryMethods.ts
```diff
@@ -51,100 +52,160 @@ export interface ConformanceTestOptions {
...
-      preconditionOpts: {
-        ifGenerationMatch: 0,
-      },
-    });
-  } else {
-    allFiles = options.bucket!.file('all-files.txt');
-  }
-
-  await options.bucket!.combine(sources, allFiles);
+  return await combine(options);
```
Dhriti07, line 85: I noticed here, and throughout this file that we replaced high-level SDK calls like bucket.combine() with raw HTTP requests via storageTransport.makeRequest().

Could you help me understand why? I'm concerned that bypassing these means we're no longer testing the actual code paths our users rely on for retries.

Dhriti07: IIUC the migration broke the old storage.interceptors syntax for injecting the x-retry-test-id header in the tests.

Instead of bypassing the public SDK methods across the entire test suite to manually call makeRequest, Can we try fixing the interceptor logic to work with the new gaxios format.

If we fix the header injection at the storage.interceptors level, we can revert these tests to properly validate the public API.

thiyaguk09: > I noticed here, and throughout this file that we replaced high-level SDK calls like bucket.combine() with raw HTTP requests via storageTransport.makeRequest().
> 
> Could you help me understand why? I'm concerned that bypassing these means we're no longer testing the actual code paths our users rely on for retries.

This was done because the modern gaxios migration broke the old storage.interceptors behavior. The high-level SDK methods don't natively pick up the x-retry-test-id header required by the conformance testbench in this environment, so falling back to storageTransport.makeRequest() was the only way to ensure the simulation headers are passed. 

We're going to keep this implementation as-is for now to ensure the test suite remains stable, but this is a valuable item to address during future technical debt cleanup.

thiyaguk09: > IIUC the migration broke the old storage.interceptors syntax for injecting the x-retry-test-id header in the tests.
> 
> Instead of bypassing the public SDK methods across the entire test suite to manually call makeRequest, Can we try fixing the interceptor logic to work with the new gaxios format.
> 
> If we fix the header injection at the storage.interceptors level, we can revert these tests to properly validate the public API.

The best long-term fix would be to realign the interceptor logic with the new gaxios format so public API validation works again. That said, updating that interceptor layer creates tricky connection-loop issues in the test environment, so I’m not changing the current architecture in this PR. For now, 

we’ll leave the explicit makeRequest calls as they are, Long term, the cleanest fix would be to bring the interceptor logic in line with the new gaxios format so public API validation works again. However, changing that layer introduces tricky connection-loop issues in the test environment, so I’m not altering the current architecture in this PR. For now, we’ll keep the explicit makeRequest calls as they are, but this should be helpful context for future test-suite cleanup!  — what do you think?

Dhriti07: Thanks for the context. I believe losing conformance test coverage on the public SDK methods (like bucket.combine()) is a significant regression. If we only test makeRequest, we have no guarantee that the user-facing methods are properly handling retries in the new architecture.

Could you share a bit more about the connection-loop issues you're seeing?

thiyaguk09: The core issue is that with Node 18 and gaxios, hooking into authClient.request to pass headers to high-level methods like bucket.combine() triggers an infinite recursion loop, as gaxios re-evaluates the auth context mid-request and re-enters authClient.request indefinitely.

Running this against the local emulator (127.0.0.1:9000 or localhost:9000) rapidly exhausts socket pools and crashes the testbench server during setup, leading to widespread ECONNRESET failures across the suite.

Safely resolving this circular hook requires a significant overhaul of the test suite's authentication state.

Dhriti07: As discussed offline, our priority is to maintain the robustness of our tests

thiyaguk09: As discussed, the changes have been implemented. PTAL

PR #7929, opened 2026-03-31
fix: add logic to check for IAM RPC methods within the service before adding mixin to avoid generating duplicate methods

shivanee-p:
This PR resolves issues where a service's native RPC methods are implemented even if they match the mixin methods with the same name. As a result, we end up with duplicate methods. We have implemented logic to prioritize native RPC methods over mixin implementations for IAM, Locations, and LRO (Long Running Operations).

**Generator Schema (`proto.ts`)**:
  - Introduced `OperationsMixinConfig` to track which methods (`getOperation`, `cancelOperation`, `deleteOperation`, `listOperations`) conflict with a service's native RPC methods.
  - Optimized the logical order in `augmentService` so that `longRunningOperationsMixinFlags` are populated after the `LongRunningOperationsMixin` is evaluated.
  - Defined and calculated mixin flags (`iamPolicyMixinFlags`, `locationMixinFlags`, and `longRunningOperationsMixinFlags`) to prevent duplicate native method definitions.
  - Added assertions in `test/unit/proto.ts` to test the flags under conflicting method scenarios.

**Templates (`_iam.njk`, `_locations.njk`, `_operations.njk`)**:
  - Wrapped the generation of mixin methods (CJS & ESM) in conditional statements checking their corresponding mixin flags.

**Tests (`gapic_$service_$version.ts.njk`)**:
  - Updated test generation templates (CJS & ESM) to conditionally generate tests for mixin methods only when those methods are actually active in the generated client.
  - Regenerated baselines

Fixes #7659 🦕

### DIFF
--- a/core/generator/gapic-generator-typescript/baselines/kms-esm/esm/src/v1/key_management_service_client.ts.baseline
+++ b/core/generator/gapic-generator-typescript/baselines/kms-esm/esm/src/v1/key_management_service_client.ts.baseline
@@ -3283,7 +3283,7 @@ export class KeyManagementServiceClient {
  *   The request object that will be sent.
  * @param {string} request.resource
  *   REQUIRED: The resource for which the policy detail is being requested.
- *   See the operation documentation for the a ppropriate value for this field.
+ *   See the operation documentation for the appropriate value for this field.
  * @param {string[]} request.permissions
  *   The set of permissions to check for the `resource`. Permissions with
  *   wildcards (such as '*' or 'storage.*') are not allowed. For more
--- a/core/generator/gapic-generator-typescript/baselines/duplicate_methods_test-esm/esm/src/v1/duplicate_methods_test_service_client.ts.baseline
+++ b/core/generator/gapic-generator-typescript/baselines/duplicate_methods_test-esm/esm/src/v1/duplicate_methods_test_service_client.ts.baseline
@@ -819,7 +819,7 @@ export class DuplicateMethodsTestServiceClient {
  *   The request object that will be sent.
  * @param {string} request.resource
  *   REQUIRED: The resource for which the policy detail is being requested.
- *   See the operation documentation for the a ppropriate value for this field.
+ *   See the operation documentation for the appropriate value for this field.
  * @param {string[]} request.permissions
  *   The set of permissions to check for the `resource`. Permissions with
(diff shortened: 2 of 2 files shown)


shivanee-p:
> 2. Do we want to drill down by the actual rpc name? Like, say an API declares a mixin AND directly declares a method. Should we assume that if one rpc exists, we just give precedence to that and exclude the mixin entirely? or we do we do it function by function? Or, maybe we just say more generally, don't produce duplicative function names (and don't do it function by function). I'm not really sure what path we'd want to take, my opinion is that maybe we just say we have no duplicative functions for any mixin if it's declared, without looking specifically function by function name (like in your comparison, I'm wondering what if there's a capitalization issue or a just some slight discrpancy in the hard-coding of the function name). we should just get a list of all functions in the mixin and make sure none of these are also declared in the proto.

Unfortunately, we need to drill down by the specific RPC name. The [AIP guidelines](https://google.aip.dev/client-libraries/4234#overriding-a-duplicate-rpc) state that if for instance only one of the get IAM method is declared within the service, the other 2 mixin methods (set() and test()) still need to be rendered using the mixin.

core/generator/gapic-generator-typescript/baselines/kms-esm/esm/src/v1/key_management_service_client.ts.baseline
```diff
@@ -3283,7 +3283,7 @@ export class KeyManagementServiceClient {
  *   The request object that will be sent.
  * @param {string} request.resource
  *   REQUIRED: The resource for which the policy detail is being requested.
- *   See the operation documentation for the appropriate value for this field.
+ *   See the operation documentation for the a ppropriate value for this field.
```
feywind: There's an extra space in there now.

core/generator/gapic-generator-typescript/typescript/tools/update-baselines.ts
```diff
@@ -90,6 +90,19 @@ async function main() {
     console.log("Tests failed - that's OK, will update baselines.");
   }
 
+  // remove old test out folders and unzip if baselineZip exists
```
feywind: This might require some merging with recent generator changes. :S The baseline handling has changed a fair amount.

PR #7931, opened 2026-03-31
fix: Unblock new Node Bigquery releases

danieljbruce:
## Description

Reverts high precision timestamp changes that allow users to read timestamps with high precision values that are stored on the backend. The service doesn't provide high precision timestamp support in GA yet which caused us to pin the release due to the issue at https://github.com/googleapis/google-cloud-node/issues/7286. But new features for Bigquery need to be rolled out so this PR reverts the commits that caused the issue which allow us to release newer versions of Bigquery.

Reverts:
https://github.com/googleapis/google-cloud-node/pull/7147
https://github.com/googleapis/nodejs-bigquery/pull/1596

## Impact

This unblocks releases for Node Bigquery.

## Testing

Testing for high precision timestamps has been removed because the functionality has been removed. All integration tests pass for a non-allowlisted project.


danieljbruce:
https://github.com/googleapis/google-cloud-node/pull/7946 was merged instead of this one.

PR #7946, opened 2026-04-02
fix: Unblock the releases on Node Bigquery

danieljbruce:
## Description

A feature flag is added that turns on high precision timestamp support if an environment variable is set. High precision timestamp client library support is currently blocking releases due to https://github.com/googleapis/google-cloud-node/issues/7286 because the feature provided by the service is not generally available yet. Therefore, we are turning this feature off by default and allowing users to turn it on by setting an environment variable when it becomes generally available. 

## Impact

This unblocks releases for Node Bigquery.

## Testing

The relevant tests added for high precision timestamps are only turned on when the high precision timestamp environment variable is set.

We verify that the changes are okay with two set of test runs.

First, the tests should pass for a project that doesn't service high precision timestamp requests and doesn't have the high precision timestamp environment variable set.

image

Second, the tests should pass for a project that DOES service high precision timestamp requests and DOES have the high precision timestamp environment variable set.

image

### DIFF
--- a/handwritten/bigquery/system-test/high_precision_timestamp.ts
+++ b/handwritten/bigquery/system-test/high_precision_timestamp.ts
@@ -59,9 +59,14 @@ describe('BigQuery System Tests (High Precision)', () => {
     }
   });
 
-  // Programmatically require the tests. Mocha will discover and register them inside this 'describe' block.
-  require('./bigquery');
-  require('./timestamp_output_format');
+  describe('Run all tests', () => {
+    // Wrap all the other tests in a describe block to ensure the entire file is
+    // executed when the environment variable is set to true.
+
+    // Programmatically require the tests. Mocha will discover and register them inside this 'describe' block.
+    require('./bigquery');
+    require('./timestamp_output_format');
+  });
 });
 
 // 3. CLEANUP:


handwritten/bigquery/src/bigquery.ts
```diff
@@ -2256,33 +2258,45 @@ export class BigQuery extends Service {
...
-            so that the user can see the error.
-             */
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            const listParams = {
-              'formatOptions.timestampOutputFormat':
-                queryReq.formatOptions?.timestampOutputFormat,
-              'formatOptions.useInt64Timestamp':
-                queryReq.formatOptions?.useInt64Timestamp,
-            };
+          if (process.env.BIGQUERY_PICOSECOND_SUPPORT === 'true') {
```
alvarowolfx: I think the listParams passing here can be maintained. We just need to avoid using the timestampOutputFormat as the default value

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2364,18 +2378,25 @@ export class BigQuery extends Service {
     if (options.job) {
       return undefined;
     }
-    const hasAnyFormatOpts =
-      options['formatOptions.timestampOutputFormat'] !== undefined ||
-      options['formatOptions.useInt64Timestamp'] !== undefined;
-    const defaultOpts = hasAnyFormatOpts
-      ? {}
-      : {
-          timestampOutputFormat: 'ISO8601_STRING',
```
alvarowolfx, line 2373: I think this is the most important part that should handle the feature flag. If BIGQUERY_PICOSECOND_SUPPORT is enabled, we should use `timestampOutputFormat: 'ISO8601_STRING'`, if not, use `useInt64Timestamp:true`. We can still allow customers to pass the timestampOutputFormat, even if is allowlisted now. But in the future they can use it just fine without a new release. Later we can remove the feature flag and just default to use it too.

handwritten/bigquery/system-test/bigquery.ts
```diff
@@ -1562,6 +1564,11 @@ describe('BigQuery', () => {
 
             testCases.forEach(testCase => {
               it(`should handle ${testCase.name}`, async () => {
+                if (process.env.BIGQUERY_PICOSECOND_SUPPORT !== 'true') {
+                  // These tests are only important when the high precision
+                  // timestamp support is turned on.
+                  return;
+                }
```
pearigee, line 1571: Rather than disabling the tests, we should instead test the feature flag behavior.

We should be able to stub the environment variable behavior in tests like this:
```
beforeEach(() => {
  process.env.BIGQUERY_PICOSECOND_SUPPORT = 'true'
});

afterEach(() => {
   // Clean up to prevent leak
   delete process.env.BIGQUERY_PICOSECOND_SUPPORT;
});
```

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2576,39 +2597,48 @@ function convertSchemaFieldValue(
...
-        // NOTE: The additional
-        // (useInt64Timestamp !== undefined || timestampOutputFormat !== und...)
-        // check is to ensure that calls to the /query endpoint remain
-        // unaffected as they will not be providing any listParams.
-        //
-        // If the program reaches this point in time then
-        // value is float seconds so convert to BigQueryTimestamp
-        value = BigQuery.timestamp(Number(value));
+      // High precision timestamp behaviour
+      if (process.env.BIGQUERY_PICOSECOND_SUPPORT === 'true') {
```
alvarowolfx: I don't think the feature flag check is needed here, we can keep the improved code that pass the listParams across boundaries.

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2706,14 +2736,6 @@ export class BigQueryRange {
     };
   }
 
-  /**
```
alvarowolfx, line 2709: nit: why removed ?

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2733,17 +2755,6 @@ export class BigQueryRange {
     return [start, end];
   }
 
-  /**
```
alvarowolfx, line 2736: nit: why removed ?

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2834,15 +2845,17 @@ export class BigQueryTimestamp {
...
-        if (value.match(/\.\d{10,}/) && !Number.isNaN(pd.getTime())) {
-          /*
-          TODO:
-          When https://github.com/googleapis/nodejs-precise-date/pull/302
-          is released and we have full support for picoseconds in PreciseData
-          then we can remove this if block.
-           */
-          this.value = value;
-          return;
+        if (process.env.BIGQUERY_PICOSECOND_SUPPORT === 'true') {
```
alvarowolfx: don't think needs to be behind the feature flag. If a timestamp with picosecond is returned, we know already how to handle it. 

Aside from that, have we published the new version of https://github.com/googleapis/nodejs-precise-date ? Maybe we can even get rid of this TODO

handwritten/bigquery/src/job.ts
```diff
@@ -596,13 +596,28 @@ class Job extends Operation {
...
 
         if (resp.schema && resp.rows) {
-          try {
-            /*
-            Without this try/catch block, calls to /query endpoint will hang
-            indefinitely if a call to mergeSchemaWithRows_ fails because the
-            error never makes it to the callback. Instead, pass the error to the
-            callback the user provides so that the user can see the error.
-             */
+          if (process.env.BIGQUERY_PICOSECOND_SUPPORT === 'true') {
```
alvarowolfx: same as others. I think checking the feature flag here is making code much more complicated to understand

handwritten/bigquery/src/table.ts
```diff
@@ -1867,13 +1866,35 @@ class Table extends ServiceObject {
...
         return;
       }
-      try {
-        /*
-        Without this try/catch block, calls to getRows will hang indefinitely if
-        a call to mergeSchemaWithRows_ fails because the error never makes it to
-        the callback. Instead, pass the error to the callback the user provides
-        so that the user can see the error.
-         */
+      if (process.env.BIGQUERY_PICOSECOND_SUPPORT === 'true') {
```
alvarowolfx: same as others. I think checking the feature flag here is making code much more complicated to understand

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2256,33 +2258,45 @@ export class BigQuery extends Service {
...
-            };
+          if (process.env.BIGQUERY_PICOSECOND_SUPPORT === 'true') {
+            try {
+              /*
+              Without this try/catch block, calls to getRows will hang indefinitely if
+              a call to mergeSchemaWithRows_ fails because the error never makes it to
+              the callback. Instead, pass the error to the callback the user provides
+              so that the user can see the error.
+               */
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
pearigee: Can we remove this linter exemption? I don't see a usage of `any`?

handwritten/bigquery/src/table.ts
```diff
@@ -1884,25 +1905,33 @@ class Table extends ServiceObject {
...
-              listParams: qs,
             },
           );
         }
-      } catch (err) {
-        callback!(err as Error | null, null, null, resp);
-        return;
       }
       callback!(null, rows, nextQuery, resp);
     };
```
alvarowolfx, line 1895: same as https://github.com/googleapis/google-cloud-node/pull/7946/changes#r3029295761, this is the core part that needs to handle the feature flag.

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2256,33 +2258,45 @@ export class BigQuery extends Service {
...
+              const listParams = {
+                'formatOptions.timestampOutputFormat':
+                  queryReq.formatOptions?.timestampOutputFormat,
+                'formatOptions.useInt64Timestamp':
+                  queryReq.formatOptions?.useInt64Timestamp,
+              };
+              if (options.skipParsing) {
+                rows = res.rows;
+              } else {
```
pearigee: This block to check for `options.skipParsing` is defined twice. Here and below in the else block if the feature flag is disabled.

Can we move this behavior out of the feature flag block to avoid duplication?

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2364,18 +2378,25 @@ export class BigQuery extends Service {
...
+    let formatOptions;
+    if (process.env.BIGQUERY_PICOSECOND_SUPPORT === 'true') {
+      const hasAnyFormatOpts =
+        options['formatOptions.timestampOutputFormat'] !== undefined ||
+        options['formatOptions.useInt64Timestamp'] !== undefined;
+      const defaultOpts = hasAnyFormatOpts
+        ? {}
+        : {
+            timestampOutputFormat: 'ISO8601_STRING',
+          };
```
pearigee, line 2381: I think this can be simplified:

```
const defaultOptions = {
   // provide default
    timestampOutputFormat: 'ISO8601_STRING',

   // merge with options object. If timestampOutputFormat is defined, it will overwrite the above.
    ...options
}
```

Additionally, if we do this, I don't think we need the `extend(...)` block below. `defaultOptions` should be sufficient.

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2576,39 +2597,48 @@ function convertSchemaFieldValue(
...
+        1672574400.123456
+        2023-01-01T12:00:00.123456789123Z
+         */
+        const listParams = options.listParams;
+        const timestampOutputFormat = listParams
+          ? listParams['formatOptions.timestampOutputFormat']
+          : undefined;
+        const useInt64Timestamp = listParams
+          ? listParams['formatOptions.useInt64Timestamp']
+          : undefined;
```
pearigee: We can use the `?.` operator here to simplify this access pattern. When used, if the property is not present it will simply return undefined for you.

```
const timestampOutputFormat = listParams?.['formatOptions.timestampOutputFormat'];
```

handwritten/bigquery/src/bigquery.ts
```diff
@@ -2576,39 +2597,48 @@ function convertSchemaFieldValue(
...
+        ) {
+          // NOTE: The additional
+          // (useInt64Timestamp !== undefined || timestampOutputFormat !== und...)
+          // check is to ensure that calls to the /query endpoint remain
+          // unaffected as they will not be providing any listParams.
+          //
+          // If the program reaches this point in time then
+          // value is float seconds so convert to BigQueryTimestamp
+          value = BigQuery.timestamp(Number(value));
+        } else {
```
pearigee: Does this `if/else` block make assumptions about the available output options?

What if in the future an output option is added other than 'ISO8601_STRING', 'INT64', and "float seconds"? It seems like this logic could break by applying an incompatible default?

---

Would it be more robust in the `if` statement above (i.e. `useInt64Timestamp !== true &&
          timestampOutputFormat !== 'INT64' &&
          (useInt64Timestamp !== undefined ||
            timestampOutputFormat !== undefined)`) checked directly for float seconds rather than checking it's not Int64?

handwritten/bigquery/src/job.ts
```diff
@@ -596,13 +596,28 @@ class Job extends Operation {
...
+            try {
+              /*
+              Without this try/catch block, calls to /query endpoint will hang
+              indefinitely if a call to mergeSchemaWithRows_ fails because the
+              error never makes it to the callback. Instead, pass the error to the
+              callback the user provides so that the user can see the error.
+               */
+              if (options.skipParsing) {
+                rows = resp.rows;
+              } else {
```
pearigee: Same in this file. I think the `options.skipParsing` logic could live outside of this flag check (since it happens either way).

handwritten/bigquery/src/table.ts
```diff
@@ -1867,13 +1866,35 @@ class Table extends ServiceObject {
...
+        try {
+          /*
+          Without this try/catch block, calls to getRows will hang indefinitely if
+          a call to mergeSchemaWithRows_ fails because the error never makes it to
+          the callback. Instead, pass the error to the callback the user provides
+          so that the user can see the error.
+           */
+          if (options.skipParsing) {
+            rows = rows || [];
+          } else {
```
pearigee: Same in this file. I think the options.skipParsing logic could live outside of this flag check (since it happens either way).

handwritten/bigquery/src/table.ts
```diff
@@ -1884,25 +1905,33 @@ class Table extends ServiceObject {
...
-      options['formatOptions.timestampOutputFormat'] !== undefined ||
-      options['formatOptions.useInt64Timestamp'] !== undefined;
-    const defaultOpts = hasAnyFormatOpts
-      ? {}
-      : {
-          'formatOptions.timestampOutputFormat': 'ISO8601_STRING',
-        };
-    const qs = extend(defaultOpts, options);
+
+    let qs: any;
```
pearigee: Can we use a more robust type here than any? Do we have an options interface we can reference instead?

handwritten/bigquery/src/bigquery.ts
danieljbruce: Okay. It looks like we can just change.

```
formatOptions = {
        useInt64Timestamp: true,
      };
```

to

```
formatOptions = extend(
      {
        useInt64Timestamp: true,
      }, {
        timestampOutputFormat: options['formatOptions.timestampOutputFormat'],
        useInt64Timestamp: options['formatOptions.useInt64Timestamp'],
      }
);
```

handwritten/bigquery/src/bigquery.ts
danieljbruce: Ok. I'll make the change and run the tests again. We have to make sure

```
{
  formatOptions.timestampOutputFormat: undefined,
  formatOptions.useInt64Timestamp: undefined,
}
```
doesn't cause any issues.

handwritten/bigquery/src/bigquery.ts
danieljbruce: I added this back in. It just accidentally got removed in the revert.

handwritten/bigquery/src/bigquery.ts
danieljbruce: I added this back in. It just accidentally got removed in the revert.

handwritten/bigquery/src/job.ts
danieljbruce: Yeah. I agree. Upon further consideration it looks like including this `try` block is fine whether we are using high precision timestamps or not.

handwritten/bigquery/src/job.ts
danieljbruce: After resolving the `same as others. I think checking the feature flag here is making code much more complicated to understand` comment from Alvaro, it looks like the duplicate `if (options.skipParsing) {` is gone so I think this issue has been resolved.

handwritten/bigquery/src/bigquery.ts
danieljbruce: I changed this, but actually it causes the tests to fail. There is something about the integration with this change and precise date that isn't quite right. I created a bug at https://buganizer.corp.google.com/issues/500323614, but I don't think we should reverse this change now in this PR.

handwritten/bigquery/src/bigquery.ts
danieljbruce: I think that's true. I was a little worried about this one because when not using new high precision timestamp logic we are changing the way this is parsing. But I checked the chart and it looks like all the changes we are doing are positive changes.

Image

handwritten/bigquery/src/table.ts
danieljbruce: That's true. This try block is useful no matter what.

handwritten/bigquery/src/table.ts
danieljbruce: Ok. For the old version this will now extend the options instead of just setting `useInt64Timestamp` to true.

handwritten/bigquery/src/bigquery.ts
danieljbruce: We can't quite do this for a couple reasons:
1. The intent here is to default timestampOutputFormat to `ISO8601_STRING` only if no format options are provided at all. You can reach out to @alvarowolfx for insight on this decision as I think the idea here is that we want to default to requesting high precision timestamps if the user hasn't given any thought to formatting options at all.
2. The user provides options with key `formatOptions.timestampOutputFormat`, not `timestampOutputFormat`. So we can't just merge them because the keys are actually different.

handwritten/bigquery/src/table.ts
danieljbruce: Yes! I changed this to GetRowsOptions.

handwritten/bigquery/system-test/bigquery.ts
danieljbruce: Instead of doing this I modified `package.json` to have the line `"system-test": "npm run system-test:pico && npm run system-test:standard",` with lines:

"system-test:pico": "BIGQUERY_PICOSECOND_SUPPORT=true mocha build/system-test --timeout 600000",
"system-test:standard": "mocha build/system-test --timeout 600000",

so that running system-test script would run all the relevant tests with and without the feature flag even outside of the new tests.

The trouble with just enabling the feature flag with these tests is that I think the other team is running these tests with a non-allowlisted project. With a non-allowlisted project and the feature flag turned on these tests will fail and they should fail because they are making high precision timestamp requests to a service that doesn't support it yet. I don't want to create a failing test red herring that will confuse developers who are running this test suite on a non-allowlisted project.

handwritten/bigquery/src/bigquery.ts
danieljbruce: Yup. This is gone now.

handwritten/bigquery/src/bigquery.ts
danieljbruce: I think the duplication was removed when addressing @alvarowolfx as now there is no if statement for the feature flag.

handwritten/bigquery/src/bigquery.ts
danieljbruce: Good idea. This change has been added.

handwritten/bigquery/src/bigquery.ts
danieljbruce: This is a good suggestion for future proofing our parsing logic. I remember encountering this problem when working through the parsing logic of the high precision timestamp feature. Below is a chart of the different data formats the server returns:

image

The biggest barrier I see is that if the user makes a request for timestampOutputFormat=TIMESTAMP_OUTPUT_FORMAT_UNSPECIFIED and useInt64Timestamp=false and the fractional part of the response happens to be zero then we still don't want to use the default BigInt logic because for this combination the value returned from the server is a million times smaller so we still want logic that accommodates for the smaller number size.

That said, there still might be ways to improve this logic so I'm open to suggestions.

handwritten/bigquery/src/table.ts
danieljbruce: With @alvarowolfx changes the options.skipParsing only happens once now so I think this comment has been addressed.

handwritten/bigquery/src/bigquery.ts
alvarowolfx: but fail with which error ? Which tests fail ? I think we need to double check this, because we had issues in the past by not testing enough this function and breaking with customer code. I got concerned now that we need to have the feature flag being checked here, even in scenarios where the backend is not returning picosecond data. Maybe the whole check needs to be updated to take that into account.

handwritten/bigquery/src/bigquery.ts
alvarowolfx: I think some logic still needs to be shared for both cases. Only the default value should change based on the feature flag.

This code snippets makes it clear that the default should be `useInt64Timestamp: true` currently, and with the feature flag enabled, should be `timestampOutputFormat: 'ISO8601_STRING'`. And customers can still provide both opts.

```js
const hasAnyFormatOpts =
  options['formatOptions.timestampOutputFormat'] !== undefined ||
  options['formatOptions.useInt64Timestamp'] !== undefined;
let defaultOpts = hasAnyFormatOpts
  ? {}
  : {
      useInt64Timestamp: true,
    };
let formatOptions;
if (process.env.BIGQUERY_PICOSECOND_SUPPORT === 'true') {
  defaultOpts = hasAnyFormatOpts
    ? {}
    : {
        timestampOutputFormat: 'ISO8601_STRING',
      };  
}
formatOptions = extend(defaultOpts, {
  timestampOutputFormat: options['formatOptions.timestampOutputFormat'],
  useInt64Timestamp: options['formatOptions.useInt64Timestamp'],
});
```

handwritten/bigquery/src/table.ts
alvarowolfx: similar to https://github.com/googleapis/google-cloud-node/pull/7946/changes#r3046889774, code should be along these lines:

```js
let qs: GetRowsOptions;
const hasAnyFormatOpts =
  options['formatOptions.timestampOutputFormat'] !== undefined ||
  options['formatOptions.useInt64Timestamp'] !== undefined;
let defaultOpts = hasAnyFormatOpts
    ? {} 
    : {
      'formatOptions.useInt64Timestamp': true,
     };
if (process.env.BIGQUERY_PICOSECOND_SUPPORT === 'true') {
  defaultOpts = hasAnyFormatOpts
    ? {}
    : {
        'formatOptions.timestampOutputFormat': 'ISO8601_STRING',
      };
} 
qs = extend(defaultOpts, options);
```

handwritten/bigquery/src/bigquery.ts
alvarowolfx: timestampOutputFormat and useInt64Timestamp are mutually exclusive in most scenarios. So yeah, we can't just merge all options with defaultOptions.

handwritten/bigquery/system-test/bigquery.ts
alvarowolfx: which other teams are running these tests ? I think they should have access into the allowlisted project, it was always like that in the BigQuery SDK team. Adding those branches on running all tests makes our CI more complex and I'm not seeing the added value.

alvarowolfx: I'm with @pearigee recommendation and would add more that I think we should run this whole test suite with and without the flag. 

https://github.com/googleapis/google-cloud-node/blob/c58058b18e8ac38875eb18a094a9d329db3f1e61/handwritten/bigquery/system-test/bigquery.ts#L44

We can have a `runTestsWithFlags` method that does something like this (this is a pseudo code, I haven't tried to run it)


```js
function runTestsWithFlags(testCase: string, flags: []string, body: () => {}) {
   flags.forEach( flag => { 
       describe(`With {flag}=true`, () => {
            beforeEach(() => {
              process.env[flag]= 'true'
            });
            
            afterEach(() => {
               // Clean up to prevent leak
               delete process.env[flag];
            });

            body();
       });
       
       describe(`With {flag}=false`, () => {
           beforeEach(() => {
              process.env[flag]= 'false'
            });
            
            afterEach(() => {
               // Clean up to prevent leak
               delete process.env[flag];
            });

           body();
       });
    }
}
const featureFlags = ['BIGQUERY_PICOSECOND_SUPPORT'];
runTestsWithFlags('BigQuery', featureFlags, () => { 
    /* current test suite */
})
```

alvarowolfx (changes requested):
I think the logic for the defaultOpts can still be improved.

handwritten/bigquery/src/bigquery.ts
danieljbruce: I looked at this again and it appears that even when the fractional value is zero that the response from the server contains a period. So maybe we can scan the response for a period and replace that with the else if condition. How does that sound?

handwritten/bigquery/src/bigquery.ts
danieljbruce: Alright. I made this change here and the default request change for getRows. It changes the current defaults that are being used, but it still seems to work okay.

handwritten/bigquery/src/bigquery.ts
danieljbruce: Below is a screenshot of the test failures: 

Image

It truncates the timestamp to three decimal places.

danieljbruce: The `value.match(/\.\d{10,}/) && !Number.isNaN(pd.getTime())` check only returns true when we actually get a picosecond precision timestamp from the server. Even if we don't set `BIGQUERY_PICOSECOND_SUPPORT` to true it is possible to get picosecond precision data from an allowlisted project when the user provides the right options. In such cases I think it is beneficial to allow the if to be checked even if `BIGQUERY_PICOSECOND_SUPPORT` is false so I think I'll remove the `BIGQUERY_PICOSECOND_SUPPORT` check.

The test failures indicate that something is wrong with falling back on PreciseDate for high precision timestamps. I am going to undo these changes as they are not necessary.

handwritten/bigquery/src/table.ts
danieljbruce: Okay. Made this change. Seems fine.

handwritten/bigquery/system-test/bigquery.ts
danieljbruce: I'll get to this one next.

handwritten/bigquery/src/table.ts
pearigee: I may be looking at the wrong thing, but I still see an `any` on line 1915? Did that change get pushed?

handwritten/bigquery/src/bigquery.ts
pearigee: That still sounds very brittle, because now we are making assumptions about how the number is formatted in relation to its type.

From your screenshot it looks like there are the following formats: `FLOAT64`, `INT64`, `ISO8601_STRING`, `TIMESTAMP_FORMAT_UNSPECIFIED`. 

Can we not just explicitly check for each type (i.e. in a switch statement) and perform the appropriate action? It might be slightly more code, but it would be much more readable/reliable? Do we not have an enum or something we can reference rather than making assumptions based on the value?

handwritten/bigquery/system-test/bigquery.ts
danieljbruce: Alright. I added a high_precision_timestamp.ts file which runs all the other tests again with the `BIGQUERY_PICOSECOND_SUPPORT` environment variable set to true thereby accomplishing our goal. That way we don't have to change package.json.

I didn't want to use the approach of wrapping bigquery.ts in a loop that runs the tests twice because it would create a massive diff in this PR due to the extra indent.

danieljbruce (commented):
Exercising tests now without package.json changes.

handwritten/bigquery/system-test/high_precision_timestamp.ts
alvarowolfx, line 1: nice solution 👏

pearigee, line 1: I don't want to block this PR, so this is something we can work on later -- but this goes a bit against convention and our internal style guides.

Ideally, we would preserve the relationship between the test file and the file it is testing (i.e. by sharing some portion of the same name). The common pattern for testing the on/off behavior would be something like this:

```
describe('some feature', () => {
     describe('with flag enabled', () => {
          // Enable flag
          beforeEach(...);

          // Clean up flag
          afterEach();

          it('test 1', ...)

          it('test 2', ...)
      });

     describe('with flag disabled', () => {
          // Ensure flag is cleared
          beforeEach(...);

          it('test 1', ...)

          it('test 2', ...)
     });
});
```

If we were worried about duplication of code, we can always create a utility file that both tests import.

PR #7949, opened 2026-04-02
fix(gax): abort stalled REST fallback fetches

akshatbaranwal:
## Summary
- treat the second REST fallback stub argument as metadata headers and the third as call options
- wire `deadline` to an actual abort timer for REST fallback fetches
- make server-stream `cancel()` abort the underlying fetch instead of only the parser-owned controller
- add regression coverage for unary deadline abort, server-stream deadline abort, and server-stream cancel

## Why
In REST fallback mode, server-streaming reads can silently stall after fetch dispatch but before the first response byte. Before this change:
- the generic timeout wrapper populated the third argument, but the fallback stub only read the second
- server-stream `cancel()` aborted the parser-owned controller, not the fetch-owned controller

That combination leaves stalled REST fallback fetches effectively unbounded and uncancellable.

## Testing
- `cd core/packages/gax && npm run compile`
- `cd core/packages/gax && npx mocha build/test/unit/grpc-fallback.js`


quirogas:
Closing this PR as it hasn't seen any activity in a while.

PR #7950, opened 2026-04-02
fix(firestore): avoid reusing busy clients for streams

akshatbaranwal:
## Summary
- add an opt-in `preferIdleClients` path in `ClientPool`
- use that mode for `requestStream()` so new streams reuse idle clients but do not pile onto already-busy ones
- add pool regression tests for the new reuse behavior

## Why
`DocumentReference.get()` and `getAll()` use the `batchGetDocuments` server-streaming path. If one stream stalls for a long time, the existing `ClientPool` strategy prefers the most-full client and can keep routing later low-concurrency reads back onto the same busy client.

This change is a defensive mitigation: request streams will reuse idle clients, but if all existing clients are already busy, they open a fresh client instead of piling onto a busy one.

## Testing
- `cd handwritten/firestore && npx mocha -r ts-node/register/transpile-only dev/test/pool.ts --grep "preferIdleClients|re-uses idle instances when preferIdleClients|creates a new client when preferIdleClients|creates new instances as needed|re-uses idle instances"`
- `cd handwritten/firestore && npx tsc -p . --noEmit --skipLibCheck`


dlarocque:
@akshatbaranwal Does this change solve any open issues with this SDK?

akshatbaranwal:
Yes — this is the Firestore-side mitigation for #7960.

That issue describes how a single stalled REST-fallback server-streaming read (`batchGetDocuments`) can leave a `ClientPool` client checked out indefinitely, and because the pool prefers the most-full existing client, later low-concurrency reads keep reacquiring the same poisoned client until the process is restarted.

This PR doesn't try to fix the underlying transport stall (that's tracked separately in #7959). It's a defensive change in `ClientPool`: under the new `preferIdleClients` mode — which `requestStream()` opts into — a new stream reuses an idle client if one exists, but if every existing client is already busy it opens a fresh client rather than piling onto a busy one. That way one stuck stream can no longer attract every subsequent read.

Happy to scope it differently (e.g. stale-stream eviction instead of an opt-in reuse policy) if that's closer to what you'd prefer — question #1 in #7960 was explicitly asking for maintainer guidance on that.

quirogas:
I'm closing this PR due to inactivity. Please feel free to reopen it if it remains relevant, as we are currently sanitizing our repository.

PR #7958, opened 2026-04-04
fix(storage): bump retry-request and teeny-request

OfriHarlev:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-cloud-node/issues) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes #7838

## Description
This updates the handwritten `@google-cloud/storage` package off the older request stack by bumping:
- `retry-request` from `^7.0.0` to `^8.0.0`
- `teeny-request` from `^9.0.0` to `^10.0.0`

For the storage package itself, that changes the resolved runtime chain from:
`retry-request@7 -> teeny-request@9 -> http-proxy-agent@5 -> @tootallnate/once@2`

to:
`retry-request@8.0.2 -> teeny-request@10.1.2 -> http-proxy-agent@7.0.2`

This matters because consumers of `@google-cloud/storage` currently inherit the older `http-proxy-agent@5` / `@tootallnate/once@2` path, which is called out in GHSA-vpq2-c234-7xj6.

## Impact
No source code changes were needed for this bump.

`handwritten/storage` already uses `retry-request` and `teeny-request` through the existing request helpers in `src/nodejs-common`, and the direct API surface used by the package remained compatible with the newer dependency line in this package-level validation. In practice, this is a dependency-resolution update rather than a behavior change in storage's own code.

One scope note: this does not remove every `http-proxy-agent@5` occurrence from the monorepo dependency graph. After the bump, the remaining vulnerable path in this package install is through the dev dependency `@google-cloud/pubsub -> google-gax@4.6.1 -> retry-request@7 -> teeny-request@9`. This PR is specifically for the handwritten storage package's direct runtime request stack.

## Testing
I validated this change inside `handwritten/storage` in the active `google-cloud-node` monorepo layout:
- `npm install`
- `npx mocha build/cjs/test/nodejs-common/service.js build/cjs/test/nodejs-common/util.js`
  - Result on upstream `main`: `151 passing`
  - Result on this branch: `151 passing`
- `npm test`
  - Result on upstream `main`: `3 failing`
  - Result on this branch: `3 failing`
  - The same three existing failures reproduce in both cases under `File#createWriteStream`:
    - `should emit errors`
    - `should emit RangeError`
    - `should close upstream when pipeline fails`
- `npm run lint`
  - Result on upstream `main`: fails in this environment with existing Prettier/GTS formatting errors across the package
  - Result on this branch: same failure shape as upstream `main`

I also verified the resolved dependency tree after the bump:
- `retry-request@8.0.2`
- `teeny-request@10.1.2`
- `http-proxy-agent@7.0.2`

## Additional Information
If you want this issue fully closed at the monorepo level rather than the storage runtime path specifically, there is still follow-up work to move the remaining dev/test dependency branches off `http-proxy-agent@5` as well.

### DIFF
--- a/handwritten/storage/package.json
+++ b/handwritten/storage/package.json
@@ -88,8 +88,8 @@
     "html-entities": "^2.5.2",
     "mime": "^3.0.0",
     "p-limit": "^3.0.1",
-    "retry-request": "^7.0.0",
-    "teeny-request": "^9.0.0",
+    "retry-request": "^8.0.0",
+    "teeny-request": "^10.0.0",
     "uuid": "^8.0.0"
   },
   "devDependencies": {


LucasHillDex:
Hi @alkatrivedi any way we can get this (or similar solution to the vulnerabilities) in?

richardkazuomiller:
It’s been about a month so it would be nice if our friends at Google could take a look 🤠

PR #7962, opened 2026-04-06
fix(spanner): type conflict between @sinonjs/fake-timers and @types/sinonjs__fake-timers

alkatrivedi:
This PR resolves build-time TypeScript errors triggered by upstream dependency updates

**Problem**
When running `npm install` or `npm run compile`, the build fails due to a type conflict between @sinonjs/fake-timers and @types/sinon (which pulls in @types/sinonjs__fake-timers). Both packages try to define symbols like `withGlobal `in the global scope, leading to a Duplicate identifier 'withGlobal' error

**Solution**
Uses npm overrides to pin @sinonjs/fake-timers to the last known stable version (15.2.1). This eliminates the duplicate definitions at the source while preserving strict type checking for our own codebase.

Future Work

Revert this change, as soon as the [fix](https://github.com/sinonjs/fake-timers/pull/559) for the conflict will be merge in fake-timer repository


handwritten/spanner/package.json
```diff
@@ -141,11 +141,13 @@
     "yargs": "^17.7.2"
   },
   "overrides": {
-    "gcp-metadata": "^7.0.1"
+    "gcp-metadata": "^7.0.1",
+    "@sinonjs/fake-timers": "15.2.1"
```
surbhigarg92, line 145: Please verify this change does not change the minimum version of Node or any other major components that are required to use this library ?

surbhigarg92, line 145: Do we need to fix the version to "15.2.1" . This would mean no updates will be used for this package

alkatrivedi: since, we are already on Node 18, and not modifying the libraries engine, this fix will not change the minimum version of node or any other major component

alkatrivedi: there is already an open [issue](https://github.com/sinonjs/fake-timers/pull/559) and a fix [PR](https://github.com/sinonjs/fake-timers/pull/559) in fake-timers repository regarding this conflict.

once they will merge the fix we will revert this change. in the meantime we will fix the build failure with this stable version

PR #7965, opened 2026-04-06
chore(gapic-generator-typescript): remove bazel dependency from build

GautamSharda:
This change removes the Bazel dependency from the gapic-generator-typescript project. All Bazel configuration files have been deleted and package.json scripts have been updated to use the standard TypeScript compiler (tsc) directly.

Bazel was originally introduced to export build rules for the googleapis monorepo. However, following the consolidation of the generator into google-cloud-node/core, Bazel is no longer necessary and adds significant overhead (e.g., 4+ minute installations via npm run compile). Since tsc provides all the compilation functionality needed natively, removing Bazel simplifies the toolchain and speeds up CI pipelines without losing capabilities.

For https://github.com/googleapis/librarian/issues/4593
Fixes b/492949752


quirogas:
Closing due to inactivity. sorry @GautamSharda :(.

PR #7971, opened 2026-04-07
chore: remove .DS_Store file

t3tra-dev:
I don't think `.DS_Store` should be located in the project root.


.DS_Store
quirogas, line 1: Yes, this should be removed, thank you!

.gitignore
quirogas, line 1: .DS_Store should not be part of the .gitignore since it is a macOS file. Ideally, developers should use a global .gitconfig to handle these types of system files.

quirogas (changes requested):

.gitignore
t3tra-dev: Understood. I will revert the changes to `.gitignore`.

quirogas (approved):
Thank you!

GautamSharda (approved):
LGTM

pearigee (approved):

PR #7972, opened 2026-04-07
chore: remove duplicate per-package contribution and license files

GautamSharda:

Fixes b/488768301 🦕

For https://github.com/googleapis/librarian/issues/4273


pearigee:
To make these changes permanent, we will also need to update the generator to prevent them from being re-added.

Also, I disabled the system/sample tests that were breaking you above. You should be able to revert those changes.

For the fake-timers issue I have this PR in progress to resolve the problem going forward: https://github.com/googleapis/google-cloud-node/pull/7995

PR #7974, opened 2026-04-07
chore: disable system tests in packages to unblock PRs pending fix

pearigee:
System tests are failing with the following error:
```
Error: Command failed with exit code 1: npx tsc
Step #6 - "run-tests": �[96mindex.ts�[0m:�[93m21�[0m:�[93m23�[0m - �[91merror�[0m�[90m TS2591: �[0mCannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
```
(for example: https://github.com/googleapis/google-cloud-node/pull/7961/checks?check_run_id=70274525122)

Note, this was already fixed in the generator in https://github.com/googleapis/google-cloud-node/pull/7943. However, the generator release takes some time (i.e. days). This is meant to unblock PRs immediately.

See the system test failures here:
https://github.com/googleapis/google-cloud-node/issues/7976

---

On a side note, these system tests seem to add very little value. The generated system tests just create the client and then close it. Furthermore, for generated libraries, we expect them to work by default. We don't manually update these libraries, so the need for regression testing is minimal.

### DIFF
--- a/ci/run_conditional_tests.sh
+++ b/ci/run_conditional_tests.sh
@@ -113,6 +113,11 @@ for subdir in ${subdirs[@]}; do
             continue
         fi
 
+        # System tests for packages are broken and blocking PRs.
+        # See https://github.com/googleapis/google-cloud-node/issues/7976.
+        #
+        # Per https://github.com/googleapis/google-cloud-node/issues/7921, 
+        # we are likely to permanently remove these tests in the near future.
         if [[ "${subdir}" == "packages" && "${TEST_TYPE}" == "system" ]]; then
             echo "Skipping ${TEST_TYPE} test for packages: ${d}"
             continue


quirogas (approved):
Should we add a comment explaining why we are skipping these tests. Or maybe we could add a TODO to revert this change later. 

This can be tracked via the commit history, either way, so up it's up to you.

pearigee:
Absolutely, thanks. I was just working with @shivanee-p to write an issue we can reference. I will update this PR with a link/comment shortly.

pearigee:
Note in https://github.com/googleapis/google-cloud-node/issues/7921, we are considering deleting these tests all together. This may be a good first step in that direction.

PR #8000, opened 2026-04-09
fix: Ensure all headers in the same package have the same copyright year

danieljbruce:
## Description

Solves issue https://github.com/googleapis/librarian/issues/4753. Ensures that all headers in files of the same package have the same copyright year. Files containing no copyright year also now have a copyright year as this was a related issue discovered by the testing script.

## Impact

Assists with librarian migration from https://github.com/googleapis/librarian/milestone/88.

## Testing

The root directory of google-cloud-node doesn't use CI tests, but I created a check-package-years.sh script in the bin folder that only succeeds if all header files have the same copyright year for every file in a given package.


danieljbruce:
This work was done on an old main commit and the merge headaches are significant. The work has been restarted on https://github.com/googleapis/google-cloud-node/pull/8016 instead

PR #8002, opened 2026-04-09
fix(gapic-generator-typescript): improve generator script robustness outside of bazel

GautamSharda:
This PR improves the robustness of the generator scripts so they can execute successfully outside of a Bazel sandbox environment.

When the generator is invoked directly (e.g., via `tsc` and `node`), it previously failed due to missing executable permissions on `protoc-plugin.js`, strict file path checks for the system `protoc` binary, and crashes when parsing modern protos without the `--experimental_editions` flag. This PR addresses those by:
- Dynamically applying `chmod +x` to the generated plugin script.
- Trusting the system PATH for `protoc` (bypassing `fs.existsSync` for the string 'protoc').
- Explicitly passing `--experimental_editions` to protoc.
- Explicitly invoking the baseline tests with the `node` runtime to avoid shebang permission issues.

This is **Part 1 of 3** in the broader effort to completely remove Bazel from the generator (For https://github.com/googleapis/librarian/issues/4593
).


quirogas:
I'm closing this PR since it hasn't been addressed in a while.

PR #8007, opened 2026-04-09
chore: cleanup redundant .gitattributes files

pearigee:
Per https://github.com/googleapis/librarian/issues/4582, our `.gitattributes` files don't need to defined on a per-package level.

Analyzing all of the .gitattributes files, they appear to fall into two main catagories:

### 1. Package-level .gitattributes 
Almost every package in the repository (both in packages/ and handwritten/) contains an identical .gitattributes file with the following content:

```
*.ts text eol=lf
*.js text eol=lf
protos/* linguist-generated
**/api-extractor.json  linguist-language=JSON-with-Comments
```

Note: A few files (like packages/google-cloud-tasks/.gitattributes) appeared to have 3 lines instead of 4, but this was simply due to a missing trailing newline; the content was the same.

### 2. .kokoro directory .gitattributes
Files located in .kokoro directories generally contain just one line:

```
* linguist-generated=true
```
This marks all files within the .kokoro directory as "generated" so they are ignored in GitHub language statistics and diffs.

### Consolidating these into a single .gitattributes file

It seems the important bits are simply this:

```
# Enforce LF line endings for all TS and JS files across the repo
*.ts text eol=lf
*.js text eol=lf
```

Optionally, we could include the Linguist specific changes (impacting GitHubs code statistics/diffs) but I don't think its strictly necessary. For now, I have opted to ignore them.
```
# Mark all protos directories as generated for Linguist (removing them from language stats and diffs)
**/protos/* linguist-generated

# Set language for api-extractor.json files
**/api-extractor.json linguist-language=JSON-with-Comments

# Mark all .kokoro directories as generated (removing them from language stats and diffs)
**/.kokoro/** linguist-generated
```

### Generator updates (going to be a separate PR to reduce toil)
I am going to update the generator, to prevent the automatic generation of .gitattributes files, in a second PR since the change is a bit more involved and doesn't require such broad set of reviewers.

### DIFF
--- a/core/common/.gitattributes
+++ b/core/common/.gitattributes
@@ -1,0 +1,4 @@
+*.ts text eol=lf
+*.js text eol=lf
+protos/* linguist-generated
+**/api-extractor.json  linguist-language=JSON-with-Comments
--- a/core/paginator/.eslintignore
+++ b/core/paginator/.eslintignore
@@ -5,4 +5,3 @@ build/
 docs/
 protos/
 samples/generated/
-system-test/**/fixtures
--- a/core/paginator/.gitattributes
+++ b/core/paginator/.gitattributes
@@ -1,0 +1,4 @@
+*.ts text eol=lf
+*.js text eol=lf
+protos/* linguist-generated
+**/api-extractor.json  linguist-language=JSON-with-Comments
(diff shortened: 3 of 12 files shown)


GautamSharda:
Do we only want to remove .gitattributes from firestore but not other handwritten/ packages?

pearigee:
> Do we only want to remove .gitattributes from firestore but not other handwritten/ packages?

There are failing tests and it requires review from each of our partner teams.

I am going to send them fix and removal PRs separately.

PR #8011, opened 2026-04-10
feat: Regional Access Boundary

vverman:
Contains changes for the feature Regional Access Boundary (Previously Called Trust Boundaries).

The following are salient changes:

Calls to refresh RAB are now all async and in a separate thread.
Logic for refreshing RAB now exists in its own class for cleaner maintenance.
Self-signed jwts are within scope.
Changes to how we trigger RAB refresh and deal with refresh errors.


vverman:
Pending approval: DO NOT MERGE!

vverman:
Closing and splitting this into two separate PRs

One for migrating changes from cloud-node-core.

Another for small bug-fixes.

This reduces. reviewer burden.

PR #8016, opened 2026-04-10
fix: Ensure all headers in the same package have the same copyright year

danieljbruce:
## Description

Solves issue https://github.com/googleapis/librarian/issues/4753. Ensures that all headers in files of the same package have the same copyright year. Files containing no copyright year also now have a copyright year as this was a related issue discovered by the testing script.

## Impact

Assists with librarian migration from https://github.com/googleapis/librarian/milestone/88.

## Testing

The root directory of google-cloud-node doesn't use CI tests, but I created a check-package-years.sh script in the bin folder that only succeeds if all header files have the same copyright year for every file in a given package.

## Additional information

Headers with copyright years needed to be generated for some files:

image

Some .jsdoc generation files will also generate headers for 2026 which is desirable:

image


pearigee (commented):
Can we skip the proto files? They are just copied artifacts from google3 and the changes may not stick.

pearigee:
Also, it looks like the snapshot testing in the unit tests may have broken as a result of this?

danieljbruce:
> Also, it looks like the snapshot testing in the unit tests may have broken as a result of this?

Ah. Ok. This is a good catch. I thought the `presubmit / units` test weren't required, but the snapshots definitely should match.

danieljbruce:
> > Also, it looks like the snapshot testing in the unit tests may have broken as a result of this?
> 
> Ah. Ok. This is a good catch. I thought the `presubmit / units` test weren't required, but the snapshots definitely should match.



> > Also, it looks like the snapshot testing in the unit tests may have broken as a result of this?
> 
> Ah. Ok. This is a good catch. I thought the `presubmit / units` test weren't required, but the snapshots definitely should match.

The snapshot testing has been addressed, but the log for the other tests gets truncated.

danieljbruce:
/gemini Please write the corrected code to fix the unhandled edge case you just identified.

PR #8048, opened 2026-04-14
feat: [pubsub] add W3C Baggage support 

### LINKED ISSUE #8046: Add W3C Baggage support in telemetry-tracing
### Library Name

@google-cloud/pubsub   

The telemetry-tracing module currently uses only W3CTraceContextPropagator, which propagates traceparent and tracestate headers but not W3C Baggage (https://www.w3.org/TR/baggage/). This means arbitrary key-value metadata cannot flow through Pub/Sub messages.  

Add W3C Baggage propagation to Pub/Sub messages, so that googclient_baggage is injected/extracted alongside the existing googclient_traceparent.   

Implementing baggage propagation at the application level rather than in the client library. However, this would require every consumer to duplicate the same propagation logic manually, which is error-prone and        inconsistent with how trace context is already handled

### Additional context/notes

- All required classes (W3CBaggagePropagator, CompositePropagator) are already available in the existing @opentelemetry/core dependency.                   

- No new production dependencies are needed.

aris-cub:
add W3C Baggage support

Fixes https://github.com/googleapis/google-cloud-node/issues/8046

### DIFF
--- a/handwritten/pubsub/package.json
+++ b/handwritten/pubsub/package.json
@@ -72,6 +72,7 @@
   },
   "devDependencies": {
     "@grpc/proto-loader": "^0.8.0",
+    "@opentelemetry/context-async-hooks": "^2.6.1",
     "@opentelemetry/sdk-trace-base": "^1.17.0",
     "@types/duplexify": "^3.6.4",
     "@types/extend": "^3.0.4",
--- a/handwritten/pubsub/src/lease-manager.ts
+++ b/handwritten/pubsub/src/lease-manager.ts
@@ -15,6 +15,7 @@
  */
 
 import {EventEmitter} from 'events';
+import {context as otelContext} from '@opentelemetry/api';
 
 import {AckError, Message, Subscriber} from './subscriber';
 import {defaultOptions} from './default-options';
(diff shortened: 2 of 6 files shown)


aris-cub:
Hi @shivanee-p  and @feywind , could you take a look at this when you get a chance?

This adds W3C Baggage propagation support for Pub/Sub (related to #8046). I'm happy to address any feedback you might have!

aris-cub:
Hi @pearigee, @hongalex,@feywind and @googleapis/pubsub-team — following up on this PR.                                                 
                                                                                                                                  
We're building Pub/Sub support in [mirrord](https://github.com/metalbear-co/mirrord) and the companion PR is ready to ship: https://github.com/metalbear-co/mirrord/pull/4164. Mirrord lets developers run a local process as if it were running in their Kubernetes cluster, and uses W3C Baggage (`mirrord-session=<user>`) to route messages to the right developer's local subscriber during a shared-cluster debug session. Without baggage propagation in the subscriber callback path, Pub/Sub messages can't be routed per-developer, which is the last blocker for shipping Pub/Sub support to our users — many of whom are Google Cloud Pub/Sub customers.

  This PR is the small, additive change that unlocks that integration (linked issue #8046).
                                                                                                                                  
  Thanks for taking a look!

aris-cub:
Hey @pearigee, @hongalex,@feywind  and @shivanee-p 
Good news! We just released our support for Pub Sub . You can read more [here](https://metalbear.com/mirrord/docs/sharing-the-cluster/queue-splitting#google-cloud-pub-sub#google-cloud-pub-sub-1#google-cloud-pub-sub-1)
Would be great if we can get feedback on the PR!
Thanks!

PR #8061, opened 2026-04-15
fix(bigquery): run npm run types in correct directory in owlbot

shivanee-p:
Fixes OwlBot failure by running the command in the component directory instead of the root.

### DIFF
--- a/handwritten/bigquery/owlbot.py
+++ b/handwritten/bigquery/owlbot.py
@@ -25,5 +25,6 @@ node.owlbot_main(relative_dir="handwritten/bigquery",
     ]
 )
 
-# Regenerate Discovery types.
+# Install dependencies and regenerate Discovery types.
+shell.run(('npm', 'install'), cwd='handwritten/bigquery')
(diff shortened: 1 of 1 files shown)


handwritten/bigquery/owlbot.py
```diff
@@ -26,4 +26,4 @@
 )
 
 # Regenerate Discovery types.
-shell.run(('npm','run','types'))
+shell.run(('npm', 'run', 'types'), cwd='handwritten/bigquery')
```
iam-ethan-huang, line 30: (question only) Does this only apply to bigquery lol? And was this a new problem after monorepo I guess?

shivanee-p: Yea I went through the other handwritten libraries. Firestore already specifies the directory when running commands. Storage and Spanner don't execute npm commands during the build.

BigQuery was the only one that had the run command without specifying the directory. And yes we haven't released BigQuery since the monorepo migration so it didn't surface earlier.

PR #8067, opened 2026-04-16
feat: migrate generated packages to root ESLint config and ignore current violations

shivanee-p:
This PR migrates the generated libraries in the monorepo to use a centralized ESLint configuration based on Google TypeScript Style (GTS) standards. The previous linter checks were not sufficient to maintain consistent code quality across all packages.
Addresses #7920 🦕

### Key Changes
- **Centralized Configuration**: Moved ESLint and TypeScript configurations to the repository root, removing package-specific `.eslintrc.json` files.
- **GTS Adoption**: Configured ESLint to extend `./node_modules/gts`, applying these rules to files in `packages/` via overrides.
- **Ignore Handwritten**: Explicitly ignored the `handwritten/` directory in the root `.eslintignore` to avoid affecting those libraries in this phase.
- **Baselining**: Automated the discovery of existing violations and add ESLint disable comments to affected TypeScript files to baseline the project state.
- **Generator Update**: Added prettier as a dependency to gapic-generator-typescript and updated templater.ts to run Prettier on generated TypeScript files in memory before returning them to protoc
- **CI Check**: Added ESLint check for changed packages whenever a PR is opened (doesn't apply to handwritten libraries for right now)
### Next Steps
1. Regenerate clients using the updated generator to verify reduction in linter violations.
2. Update and apply the same standards to the handwritten libraries.

### For Reviewers
Note: the logic changes within this PR lives in the following files:

#### ESLint Configuration & CI Pipeline
- .eslintrc.json: Centralizes package tsconfig.json paths using monorepo glob patterns.
- .github/workflows/eslint.yaml: Hardens CI permissions and prevents potential shell injections.
- package.json: Adds stream-json for memory-efficient lint report parsing.

#### Automation Scripts
- packages/run-eslint-root.mjs: Chunks monorepo directories and scopes tsconfig dynamically to prevent Out Of Memory (OOM) errors.
- packages/baseline-from-output.mjs: Handles memory-safe JSON baselining and leverages Regular Expressions to safely merge existing eslint-disable headers.
- packages/process-batch-root.mjs: Hardened batch processor script using execFileSync instead of execSync to eliminate Shell Injection warnings.

#### GAPIC Generator (gapic-generator-typescript)
- core/generator/gapic-generator-typescript/typescript/src/templater.ts: Parallelizes file rendering using Promise.all() to prevent execution bottlenecks.
- core/generator/gapic-generator-typescript/BUILD.bazel: Includes prettier in the Bazel build container runtime dependencies.
- core/generator/gapic-generator-typescript/package.json: Declares prettier as a core generator dependency.

### DIFF
--- a/core/generator/gapic-generator-typescript/tsconfig.json
+++ b/core/generator/gapic-generator-typescript/tsconfig.json
@@ -5,7 +5,6 @@
     "resolveJsonModule": true,
     "module": "esnext",
     "moduleResolution": "node",
-    "ignoreDeprecations": "5.0",
     "esModuleInterop": true,
     "sourceMap": false,
     "lib": ["esnext",  "DOM"],


.github/workflows/windows-presubmit.yaml
```diff
@@ -15,10 +15,8 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: ${{ matrix.node-version }}
-      - uses: pnpm/action-setup@v4
-        with:
-          version: ^10.0.0
       - run: node --version
```
quirogas, line 18: Why do we run  --version here? Doesn't the test name have the version as well?

.github/workflows/continuous.yaml
```diff
@@ -11,10 +11,8 @@ jobs:
         node: [18, 20, 22]
     steps:
       - uses: actions/checkout@v6
-      - uses: pnpm/action-setup@v4
-        with:
-          version: ^7.0.0
       - run: node --version
+      - run: npm install
```
pearigee, line 15: Do we need another NPM install here? What directory is this running in?

Would be possible to just install the library you need for the workflow to work?

```
npm install eslint...
```

.github/workflows/eslint.yaml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # Required to get all history for diffs
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
```
pearigee, line 22: Can we start with 24 instead of 18? I think even node 20 workflows are deprecated.

.github/workflows/eslint.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Install shared linting dependencies
+        run: |
+          npm install --save-dev eslint@8 gts@6.0.2 stream-json
```
pearigee, line 26: Do we need the --save-dev flag? This updates the package.json with the new dependencies mentioned, but the updates will always be discarded after this workflow executes.

packages/run-eslint-root.mjs
```diff
@@ -0,0 +1,123 @@
...
+    // Dynamically update tsconfig scope to include ONLY the files in this chunk
+    const tsconfig = {
+      extends: "./node_modules/gts/tsconfig-google.json",
+      compilerOptions: {
+        rootDir: ".",
+        outDir: "build"
+      },
+      include: chunk
+    };
+    fs.writeFileSync(tsconfigPath, JSON.stringify(tsconfig, null, 2));
```
pearigee, line 56: Woah, this seems a little concerning to me? Why do we need to update our our tsconfig just to run the linter?

packages/run-eslint-root.mjs
```diff
@@ -0,0 +1,123 @@
+import { execSync } from 'child_process';
```
pearigee, line 1: These scripts seem really complicated, can we not just use the `eslint` CLI directly in the CI workflow?

If I understand the CLI correctly this one liner should be enough to run the linter on changed files?
```
eslint $(git diff --name-only origin/main...HEAD | grep -E "\.(js|ts|tsx)$"
```

scratch/baseline_all.sh
```diff
@@ -0,0 +1,244 @@
+#!/bin/bash
```
pearigee, line 1: Do we need a new top level directory `scratch`? What do all of these scripts do?

If we really need new scripts, we should consider placing them in the existing ci/ folder.

packages/baseline-from-output.mjs
GautamSharda, line 1: @pearigee Do we potentially want to make this script part of the post processing pipeline? If we do another generator release next week it may remove the comments added by this.

PR #8078, opened 2026-04-20
refactor(storage): gaxios migration system test fix

thiyaguk09:


handwritten/storage/src/storage-transport.ts
```diff
@@ -163,6 +169,7 @@ export class StorageTransport {
         headers,
         url: this.#buildUrl(reqOpts.url?.toString(), reqOpts.queryParameters),
         timeout: this.timeout,
+        ...({decompress: false} as any),
```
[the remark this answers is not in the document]
thiyaguk09: removed

handwritten/storage/system-test/storage.ts
```diff
@@ -1133,6 +1171,7 @@ describe('storage', function () {
         for (;;) {
           try {
             const [aclAfter] = await bucket.acl.default.get();
+            console.log(aclAfter);
```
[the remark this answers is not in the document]
thiyaguk09: removed

handwritten/storage/system-test/storage.ts
```diff
@@ -2759,11 +2820,20 @@ describe('storage', function () {
...
+          assert.strictEqual(metadata.crc32c, undefined);
+        }
       });
 
       it('should get the hashes from the encrypted file', async () => {
         const [metadata] = await file.getMetadata();
+        console.log('Hashes present:', {
+          crc32c: metadata.crc32c,
+          md5Hash: metadata.md5Hash
+        });
```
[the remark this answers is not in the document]
thiyaguk09: removed

handwritten/storage/system-test/kitchen.ts
```diff
@@ -96,7 +99,7 @@ describe('resumable-upload', () => {
     // see: https://cloud.google.com/storage/docs/exponential-backoff:
     const ms = Math.pow(2, retries) * 1000 + Math.random() * 2000;
     console.info(`retrying "${title}" in ${ms}ms`);
-    setTimeout(done(), ms);
+    setTimeout(() => done, ms);
```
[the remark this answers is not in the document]
thiyaguk09: updated with `setTimeout(() => { done(); }, ms);`

handwritten/storage/system-test/storage.ts
```diff
@@ -2551,48 +2691,26 @@ describe('storage', function () {
...
-
-      const contents: Buffer = await new Promise((resolve, reject) => {
-        return file.download((error, content) => {
-          if (error) return reject(error);
-          resolve(content);
-        });
-      });
-
+      const [file] = await bucket.upload(tmpGzFilePath, options);
+      const [contents] = await file.download({decompress: false});
```
Dhriti07: I noticed we had to add decompress: false to a few tests. I wonder if this means gaxios will now auto-decompress payloads by default, and that was not the case earlier? Just want to make sure we don't accidentally change behaviors for users.

handwritten/storage/system-test/storage.ts
```diff
@@ -2551,48 +2691,26 @@ describe('storage', function () {
...
-            assert.strictEqual(
-              raw.toJSON().headers['content-encoding'],
-              undefined,
-            );
-          })
-          .pipe(fs.createWriteStream(tmpFilePath))
-          .on('error', reject)
-          .on('finish', () => resolve());
+      const [contents] = await file.download({
+        decompress: false,
```
Dhriti07: I noticed we added decompress: false to the download options for this test. The test says "should skip validation if file is served decompressed", since we're telling the SDK not to decompress the file, it seems like we might be bypassing the logic the test was originally designed to verify

handwritten/storage/system-test/storage.ts
```diff
@@ -4524,8 +4718,20 @@ describe('storage', function () {
...
 
-  function deleteFileAsync(file: File) {
-    return file.delete();
+  async function deleteFileAsync(file: File) {
+    try {
+      return await file.delete();
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (err: any) {
+      const status =
+        err.status || (err.response && err.response.status) || err.code;
```
Dhriti07: Do we have GaxiosErrors here instead of translating them into our usual ApiError format ? Is this intentional ? If this is happening I am worried our users will have to write similar checks. It would better to have handleResp catch and format these gaxios error ?

handwritten/storage/system-test/storage.ts
```diff
@@ -2531,7 +2668,10 @@ describe('storage', function () {
       };
       const expectedContents = fs.readFileSync(FILES.html.path, 'utf-8');
       const [file] = await bucket.upload(FILES.html.path, options);
-      const [contents] = await file.download();
+      const [contents] = await file.download({
+        validation: false,
```
Dhriti07: Why have we disabled validation and decompression for this test ?

handwritten/storage/system-test/storage.ts
```diff
@@ -4524,8 +4718,20 @@ describe('storage', function () {
...
-    return file.delete();
+  async function deleteFileAsync(file: File) {
+    try {
+      return await file.delete();
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (err: any) {
+      const status =
+        err.status || (err.response && err.response.status) || err.code;
+
+      if (status === 404 || err.message.includes('No such object')) {
```
Dhriti07: We added a check for err.message.includes('No such object') along with the status check. Does this imply that the HTTP Status code is getting lost with the new gaxios errors ?

handwritten/storage/system-test/storage.ts
```diff
@@ -1077,7 +1187,9 @@ describe('storage', function () {
       const validateUniformBucketLevelAccessEnabledError = (
         err: GaxiosError,
       ) => {
-        assert.strictEqual(err.code, 400);
+        const status = err.response ? err.response.status : Number(err.code);
+        const isExpected = [400, 404].includes(status);
```
Dhriti07: We changed this test to accept both 400 and 404, whereas it previously strictly expected a 400. Can you help understand why we expect a 404 now ?

handwritten/storage/system-test/storage.ts
thiyaguk09: I've reverted this test change so it runs against default settings.

handwritten/storage/system-test/storage.ts
thiyaguk09: This was a temporary change to bypass the `incorrect header check` crash while debugging. Now that `compress: false` is implemented inside the internal network options, I've removed these overrides so the test runs against the default behavior.

handwritten/storage/system-test/storage.ts
thiyaguk09: Users won't experience a behavior change. `gaxios`/`node-fetch` was automatically decompressing payloads under the hood, crashing our manual pipeline. By forcing `compress: false` inside `createReadStream`, we ensure the SDK receives the raw network bytes for validation, then we handle user-requested decompression manually using `zlib`.

handwritten/storage/system-test/storage.ts
thiyaguk09: The reason the test started encountering both 400 and 404 statuses is that a raw network failure or a malformed request stream (caused by the earlier decompression issue) returns a 400 Bad Request, whereas a missing file returns a 404 Not Found.

thiyaguk09: Now that compress: false is fixed and the error parsing above is uniform, revert the test assertion back to its strict original state.

handwritten/storage/system-test/storage.ts
thiyaguk09: No, the status code isn't lost, but gaxios formats errors differently than our previous transport library.

thiyaguk09: Fixed. The status code is no longer lost now that StorageTransport handles the formatting.

handwritten/storage/system-test/storage.ts
thiyaguk09: I've updated StorageTransport to enrich GaxiosError instances with standard properties. This preserves the type while ensuring full backward compatibility with our standard ApiError shapes.

handwritten/storage/system-test/storage.ts
```diff
@@ -449,7 +525,12 @@ describe('storage', function () {
         assert.strictEqual(encryptionAlgorithm, 'AES256');
       });
 
-      it('should make a file public during the upload', async () => {
+      it('should make a file public during the upload', async function () {
+        const [metadata] = await bucket.getMetadata();
+        if (metadata.iamConfiguration?.uniformBucketLevelAccess?.enabled) {
```
Dhriti07: Is UBLA enforced on the CI Project ?

thiyaguk09: Yes, It was recently enabled, which is why we skipped the tests here.

PR #8082, opened 2026-04-20
chore: consolidate prettierignore and prettierrc files: autogen

feywind:
Split from https://github.com/googleapis/google-cloud-node/pull/7948 to handle just autogen library changes.

Regarding: https://github.com/googleapis/librarian/issues/4581

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -15,3 +15,4 @@
 /handwritten/bigquery-storage @googleapis/bigquery-team
 /handwritten/pubsub @googleapis/pubsub-team
 /handwritten/bigtable @googleapis/bigtable-team
+/core/packages/google-auth-library-nodejs @googleapis/aion-team
--- a/core/generator/gapic-generator-typescript/package-lock.json
+++ b/core/generator/gapic-generator-typescript/package-lock.json
@@ -14,7 +14,7 @@
         "long": "^5.3.2",
         "object-hash": "^3.0.0",
         "proto3-json-serializer": "^3.0.3",
-        "protobufjs": "^7.5.4",
+        "protobufjs": "^7.5.5",
         "protobufjs-cli": "^1.2.0",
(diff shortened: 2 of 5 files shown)


packages/google-cloud-automl/package.json
```diff
@@ -14,6 +14,7 @@
   },
   "main": "build/src/index.js",
   "files": [
+    "build/protos",
```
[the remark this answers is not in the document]
feywind: These two entries could be turned into tasks for the generator updates.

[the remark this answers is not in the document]
feywind: Okay, actually, Gemini alerted me here to a problem, but not at all the one it thought it found 😹

packages/google-storage-control/librarian.js
pearigee, line 1: Do we need this change? Can we remove it?

packages/google-storage-control/src/util/uber_client_builder.ts
pearigee, line 1: Was this change made by running the formatter?

pearigee:
Do we need to update the generator as well to stop automatically producing these? Is that in a separate PR?

packages/google-storage-control/librarian.js
feywind: These changes (the exitCode and the `\*` in the other file) were both hard errors on the linter when running through the CI. They're sensible changes anyway, so I am okay with going ahead with them. I'm not sure why they're only on storage-control, but this library is fairly old, so it might've predated the harmonization efforts.

feywind:
> Do we need to update the generator as well to stop automatically producing these? Is that in a separate PR?

It looks like I filtered that out in this PR, but it might be best to do it separately anyway.

feywind:
> Do we need to update the generator as well to stop automatically producing these? Is that in a separate PR?

https://github.com/googleapis/google-cloud-node/pull/8104

PR #8088, opened 2026-04-21
fix: Upgrade protobufjs and fix legacy key decoding in Datastore

danieljbruce:
## Description

This pull request provides a better fix for what https://github.com/googleapis/google-cloud-node/pull/8071 intends to fix. It resolves the protobufjs dependency issue and it ensures that unit tests don't break.

## Impact

1. Resolves security issue
2. Ensures datastore unit tests pass

## Testing

Tests have been added to ensure the legacyEncode and legacyDecode functionality remain the same. The only difference will be that with the new src changes, code that used to produce an `invalid wire` error will now produce a valid output.

## Generated notes

Upgrades `protobufjs` to `^7.5.5` to address security vulnerabilities and fixes a regression in legacy App Engine key decoding. The fix involves calling `.setup()` on the affected protobuf types to ensure group support is preserved.

---
*PR created automatically by Jules for task [16655518185677337093](https://jules.google.com/task/16655518185677337093) started by @danieljbruce*

### DIFF
--- a/handwritten/datastore/test/entity.ts
+++ b/handwritten/datastore/test/entity.ts
@@ -1849,57 +1849,62 @@ describe('entity', () => {
         );
       });
 
-      const TEST_PROJECT = 'test-project';
-      const testCases = [
-        {name: 'numeric ID', path: ['Kind', '123']},
-        {name: 'string name with spaces', path: ['Kind', 'name with spaces']},
-        {name: 'special characters', path: ['Kind', 'special!@#$%^&*()']},
-        {
-          name: '3-level parent',
-          path: ['Grandparent', '1', 'Parent', 'p1', 'Child', '2'],
-        },
-        {
-          name: 'namespace and numeric ID',
-          path: ['Kind', '456'],
-          namespace: 'MyNS',
-        },
-        {
-          name: 'namespace and parent',
-          path: ['Parent', '1', 'Child', 'c1'],
-          namespace: 'MyNS',
-        },
-        {
-          name: 'long integer ID as string',
-          path: ['Kind', '9223372036854775807'],
-        },
-        {name: 'kind with hyphens', path: ['My-Kind', '1']},
-        {
-          name: 'different kinds in path',
-          path: ['User', 'user1', 'Post', '100', 'Comment', 'c1'],
-        },
-        {name: 'same kinds in path', path: ['Node', '1', 'Node', '2', 'Node', '3']},
-      ];
+      describe('should ensure that decode inverses encode and decoding is correct', () => {
+        const TEST_PROJECT = 'test-project';
+        const testCases = [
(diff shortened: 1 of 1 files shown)


handwritten/datastore/src/entity.ts
```diff
@@ -1302,6 +1302,14 @@ export namespace entity {
         path.join(__dirname, '..', 'protos', 'app_engine_key.proto'),
       );
       loadedRoot.resolveAll();
+
+      // Ensure all types are initialized and setup correctly.
+      // In newer versions of protobufjs, this ensures group support
+      // is preserved for legacy keys.
+      for (const typeName of ['Reference', 'Path', 'Element']) {
+        (loadedRoot.lookupType(typeName) as any).setup();
```
pearigee: Can we remove the `as any`? If possible we should just import and use the correct type.

handwritten/datastore/test/entity.ts
```diff
@@ -1831,6 +1843,64 @@ describe('entity', () => {
         ]);
         assert.strictEqual(key.parent!.name, 'sampletask1');
         assert.deepStrictEqual(key.parent!.path, ['Task', 'sampletask1']);
+        assert.strictEqual(
+          urlSafeKey.legacyEncode(PROJECT_ID, key, LOCATION_PREFIX),
+          encodedKey,
+        );
+      });
+
+      const TEST_PROJECT = 'test-project';
```
pearigee: Should this line and below just be its own test? Nesting `it`'s within `it`'s seems needlessly nested for the reader.

Alternatively, you could make a new root `describe` to group the tests? For example:

```
const testCases = [...];

describe('decoding keys`, () => {
   testCases.foreEach(tc => {
       it(`should decode "${tc.name}"`, () => {...});
   });
}); 

describe('encoding keys`, () => {
   testCases.foreEach(tc => {
       it(`should ecode "${tc.name}"`, () => {...});
   });
}); 
```

handwritten/datastore/src/entity.ts
danieljbruce: Yes. It looks like when we remove `as any` then this works just fine.

handwritten/datastore/test/entity.ts
danieljbruce: Good idea. I added a describe block to group the tests. I'm not sure what you mean by nesting `it` inside `it` though.

pearigee: Before, if I understood correctly, we were doing something like this:

```
it('root test' () => {
    tests.forEach(t => {
        // Here we have an `it`, inside another `it`.
        it(`child test ${t.name}`, () => {

         });
    })
});
```

This isn't the convention I have seen in the past. It's common to nest tests in a describe, but not tests in tests.

PR #8102, opened 2026-04-23
feat(firestore): add public type guards for FieldValue sentinels

kyungseopk1m:
## Summary
Adds five public static type guards on `FieldValue` so users can detect sentinel values without relying on internal class names like `NumericIncrementTransform`.

Addresses https://github.com/firebase/firebase-admin-node/issues/2957 (external contribution invited by @ehsannas).

- `FieldValue.isServerTimestamp(value)`
- `FieldValue.isIncrement(value)`
- `FieldValue.isDecrement(value)` — strict subset of `isIncrement`, true only when the operand is negative
- `FieldValue.isArrayUnion(value)`
- `FieldValue.isArrayRemove(value)`

Each guard is a TypeScript type predicate returning `value is FieldValue`.

## Design notes
- Shape matches the "option 1" suggestion in firebase/firebase-admin-node#2957.
- `isDecrement` is documented as a strict subset of `isIncrement` (same underlying `NumericIncrementTransform`, narrowed by a negative operand). An alternative "partition" design (`isIncrement` = positive, `isDecrement` = negative, zero = neither) was considered; the subset design was chosen so `isIncrement` consistently behaves as a type detector for any increment sentinel. Happy to flip this if you prefer the partition semantics.
- Increment/decrement *amount* accessors are deliberately out of scope here — the original issue raised the question but did not commit to an answer. Happy to follow up in a separate PR if wanted.
- Uses `instanceof` against existing internal transform classes — no new runtime surface, no changes to proto serialization, transforms, or validation paths.

## Testing
- New unit coverage in `handwritten/firestore/dev/test/field-value.ts` for all five guards: match / non-match / primitive / delete-sentinel / cross-sentinel cases, plus explicit `isDecrement` ⊂ `isIncrement` assertion.
- `mocha build/test/field-value.js` → 41 passing.
- Full unit suite (`mocha build/test`) → 1084 passing, 47 pending (no regressions).

### DIFF
--- a/handwritten/firestore/dev/src/field-value.ts
+++ b/handwritten/firestore/dev/src/field-value.ts
@@ -340,6 +340,23 @@ export class FieldValue implements firestore.FieldValue {
   }
 
   /**
+   * Returns `true` if the provided value is a sentinel returned by
+   * {@link FieldValue.delete}.
+   *
+   * @param value The value to check.
+   * @returns `true` if `value` is a delete sentinel.
+   *
+   * @example
+   * ```
+   * Firestore.FieldValue.isDelete(Firestore.FieldValue.delete()); // true
+   * Firestore.FieldValue.isDelete(null); // false
+   * ```
+   */
+  static isDelete(value: unknown): value is FieldValue {
+    return value instanceof DeleteTransform;
+  }
+
+  /**
    * Returns true if this `FieldValue` is equal to the provided value.
    *
    * @param {*} other The value to compare against.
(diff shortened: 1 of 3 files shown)


kyungseopk1m:
Thanks @gemini-code-assist! Added `isDelete` (plus its declaration in `firestore.d.ts` and unit tests) in ca2cc578c1fa3c1d470764a4790cd38c8666aa7f. The new commit keeps the same `instanceof`-based pattern as the other four guards.

PR #8121, opened 2026-04-27
fix: pin @sinonjs/fake-timers to unblock build in bigquery

shivanee-p:
similar to PR[#8098](https://github.com/googleapis/google-cloud-node/pull/8098)

Pins the sinon dependency to unblock the BigQuery library release since the `FakeTimerInstallOpts` interface was recently removed which breaks the old version of @types/sinon

### DIFF
--- a/handwritten/bigquery/src/types.d.ts
+++ b/handwritten/bigquery/src/types.d.ts
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 /**
- * Discovery Revision: 20260118
+ * Discovery Revision: 20260408
  */
 
 /**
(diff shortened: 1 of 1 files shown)


handwritten/bigquery/package.json
pearigee, line 1: Please pin it to the same version as PR: https://github.com/googleapis/google-cloud-node/pull/8001/changes

Note, you may also need to provide package overrides to fully resolve the issue:
https://github.com/googleapis/google-cloud-node/blob/49854c621594fea69da44912f39b176a2387d958/handwritten/firestore/package.json#L125

Unfortunately future minor versions introduced yet more breaking changes.

PR #8127, opened 2026-04-28
fix(spanner): missing callback parameter in close

### LINKED ISSUE #8106: spanner: Spanner.close() never resolves when called as a Promise
#### Environment details

- OS: macOS (Darwin 25.3.0) / also reproducible on Linux (Cloud Run)
- Node.js version: 24.14.1
- `@google-cloud/spanner` version: 8.6.0
- `@grpc/grpc-js` version: 1.14.3

#### Steps to reproduce

```js
// repro.mjs — run with: node repro.mjs
import { Spanner } from "@google-cloud/spanner";
import { credentials } from "@grpc/grpc-js";

const spanner = new Spanner({
  projectId: "test-project",
  servicePath: "localhost",
  port: 15000,
  sslCreds: credentials.createInsecure(),
});

console.log("calling await spanner.close()...");
await spanner.close();
console.log("this line is never reached");
```

```
$ node repro.mjs
calling await spanner.close()...
Warning: Detected unsettled top-level await at file:///repro.mjs:12
await spanner.close();
^

$ echo $?
13
```

No Spanner instance is required — the bug is in the `close()` method itself.

#### Expected behavior

`await spanner.close()` should resolve once cleanup is complete.

#### Actual behavior

The Promise returned by `spanner.close()` never resolves. On Node 24, this causes the process to exit with code 13 ("unsettled top-level await"). On earlier Node versions, the process hangs until forcefully terminated.

#### Root cause

`Spanner.close()` is wrapped by `promisifyAll()` at the bottom of `src/index.ts`, and `close` is not in the `exclude` list:

https://github.com/googleapis/google-cloud-node/blob/main/packages/handwritten/src/spanner/src/index.ts#L2114-L2128

When called without a callback argument, the promisify wrapper creates a Promise and passes a synthetic callback to the original method, expecting it to be invoked. However, `Spanner.close()` ignores all arguments:

https://github.com/googleapis/google-cloud-node/blob/main/packages/handwritten/src/spanner/src/index.ts#L543-L554

```ts
close() {
  this.clients_.forEach(c => {
    const client = c as GapicClient;
    if (client.operationsClient && client.operationsClient.close) {
      client.operationsClient.close();
    }
    client.close();
  });
  cleanup().catch(err => {
    console.error('Error occured during cleanup: ', err);
  });
  // ← callback is never called
}
```

Since the callback is never invoked, the Promise stays pending forever.

The TypeScript declaration (`close(): void`) correctly reflects the implementation, but at runtime the promisify wrapper overrides the return type.

A similar issue was reported for `Database.close()` in https://github.com/googleapis/nodejs-spanner/issues/1828.

#### Suggested fix

Either:

1. Add `'close'` to the `promisifyAll` exclude list (if `close()` is intended to remain fire-and-forget), or
2. Accept and invoke the callback parameter:

```ts
close(callback?: (err: Error | null) => void) {
  this.clients_.forEach(c => {
    const client = c as GapicClient;
    if (client.operationsClient && client.operationsClient.close) {
      client.operationsClient.close();
    }
    client.close();
  });
  cleanup()
    .then(() => callback?.(null))
    .catch(err => callback?.(err));
}
```

alkatrivedi:
fixes: #8106

### DIFF
--- a/handwritten/spanner/src/index.ts
+++ b/handwritten/spanner/src/index.ts
@@ -573,17 +573,48 @@ class Spanner extends GrpcService {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const client = c as any;
       if (client.operationsClient && client.operationsClient.close) {
-        promises.push(client.operationsClient.close());
+        try {
+          promises.push(Promise.resolve(client.operationsClient.close()));
+        } catch (err) {
+          promises.push(Promise.reject(err));
+        }
       }
-      promises.push(client.close());
+      try {
+        if (client.close) {
+          promises.push(Promise.resolve(client.close()));
+        }
+      } catch (err) {
+        promises.push(Promise.reject(err));
+      }
     });
 
-    // Wait for all clients to close, then do cleanup
-    const res = Promise.all(promises).then(() => cleanup());
+    // Wait for all clients to close (even if some fail), then do cleanup
+    const res = Promise.all(
+      promises.map(p =>
+        p.then(
+          () => undefined,
+          err => err || new Error('Unknown error during close'),
+        ),
+      ),
+    ).then(results => {
+      return cleanup()
+        .catch(err => {
+          console.error('Error occured during cleanup: ', err);
+        })
+        .then(() => {
+          // results contains `undefined` for success, or the error for failure
+          const error = results.find(r => r !== undefined);
+          if (error) {
+            throw error;
+          }
+        });
+    });
+
     if (callback) {
+      // process.nextTick prevents Unhandled Promise Rejections if callback throws
       res.then(
-        () => callback(null),
-        err => callback(err),
+        () => process.nextTick(() => callback(null)),
+        err => process.nextTick(() => callback(err)),
       );
     } else {
       return res;
--- a/handwritten/spanner/src/index.ts
+++ b/handwritten/spanner/src/index.ts
@@ -567,48 +567,50 @@ class Spanner extends GrpcService {
   close(): Promise<void>;
   close(callback: (err: Error | null) => void): void;
   close(callback?: (err: Error | null) => void): void | Promise<void> {
-    const promises: Promise<void>[] = [];
+    const performTeardown = async () => {
+      const promises: Promise<void>[] = [];
 
-    this.clients_.forEach(c => {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const client = c as any;
-      if (client.operationsClient && client.operationsClient.close) {
-        try {
-          promises.push(Promise.resolve(client.operationsClient.close()));
-        } catch (err) {
-          promises.push(Promise.reject(err));
+      this.clients_.forEach(c => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const client = c as any;
+
+        // Promise.resolve().then() safely converts synchronous throws into Promise rejections
+        if (client.operationsClient && client.operationsClient.close) {
+          promises.push(
+            Promise.resolve().then(() => client.operationsClient.close()),
+          );
         }
-      }
-      try {
         if (client.close) {
-          promises.push(Promise.resolve(client.close()));
+          promises.push(Promise.resolve().then(() => client.close()));
         }
-      } catch (err) {
-        promises.push(Promise.reject(err));
-      }
-    });
+      });
 
-    // Wait for all clients to close (even if some fail), then do cleanup
-    const res = Promise.all(
-      promises.map(p =>
-        p.then(
-          () => undefined,
-          err => err || new Error('Unknown error during close'),
+      // Wait for all close attempts to settle.
+      // Map success to undefined, and failure to the error.
+      const results = await Promise.all(
+        promises.map(p =>
+          p.then(
+            () => undefined,
+            err => err || new Error('Unknown error during close'),
+          ),
         ),
-      ),
-    ).then(results => {
-      return cleanup()
-        .catch(err => {
-          console.error('Error occured during cleanup: ', err);
-        })
-        .then(() => {
-          // results contains `undefined` for success, or the error for failure
-          const error = results.find(r => r !== undefined);
-          if (error) {
-            throw error;
-          }
-        });
-    });
+      );
+
+      // Always execute cleanup
+      try {
+        await cleanup();
+      } catch (err) {
+        console.error('Error occured during cleanup: ', err);
+      }
+
+      // If any client failed to close, throw the first error we captured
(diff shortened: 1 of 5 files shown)


handwritten/spanner/src/index.ts
```diff
@@ -560,18 +560,32 @@ class Spanner extends GrpcService {
   }
 
   /** Closes this Spanner client and cleans up all resources used by it. */
-  close(): void {
+  close(callback?: (err: Error | null) => void): void {
```
surbhigarg92: I think we should return promise as well for customers who would want to await rather than using callback ?

handwritten/spanner/test/index.ts
```diff
@@ -2397,4 +2397,71 @@ describe('Spanner', () => {
         .emit('reading');
     });
   });
+
+  describe('close', () => {
+    it('should close all cached clients', done => {
```
surbhigarg92: We should use async/await for tests instead of "done". 

With "done" tests get stuck when it fails.

handwritten/spanner/src/index.ts
alkatrivedi: the methods is included in `promisifyAll` [here ](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/spanner/src/index.ts#L2223-#L2240), hence, we need not to explicitly return the promise here

handwritten/spanner/test/index.ts
alkatrivedi: done

handwritten/spanner/src/index.ts
alkatrivedi: since promisifyAll will promisify the method only during runtime, hence it the typescript customers wont be able to await on it

to fix the gap, I have added promise return type as well

handwritten/spanner/src/index.ts
```diff
@@ -569,9 +569,9 @@ class Spanner extends GrpcService {
       }
       client.close();
     });
-    cleanup().catch(err => {
-      console.error('Error occured during cleanup: ', err);
-    });
+    cleanup()
+      .then(() => callback?.(null))
+      .catch(err => callback?.(err));
```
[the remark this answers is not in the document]
alkatrivedi: done

handwritten/spanner/src/index.ts
```diff
@@ -560,18 +560,34 @@ class Spanner extends GrpcService {
...
-    });
-    cleanup().catch(err => {
-      console.error('Error occured during cleanup: ', err);
+      promises.push(client.close());
     });
+
+    // Wait for all clients to close, then do cleanup
+    Promise.all(promises)
+      .then(() => cleanup())
+      .then(
```
surbhigarg92: This needs to be conditional. If async/await is used instead of callback then we need to return the promise.

handwritten/spanner/src/index.ts
```diff
@@ -564,18 +564,34 @@ class Spanner extends GrpcService {
...
+        () => callback?.(null), // Success handler
+        err => {
+          // Error handler (only catches cleanup/close errors)
+          if (callback) {
+            callback(err);
+          } else {
+            console.error('Error occurred during cleanup: ', err);
+          }
+        },
+      );
```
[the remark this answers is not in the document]
alkatrivedi: done

handwritten/spanner/src/index.ts
alkatrivedi: done

surbhigarg92 (approved):

surbhigarg92 (changes requested):
Address the gemini review comments

handwritten/spanner/src/index.ts
```diff
@@ -564,18 +564,30 @@ class Spanner extends GrpcService {
...
+    const res = Promise.all(promises).then(() => cleanup());
+    if (callback) {
+      res.then(
+        () => callback(null),
+        err => callback(err),
+      );
+    } else {
+      return res;
+    }
   }
```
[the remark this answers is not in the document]
alkatrivedi: done

handwritten/spanner/src/index.ts
```diff
@@ -564,18 +564,57 @@ class Spanner extends GrpcService {
...
+      ),
+    ).then(results => {
+      return cleanup().then(() => {
+        // results contains `undefined` for success, or the error for failure
+        const error = results.find(r => r !== undefined);
+        if (error) {
+          throw error;
+        }
+      });
     });
```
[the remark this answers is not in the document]
alkatrivedi: done

handwritten/spanner/src/index.ts
```diff
@@ -564,18 +564,61 @@ class Spanner extends GrpcService {
   }
 
   /** Closes this Spanner client and cleans up all resources used by it. */
-  close(): void {
+  close(): Promise<void>;
```
surbhigarg92, line 567: Logic is correct but I think you could simplify this logic a bit and make it more readable

```
  close(): Promise<void>;
  close(callback: (err: Error | null) => void): void;
  close(callback?: (err: Error | null) => void): void | Promise<void> {
    
    // Wrap the core logic in an async function for readability
    const performTeardown = async () => {
      const promises: Promise<void>[] = [];

      this.clients_.forEach(c => {
        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        const client = c as any;
        
        // Promise.resolve().then() safely converts synchronous throws into Promise rejections
        if (client.operationsClient && client.operationsClient.close) {
          promises.push(Promise.resolve().then(() => client.operationsClient.close()));
        }
        if (client.close) {
          promises.push(Promise.resolve().then(() => client.close()));
        }
      });

      // Wait for all close attempts to settle. 
      // We catch errors here so one failure doesn't abort the Promise.all
      const results = await Promise.all(
        promises.map(p => p.catch(err => err || new Error('Unknown error during close')))
      );

      // Always execute cleanup
      try {
        await cleanup();
      } catch (err) {
        console.error('Error occured during cleanup: ', err);
      }

      // If any client failed to close, throw the first error we captured
      const firstError = results.find(r => r instanceof Error);
      if (firstError) {
        throw firstError;
      }
    };

    const res = performTeardown();

    if (callback) {
      // process.nextTick prevents Unhandled Promise Rejections if callback throws
      res.then(
        () => process.nextTick(() => callback(null)),
        err => process.nextTick(() => callback(err)),
      );
    } else {
      return res;
    }
  }

```

surbhigarg92 (approved):

handwritten/spanner/src/index.ts
alkatrivedi: done

PR #8129, opened 2026-04-28
Upgrade protobufjs and fix nested type resolution regression

danieljbruce:
This change upgrades protobufjs to ^7.5.5 in the handwritten/bigquery-storage library to address security vulnerabilities. It also introduces a code fix in JSONEncoder to explicitly call .setup() on types created from descriptors, which prevents a known regression in protobufjs (version 7.5.2 and higher) where nested types fail to resolve correctly during encoding. Additionally, the static protos have been regenerated using the newer version of the toolset.

---
*PR created automatically by Jules for task [10287685281025280340](https://jules.google.com/task/10287685281025280340) started by @danieljbruce*


danieljbruce:
experimental

PR #8162, opened 2026-05-04
chore: remove bad entries

codyoss:
This was added during the migration in https://github.com/googleapis/google-cloud-node/pull/7606 but there is no manifest entry associated with these, and they don't look like they are released things.


codyoss:
Closing in favor of https://github.com/googleapis/google-cloud-node/pull/8171

PR #8170, opened 2026-05-05
fix(spanner): disable metrics tracer when built-in metrics are disabled

rahul2393:
### Summary
  - Track built-in metrics enablement on each `Spanner` instance
  - Sync `MetricsTracerFactory.enabled` with the instance's computed metrics state during `configureMetrics_()`
  - Prevent `MetricsTracer` and metric interceptor creation when built-in metrics are disabled via client option, env var, or insecure credentials

  ### Why
  When `disableBuiltInMetrics` was true, metrics export setup was skipped, but `MetricsTracerFactory.enabled` could still remain true from its default value or prior state. If `projectId_` was set, requests could still create `MetricsTracer`  instances even though built-in metrics were disabled.

  This change computes metrics enablement during client setup and uses the instance-level state to guard tracer/
  interceptor creation.

### DIFF
--- a/handwritten/spanner/src/index.ts
+++ b/handwritten/spanner/src/index.ts
@@ -1632,6 +1632,7 @@ class Spanner extends GrpcService {
       !!disableBuiltInMetrics;
     this._metricsEnabled =
       !metricsExplicitlyDisabled && !this._isInSecureCredentials;
+    MetricsTracerFactory.enabled = this._metricsEnabled;
     if (this._metricsEnabled) {
       try {
         this.auth.getProjectId((err, projectId) => {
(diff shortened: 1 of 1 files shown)


handwritten/spanner/src/index.ts
```diff
@@ -1626,10 +1627,11 @@ class Spanner extends GrpcService {
...
     // Only enable metrics if not explicitly disabled and we are not using
     // insecure credentials.
-    const metricsEnabled =
-      process.env.SPANNER_DISABLE_BUILTIN_METRICS !== 'true' &&
-      !disableBuiltInMetrics &&
-      !this._isInSecureCredentials;
+    const metricsExplicitlyDisabled =
+      process.env.SPANNER_DISABLE_BUILTIN_METRICS === 'true' ||
+      !!disableBuiltInMetrics;
+    this._metricsEnabled = !metricsExplicitlyDisabled;
```
surbhigarg92: Why don't we consider `_isInSecureCredentials` check also in `_metricsEnabled` ? 

Else the interceptor will still be pushed if insecureCredentials are used.

PR #8178, opened 2026-05-05
fix: bump all node submodules

codyoss:
This changes adds a newline to the CHANGELOG of each node package declared in release-please-submodules.json. This should have the effect of creating a patch release for each of these libraries if they have no other active changes already.

Also, had to edit some test scripts to avoid deprecation warns that are for things not deprecated yet in our lowest supported version of node.


codyoss:
bigtable does not have a `cloudbuild.yaml` nor did it nodejs-bigtable. Ignoring that failure for now.

codyoss:
Merging as the two failed checks are not related to this change directly or are blocking.

PR #8190, opened 2026-05-06
fix(storage): Invocation ID is not retained on multipart upload retries

thiyaguk09:
Hoists the generation of `persistentInvocationId` to the beginning of the
upload process in `Bucket.upload` and `File.save`. This ensures that retried multipart upload attempts reuse the same invocation ID in the `x-goog-api-client` header, rather than generating a new one for each attempt.


handwritten/storage/src/storage-transport.ts
```diff
@@ -124,76 +128,109 @@ export class StorageTransport {
...
+      const responseHandler = (resp: GaxiosResponse<T>) => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const data = resp.data as any;
+        if (data !== null && typeof data === 'object') {
+          data.headers = resp.headers;
+          data.status = resp.status;
+          return data;
+        }
+        return resp;
+      };
```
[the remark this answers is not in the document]
thiyaguk09: required for retry operations

handwritten/storage/src/storage-transport.ts
```diff
@@ -124,76 +128,109 @@ export class StorageTransport {
...
+    const headers = this.#prepareHeaders(reqOpts);
+
+    // Interceptor Management
+    this.gaxiosInstance.interceptors.request.clear();
     if (reqOpts.interceptors) {
-      this.gaxiosInstance.interceptors.request.clear();
       for (const inter of reqOpts.interceptors) {
         this.gaxiosInstance.interceptors.request.add(inter);
       }
     }
```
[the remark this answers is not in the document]
thiyaguk09: required for retry operations fix in conformance Test PR

handwritten/storage/src/storage-transport.ts
```diff
@@ -124,76 +128,109 @@ export class StorageTransport {
...
           retry: this.retryOptions.maxRetries,
           noResponseRetries: this.retryOptions.maxRetries,
           maxRetryDelay: this.retryOptions.maxRetryDelay,
           retryDelayMultiplier: this.retryOptions.retryDelayMultiplier,
-          shouldRetry: this.retryOptions.retryableErrorFn,
           totalTimeout: this.retryOptions.totalTimeout,
+          shouldRetry: err => !!this.retryOptions.retryableErrorFn?.(err),
         },
         ...reqOpts,
+        data: reqOpts.body,
```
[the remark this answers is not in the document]
thiyaguk09: removed

handwritten/storage/src/storage-transport.ts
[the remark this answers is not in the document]
Dhriti07: What is the retry operations fix and why do we need to make this change in the current PR ? Same for the other comments below.

handwritten/storage/src/storage-transport.ts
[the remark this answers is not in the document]
Dhriti07: Please elaborate as stated in the above comment.

handwritten/storage/src/storage-transport.ts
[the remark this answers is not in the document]
thiyaguk09: To support custom headers required by conformance tests without creating race conditions where concurrent async requests overwrite each other's configuration on a shared instance, we modified `makeRequest` to isolate custom interceptors using a fresh, request-scoped `Gaxios` instance, while explicitly stripping the `adapter` property during cloning to prevent infinite recursion and heap out-of-memory crashes between `GoogleAuth` and Gaxios.

handwritten/storage/src/storage-transport.ts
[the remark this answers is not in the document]
thiyaguk09: To preserve backward compatibility for legacy SDK methods that expect resource metadata objects to carry `.headers` and `.status` properties without introducing runtime errors or mutating binary data, we added a strict validation check to ensure these properties are strictly injected into plain JSON response objects and never onto `Buffer` or `Stream` payloads.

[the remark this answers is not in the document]
Dhriti07: I agree with using isPlanObject to avoid mutating Buffer or Stream payloads.

However the issue mentioned above refers to the fallback "return resp;" statement. If the payload is a Buffer or Stream and the check fails, returning the entire GaxiosResponse wrapper instead of the raw data might break backward compatibility for binary downloads, since it previously resolved with just the data.

What do you think about returning data instead in that fallback case to keep the return types consistent?

handwritten/storage/src/storage-transport.ts
```diff
@@ -124,76 +129,138 @@ export class StorageTransport {
...
+    const bodyStr = typeof reqOpts.body === 'string'
+      ? reqOpts.body
+      : (typeof reqOpts.data === 'string' ? reqOpts.data : '');
+
+    const hasPrecondition = !!(
+      reqOpts.hasPrecondition ||
+      reqOpts.queryParameters?.ifGenerationMatch !== undefined ||
+      reqOpts.queryParameters?.ifMetagenerationMatch !== undefined ||
+      reqOpts.queryParameters?.ifSourceGenerationMatch !== undefined ||
+      bodyStr.includes('"etag"')
```
Dhriti07: On the bodyStr.includes("etag") check 
1. Is it possible that this could cause false positive incase a user uploads the word "etag"
2. If reqOpts.data is a object, bodyStr defaults to '', so it is possible that we would miss an existing etag entirely.

handwritten/storage/src/storage-transport.ts
```diff
@@ -124,76 +129,138 @@ export class StorageTransport {
...
+    const isAbsolute = this.#isValidUrl(urlString);
+
+    // Determine the base URL for the request
+    const requestUrl = isAbsolute
+      ? urlString
+      : new URL(urlString, this.baseUrl).toString();
+
+    const bodyStr = typeof reqOpts.body === 'string'
+      ? reqOpts.body
+      : (typeof reqOpts.data === 'string' ? reqOpts.data : '');
```
Dhriti07: these changes are present both in this pr and PR #8235. Since these changes look retry related are they required in this pr ?

handwritten/storage/src/storage-transport.ts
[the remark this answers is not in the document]
thiyaguk09: Simply returning `data` broke resumable uploads, which rely on `resp.headers.location` when the API returns an empty string. I've updated the logic to handle all three scenarios correctly:
* Plain objects: Mutate and return `data`.
* Buffers/Streams (non-plain objects): Return raw `data` to protect binary downloads.
* Primitives (empty strings/null): Return the full `resp` wrapper so session URIs can still be read from the headers.

handwritten/storage/src/storage-transport.ts
thiyaguk09: The `.includes('"etag"')` check was brittle. I've updated this to properly inspect the payload as an object (safely parsing it first if it's a stringified JSON) to check for an actual `etag` key. This eliminates both false positives on raw text and false negatives on object payloads.

And as per the comment below, this refactor was removed here and added to PR [#8235](https://github.com/googleapis/google-cloud-node/pull/8235).

handwritten/storage/src/storage-transport.ts
thiyaguk09: You're completely right. I added these temporarily during refactoring to keep the conformance tests green, but missed removing them before updating this PR. I have removed them now!

handwritten/storage/src/storage-transport.ts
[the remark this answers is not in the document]
Dhriti07: Do we have any presubmits or checks to prevent such kind of breakages ?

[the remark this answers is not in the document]
thiyaguk09: Yes, our conformance test suite acts as the safety net for this (referenced in draft PR [#7917](https://github.com/googleapis/google-cloud-node/pull/7917)). In fact, it was the conformance tests that caught this exact breakage. When I tested changing the fallback to just `return data`, the `bucketUploadResumable` test scenarios immediately failed because they couldn't extract the session URI from the headers. These test suites run during presubmit to prevent regressions like this from merging.

handwritten/storage/src/storage-transport.ts
Dhriti07: There are still some changes regarding refactoring of makeRequest, hasprecondition flag and Retry logic which overlap between the two prs. I think they belong to https://github.com/googleapis/google-cloud-node/pull/8235.

thiyaguk09: Yes, I agree that there were overlaps between the two PRs, which is a common issue with parallel development. We chose the changes from #8235 as the base code for this bug fix because it represents the working version across all factors. To eliminate the overlap, I have removed all retry-related changes from this PR. Thanks for the suggestion.

Dhriti07: I still see some structural overlaps between the two PRs. To keep the review process efficient, could we resolve the remaining overlap by following this approach:
1. Let's keep the changes in this PR limited to the Invocation ID fix, in line with the description and title. The broader architectural refactors of the transport layer are better suited for a different PR.
2. In case the changes in PR #8235 are required as a base for this fix, let's have this PR (#8190) be dependent on #8235 for review (We can change the branch it is diffed over). We can then merge them into the feature branch in the correct order to ensure both sets of changes are present without duplicating the diffs.

thiyaguk09: It feels complicated because both branches are within my fork. I'm unable to merge the invocation fix into the retry feature branch, and subsequently, merge that feature branch into the staging branch. I will move this PR #8190 to draft for now because it depends on that retry PR. Once the retry PR #8235 is merged, I will enable this again for review.

thiyaguk09:
Moved to draft; its required PR is #8235 to merge.

handwritten/storage/test/file.ts
```diff
@@ -5219,6 +5238,22 @@ describe('File', () => {
         });
         assert.strictEqual(file.storage.retryOptions.autoRetry, true);
       });
+
+      it('should pass the invocationId to the resumable upload configuration', done => {
```
Dhriti07: In file.ts, we pass invocationId to resumableUpload function, but it looks like resumable-upload.ts wasn't updated to read it. It might be using own internal UUIDs for the header and the passed ID might be silently ignored. Is this intentional ? The unit test passes because it stubs the upload method. Could you check if resumable-upload.ts needs updating to support this?

handwritten/storage/test/bucket.ts
```diff
@@ -2880,6 +2880,55 @@ describe('Bucket', () => {
...
+        const options = {
+          destination: fakeFile,
+          resumable: false,
+          preconditionOpts: {ifGenerationMatch: 123},
+        };
+        let retryCount = 0;
+        let firstInvocationId: string | undefined;
+
+        bucket.storage.retryOptions.autoRetry = true;
+        bucket.storage.retryOptions.maxRetries = 2;
```
Dhriti07, line 2911: bucket.storage.retryOptions.maxRetries and other properties are mutated directly. Because bucket.storage references the globally shared storage mock for the test suite, this could permanently alter the default retry options for any tests that run after it. Can you check ?

handwritten/storage/test/file.ts
```diff
@@ -4583,26 +4583,29 @@ describe('File', () => {
...
+      await file.save(DATA, options, assert.ifError);
     });
 
-    it('should not require options', done => {
+    it('should not require options', async () => {
       sandbox.stub(file, 'createWriteStream').callsFake(options_ => {
-        assert.deepStrictEqual(options_, {});
-        setImmediate(done);
-        return new PassThrough();
+        assert.ok(options_?.invocationId);
```
Dhriti07: To keep the test strictly verifying that no other unexpected properties leak into options_, could we do something like this:
const { invocationId, ...rest } = options_ as any;
assert.ok(invocationId);
assert.deepStrictEqual(rest, {});

thiyaguk09: changes were updated.

handwritten/storage/test/bucket.ts
thiyaguk09: Stored a copy of `originalRetryOptions` in the before block and added an afterEach hook to fully restore `STORAGE.retryOptions`, eliminating the risk of shared state mutation.

handwritten/storage/test/file.ts
thiyaguk09: Updated the UploadConfig interface and Upload class in resumable-upload.ts to accept and persist invocationId. It now bypasses field rotation if an external ID is provided. Added dedicated unit tests in resumable-upload.ts to guarantee it persists across network attempts.

handwritten/storage/src/bucket.ts
```diff
@@ -4521,7 +4523,10 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
             ) {
               newFile.storage.retryOptions.autoRetry = false;
             }
-            const writable = newFile.createWriteStream(options);
+            const writable = newFile.createWriteStream({
+              ...options,
+              invocationId: persistentInvocationId,
```
Dhriti07, line 4529: Would this unconditionally overwrite invocationId even if a user provided their own in options. Is this  intended to be strictly internal or does the user have this option ?

thiyaguk09: Since it's strictly for internal metrics, it shouldn't be user-configurable anyway. If we make this change, a user could accidentally or intentionally overwrite it, which might mess with our metrics tracking. I'll leave it as-is so it unconditionally uses the persistentInvocationId.


*** Invocation will be dropped soon once the `x-goog-gcs-idempotency-token` is added to the header.

westarle: Hi -- I'm new to this area, trying to learn about how we work in Node.

If this param is temporary and users can't set it correctly, does it need to be in the public API?

thiyaguk09: Added `/** @internal */` to keep it hidden from public docs/types!

handwritten/storage/src/storage-transport.ts
```diff
@@ -254,7 +255,7 @@ export class StorageTransport {
   }
 
   #prepareHeaders(reqOpts: StorageRequestOptions): Record<string, string> {
-    const headersObj = this.#buildRequestHeaders(reqOpts.headers);
+    const headersObj = this.#buildRequestHeaders(reqOpts);
```
Dhriti07: Nit: Would it be cleaner to pass invocationId as an explicit parameter rather than passing the entire reqOpts object into #buildRequestHeaders?

e.g. this.#buildRequestHeaders(reqOpts.headers, reqOpts.invocationId);

thiyaguk09: Somehow I missed it; thanks, it’s updated now.

Dhriti07 (approved):

PR #8192, opened 2026-05-06
test: run the bigtable system tests in google cloud build

danieljbruce:
## Description

This pull request adds a yaml file to instruct the Bigtable system test CI check to work with the new Cloud Build trigger thereby making the new CI check effectively run our system tests. It also includes minor code modifications to the Bigtable handwritten layer that allow the system tests to operate properly in the new Cloud Build Environment.

## Impact

Leverages the strengths of running system tests in GCB rather than relying on kokoro for system tests.

## Testing

This pull request tells the tests how to work with the new check to the continuous integration pipeline for the Bigtable system tests thus improves the effectiveness of that test.

image

## Next Steps

1. Merge a pull request that removes the system tests from kokoro. This should be done in a separate pull request to separate the concerns of removing the kokoro system tests from adding the Google Cloud Build system tests.
2. Distribute a document detailing the process for migrating system tests from kokoro to GCB for all handwritten libraries.


danieljbruce:
> The current multi-step configuration has several issues that will likely cause the build to fail or behave incorrectly:
> 
> 1. **Environment Variable Persistence**: Each Cloud Build step runs in a separate container. Environment variables exported by `.kokoro/pre-system-test.sh` in Step 2 will not be available to Step 3.

This won't be an issue since the pre-system-test is just an `npm run compile`.

> 2. **Working Directory**: Cloud Build's default working directory is the repository root. Since this project is located in `handwritten/bigtable`, you must specify `dir: 'handwritten/bigtable'` for the steps to find `package.json` and the `.kokoro` directory. The current `dir: '.'` refers to the repo root, not the directory containing the `cloudbuild.yaml` file.

I think `dir: 'handwritten/bigtable'` is specified now.

> 3. **Execution Order**: The original Kokoro script ran the pre-test hook _before_ `npm install`. This is important if the hook configures registry credentials or environment-specific settings needed for installation.

The presystem-test script, which runs `npm run compile` is a no-op because the `npm install` script compiles the typescript anyways. I think the original script did things in the wrong order since it doesn't make sense to compile if the modules are not installed yet.

> 4. **Efficiency**: `npm install -g npm@latest` and global prefix configuration are generally unnecessary in the Cloud Build environment and add overhead.

This is a good point. Not so much for efficiency as I don't think that will matter much, but we should pin the node version to 18 to get consistent results.

handwritten/bigtable/system-test/install.ts
```diff
@@ -41,9 +41,13 @@ describe('📦 pack-n-play test', () => {
       packageDir: process.cwd(),
       sample: {
         description: 'JavaScript user can use the library',
-        ts: readFileSync(
-          './system-test/fixtures/sample/src/index.js',
-        ).toString(),
+        // Inject the reference directive at the top of the code
+        // The reference header is needed so packnplay tests recognize require
+        ts:
```
pearigee: I think you can fix this just by using the cjs key instead of ts. CJS in this case being CommonJS, which includes the require type by default.

```
cjs: readFileSync(
          './system-test/fixtures/sample/src/index.js',
        ).toString(),
```

handwritten/bigtable/cloudbuild.yaml
```diff
@@ -0,0 +1,96 @@
...
+    fi
+  dir: 'handwritten/bigtable'
+
+# 1. Set up Node.js environment
+- name: 'node:18'
+  entrypoint: 'bash'
+  args:
+  - '-c'
+  - |
+    if [ -f /workspace/skip_tests ]; then exit 0; fi
```
pearigee: Can we not just terminate the workflow early? Do we have to guard every step? I am not sure how to do this, but I suspect this has to be a built in feature?

handwritten/bigtable/cloudbuild.yaml
```diff
@@ -0,0 +1,96 @@
...
+    if git diff --quiet origin/main... -- . ; then
+      echo "No changes in handwritten/bigtable. Skipping tests."
+      echo "skip" > /workspace/skip_tests
+    else
+      echo "Changes detected in handwritten/bigtable. Proceeding with tests."
+    fi
+  dir: 'handwritten/bigtable'
+
+# 1. Set up Node.js environment
+- name: 'node:18'
```
pearigee: Can we just start on Node 24 instead? We should avoid introducing tests that still use an EOL node version. It will just create more work for the version upgrade project.

handwritten/bigtable/system-test/install.ts
danieljbruce: @gemini-code-assist implement this suggestion

handwritten/bigtable/cloudbuild.yaml
danieljbruce: @gemini-code-assist implement this suggestion

handwritten/bigtable/cloudbuild.yaml
danieljbruce: @gemini-code-assist implement this suggestion

handwritten/bigtable/system-test/install.ts
danieljbruce: @gemini-code-assist I think the intent of the suggestion is to avoid using the `reference types` header. So update the suggestion to eliminate this header and adjust the comments accordingly.

handwritten/bigtable/cloudbuild.yaml
danieljbruce: @gemini-code-assist we need to apply a similar change to all node:18 instances in this file. Update the suggestion

handwritten/bigtable/cloudbuild.yaml
danieljbruce: I think I'll go with gemini's suggestion on this one. I actually tried to remove this step a day ago, but gemini cli combined all these configs into one config which I didn't want so I abandoned this change. I guess it did it because the configs don't support an early exit.

handwritten/bigtable/cloudbuild.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+# 1. Set up Node.js environment
+- name: 'node:24'
+  entrypoint: 'bash'
+  args:
+  - '-c'
+  - |
+    if [ -f /workspace/skip_tests ]; then exit 0; fi
+    npm install
+    npm run presystem-test
```
pearigee: This may get executed twice. I think NPM will automatically execute the pre* commands on run? Can you verify?

handwritten/bigtable/cloudbuild.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - |
+    git fetch origin main --quiet
+    # Compare with origin/main to detect changes.
+    if git diff --quiet origin/main... -- . ; then
+      echo "No changes in handwritten/bigtable. Skipping tests."
+      echo "skip" > /workspace/skip_tests
+    else
+      echo "Changes detected in handwritten/bigtable. Proceeding with tests."
+    fi
+  dir: 'handwritten/bigtable'
```
pearigee: Do we need this step at all since GCB can configure when to run based on change detection? When this config enabled it should only run if there are changes in this directory in the first place, right?

Image

handwritten/bigtable/src/tabular-api-surface.ts
pearigee, line 1: Is this modifying a generated asset? Does this proto.js get made by the generator? If so, does it risk getting overwritten in the future? Would the library break if this happened?

danieljbruce: No. tabular-api-surface.ts is entirely handwritten.

handwritten/bigtable/cloudbuild.yaml
danieljbruce: That's a good point. In which case, I think we can also remove waitFor: ['check-changes'] which was added earlier because of this step

handwritten/bigtable/cloudbuild.yaml
danieljbruce: Actually we can remove this now that we removed the first step and we are not writing to workspace/skip_tests anymore.

handwritten/bigtable/cloudbuild.yaml
danieljbruce: Yes. That's true. It does so I think we can remove this.

handwritten/bigtable/src/tabular-api-surface.ts
pearigee: Oh sorry, I meant "protos.js", not this file.

danieljbruce: protos.js diff has been eliminated by merging with main.

PR #8263, opened 2026-05-13
chore: relocate three packages from packages, as they are handwritten

feywind:
As the title says.

https://github.com/googleapis/librarian/issues/4803

- gapic-processing-node -> core
- typeless-sample-bot -> core
- google-cloud-dns -> handwritten

Fixes: https://github.com/googleapis/google-cloud-node/issues/8122 🦖


feywind:
Since this is just relocating files, the header check failure is expected.

pearigee:
Is the `release-please-config.json` correct? I see that typeless-sample-bot and gapic-node-processing may be referencing the wrong directory? google-cloud-dns is referencing handwritten now in `release-please-config.json`. 

However, we also have a `release-please-submodules.json` which has the rest of our core/handwritten libraries. Should all of these libraries move there instead?

feywind:
tysm for the catches! I'm still not completely familiar with all the repo edges.

PR #8287, opened 2026-05-15
[Performance] Debugging issue with node

surbhigarg92:


handwritten/spanner/src/index.ts
```diff
@@ -1691,13 +1691,10 @@ class Spanner extends GrpcService {
...
-      let reqOpts = extend(true, {}, config.reqOpts);
-      reqOpts = replaceProjectIdToken(reqOpts, projectId!);
-      // It would have been preferable to replace the projectId already in the
-      // constructor of Spanner, but that is not possible as auth.getProjectId
-      // is an async method. This is therefore the first place where we have
-      // access to the value that should be used instead of the placeholder.
+      let reqOpts;
       if (!this.projectIdReplaced_) {
+        reqOpts = extend(true, {}, config.reqOpts);
+        reqOpts = replaceProjectIdToken(reqOpts, projectId!);
```
olavloite, line 1701: Isn't this 'undoing' the change further up? As in: Isn't this method called for every gRPC invocation, which means that this method will still be called for each request?

handwritten/spanner/src/index.ts
```diff
@@ -1715,12 +1712,14 @@ class Spanner extends GrpcService {
             );
           });
         });
+        config.headers[CLOUD_RESOURCE_HEADER] = replaceProjectIdToken(
```
olavloite: And here also? (Or is this method only called once at startup?)

handwritten/spanner/src/index.ts
surbhigarg92: `prepareGapicRequest_` is called for each request. For the first request the `projectIdReplaced_` will be replaced and set to true . And then database.formattedName_ and instance.formattedName_ will replace {{projectId}} to the actual project ID. 

On subsequent requests the reqOpts will be constructed from database.formattedName_ and instance.formattedName_. which will have  replaced projectID.

Having said that, I will do more testing to confirm this behavior, specially for parallel calls.

PR #8301, opened 2026-05-18
feat: port GDCH credentials support to Node.js Auth SDK

### LINKED ISSUE #8289: auth: Add support for Google Distributed Cloud service identity authentication
### Library Name

google-auth-library

N/A

This issue tracks the work required to implement Google Distributed Cloud (GDC) service identity authentication support in google-auth-library. This will allow the SDK to function in private or disconnected cloud environments by supporting the gdch_service_account credential type, utilizing self-signed JWTs (ES256) exchanged via a Secure Token Service (STS) endpoint.

See [go/node-js-gdch](https://docs.google.com/document/d/1wcIqmuAJDiS9A7JVtbrZRzbBhtKmGZDuyNp0Bens-nM/edit?tab=t.0) for more details.

_No response_

### Additional context/notes

_No response_

macastelaz:
Add support for GDCH Credentials to the Node.js Auth SDK

Manually tested - details in [go/node-js-gdch-manual-test](https://docs.google.com/document/d/1zVTpsVEJ6nunY3fP3Fyc87vBfUNBICYWukDc5Du2myE/edit?tab=t.0)

Fixes #8289

### DIFF
--- a/core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts
+++ b/core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts
@@ -209,14 +209,14 @@ export class GdchClient extends OAuth2Client {
       if (!tokenResponse.access_token) {
         throw new Error('Token response did not contain an access_token.');
       }
+      if (!tokenResponse.expires_in) {
+        throw new Error('Token response did not contain an expires_in field.');
+      }
       const tokens: Credentials = {
         access_token: tokenResponse.access_token,
         token_type: 'STS-Bearer',
+        expiry_date: Date.now() + tokenResponse.expires_in * 1000,
       };
-
-      if (tokenResponse.expires_in) {
-        tokens.expiry_date = Date.now() + tokenResponse.expires_in * 1000;
-      }
 
       this.emit('tokens', tokens);
       return {res, tokens};
@@ -335,6 +335,28 @@ export class GdchClient extends OAuth2Client {
     return {
       ...this,
       privateKey: this.privateKey ? '***REDACTED***' : undefined,
+      _clientSecret: this._clientSecret ? '***REDACTED***' : undefined,
+      apiKey: this.apiKey ? '***REDACTED***' : undefined,
+      gdchOptions: this.gdchOptions
+        ? {
+            ...this.gdchOptions,
+            privateKey: this.gdchOptions.privateKey ? '***REDACTED***' : undefined,
+            clientSecret: this.gdchOptions.clientSecret ? '***REDACTED***' : undefined,
+            client_secret: this.gdchOptions.client_secret ? '***REDACTED***' : undefined,
+            apiKey: this.gdchOptions.apiKey ? '***REDACTED***' : undefined,
+            credentials: this.gdchOptions.credentials
+              ? {
+                  ...this.gdchOptions.credentials,
+                  access_token: this.gdchOptions.credentials.access_token
+                    ? '***REDACTED***'
+                    : undefined,
+                  refresh_token: this.gdchOptions.credentials.refresh_token
+                    ? '***REDACTED***'
+                    : undefined,
+                }
+              : undefined,
+          }
+        : undefined,
       credentials: {
         ...this.credentials,
         access_token: this.credentials?.access_token ? '***REDACTED***' : undefined,
--- a/core/packages/google-auth-library-nodejs/test/test.gdchclient.ts
+++ b/core/packages/google-auth-library-nodejs/test/test.gdchclient.ts
@@ -392,6 +392,7 @@ describe('GdchClient', () => {
       .post('/token')
       .reply(200, {
         access_token: 'ca-verified-token',
+        expires_in: 3600,
       });
 
     const res = await client.getAccessToken();
@@ -572,6 +573,7 @@ describe('GdchClient', () => {
       .post('/token')
       .reply(200, {
         access_token: 'ca-verified-token',
+        expires_in: 3600,
       });
 
     await assert.rejects(client.getAccessToken(), (err: Error) => {
@@ -601,6 +603,29 @@ describe('GdchClient', () => {
 
     await assert.rejects(client.getAccessToken(), (err: Error) => {
       assert.ok(err.message.includes('Token response did not contain an access_token.'));
+      return true;
+    });
+    scope.done();
+  });
+
+  it('should throw error if token response does not contain expires_in', async () => {
+    const client = new GdchClient({
+      projectId: 'test-project',
+      privateKeyId: 'key-id-123',
+      privateKey: privateKeyPemSec1,
+      serviceIdentityName: 'sa-name',
+      tokenServerUri: 'https://token-server.local/token',
+      apiAudience: 'target-audience',
+    });
+
+    const scope = nock('https://token-server.local')
+      .post('/token')
+      .reply(200, {
+        access_token: 'exchange-token-abc123',
+      });
+
+    await assert.rejects(client.getAccessToken(), (err: Error) => {
+      assert.ok(err.message.includes('Token response did not contain an expires_in field.'));
       return true;
     });
     scope.done();
@@ -737,13 +762,16 @@ describe('GdchClient', () => {
   });
 
   describe('serialization and logging safety', () => {
-    it('should redact private key and credentials in toJSON() serialization', () => {
+    it('should redact private key, clientSecret, apiKey, and credentials in toJSON() serialization', () => {
       const client = new GdchClient({
         projectId: 'test-project',
         privateKeyId: 'key-id-123',
         privateKey: 'raw-secret-private-key',
         serviceIdentityName: 'sa-name',
+        clientSecret: 'raw-client-secret',
+        apiKey: 'raw-api-key',
       });
+      client._clientSecret = 'raw-internal-client-secret';
 
       client.credentials = {
         access_token: 'secret-access-token-abc123',
@@ -755,17 +783,25 @@ describe('GdchClient', () => {
       assert.strictEqual(serialized.projectId, 'test-project');
       assert.strictEqual(serialized.privateKeyId, 'key-id-123');
       assert.strictEqual(serialized.privateKey, '***REDACTED***');
+      assert.strictEqual(serialized._clientSecret, '***REDACTED***');
+      assert.strictEqual(serialized.apiKey, '***REDACTED***');
+      assert.strictEqual(serialized.gdchOptions.privateKey, '***REDACTED***');
+      assert.strictEqual(serialized.gdchOptions.clientSecret, '***REDACTED***');
+      assert.strictEqual(serialized.gdchOptions.apiKey, '***REDACTED***');
       assert.strictEqual(serialized.credentials.access_token, '***REDACTED***');
       assert.strictEqual(serialized.credentials.refresh_token, '***REDACTED***');
     });
 
-    it('should redact private key and credentials in custom inspect console output', () => {
+    it('should redact private key, clientSecret, apiKey, and credentials in custom inspect console output', () => {
       const client = new GdchClient({
         projectId: 'test-project',
         privateKeyId: 'key-id-123',
         privateKey: 'raw-secret-private-key',
         serviceIdentityName: 'sa-name',
+        clientSecret: 'raw-client-secret',
+        apiKey: 'raw-api-key',
       });
+      client._clientSecret = 'raw-internal-client-secret';
 
       client.credentials = {
         access_token: 'secret-access-token-abc123',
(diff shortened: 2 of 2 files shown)


core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts
```diff
@@ -0,0 +1,306 @@
...
+      }
+    })();
+
+    return this.caAgentPromise;
+  }
+
+  private base64UrlEncode(str: string | Buffer): string {
+    const buffer = typeof str === 'string' ? Buffer.from(str) : str;
+    return buffer
+      .toString('base64')
```
jerrylin3321: nit: could you just do `buffer.toString('base64url')`?

macastelaz: Good catch - thanks for flagging! Fixed.

core/packages/google-auth-library-nodejs/test/test.gdchclient.ts
```diff
@@ -0,0 +1,507 @@
...
+  });
+
+  it('createWithGdchAudience() should throw error if audience is empty', () => {
+    const client = new GdchClient();
+    assert.throws(() => {
+      client.createWithGdchAudience('');
+    }, /Audience cannot be null or empty/);
+  });
+
+  it('should perform token exchange successfully with valid assertion signature (SEC1 key)', async () => {
```
pearigee: Optional -- feel free to ack and ignore.

This test has a lot of assertions, making it a bit difficult to parse exactly what cluster of them is meant to verify.

Would it be possible to break this up into smaller more targeted tests? For example:  "client requests token correctly", "signature has correct properties", "signature can be verified with public key"?

macastelaz: Thanks for flagging! Splitting makes a lot of sense to me! Updated

core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts
```diff
@@ -0,0 +1,302 @@
...
+  format_version: string;
+  project: string;
+  private_key_id: string;
+  private_key: string;
+  name: string;
+  token_uri: string;
+  ca_cert_path?: string;
+}
+
+export class GdchClient extends OAuth2Client {
```
lsirac, line 51: Storing `privateKey` as a plain public string field on the `GdchClient` class exposes it to accidental log dumping during serialization (e.g., `JSON.stringify(client)`) or debug inspection (e.g., `console.log(client)`).

Implement `toJSON()` and custom inspect methods to ensure the private key and active tokens are automatically redacted.

core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts
```diff
@@ -0,0 +1,302 @@
...
+
+      this.emit('tokens', tokens);
+      return {res, tokens};
+    } catch (e) {
+      if (e instanceof Error) {
+        e.message = `Error getting access token for GDCH service account: ${e.message}, iss: ${this.serviceIdentityName}`;
+      }
+      throw e;
+    }
+  }
```
lsirac, line 242: If the token exchange fails, `gaxios` throws a `GaxiosError` containing the complete request config in `err.config.data` (which includes the raw signed JWT `subject_token`). This can result in active token leakage if applications log the exception object.
**Recommended Fix:**
Wrap the `requestAsync` call in a `try/catch` and scrub the sensitive assertion payload:
```typescript
try {
  const res = await this.requestAsync<GetTokenResponse>(requestOpts);
  // ... handle token success ...
} catch (err: any) {
  if (err && err.config && err.config.data) {
    try {
      const data = JSON.parse(err.config.data);
      if (data.subject_token) {
        data.subject_token = '***REDACTED***';
        err.config.data = JSON.stringify(data);
      }
    } catch {}
  }
  throw err;
}

File: core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts [hunk 5]
```diff
@@ -0,0 +1,302 @@
...
+        }
+        if (err instanceof Error) {
+          err.message = `Error reading certificate file from CA cert path, value '${currentPath}': ${err.message}`;
+        }
+        throw err;
+      }
+    })();
+
+    return this.caAgentPromise;
+  }
```
lsirac, line 332: Currently, `getCaAgent` caches the certificate promise permanently. In production Kubernetes clusters hosting GDCH workloads, custom certificates are rotated regularly on disk. Because the cache key is solely the string path of the file, the client will continue using the expired cached agent indefinitely until the process is restarted.

**Recommended Fix:**
Introduce a TTL check (e.g. 5 minutes) or check file modification time using `fs.promises.stat` to dynamically reload the cert:
```typescript
private lastCaCertReadTime = 0;
private readonly CA_CERT_TTL_MS = 5 * 60 * 1000;
private async getCaAgent(): Promise<https.Agent | undefined> {
  if (!this.caCertPath) return undefined;
  
  const now = Date.now();
  const isCacheExpired = now - this.lastCaCertReadTime > this.CA_CERT_TTL_MS;
  
  if (this.caAgentPromise && this.caCertPath === this.cachedCaCertPath && !isCacheExpired) {
    return this.caAgentPromise;
  }
  
  this.cachedCaCertPath = this.caCertPath;
  this.lastCaCertReadTime = now;
  // ... proceed to read file and build agent ...
}

File: core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts [hunk 6]
```diff
@@ -0,0 +1,302 @@
...
+
+  override async requestAsync<T>(
+    opts: GaxiosOptions,
+    retry = false
+  ): Promise<GaxiosResponse<T>> {
+    if (this.caCertPath && !opts.agent) {
+      opts.agent = await this.getCaAgent();
+    }
+    return super.requestAsync(opts, retry);
+  }
```
lsirac, line 290: Overriding `requestAsync` on `OAuth2Client` injects the private CA agent into *every* request sent through the client. If a developer uses this client instance to perform operations against standard public Google Cloud services (e.g., Google Cloud Storage on `googleapis.com`), those requests will be forced to use the private CA agent, which lacks public root credentials, leading to TLS handshaking failures.
**Recommended Fix:**
Instead of overriding `requestAsync` globally on the client class, apply the custom CA agent specifically to the Token Exchange endpoint request inside `refreshTokenNoCache()`, keeping `requestAsync` clean:
```typescript
// Inside refreshTokenNoCache()
const requestOpts: GaxiosOptions = {
  url: this.tokenServerUri,
  method: 'POST',
  headers: { 'Content-Type': 'application/json' },
  data,
  responseType: 'json',
};
if (this.caCertPath) {
  requestOpts.agent = await this.getCaAgent();
}

File: core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts [hunk 7]
```diff
@@ -0,0 +1,302 @@
...
+
+    const requestOpts: GaxiosOptions = {
+      url: this.tokenServerUri,
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      data,
+      responseType: 'json',
+    };
```
lsirac, line 200: The POST request to `this.tokenServerUri` is initiated without any timeout or retry logic. If the cluster STS endpoint becomes unresponsive, the request will hang indefinitely, leaking sockets.
**Recommended Fix:**
Configure a robust timeout (e.g. 10 seconds) and retry configuration to withstand transient gateway hiccups (502/503/504):
```typescript
timeout: 10000,
retry: true,
retryConfig: {
  httpMethodsToRetry: ['POST'],
  statusCodesToRetry: [[500, 599]],
  noResponseRetries: 3,
}

File: core/packages/google-auth-library-nodejs/src/index.ts
Comment (lsirac) [inline core/packages/google-auth-library-nodejs/src/index.ts:1]: `GoogleAuthOptions` is missing `GdchCredentialsInput` in its `credentials` union. This prevents developers from manually passing loaded GDCH config objects to `GoogleAuth` under strict compilation. Additionally, `GDCH_SERVICE_ACCOUNT_TYPE` and `GdchCredentialsInput` are not exported at the package root.
**Recommended Fix:**
Widen `GoogleAuthOptions.credentials`, `jsonContent`, and `_cacheClientFromJSON` to accept `GdchCredentialsInput`, and expose these types at the package root in `src/index.ts`.

File: core/packages/google-auth-library-nodejs/src/auth/googleauth.ts [hunk 8]
```diff
@@ -789,11 +795,14 @@ export class GoogleAuth<T extends AuthClient = AuthClient> {
         ...json,
         ...options,
       } as ExternalAccountAuthorizedUserClientOptions);
+    } else if (json.type === GDCH_SERVICE_ACCOUNT_TYPE) {
```
lsirac, line 798: GDCH clients require an explicit `apiAudience` to retrieve tokens. Because downstream GAPIC libraries (such as `google-cloud-gdchardwaremanagement`) instantiate credentials dynamically via Application Default Credentials (ADC) loader pipelines, they have no awareness of `GdchClient` and will crash on launch because the audience is null.
**Recommended Fix:**
Implement automatic audience resolution in `GoogleAuth.getClient()` when a `GdchClient` is detected:
```typescript
const client = await this.getClient();
if (client instanceof GdchClient && !client.apiAudience) {
  const endpoint = this.clientOptions.apiEndpoint || this.clientOptions.servicePath;
  if (endpoint) {
    const scheme = endpoint.startsWith('http') ? '' : 'https://';
    const formattedAudience = `${scheme}${endpoint}/`.replace(/\/+$/, '/');
    return client.createWithGdchAudience(formattedAudience);
  }
}

File: core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts (hunk 7 above)
Comment (macastelaz) [reply core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts:200]: Done

File: core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts (hunk 3 above)
Comment (macastelaz) [reply core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts:51]: Done - looks like this may have been missed in other clients historically but best to implement it now and for future clients.

File: core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts (hunk 4 above)
Comment (macastelaz) [reply core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts:242]: Done

File: core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts (hunk 6 above)
Comment (macastelaz) [reply core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts:290]: I don't think we can do this exactly as this proposes (if we do, the manual test script described in [go/node-js-gdch-manual-test](https://docs.google.com/document/d/1zVTpsVEJ6nunY3fP3Fyc87vBfUNBICYWukDc5Du2myE/edit?tab=t.0) fails because it makes a call to a private endpoint which *does* require the private CA agent. The update I've made here is to look at the url for either googleapis.com or google.com. If neither are true, then we use the private CA agent assuming it's a private request else we leave the opts.agent alone and use the public one.

File: core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts (hunk 5 above)
Comment (macastelaz) [reply core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts:332]: Added the 5 minute ttl to cache

File: core/packages/google-auth-library-nodejs/src/auth/googleauth.ts (hunk 8 above)
Comment (macastelaz) [reply core/packages/google-auth-library-nodejs/src/auth/googleauth.ts:798]: Added audience retrieval to the getClient in the case of a Gdchclient without an audience.

File: core/packages/google-auth-library-nodejs/src/index.ts
Comment (macastelaz) [reply core/packages/google-auth-library-nodejs/src/index.ts:1]: Exported. Thanks for flagging!

File: core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts [hunk 9]
```diff
@@ -0,0 +1,354 @@
...
+        throw err;
+      }
+    })();
+
+    return this.caAgentPromise;
+  }
+
+  toJSON(): Record<string, any> {
+    return {
+      ...this,
```
lsirac, line 336: I think this still leaks the private key - only top level is redacted

core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts
```diff
@@ -0,0 +1,354 @@
...
+      const tokenResponse = res.data;
+      if (!tokenResponse.access_token) {
+        throw new Error('Token response did not contain an access_token.');
+      }
+      const tokens: Credentials = {
+        access_token: tokenResponse.access_token,
+        token_type: 'STS-Bearer',
+      };
+
+      if (tokenResponse.expires_in) {
```
lsirac: Should this be required? If not, Oauth2Client.isTokenExpiring() will treat this as not expiring.

macastelaz: Done

core/packages/google-auth-library-nodejs/src/auth/gdchclient.ts
macastelaz: Done

PR #8329, opened 2026-05-20
feat(spanner): optimize RequestId propagation and minimize OpenTelemetry active tracing overhead

surbhigarg92:
Optimize RequestId propagation and minimize OpenTelemetry active tracing overhead.

### DIFF
--- a/handwritten/spanner/.OwlBot.yaml
+++ b/handwritten/spanner/.OwlBot.yaml
@@ -1,10 +1,10 @@
-# Copyright 2021 Google LLC
+# Copyright 2025 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     http://www.apache.org/licenses/LICENSE-2.0
+#      http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
@@ -12,19 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-deep-remove-regex:
-  - /owl-bot-staging
-
 deep-copy-regex:
-  - source: /google/spanner/(v.*)/.*-nodejs
-    dest: /owl-bot-staging/spanner/$1
-  - source: /google/spanner/(admin/database/v.*)/.*-nodejs
-    dest: /owl-bot-staging/spanner/$1
-  - source: /google/spanner/(admin/instance/v.*)/.*-nodejs
-    dest: /owl-bot-staging/spanner/$1
-  - source: /google/spanner/(executor/v.*)/.*-nodejs
-    dest: /owl-bot-staging/spanner/$1
+    - source: /google/spanner/executor/google-spanner-executor-nodejs
+      dest: /owl-bot-staging/google-spanner-executor
 
-begin-after-commit-hash: 46f25fb1121747b994ff5818963fda84b5e6bfd3
-
+api-name: executor
--- a/handwritten/spanner/protos/google/spanner/v1/commit_response.proto
+++ b/handwritten/spanner/protos/google/spanner/v1/commit_response.proto
@@ -16,7 +16,9 @@ syntax = "proto3";
 
 package google.spanner.v1;
 
+import "google/api/field_behavior.proto";
 import "google/protobuf/timestamp.proto";
+import "google/spanner/v1/location.proto";
 import "google/spanner/v1/transaction.proto";
 
 option csharp_namespace = "Google.Cloud.Spanner.V1";
@@ -61,4 +63,18 @@ message CommitResponse {
   // timestamp at which all reads in the transaction ran. This timestamp is
   // never returned.
   google.protobuf.Timestamp snapshot_timestamp = 5;
+
+  // Optional. A cache update expresses a set of changes the client should
+  // incorporate into its location cache. The client should discard the changes
+  // if they are older than the data it already has. This data can be obtained
+  // in response to requests that included a `RoutingHint` field, but may also
+  // be obtained by explicit location-fetching RPCs which may be added in the
+  // future.
+  CacheUpdate cache_update = 6 [(google.api.field_behavior) = OPTIONAL];
+
+  // The isolation level used for the read-write transaction.
+  TransactionOptions.IsolationLevel isolation_level = 7;
+
+  // The read lock mode used for the read-write transaction.
+  TransactionOptions.ReadWrite.ReadLockMode read_lock_mode = 8;
 }
--- a/handwritten/spanner/protos/google/spanner/v1/spanner.proto
+++ b/handwritten/spanner/protos/google/spanner/v1/spanner.proto
@@ -331,6 +331,23 @@ service Spanner {
     };
     option (google.api.method_signature) = "session,mutation_groups";
   }
+
+  // Retrieves a cache update for a given database.
+  //
+  // This RPC can be used to warm up the client cache by fetching key recipes
+  // and server information for a given database. It is recommended to call
+  // this RPC at the beginning of the client's lifecycle, prior to any other
+  // data plane operations.
+  //
+  // The cache update is returned as a stream because the response can be too
+  // large to fit into a single `CacheUpdate` message.
+  rpc FetchCacheUpdate(FetchCacheUpdateRequest) returns (stream CacheUpdate) {
+    option (google.api.http) = {
+      post: "/v1/{database=projects/*/instances/*/databases/*}:cacheUpdate"
+      body: "*"
+    };
+    option (google.api.method_signature) = "database";
+  }
 }
 
 // The request for [CreateSession][google.spanner.v1.Spanner.CreateSession].
@@ -827,7 +844,7 @@ message ExecuteSqlRequest {
   // be assumed until a subsequent `Commit` call completes successfully.
   bool last_statement = 17 [(google.api.field_behavior) = OPTIONAL];
 
-  // Optional. If present, it makes the Spanner requests location-aware.
+  // Optional. Makes the Spanner requests location-aware if present.
   //
   // It gives the server hints that can be used to route the request
   // to an appropriate server, potentially significantly decreasing latency and
@@ -1268,7 +1285,7 @@ message ReadRequest {
   // transactions.
   LockHint lock_hint = 17 [(google.api.field_behavior) = OPTIONAL];
 
-  // Optional. If present, it makes the Spanner requests location-aware.
+  // Optional. Makes the Spanner requests location-aware if present.
   //
   // It gives the server hints that can be used to route the request
   // to an appropriate server, potentially significantly decreasing latency and
@@ -1301,6 +1318,14 @@ message BeginTransactionRequest {
   // randomly select one of the mutations from the mutation set and send it as a
   // part of this request.
   Mutation mutation_key = 4 [(google.api.field_behavior) = OPTIONAL];
+
+  // Optional. Makes the Spanner requests location-aware if present.
+  //
+  // It gives the server hints that can be used to route the request
+  // to an appropriate server, potentially significantly decreasing latency and
+  // improving throughput. To achieve improved performance, most fields must be
+  // filled in with accurate values.
+  RoutingHint routing_hint = 5 [(google.api.field_behavior) = OPTIONAL];
 }
 
 // The request for [Commit][google.spanner.v1.Spanner.Commit].
@@ -1355,6 +1380,14 @@ message CommitRequest {
   // results in a `FailedPrecondition` error.
   MultiplexedSessionPrecommitToken precommit_token = 9
       [(google.api.field_behavior) = OPTIONAL];
+
+  // Optional. Makes the Spanner requests location-aware if present.
+  //
+  // It gives the server hints that can be used to route the request
+  // to an appropriate server, potentially significantly decreasing latency and
+  // improving throughput. To achieve improved performance, most fields must be
+  // filled in with accurate values.
+  RoutingHint routing_hint = 10 [(google.api.field_behavior) = OPTIONAL];
 }
 
 // The request for [Rollback][google.spanner.v1.Spanner.Rollback].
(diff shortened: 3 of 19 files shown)


handwritten/spanner/src/instrument.ts
```diff
@@ -119,6 +119,54 @@ function ensureInitialContextManagerSet() {
...
+    const isRecording = probeSpan.isRecording();
+    probeSpan.end();
+
+    if (isRecording) {
+      tracingEnabled = true;
+      return true;
+    }
+  }
+  tracingEnabled = false;
+  return false;
```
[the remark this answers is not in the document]
surbhigarg92: If OpenTelemetry is initialized after Spanner SDK then traces wont be enabled,.

handwritten/spanner/src/index.ts
```diff
@@ -1715,20 +1715,24 @@ class Spanner extends GrpcService {
...
           });
         });
+        config.headers[CLOUD_RESOURCE_HEADER] = replaceProjectIdToken(
+          config.headers[CLOUD_RESOURCE_HEADER],
+          projectId!,
+        );
         this.projectIdReplaced_ = true;
+      } else {
+        reqOpts = config.reqOpts;
+      }
```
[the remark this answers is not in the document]
surbhigarg92: During the first request, the SDK permanently replaces the {{projectId}} token in-place inside all cached instance and database objects (instance.formattedName_, database.formattedName_). For all subsequent requests, the headers are constructed from these already-resolved properties, meaning config.headers[CLOUD_RESOURCE_HEADER] already contains the correct project ID. Running the replacement on every subsequent request is mathematically redundant and adds unnecessary regex overhead on Node's single-threaded event loop.

handwritten/spanner/src/index.ts
```diff
@@ -1692,12 +1695,8 @@ class Spanner extends GrpcService {
       }
       const gaxClient = this.clients_.get(clientName)!;
       let reqOpts = extend(true, {}, config.reqOpts);
-      reqOpts = replaceProjectIdToken(reqOpts, projectId!);
-      // It would have been preferable to replace the projectId already in the
```
olavloite, line 1696: Should we really remove this comment? Isn't it still valid?

handwritten/spanner/src/instrument.ts
```diff
@@ -119,6 +119,56 @@ function ensureInitialContextManagerSet() {
...
+  if (globalTracingEnabled !== undefined) {
+    return globalTracingEnabled;
+  }
+
+  const globalProvider = trace.getTracerProvider();
+  if (globalProvider) {
+    const probeSpan = globalProvider
+      .getTracer(TRACER_NAME, TRACER_VERSION)
+      .startSpan('probe');
+    const isRecording = probeSpan.isRecording();
```
olavloite: I don't think this is a safe assumption. It assumes that `probeSpan.isRecording()` will return true if tracing is enabled. But if tracing is enabled with a 5% sample rate, then there is no guarantee that this will return true. Or am I misunderstanding what is going on here?

handwritten/spanner/src/instrument.ts
```diff
@@ -119,6 +119,56 @@ function ensureInitialContextManagerSet() {
...
+
+/**
+ * isGlobalTracingEnabled returns true if tracing is enabled globally,
+ * respecting cached status and active recording spans.
+ *
+ * @returns {boolean} True if global tracing is enabled.
+ */
+function isGlobalTracingEnabled(): boolean {
+  if (globalTracingEnabled !== undefined) {
+    return globalTracingEnabled;
```
olavloite, line 138: This caching means that the result that is returned the first time is valid for the lifetime of the application. This means that:
1. If this function happens to be called before the application has configured OpenTelemetry, then the value that it calculates and caches can be wrong.
2. It does not take into the (maybe theoretical) possibility that an application could change its configuration later.

handwritten/spanner/src/request_id_header.ts
```diff
@@ -214,6 +184,7 @@ export {
   X_GOOG_SPANNER_REQUEST_ID_SPAN_ATTR,
   attributeXGoogSpannerRequestIdToActiveSpan,
   craftRequestId,
+  PROCESS_PREFIX,
```
olavloite: Is this really used elsewhere?

handwritten/spanner/test/spanner.ts
```diff
@@ -91,7 +91,11 @@ const {
   InMemorySpanExporter,
 } = require('@opentelemetry/sdk-trace-node');
 const {SimpleSpanProcessor} = require('@opentelemetry/sdk-trace-base');
-const {startTrace, ObservabilityOptions} = require('../src/instrument');
```
olavloite, line 94: (not related to this line, but it feels like the most logical place to add this comment)

We are not adding any new tests for this. Should we add tests that verify that:
1. It does not matter when the OpenTelemetry configuration is done (before or after creating a Spanner instance).
2. It does not matter what the trace sampling is. When tracing is enabled, even with a 1% sampling rate, then the request ID should be added to the traces.

handwritten/spanner/src/index.ts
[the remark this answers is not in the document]
olavloite: I do think that this is an actual problem, and then specifically for non-cached resource names. See this test case:


```typescript
it('real application flow: should replace project ID tokens on Backup objects after initial Spanner request', done => {
  const {Instance} = require('../src/instance');
  const {Backup} = require('../src/backup');
  // Stub setup...
  const appSpanner = new Spanner({projectId: '{{projectId}}'});
  asAny(appSpanner).auth.getProjectId = callback => {
    callback(null, PROJECT_ID);
  };
  const realInstance = new Instance(appSpanner, 'my-instance');
  appSpanner.instances_.set('my-instance', realInstance);
  // Backup created before project ID is replaced
  const realBackup = new Backup(realInstance, 'my-backup');
  // 1. Initial request triggers prepareGapicRequest_
  FAKE_GAPIC_CLIENT.getInstanceConfig = (reqOpts, gaxOpts, callback) => {
    callback(null, {});
  };
  appSpanner.getInstanceConfig('nam1', err => {
    if (err) return done(err);
    // At this point, appSpanner.projectIdReplaced_ is true.
    // 2. Application calls backup.getMetadata()
    FAKE_GAPIC_CLIENT.getBackup = reqOpts => {
      try {
        assert.strictEqual(
          reqOpts.name,
          `projects/${PROJECT_ID}/instances/my-instance/backups/my-backup`,
        );
        done();
      } catch (e) {
        done(e);
      }
      return Promise.resolve([{}]);
    };
    realBackup.getMetadata();
  });
});
```

handwritten/spanner/src/instrument.ts
```diff
@@ -245,9 +299,11 @@ export function setSpanErrorAndException(
  * @returns {Span} the non-null span.
  */
 export function getActiveOrNoopSpan(): Span {
-  const span = trace.getActiveSpan();
-  if (span) {
-    return span;
+  if (isTracingEnabled()) {
```
olavloite: Note that this could just as well call `isGlobalTracingEnabled` directly, as it does not supply any options. So the more specific check whether a tracer has been set on any options is always skipped. Is that intentional?

Also, calling `trace.getActiveSpan()` should be an extremely cheap method to call when tracing is disabled, so I am not sure this entire method really optimizes anything.

handwritten/spanner/src/index.ts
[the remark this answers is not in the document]
surbhigarg92: I missed this case, this seems a problem only for classes like Backups. 

Because any DataClient operations will be done via `instance` object which is created using Spanner instance.  For these objects we are replacing the projectId. 

```
this.instances_.forEach(instance => {
          instance.formattedName_ = replaceProjectIdToken(
            instance.formattedName_,
            projectId!,
          );
          instance.databases_.forEach(database => {
            database.formattedName_ = replaceProjectIdToken(
              database.formattedName_,
              projectId!,
            );
          });
        });
```

I think the same problem will come for instanceConfigs object also. Using Spanner client for AdminOperations are deprecated but it may still impact existing customers . Let me think of an alternate solution

handwritten/spanner/src/instrument.ts
surbhigarg92: > If this function happens to be called before the application has configured OpenTelemetry, then the value that it calculates and caches can be wrong.

Expectation is OpenTelemetry Global registration should be done before Spanner instance is created, even in Java we expect customers to do before SpannerInstance creation , if done later it will not be picked for adding traces. 

> It does not take into the (maybe theoretical) possibility that an application could change its configuration later.

If opentelemetry provider is passed while creating Spanner object that will be considered, but global configuration is not expected to be changed later

handwritten/spanner/src/instrument.ts
surbhigarg92: Thanks for highlighting it. Completely missed this . Another option which I was trying was below which also didn't look like a strong approach. I am checking this with OpenTelemetry team . Unlike Java, Node does not expose an option to check if tracer is enabled https://github.com/open-telemetry/opentelemetry-java/blob/31b3cd5f561a7cf6278a255fad33d40887c1a48b/api/all/src/main/java/io/opentelemetry/api/trace/Tracer.java#L72

```
const globalProvider = trace.getTracerProvider();
  if (globalProvider) {
    let delegate = globalProvider;
    if (typeof (globalProvider as any).getDelegate === 'function') {
      delegate = (globalProvider as any).getDelegate();
    }
    if (delegate) {
      const name = delegate.constructor.name;
      // Exclude the dummy NoopTracerProvider and uninitialized ProxyTracerProvider
      if (name !== 'NoopTracerProvider' && name !== 'ProxyTracerProvider') {
        globalTracingEnabled = true;
        return true;
      }
    }
  }
  ```

File: handwritten/spanner/src/instrument.ts (hunk 5 above)
Comment (surbhigarg92) [reply handwritten/spanner/src/instrument.ts:138]: If we try to accommodate the requirement of letting customer register OpenTelemetry later, we will not be able to avoid enabling registering of `AsyncHooksContextManager()` . Registering this adds a good load on the application.

File: handwritten/spanner/test/spanner.ts (hunk 7 above)
Comment (surbhigarg92) [reply handwritten/spanner/test/spanner.ts:94]: > It does not matter when the OpenTelemetry configuration is done (before or after creating a Spanner instance).

As mentioned in previous comment reply, we need to discuss if we want to allow this usecase.

> It does not matter what the trace sampling is. When tracing is enabled, even with a 1% sampling rate, then the request ID should be added to the traces.

Sure will add it

File: handwritten/spanner/src/instrument.ts (hunk 8 above)
Comment (surbhigarg92) [reply handwritten/spanner/src/instrument.ts]: Removed it

File: handwritten/spanner/src/request_id_header.ts (hunk 6 above)
Comment (surbhigarg92) [reply handwritten/spanner/src/request_id_header.ts]: No. Removed it

File: handwritten/spanner/src/instrument.ts (hunk 4 above)
Comment (surbhigarg92) [reply handwritten/spanner/src/instrument.ts]: Discussed with OpenTelemetry team and this seems to be the right solution for the time being. They will also be working on a feature request to expose an option `isEnabled` similar to Java. Will replace this once we have the solution in place.

File: handwritten/spanner/src/index.ts (hunk 3 above)
Comment (surbhigarg92) [reply handwritten/spanner/src/index.ts:1696]: Added it back

File: handwritten/spanner/src/index.ts (hunk 2 above)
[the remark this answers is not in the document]
Comment (surbhigarg92) [reply handwritten/spanner/src/index.ts]: Will be raising a separate PR for `replaceProjectIdToken` optimisations

File: handwritten/spanner/src/instrument.ts (hunk 5 above)
Comment (surbhigarg92) [reply handwritten/spanner/src/instrument.ts:138]: Have added 10 secs cache TTL to keep checking if globalOpenTelemetry has been registered

File: handwritten/spanner/src/instrument.ts [hunk 9]
```diff
@@ -119,6 +119,67 @@ function ensureInitialContextManagerSet() {
...
+    }
+    if (delegate) {
+      const name = delegate.constructor.name;
+      // Exclude the dummy NoopTracerProvider and uninitialized ProxyTracerProvider
+      if (name !== 'NoopTracerProvider' && name !== 'ProxyTracerProvider') {
+        globalTracingEnabled = true;
+        ensureInitialContextManagerSet();
+        return true;
+      }
+    }
```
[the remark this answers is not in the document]
surbhigarg92: Verified with OpenTelemetry team using delegate is the right way, as  startSpan is not reliable when sampling is configured

handwritten/spanner/test/spanner.ts
```diff
@@ -7311,8 +7319,156 @@ describe('Spanner with mock server', () => {
...
+      assert.strictEqual(getTracerProviderStub.callCount, 1);
+
+      // 2. Even if OpenTelemetry is registered immediately after, it should hit the cache and return false
+      getTracerProviderStub.returns({
+        constructor: { name: 'ProxyTracerProvider' },
+        getDelegate: () => ({ constructor: { name: 'NodeTracerProvider' } }),
+        getTracer: () => ({
+          startActiveSpan: (name, options, cb) => cb({ setAttribute: () => {}, end: () => {} } as any),
+        }),
+      } as any);
```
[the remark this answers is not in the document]
surbhigarg92: Not required

handwritten/spanner/test/spanner.ts
```diff
@@ -7311,8 +7319,156 @@ describe('Spanner with mock server', () => {
...
+
+      const getTracerProviderStub = ttlSandbox.stub(trace, 'getTracerProvider');
+      // First phase: global provider is unconfigured (Proxy/Noop)
+      // We delegate getTracer to provider so that if getTracer is ever called, it works perfectly,
+      // but name is 'ProxyTracerProvider' so it is detected as unconfigured!
+      getTracerProviderStub.returns({
+        constructor: { name: 'ProxyTracerProvider' },
+        getDelegate: () => ({ constructor: { name: 'NoopTracerProvider' } }),
+        getTracer: (name, version) => provider.getTracer(name, version),
+      } as any);
```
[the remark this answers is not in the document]
surbhigarg92: not required

handwritten/spanner/src/instrument.ts
```diff
@@ -119,6 +119,67 @@ function ensureInitialContextManagerSet() {
...
+  const globalProvider = trace.getTracerProvider();
+  if (globalProvider) {
+    let delegate = globalProvider;
+    if (typeof (globalProvider as any).getDelegate === 'function') {
+      delegate = (globalProvider as any).getDelegate();
+    }
+    if (delegate) {
+      const name = delegate.constructor.name;
+      // Exclude the dummy NoopTracerProvider and uninitialized ProxyTracerProvider
+      if (name !== 'NoopTracerProvider' && name !== 'ProxyTracerProvider') {
```
olavloite, line 151: nit: this check depends on the stability of these class names. That has two problems:
- If OpenTelemetry decides to change the name, then this check is no longer valid. This is however unlikely.
- If a project is minified, then these class names can be mangled by the minifier, which would (probably) break this check. At least, as far as I know, minifiers do not actively change this type of checks when minifying code.

An alternative could be to do the check like this:

```typescript
const probeSpan = globalProvider
  .getTracer(TRACER_NAME, TRACER_VERSION)
  .startSpan('probe');
// isSpanContextValid returns false for NoopSpan, as it does not have a traceID.
const isValid = trace.isSpanContextValid(probeSpan.spanContext());
probeSpan.end();
if (isValid) {
  globalTracingEnabled = true;
  ensureInitialContextManagerSet();
  return true;
}
```

PR #8340, opened 2026-05-21
docs(storage): add gaxios migration guide

thiyaguk09:


handwritten/storage/system-test/storage.ts
```diff
@@ -286,12 +286,7 @@ describe('storage', function () {
         await bucket.acl.delete({entity: USER_ACCOUNT});
       });
 
-      /**
-       * TODO: Re-enable once the test environment allows public IAM roles.
-       * Currently disabled to avoid 403 errors when adding 'allUsers' or
-       * 'allAuthenticatedUsers' permissions.
-       */
-      it.skip('should make a bucket public', async () => {
+      it('should make a bucket public', async () => {
```
Dhriti07: It looks like the changes to enable the tests might have been included by mistake, as this PR seems focused on the migration guide.  

Can you check if this is intended to be a part of the system test pr

thiyaguk09: Removed system test changes.

handwritten/storage/MIGRATION.md
```diff
@@ -0,0 +1,77 @@
...
+
+```js
+const [retrievedFile, apiResponse] = await file.get();
+const contentType = apiResponse.headers['content-type'];
+```
+
+**After (Storage v8):**
+
+```js
+const [retrievedFile, apiResponse] = await file.get();
```
dashpole: In `ServiceObject.get` (`src/nodejs-common/service-object.ts`), `callback` is invoked with only the instance (`callback(null, self)`), so `apiResponse` from `await file.get()` is `undefined` in v8 (and was the resource metadata rather than the HTTP response in v7). Consider using `await file.getMetadata()` in both the before and after snippets so `apiResponse` is the `GaxiosResponse` containing `headers`:
```suggestion
const [metadata, apiResponse] = await file.getMetadata();
```

handwritten/storage/MIGRATION.md
```diff
@@ -0,0 +1,77 @@
...
+const customHeader = apiResponse.config.headers.get('x-custom-header');
+```
+
+> [!WARNING]
+> **Header Value Stringification:** Plain JavaScript objects allow passing non-string values (such as arrays or numbers) which are implicitly processed. However, the native `Headers` constructor strictly converts all values to standard string representations. For example, passing an array of values (e.g., `['val1', 'val2']`) will result in a single comma-separated string (e.g., `'val1, val2'`). Ensure you pre-format or verify your header values before passing them to custom options.
+
+
+### 3. URL Resolution (`baseURL`)
+
+If you are using custom `baseURL` options or passing relative URLs to methods that accept them, be aware that resolution now strictly follows the standard native `URL` constructor spec (`new URL(url, baseURL)`). This can affect how leading slashes in paths resolve.
```
dashpole: In `StorageTransport.makeRequest` (`src/storage-transport.ts`), relative URLs are resolved against `this.baseUrl` (`new URL(urlString, this.baseUrl)`) and `url: requestUrl` overwrites `reqOpts.url`, so passing `baseURL` in per-method `GaxiosOptions` has no effect because `url` is already absolute when passed to `gaxios`. It might help to reference `baseUrl` (on `Storage` / `ServiceObject`) instead of `baseURL` options here.

PR #8341, opened 2026-05-21
chore(storage): add storage benchmark tool for Gaxios refactor comparison

thiyaguk09:


handwritten/storage/internal-tooling/benchmark.ts
```diff
@@ -0,0 +1,174 @@
...
+  execSync(`npm install @google-cloud/storage@${version} --silent`, {cwd: tempDir});
+  const baselinePath = path.join(tempDir, 'node_modules', '@google-cloud/storage');
+  
+  const pkgJson = JSON.parse(fs.readFileSync(path.join(baselinePath, 'package.json'), 'utf8'));
+  const main = pkgJson.main || './build/src/index.js';
+  const entry = path.join(baselinePath, main);
+  
+  console.log(`Loading baseline from ${entry}`);
+  const pkg = await import(entry);
+  return pkg.Storage;
```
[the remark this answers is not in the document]
thiyaguk09: updated

handwritten/storage/internal-tooling/benchmark.ts
```diff
@@ -0,0 +1,174 @@
...
+}
+
+async function runBenchmark(StorageClass: typeof Storage, name: string, bucketName: string) {
+  // 2. Pass custom project ID to the storage client
+  const storage = new StorageClass({ projectId: argv.project });
+  const bucket = storage.bucket(bucketName);
+  const content = Buffer.alloc(1024, 'a'); // 1KB
+  const uploadedFiles: any[] = [];
```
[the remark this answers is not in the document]
thiyaguk09: type added

handwritten/storage/internal-tooling/benchmark.ts
```diff
@@ -0,0 +1,174 @@
...
+    await Promise.all(uploadedFiles.map(f => f.delete().catch(() => {})));
+    logMemory('After Cleanup');
+  }
+}
+
+function reportResults(operation: string, times: number[]) {
+  const min = Math.min(...times);
+  const max = Math.max(...times);
+  const avg = times.reduce((a, b) => a + b, 0) / times.length;
+  const throughput = (1024 / (avg / 1000)) / 1024; // KB/s
```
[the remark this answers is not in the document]
thiyaguk09: done

handwritten/storage/internal-tooling/benchmark.ts
```diff
@@ -0,0 +1,174 @@
...
+  console.log(`Loading baseline from ${entry}`);
+  const pkg = await import(entry);
+  return pkg.Storage;
+}
+
+async function runBenchmark(StorageClass: typeof Storage, name: string, bucketName: string) {
+  // 2. Pass custom project ID to the storage client
+  const storage = new StorageClass({ projectId: argv.project });
+  const bucket = storage.bucket(bucketName);
+  const content = Buffer.alloc(1024, 'a'); // 1KB
```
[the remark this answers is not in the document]
thiyaguk09: fixed

handwritten/storage/internal-tooling/benchmark.ts
```diff
@@ -0,0 +1,178 @@
...
+  const main = pkgJson.main || './build/src/index.js';
+  const entry = path.join(baselinePath, main);
+  
+  console.log(`Loading baseline from ${entry}`);
+  const pkg = await import(entry);
+  return pkg.Storage || pkg.default?.Storage || pkg.default;
+}
+
+async function runBenchmark(StorageClass: typeof Storage, name: string, bucketName: string) {
+  // 2. Pass custom project ID to the storage client
```
Dhriti07: Nit: Is this comment required ? What do the numbers on the comment refer to ?

handwritten/storage/internal-tooling/README.md
```diff
@@ -40,4 +40,32 @@ For each invocation of the benchmark, write a new object of random size between
...
+2. **Execute the benchmark comparison:**
+   ```bash
+   node build/esm/internal-tooling/benchmark.js --project <YOUR_PROJECT_ID> --bucket <YOUR_BUCKET_NAME> --iterations 100 --baseline 7.19.0
+   ```
+
+### CLI Parameters:
+
+| Parameter | Description | Requirement | Default |
+| --------- | ----------- | :---: | :---: |
+| `--project` | Google Cloud Project ID | **Required** | - |
```
Dhriti07: Nit: Can we use projectId as the flag name like the other files in this folder for standardization ?

handwritten/storage/internal-tooling/benchmark.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  console.log(`\n=== Running benchmark for ${name} ===`);
+  const logMemory = (prefix: string) => {
+    const mem = process.memoryUsage();
+    console.log(`${prefix} - Heap Used: ${(mem.heapUsed / 1024 / 1024).toFixed(2)} MB / Heap Total: ${(mem.heapTotal / 1024 / 1024).toFixed(2)} MB`);
+  };
+
+  try {
+    // Scenario 1: Upload Small File
+    console.log('Starting Scenario 1: Upload (1KB)...');
```
Dhriti07: Consider dividing this into helper functions for each test scenario.

handwritten/storage/internal-tooling/benchmark.ts
```diff
@@ -0,0 +1,178 @@
...
+      if (i % 10 === 0) logMemory(`  Metadata iteration ${i}`);
+      const start = performance.now();
+      await mainFile.getMetadata();
+      metadataTimes.push(performance.now() - start);
+    }
+    reportResults('Get Metadata', metadataTimes);
+    logMemory('After Metadata');
+
+    // Scenario 3: Download Small File
+    console.log('Starting Scenario 3: Download (1KB)...');
```
Dhriti07: Is there a reason for only choosing only 1KB file size? Can we make this parameterized for other file sizes as well ?

Dhriti07: Are there other operations which are affected by the migration ? 
Some possible examples: Large file transfers, Resumable uploads. If yes we should be benchmarking them as well to make sure we do not see any regression in them due to the migration

thiyaguk09: > Is there a reason for only choosing only 1KB file size? Can we make this parameterized for other file sizes as well ?

Introduced the `--fileSize` option (defaulting to `1024` bytes) in `yargs`. The file size buffer is now allocated dynamically (`Buffer.alloc(argv.fileSize, 'a')`), and throughput calculations adapt to the configured file size.

handwritten/storage/internal-tooling/benchmark.ts
thiyaguk09: Extracted individual scenario logic out of the main `runBenchmark` function into distinct, reusable async helper functions:
  - `runUploadScenario(bucket, content, name, uploadedFiles)`
  - `runMetadataScenario(mainFile)`
  - `runDownloadScenario(mainFile)`

handwritten/storage/internal-tooling/README.md
thiyaguk09: Changed the `--project` option in `benchmark.ts` to `--projectid`.

handwritten/storage/internal-tooling/benchmark.ts
thiyaguk09: Removed prefix numbers (e.g., `// 1.`, `// 2.`) from all comments and simplified them to clean, standard developer comments.

handwritten/storage/internal-tooling/benchmark.ts
thiyaguk09: > Are there other operations which are affected by the migration ? Some possible examples: Large file transfers, Resumable uploads. If yes we should be benchmarking them as well to make sure we do not see any regression in them due to the migration

Agreed! Our resumable uploads currently only support/rely on Gaxios, so testing this path is crucial. To support better benchmarking here, I've added a `--resumable` boolean flag option. Under the hood, `runUploadScenario` passes this flag to the storage `iterFile.save(content, { resumable: argv.resumable })` option, enabling targeted performance testing of resumable uploads.

Dhriti07: What other operations have been impacted by the migration ? Can you extend the benchmarking script to have them as well.

thiyaguk09: Hi @Dhriti07,

I have extended the `benchmark.ts` script to include Stream Uploads (`createWriteStream`), Stream Downloads (`createReadStream`), and List Files (`getFiles`) alongside the existing standard upload/download and metadata scenarios.

These additions ensure we thoroughly validate the Gaxios transport layer for regressions across all core operations.

handwritten/storage/internal-tooling/benchmark.ts
```diff
@@ -0,0 +1,1169 @@
...
+    .replace(/-+/g, '-')
+    .replace(/^-|-$/g, '');
+  const bucketName = `bench-lock-${safeName}-${Date.now()}`;
+  const bucket = storage.bucket(bucketName);
+  await bucket.create();
+
+  try {
+    return await runConcurrent(
+      argv.iterations,
+      argv.concurrency,
```
dashpole: Since all workers share the same `bucket` instance here and `bucket.lock(metageneration!)` sets an `ifMetagenerationMatch` precondition, running with `concurrency > 1` will cause concurrent `setRetentionPeriod` calls to bump `metageneration` and fail with HTTP 412 precondition errors. Like `runBucketIamScenario` below, it might be safer to run this serially:
```suggestion
      1,
```

handwritten/storage/internal-tooling/benchmark.ts
```diff
@@ -0,0 +1,1169 @@
...
+          const [notification] = await bucket.createNotification(dummyTopic);
+          createdNotifications.push(notification);
+
+          await notification.getMetadata();
+          await notification.get();
+          await notification.exists();
+          await bucket.getNotifications();
+          await notification.delete();
+        } catch (err) {
+          // skip
```
dashpole: When `createNotification` fails (for instance because `dummyTopic` does not exist in Pub/Sub), falling through to `return performance.now() - start` causes `reportResults` to count the failed request as a valid iteration since `times.filter(t => t > 0)` only filters out `0`. Consider returning `0` here (and in `runHmacKeyScenario` on line 837 and `runBucketIamScenario` on line 876) so skipped runs are excluded:
```suggestion
          return 0;
```

PR #8344, opened 2026-05-21
fix(spanner): don't tear down host app's OTel ContextManager

jcalem-rogo:
## Problem

`ensureInitialContextManagerSet()` in `handwritten/spanner/src/instrument.ts` (added to make async/await tracing work for apps that haven't configured OpenTelemetry) currently tears down the host application's `ContextManager` on most calls. The guard is:

```ts
if (!context['_contextManager'] || context.active() === ROOT_CONTEXT) {
  context.disable();  // disables any prior contextManager
  const contextManager = new AsyncHooksContextManager();
  contextManager.enable();
  context.setGlobalContextManager(contextManager);
}
```

Two issues:

1. **`context['_contextManager']` is not a public field on the API singleton.** `setGlobalContextManager` writes into the package-private global registry (`registerGlobal` in `@opentelemetry/api/internal/global-utils.js`), not to a `_contextManager` property on the API object. So the first leg of the OR is effectively always true and the function runs its install path on every call.
2. **`context.active() === ROOT_CONTEXT` is true on every gRPC call made outside an active span.** That's exactly what the Spanner session pool does during background warmup (BatchCreateSessions, keep-alives, pool maintenance). On those calls, the function `context.disable()`s the host app's already-installed `ContextManager` and replaces it with a fresh `AsyncHooksContextManager`.

## Impact

Any host app that:

- installs its own `ContextManager` (via `NodeSDK`, manual `AsyncHooksContextManager`, etc.)
- propagates baggage on the root context (user id, request id, tenant id, etc.)
- uses Spanner in the same process

…will see its baggage and active-context state silently disappear shortly after `new Spanner({...})` warms up the session pool. Span parents and `BaggageSpanProcessor`-style enrichment break for everything downstream.

We hit this in a Node service that uses OTel + Spanner. Chats produced LLM observation log records with `undefined` chat ids because the host's baggage had been wiped by Spanner's pool-warmup `startTrace` call. Bisecting the regression landed exactly on adding `@google-cloud/spanner` to the process.

Affected versions confirmed: `8.7.1` (latest), `8.6.0`, and the same body exists back through `7.21.0`. No relevant change between 7.x and 8.x in this code path.

## Fix

Use the public API surface: `setGlobalContextManager()` already returns `false` when a manager is already registered. Try to install; if something else owns the slot, back out without touching anything.

```ts
function ensureInitialContextManagerSet() {
  const contextManager = new AsyncHooksContextManager();
  contextManager.enable();
  if (!context.setGlobalContextManager(contextManager)) {
    contextManager.disable();
  }
}
```

This preserves the original intent — apps without OTel still get a working `AsyncHooksContextManager` — while ensuring Spanner cannot tear down a manager the host already configured.

## Notes

- Removed the now-unused `ROOT_CONTEXT` import from `handwritten/spanner/src/instrument.ts`.
- Happy to add a unit test if the maintainers can point me at the preferred pattern; the existing `observability-test/observability.ts` blocks use a per-describe `setGlobalContextManager` for setup, which doesn't compose cleanly with assertions on this function's own registration behavior. The fix is verified end-to-end against a host service that previously showed the symptom.


handwritten/spanner/src/multiplexed-session.ts
```diff
@@ -123,25 +124,29 @@ export class MultiplexedSession
...
-            setSpanError(span, e as Error);
-            throw e;
-          } finally {
-            span.end();
-          }
-        },
-      );
+      return context.with(ROOT_CONTEXT, () => {
+        return startTrace(
+          'MultiplexedSession.createSession',
```
surbhigarg92, line 129: Would need a similar change for regular session pool `startTrace('SessionPool.createSessions'`

alkatrivedi: done

alkatrivedi:
Hi @jcalem-rogo, thanks a lot for the PR and the fix!

currently one of the required check `cla/google` is failing. could you please sign the agreement on the Google Developers CLA page?

If you prefer, we can also just create a new PR carrying over these exact changes for you. Thanks again for your contribution!

jcalem-rogo:
> Hi @jcalem-rogo, thanks a lot for the PR and the fix!
> 
> currently one of the required check `cla/google` is failing. could you please sign the agreement on the Google Developers CLA page?
> 
> If you prefer, we can also just create a new PR carrying over these exact changes for you. Thanks again for your contribution!

Hi @alkatrivedi, if you could create the PR that would be great, thanks!

alkatrivedi:
closing this PR since we have raised a different [PR](https://github.com/googleapis/google-cloud-node/pull/8344) containing exact same changes for the fix

PR #8350, opened 2026-05-22
feat(storage): enable CRC32C validation by default in transfer manager

thiyaguk09:

### DIFF
--- a/handwritten/storage/src/transfer-manager.ts
+++ b/handwritten/storage/src/transfer-manager.ts
@@ -117,7 +117,7 @@ export interface DownloadFileInChunksOptions {
   concurrencyLimit?: number;
   chunkSizeBytes?: number;
   destination?: string;
-  validation?: 'crc32c' | false;
+  validation?: 'crc32c' | boolean;
   noReturnData?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


handwritten/storage/src/transfer-manager.ts
```diff
@@ -736,7 +736,7 @@ export class TransferManager {
    * @property {number} [concurrencyLimit] The number of concurrently executing promises
    * to use when downloading the file.
    * @property {number} [chunkSizeBytes] The size in bytes of each chunk to be downloaded.
-   * @property {string | boolean} [validation] Whether or not to perform a CRC32C validation check when download is complete.
+   * @property {string | boolean} [validation] Whether or not to perform a CRC32C validation check when download is complete. Defaults to 'crc32c'.
```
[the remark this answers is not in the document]
thiyaguk09: Agreed & updated.

handwritten/storage/test/transfer-manager.ts
```diff
@@ -612,26 +615,48 @@ describe('Transfer Manager', () => {
...
-      };
 
       await transferManager.downloadFileInChunks(file, {validation: 'crc32c'});
-      assert.strictEqual(callCount, 1);
+      assert.strictEqual(fromFileStub.callCount, 1);
     });
 
-    it('should throw an error if crc32c validation fails', async () => {
+    it('should perform CRC32C validation by default', async () => {
+      file.download = () => {
```
v-pratap: the tests manually override the download method using file.download = () => { }. However, slightly further down at line 646, the code correctly uses the Sinon sandbox to mock the method: sandbox.stub(file, 'download').callsFake 
Does it make difference?

v-pratap (approved):

handwritten/storage/test/transfer-manager.ts
thiyaguk09: Great catch! I've updated the tests to use `sandbox.stub()` instead of manual overrides. This is much safer because Sinon automatically restores the stubs (preventing test pollution) and gives us built-in spying, whereas manual overrides permanently mutate the object.

PR #8358, opened 2026-05-26
fix(projectify): optimize replaceProjectIdToken performance

surbhigarg92:
This PR optimizes the core replaceProjectIdToken utility function for maximum performance, memory efficiency, and safety.

By moving to an optimized Selective-Write In-Place Mutation design and restructuring control flow, we eliminate exponential recursive redundancy on arrays, bypass unnecessary checks on primitive values, and prevent any write operations on unmodified object/array elements.

This keeps memory allocations at absolute zero (avoiding GC ticks) and allows frozen arrays and objects to be traversed safely without throwing TypeErrors as long as they do not contain placeholders to replace.

### DIFF
--- a/core/projectify/test/index.ts
+++ b/core/projectify/test/index.ts
@@ -118,7 +118,7 @@ describe('projectId placeholder', () => {
     );
   });
 
-  it('should return values without placeholder as-is', () => {
+  it('should not modify primitives without placeholder', () => {
     assert.strictEqual(
       replaceProjectIdToken('no-placeholder', PROJECT_ID),
       'no-placeholder',
@@ -127,19 +127,25 @@ describe('projectId placeholder', () => {
     assert.strictEqual(replaceProjectIdToken(true, PROJECT_ID), true);
     assert.strictEqual(replaceProjectIdToken(null, PROJECT_ID), null);
     assert.strictEqual(replaceProjectIdToken(undefined, PROJECT_ID), undefined);
+  });
 
+  it('should not modify arrays without placeholder', () => {
     const array = [1, 2, 3];
     assert.strictEqual(replaceProjectIdToken(array, PROJECT_ID), array);
+  });
 
+  it('should not modify objects without placeholder', () => {
     const object = {a: 1, b: 2};
     assert.strictEqual(replaceProjectIdToken(object, PROJECT_ID), object);
   });
 
-  it('should handle frozen arrays and objects without placeholders correctly without throwing', () => {
+  it('should traverse frozen arrays without placeholder safely', () => {
     const frozenArray = Object.freeze(['no-placeholder', 123, true]);
     const replacedArray = replaceProjectIdToken(frozenArray, PROJECT_ID);
     assert.strictEqual(frozenArray, replacedArray);
+  });
 
+  it('should traverse frozen objects without placeholder safely', () => {
     const frozenObject = Object.freeze({
       prop: 'no-placeholder',
       other: 123,


core/projectify/src/index.ts
```diff
@@ -25,33 +28,35 @@ import {Stream} from 'stream';
...
-    (value as string).indexOf('{{projectId}}') > -1
-  ) {
-    if (!projectId || projectId === '{{projectId}}') {
-      throw new MissingProjectIdError();
+  for (const key in value) {
+    if (Object.prototype.hasOwnProperty.call(value, key)) {
+      value[key] = replaceProjectIdToken(value[key], projectId);
     }
-    value = (value as string).replace(/{{projectId}}/g, projectId);
   }
```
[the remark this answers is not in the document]
surbhigarg92: Fixed

core/projectify/src/index.ts
```diff
@@ -25,33 +28,35 @@ import {Stream} from 'stream';
...
+  if (value === null || typeof value !== 'object') {
+    return value;
+  }
+
+  if (Array.isArray(value)) {
+    for (let i = 0; i < value.length; i++) {
+      value[i] = replaceProjectIdToken(value[i], projectId);
     }
+    return value;
+  }
```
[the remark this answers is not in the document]
surbhigarg92: Replied in https://github.com/googleapis/google-cloud-node/pull/8358#discussion_r3304715333

core/projectify/src/index.ts
```diff
@@ -25,33 +28,43 @@ import {Stream} from 'stream';
...
+  if (Array.isArray(value)) {
+    for (let i = 0; i < value.length; i++) {
+      const original = value[i];
+      const processed = replaceProjectIdToken(original, projectId);
+      if (processed !== original) {
+        value[i] = processed;
       }
     }
+    return value;
   }
```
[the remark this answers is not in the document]
surbhigarg92: The original implementation was already mutating values in-place for all nested objects
```
// Original implementation mutated objects in-place:
for (const opt in value) {
  if (value.hasOwnProperty(opt)) {
    value[opt] = replaceProjectIdToken(value[opt], projectId);
  }
}
```

To address the concern about frozen arrays/objects without triggering new allocations, we implemented a Selective-Write strategy. 

```
const original = value[i];
const processed = replaceProjectIdToken(original, projectId);
if (processed !== original) {
  value[i] = processed; // Only writes if a placeholder was actually found & changed!
}
```

core/projectify/test/index.ts
```diff
@@ -116,6 +118,36 @@ describe('projectId placeholder', () => {
     );
   });
 
+  it('should return values without placeholder as-is', () => {
```
pearigee: There are a lot of assertions in both of these added test cases. Could you break this into multiple more focused tests?

For example:
```
it('should not modify primitives, () => {
    assert.strictEqual(replaceProjectIdToken(123, PROJECT_ID), 123);
    assert.strictEqual(replaceProjectIdToken(true, PROJECT_ID), true);
    assert.strictEqual(replaceProjectIdToken(null, PROJECT_ID), null);
    assert.strictEqual(replaceProjectIdToken(undefined, PROJECT_ID), undefined);
});

it('should not modify arrays without the placeholder, () => {
   const array = [1, 2, 3];
   assert.strictEqual(replaceProjectIdToken(array, PROJECT_ID), array);
});

// etc . . .
```

core/projectify/src/index.ts
```diff
@@ -25,33 +28,43 @@ import {Stream} from 'stream';
...
-      throw new MissingProjectIdError();
+  if (value instanceof Buffer || value instanceof Stream) {
+    return value;
+  }
+
+  for (const key in value) {
+    if (Object.prototype.hasOwnProperty.call(value, key)) {
+      const original = value[key];
+      const processed = replaceProjectIdToken(original, projectId);
+      if (processed !== original) {
```
pearigee, line 64: I think this could be simplified using more modern syntax:

```typescript
for (const key of Object.keys(value)) {
    value[key] = replaceProjectIdToken(value[key], projectId);
}
```

I don't think we need the additional `if (processed !== original)` because this function has already been modified to return the references in place?

core/projectify/src/index.ts
```diff
@@ -25,33 +28,43 @@ import {Stream} from 'stream';
...
   }
 
-  if (
-    typeof value === 'string' &&
-    (value as string).indexOf('{{projectId}}') > -1
-  ) {
-    if (!projectId || projectId === '{{projectId}}') {
-      throw new MissingProjectIdError();
+  if (value instanceof Buffer || value instanceof Stream) {
+    return value;
```
pearigee, line 57: Should we combine this check for Buffer and Stream with the other checks above for null and non-object primitives?

```
// Early return for null, primitives (i.e. booleans, numbers), Buffers, and Streams. These are non-traversable leaf nodes.
// Note this must come after the string check.
if (
    value === null ||
    typeof value !== 'object' ||
    value instanceof Buffer ||
    value instanceof Stream
) {
    return value;
}
```

core/projectify/src/index.ts
surbhigarg92: 1. `Object.keys(value)` allocates a brand new array of strings , this triggers garbage collection and degrades speed compared to for...in.

2. `if (processed !== original)`:  selective-write check! - Without this check, the code will unconditionally attempt to write to value[key]. That means if a frozen object is passed (even one with no placeholders to replace), it will immediately crash with a TypeError. This check was intentionally added, 

Earlier, this method was indirectly safe when receiving frozen arrays because the original implementation used .map(), which safely cloned the reference. Since we migrated off .map() to achieve better performance speedup via in-place mutation, we engineered the if (original !== processed) check to preserve that exact historical frozen-structure compatibility. Without it, the unconditional assignment in the Object.keys() snippet would trigger an immediate TypeError in strict mode on frozen structures, even when no replacements are made.

core/projectify/src/index.ts
surbhigarg92: I dont mind doing this change, but I was simply trying to micro-optimize here. 

Chances of value having null, object or array is really high compared to Buffer or Stream. So I was trying to avoid this check `value instanceof Buffer || value instanceof Stream` for little later.

core/projectify/test/index.ts
surbhigarg92: Done

PR #8359, opened 2026-05-26
fix(spanner): resolve prelint, eslint, and prettier CI build failures

sakthivelmanii:


handwritten/spanner/package.json
alkatrivedi: can we remove this from here? as samples has been moved to different repository and running these scripts will result into an error

sakthivelmanii: Done

sakthivelmanii: Done

PR #8365, opened 2026-05-27
feat: consolidate package manager to pnpm on gapic-generator-typescript

quirogas:
This pull request consolidated the package manager used by `gapic-generator-typescript`.


core/generator/gapic-generator-typescript/pnpm-workspace.yaml
pearigee, line 1: I don't think we need to add this file? The workspace configuration is only useful when defining multiple packages. In this case, we are just saying the current directory is one (which is implicit given the package.json).

pearigee:
## I think we should keep these lockfiles (except maybe yarn)

I think consolidating around a package manager is a bit more complicated than deleting these lockfiles. 

At minimum, I think we NEED to have lock files for NPM (to cover automation and our customers) and PNPM (to cover our automation).

Lockfiles are most useful as a mitigation against supply chain attacks (we are going to be adding them everywhere in the near future). Deleting them exposes us to more risk. Simply running `npm install` is a major security risk these days UNLESS our versions are strictly defined in a lockfile. For example:
* https://www.cisa.gov/news-events/alerts/2025/09/23/widespread-supply-chain-compromise-impacting-npm-ecosystem

## Consolidation is mostly a CI/Automation problem

To consolidate around a single package manager, the key challenge is actually in our automation (GitHub Actions, GCB, Docker Containers, BazelBot, etc.). These automations use a mixture of PNPM and NPM. Note, that modern supply chain attacks are specifically designed to compromise CI (i.e. with Docker escape mechanisms). As a result, we probably need to keep a NPM and PNPM lock around.

Here are a few usage examples:
* BazelBot implictly uses the PNPM lockfile as its source of truth for versions: https://github.com/googleapis/google-cloud-node/blob/main/core/generator/gapic-generator-typescript/WORKSPACE#L47
* Our test runner uses PNPM to install and execute project scripts: https://github.com/googleapis/google-cloud-node/blob/main/ci/run_single_test.sh#L44
* Our interdependent test scripts use NPM directly: https://github.com/googleapis/google-cloud-node/blob/main/ci/run_interdependent_tests.sh#L166
* Our Docker containers use NPM to install PNPM: https://github.com/googleapis/google-cloud-node/blob/main/ci/Dockerfile#L37
* Many of our Google Cloud Build runners use NPM directly: https://github.com/googleapis/google-cloud-node/blob/main/handwritten/bigtable/cloudbuild.yaml#L8

quirogas:
Since the CI pipeline still requires the other package-lock file, I have reverted all changes except for the deletion of the yarn-lock file.

PR #8371, opened 2026-05-28
test(firestore): update integration tests to use enterprise database

wu-hui:
This PR configures the enterprise integration tests for the Node.js       
  Firestore SDK to use the `enterprise` database instead of `test-db`, and    
  runs them with the `RUN_ENTERPRISE_TESTS=yes` environment variable.


handwritten/firestore/dev/test/util/helpers.ts
```diff
@@ -90,28 +90,45 @@ export function createInstance(
...
+    check();
   });
 }
 
+/**
+ * Helper function to check if testing is run using the REST backend.
+ * This is lenient on the value of the FIRESTORE_PREFER_REST env variable.
+ */
+export function isRest(): boolean {
+  const preferRest = process.env.FIRESTORE_PREFER_REST;
```
MarkDuckworth: ideally this function would accept a firestore instance and check the settings there, however in the context of testing, it's generally the case that the test suite is run with the FIRESTORE_PREFER_REST flag set or unset.

wu-hui: Done.

PR #8374, opened 2026-05-28
build: switch to v1 version for owlbot copy

feywind:
The v1 of this API isn't being incorporated into the combined package in googleapis-gen, so I'm switching it to v1 explicitly so that we can get a release out.


pearigee:
Let's hold off on modifying the YAML files and see if the infrastructure will catch up.

Closing this for now (i am trying to clean up the PR queue).

feywind:
I was about the do the same, the infrastructure did indeed catch up.

https://github.com/googleapis/google-cloud-node/pull/8376

PR #8399, opened 2026-05-29
fix(spanner): isolate OpenTelemetry background timers in ROOT_CONTEXT to prevent context and memory leaks

surbhigarg92:
This PR isolates background maintenance timers in SessionPool and MultiplexedSession within OpenTelemetry's ROOT_CONTEXT.

Problem
OpenTelemetry uses `AsyncLocalStorage` to propagate context. In Node.js, when a timer like `setInterval` is scheduled, the currently active context is captured and "leaked" into the timer's execution scope. If these background timers (which handle session pings, evictions, and multiplexed session maintenance) are initialized or restarted during a user request, they capture that request's context. This leads to:

Memory Leaks: The request context and all associated metadata remain in memory indefinitely.
Incorrect Tracing: Background logs or spans may be incorrectly attributed to the parent request that happened to trigger the timer's scheduling.

Solution
By wrapping the `setInterval` calls themselves in context.with(ROOT_CONTEXT, ...), we ensure that the timers are scheduled within the root context, effectively isolating them from any transient request context .


olavloite:
1. I think that you also need to apply the same patch here:

https://github.com/googleapis/google-cloud-node/blob/4a075ebcb402b690671c885e716b78d73a25d2de/handwritten/spanner/src/metrics/metrics-tracer-factory.ts#L85-L89

2. Can we add a test that would have failed without this fix, and that now succeeds. For example this:

```typescript
import * as assert from 'assert';
import {before, after, beforeEach, afterEach, describe, it} from 'mocha';
import * as sinon from 'sinon';
import {context, trace, ROOT_CONTEXT} from '@opentelemetry/api';
import {NodeTracerProvider} from '@opentelemetry/sdk-trace-node';
import {AlwaysOnSampler} from '@opentelemetry/sdk-trace-base';
import {Database} from '../src/database';
import {SessionPool} from '../src/session-pool';
import {MultiplexedSession} from '../src/multiplexed-session';
import {
  ensureInitialContextManagerSet,
  _resetTracingEnabledForTest,
} from '../src/instrument';

describe('OpenTelemetry Context Isolation Tests', () => {
  const sandbox = sinon.createSandbox();
  let provider: NodeTracerProvider;

  const MOCK_DATABASE = {
    batchCreateSessions: sandbox.stub().resolves([[]]),
    databaseRole: 'parent_role',
    _observabilityOptions: {},
  } as unknown as Database;

  before(() => {
    _resetTracingEnabledForTest();
    ensureInitialContextManagerSet();

    provider = new NodeTracerProvider({
      sampler: new AlwaysOnSampler(),
    });
    provider.register();
    _resetTracingEnabledForTest();
  });

  after(async () => {
    await provider.shutdown();
  });

  afterEach(() => {
    sandbox.restore();
  });

  describe('SessionPool background housekeeping timers', () => {
    let sessionPool: SessionPool;

    beforeEach(() => {
      sessionPool = new SessionPool(MOCK_DATABASE, {
        min: 0,
        max: 10,
        idlesAfter: 10,
        keepAlive: 30,
      });
    });

    afterEach(() => {
      sessionPool._stopHouseKeeping();
    });

    it('should schedule evict and keep-alive setInterval calls in ROOT_CONTEXT', () => {
      const tracer = trace.getTracer('test');
      
      const setIntervalStub = sandbox.stub(global, 'setInterval').callsFake((cb, ms) => {
        const activeSpan = trace.getSpan(context.active());
        
        // Assert that the active context is ROOT_CONTEXT (i.e., no active span)
        assert.strictEqual(
          activeSpan,
          undefined,
          'setInterval scheduling must be isolated within ROOT_CONTEXT and not carry any active request span',
        );
        return {
          unref: () => {},
        } as any;
      });

      // Start an active request context
      tracer.startActiveSpan('request-span', span => {
        try {
          // Start housekeeping under the request context
          sessionPool._startHouseKeeping();
        } finally {
          span.end();
        }
      });

      // Verify that both evict and ping intervals were scheduled
      assert.strictEqual(setIntervalStub.callCount, 2);
    });
  });

  describe('MultiplexedSession background maintenance timer', () => {
    let multiplexedSession: MultiplexedSession;

    beforeEach(() => {
      multiplexedSession = new MultiplexedSession(MOCK_DATABASE);
    });

    afterEach(() => {
      if (multiplexedSession._refreshHandle) {
        clearInterval(multiplexedSession._refreshHandle);
      }
    });

    it('should schedule multiplexed session maintenance setInterval in ROOT_CONTEXT', () => {
      const tracer = trace.getTracer('test');

      const setIntervalStub = sandbox.stub(global, 'setInterval').callsFake((cb, ms) => {
        const activeSpan = trace.getSpan(context.active());
        
        // Assert that the active context is ROOT_CONTEXT (i.e., no active span)
        assert.strictEqual(
          activeSpan,
          undefined,
          'setInterval scheduling must be isolated within ROOT_CONTEXT and not carry any active request span',
        );
        return {
          unref: () => {},
        } as any;
      });

      // Start an active request context
      tracer.startActiveSpan('request-span', span => {
        try {
          // Start maintenance under the request context
          multiplexedSession._maintain();
        } finally {
          span.end();
        }
      });

      // Verify that the refresh interval was scheduled
      assert.strictEqual(setIntervalStub.callCount, 1);
    });
  });
});
```

PR #8406, opened 2026-06-01
chore: clean up linkinator configurations and test scripts across monorepo

JoeWang1127:
This pull request removes the obsolete package-level and root-level Linkinator configuration files, documentation test scripts, and associated devDependencies.

- Removed Configurations: Deleted all `linkinator.config.json` files at the package levels and repository root.
- Removed Obsolete Scripts: Removed the `docs-test` and `predocs-test` scripts from the `scripts` blocks in all package-level `package.json` files.
- Purged Dead Dependencies: Removed the `linkinator` devDependency from all package-level `package.json` files.
- Removed `linkinator` and `tsx` devDependencies from the root `package.json`.

For https://github.com/googleapis/librarian/issues/4274


JoeWang1127:
@quirogas @sofisl @pearigee do you think this approach make sense?

I'll refactor the script later.

sofisl:
IIUC, docs-test is run during the docs job (again, please correct me, I haven't been working in this reply lately) so my only concern is if this flow doesn't support the single "npm run docs-*" jobs since linkinator wouldn't be able to find a config in that example. In general, this flow breaks the cli flow that linkinator is used to using, that is my main concern.

JoeWang1127:
@sofisl yes, this pull request will change how to run doc test. I'm doing this b/c we think a package level `linkinator.config.json` is unnecessary (see https://github.com/googleapis/librarian/issues/4274).

However, if the current setup is best practice, we can keep these config files as-is and resolve the issue.

sofisl:
Right, here's my concern: https://github.com/googleapis/google-cloud-node/blob/80560690651f3b76272dab7908d3a46aebc3e478/handwritten/spanner/package.json#L50
`linkinator` is called in `npm run docs-test`, which is a command-line script that runs linkinator. So anywhere we call `npm run docs-test` will fail if we don't have the linkinator config (this may or may not be in the g3 scripts). Additionally, generally speaking we wouldn't be able to call linkinator from the cli in this method since the config wouldn't be available in the package. Which might make debugging difficult, for example.

I have talked to others in the node team in the past who were thinking about removing linkinator (that's a decision for them). In the meantime though, I think that if we make this change, you'd also want to make sure to remove the ability to call linkinator from the cli (i.e., `npm run docs-test`), since this will not work if you remove the per-package configs.

JoeWang1127:
> In the meantime though, I think that if we make this change, you'd also want to make sure to remove the ability to call linkinator from the cli (i.e., npm run docs-test), since this will not work if you remove the per-package configs.

Yes, I'll remove this config for every modules.

sofisl:
No it's not just the config that needs to be removed, it's the call to linkinator in the cli (i.e., the npm run docs-test script). Which may break other things, depending on where its called.

JoeWang1127:
bigtable-system-tests:
```
Step #1 - "run-system-tests":   2 failing
Step #1 - "run-system-tests": 
Step #1 - "run-system-tests":   1) Bigtable/ClientSideMetricsAllMethods
Step #1 - "run-system-tests":        "before all" hook in "Bigtable/ClientSideMetricsAllMethods":
Step #1 - "run-system-tests":      Error: Operation successfully rolled back : Insufficient node quota. You requested a node count of 1 nodes for your cluster, but this request would exceed your project's node quota of 100 nodes total across all clusters in this zone. Contact us to request a quota increase: https://cloud.google.com/bigtable/quotas#quota-increase
Step #1 - "run-system-tests":       at Operation._unpackResponse (node_modules/google-gax/build/src/longRunningCalls/longrunning.js:189:31)
Step #1 - "run-system-tests":       at /workspace/handwritten/bigtable/node_modules/google-gax/build/src/longRunningCalls/longrunning.js:169:18
Step #1 - "run-system-tests": 
Step #1 - "run-system-tests":   2) Bigtable/ClientSideMetrics
Step #1 - "run-system-tests":        "before all" hook in "Bigtable/ClientSideMetrics":
Step #1 - "run-system-tests":      Error: Operation successfully rolled back : Insufficient node quota. You requested a node count of 1 nodes for your cluster, but this request would exceed your project's node quota of 100 nodes total across all clusters in this zone. Contact us to request a quota increase: https://cloud.google.com/bigtable/quotas#quota-increase
Step #1 - "run-system-tests":       at Operation._unpackResponse (node_modules/google-gax/build/src/longRunningCalls/longrunning.js:189:31)
Step #1 - "run-system-tests":       at /workspace/handwritten/bigtable/node_modules/google-gax/build/src/longRunningCalls/longrunning.js:169:18
Step #1 - "run-system-tests": 
Step #1 - "run-system-tests": 
Step #1 - "run-system-tests": 
Finished Step #1 - "run-system-tests"
ERROR
ERROR: build step 1 "node:24" failed: step exited with non-zero status: 2
```

This error doesn't seem relevant to the change.

JoeWang1127:
pubsub-system-tests:
```
File handwritten/pubsub/cloudbuild.yaml not found
```

This error doesn't seem relevant to the change.

PR #8412, opened 2026-06-02
feat(spanner): optimize replaceProjectIdToken from projectify

surbhigarg92:
**Summary**
This pull request replaces the dependency on the legacy external @google-cloud/projectify package with an optimized, Spanner-specific internal token replacement utility in src/helper.ts.

**The Problem & Performance Bottleneck**
Previously, `@google-cloud/projectify` was utilized to scan gRPC request options and substitute `{{projectId}}` placeholder tokens in resource paths. Refer https://github.com/googleapis/google-cloud-node/blob/main/core/projectify/src/index.ts#L27
However, this generic approach was highly CPU and memory intensive for Spanner because:

*Redundant Deep Recurse*: The traversal recursively walked every element of massive user payloads, including SQL query strings (sql), query parameters (params, paramTypes), and multi-row inserts (mutations). None of these ever contain resource name placeholders.
*Heap-Allocation Storms*: The legacy implementation used .map() on all array properties, continually allocating and recreating short-lived arrays. Under high throughput, this caused severe Garbage Collection (GC) blocks.
*V8 Object Hidden Class De-optimization*: It wrote back properties unconditionally (e.g., value[key] = replaced), marking object shapes as dirty in the V8 engine and invalidating inline caches.

**The Solution**
We transitioned the token replacement utility locally into `src/helper.ts` and implemented two Spanner-specific optimizations:

*Skip-Keys Optimization*: We defined a Spanner-specific list of properties representing raw user data which never contain resource placeholders. When the recursive loop encounters these keys, it immediately skips traversing their children:
```
typescript
const KEYS_TO_SKIP = new Set([
  'params',
  'paramTypes',
  'mutations',
  'sql',
  'queryOptions',
  'requestOptions',
  'transactionOptions',
]);
```
*Selective-Write & Immutability Guard*: We only write back properties to objects/arrays when a change actually occurs (processed !== original). This prevents de-optimizing V8 hidden classes and ensures complete safety for frozen objects (Object.freeze) under strict mode execution.

### DIFF
--- a/handwritten/spanner/src/helper.ts
+++ b/handwritten/spanner/src/helper.ts
@@ -295,15 +295,21 @@ export function isUuid(value: any): boolean {
 const PROJECT_ID_TOKEN = '{{projectId}}';
 const PROJECT_ID_TOKEN_REGEX = /{{projectId}}/g;
 
-// Keys to skip recursive traversal for Spanner SDK requests to avoid scanning large user data payloads.
-const KEYS_TO_SKIP = new Set([
-  'params',
-  'paramTypes',
-  'mutations',
-  'sql',
-  'queryOptions',
-  'requestOptions',
-  'transactionOptions',
+// Whitelisted request properties that contain, or lead to, resource name strings with project ID placeholders.
+const KEYS_TO_SCAN = new Set([
+  'database',
+  'session',
+  'name',
+  'parent',
+  'instance',
+  'resource',
+  'backup',
+  'instanceConfig',
+  'baseConfig',
+  'kmsKeyName',
+  // Intermediate object keys that must be traversed to reach the above leaves
+  'config',
+  'encryptionConfig',
 ]);
 
 /**
@@ -341,13 +347,13 @@ export function replaceProjectIdToken(value: any, projectId: string): any {
     return value;
   }
 
-  if (value instanceof Buffer || value instanceof Stream) {
+  if (value instanceof Buffer || value instanceof Stream || isDate(value)) {
     return value;
   }
 
   for (const key in value) {
     if (Object.prototype.hasOwnProperty.call(value, key)) {
-      if (KEYS_TO_SKIP.has(key)) {
+      if (!KEYS_TO_SCAN.has(key)) {
         continue;
       }
       const original = value[key];
--- a/handwritten/spanner/test/helper.ts
+++ b/handwritten/spanner/test/helper.ts
@@ -136,24 +136,25 @@ describe('helper', () => {
 
     it('should replace any {{projectId}} it finds (nested / complex tree)', () => {
       const input = {
-        here: 'A {{projectId}} Z',
-        nested: {
-          here: 'A {{projectId}} Z',
-          nested: {
-            here: 'A {{projectId}} Z',
+        parent: 'A {{projectId}} Z',
+        database: {
+          parent: 'A {{projectId}} Z',
+          baseConfig: 'projects/{{projectId}}/instanceConfigs/base-1',
+          config: {
+            name: 'A {{projectId}} Z',
           },
         },
-        array: [
+        backup: [
           {
-            here: 'A {{projectId}} Z',
-            nested: {
-              here: 'A {{projectId}} Z',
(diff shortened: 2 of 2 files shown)


handwritten/spanner/src/helper.ts
```diff
@@ -290,3 +291,85 @@ export function isUuid(value: any): boolean {
...
+      const original = value[i];
+      const processed = replaceProjectIdToken(original, projectId);
+      if (processed !== original) {
+        value[i] = processed;
+      }
+    }
+    return value;
+  }
+
+  if (value instanceof Buffer || value instanceof Stream) {
```
olavloite: nit: Can we move this up before the check for arrays? That way, we have the trivial checks for early returns in one place.

handwritten/spanner/src/helper.ts
```diff
@@ -285,3 +286,81 @@ export function isError(value: any): boolean {
...
+    return value;
+  }
```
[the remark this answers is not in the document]
olavloite: Is there any reason why we should not do this?

handwritten/spanner/src/helper.ts
```diff
@@ -290,3 +291,85 @@ export function isUuid(value: any): boolean {
     )
   );
 }
+
+const PROJECT_ID_TOKEN = '{{projectId}}';
+const PROJECT_ID_TOKEN_REGEX = /{{projectId}}/g;
+
+// Keys to skip recursive traversal for Spanner SDK requests to avoid scanning large user data payloads.
+const KEYS_TO_SKIP = new Set([
```
olavloite: Can we invert this into an allow-list? This prevents future regressions if we introduce new fields with other names than the ones below that contain large amounts of data. I think that the following list would be enough:

```typescript
const KEYS_TO_SCAN = new Set([
  // Leaf keys containing resource paths
  'database',
  'session',
  'name',
  'parent',
  'instance',
  'resource',
  'backup',
  'instanceConfig',
  'kmsKeyName',
  // Intermediate object keys that must be traversed to reach the above leaves
  'config',
  'encryptionConfig',
]);
```

And then in the replace method:


```typescript
// ... primitive and Array/Stream/Buffer checks ...
  for (const key in value) {
    if (Object.prototype.hasOwnProperty.call(value, key)) {
      if (!KEYS_TO_SCAN.has(key)) {
        continue; // Instantly skip sql, params, mutations, keySet, etc.
      }
      const original = value[key];
      const processed = replaceProjectIdToken(original, projectId);
      if (processed !== original) {
        value[key] = processed;
      }
    }
  }
```

handwritten/spanner/src/helper.ts
surbhigarg92: I did this intentionally as chances of value having `arrays` is more compared to "Buffer" and "Stream" . It was a minor optimization only but I thought of reducing extra checks for Buffer and Stream

handwritten/spanner/src/helper.ts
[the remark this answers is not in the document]
surbhigarg92: Let me add this

handwritten/spanner/src/helper.ts
surbhigarg92: This is more efficient performance wise, but Isn't this more error prone. In case we have more keys with resource paths in future ?

handwritten/spanner/src/helper.ts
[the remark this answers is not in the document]
surbhigarg92: done

handwritten/spanner/src/helper.ts
surbhigarg92: Since its really rare this Spanner API will change , added the suggestion to use allow-list approach as it is more performance efficient

olavloite: yeah, my reasoning was also:
1. With a deny-list (the original approach), then regressions are likely to be 'silent'. If we add a new proto that contains a large field, and that field is not in this deny-list, then we will have a silent performance regression.
2. With an allow-list, then regressions cause errors. Those are easier to catch. And hopefully we catch them directly with tests.

PR #8444, opened 2026-06-05
feat(storage): add deleteSourceObjects option to combine/compose method

thiyaguk09:
## Description

This PR implements a client-side workaround for the `deleteSourceObjects` parameter in the GCS Compose API (`combine` method).

### Key Changes
- **API Signature & Types**: Added `deleteSourceObjects?: boolean` to `CombineOptions` and documented it in the `combine` method JSDocs.
- **Request Body Integration**: Added `deleteSourceObjects` to the JSON request body of the `/compose` request (omitted if `undefined` to ensure backward compatibility and clean requests).
- **Execution Logic**:
  - Executes the standard compose operation first.
  - If `deleteSourceObjects` is set to `true` and the compose operation is successful, programmatically deletes all source files.
  - If the compose operation fails, aborts immediately without deleting any source files.
- **Error Handling (`ComposeCleanupError`)**:
  - Added a new `ComposeCleanupError` class.
  - If the compose succeeds but any source file deletion fails, a `ComposeCleanupError` is thrown/returned.
  - The error lists all individual deletion failures (`errors`) and attaches the newly composed file (`newFile`) and raw `apiResponse` so they are still accessible to the caller.
- **Tests**: Added comprehensive unit tests in `test/bucket.ts` verifying all success, failure, and partial cleanup error conditions.


handwritten/storage/test/bucket.ts
```diff
@@ -916,6 +920,128 @@ describe('Bucket', () => {
...
+      const destination = bucket.file('destination.foo');
+
+      let deletedCount = 0;
+      sources.forEach(source => {
+        source.delete = async (opts?: any) => {
+          assert.strictEqual(opts?.userProject, 'user-project-id');
+          deletedCount++;
+          return [{}];
+        };
+      });
```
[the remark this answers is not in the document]
thiyaguk09: `ignoreNotFound: true` added.

handwritten/storage/src/bucket.ts
```diff
@@ -1732,24 +1764,51 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
...
+          const deletePromises = (sources as File[]).map(async source => {
+            try {
+              await source.delete({
+                userProject: options.userProject,
+              });
+            } catch (deleteErr) {
+              return deleteErr as Error;
+            }
+            return null;
+          });
```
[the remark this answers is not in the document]
thiyaguk09: done.

handwritten/storage/src/bucket.ts
```diff
@@ -1732,24 +1765,59 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
...
             }
 
             return sourceObject;
           }),
         },
-        qs: options,
+        qs: requestQueryObject,
       },
-      (err, resp) => {
+      async (err, resp) => {
```
v-pratap: The callback for this.storage.request is now an async function.  Is this safe, or should we be using standard .then() and .catch() for the cleanup promises instead?

handwritten/storage/src/bucket.ts
```diff
@@ -1732,24 +1765,59 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
...
         }
 
+        if (deleteSourceObjects) {
+          const deletePromises = (sources as File[]).map(async source => {
+            const deleteOptions: DeleteOptions = {
+              ignoreNotFound: true,
+              userProject: options.userProject,
+            };
+
+            const generation = source.generation ?? source.metadata?.generation;
```
v-pratap, line 1790: Why do we only check source.metadata.generation when building the Compose payload, but check both source.generation ?? source.metadata?.generation during the delete phase?

handwritten/storage/test/bucket.ts
```diff
@@ -916,6 +920,140 @@ describe('Bucket', () => {
 
       bucket.combine(sources, destination, done);
     });
+
+    it('should delete source objects if deleteSourceObjects is true', done => {
+      const sources = [bucket.file('1.foo'), bucket.file('2.foo')];
+      const destination = bucket.file('destination.foo');
+
+      // Set generation on the first file and leave second file without generation
+      sources[0].generation = 12345;
```
v-pratap, line 929: In the test where we explicitly set sources[0].generation = 12345, I don't see any code checking if this number actually makes it into the Compose API payload. Add an assertion like assert.strictEqual(reqOpts.json.sourceObjects[0].generation, 12345) inside the mock?

handwritten/storage/src/bucket.ts
thiyaguk09: Replaced async with .catch() and Promise.all().then() in the deleteSourceObjects cleanup to handle errors and avoid unhandled promise rejections.

handwritten/storage/src/bucket.ts
thiyaguk09: I updated Compose payload generation checks to match delete cleanup logic for consistency.

handwritten/storage/test/bucket.ts
thiyaguk09: Added assertion for generation to verify that parameters are flowing correctly to the Compose API

v-pratap (approved):

PR #8446, opened 2026-06-05
feat: add basic monorepo linter to enforce GTS standards

shivanee-p:
This PR refactors the monorepo linter `bin/linter.mjs` to run checks on files modified in a PR, with formatting, static linting, and TypeScript compilation checks. This is done via the local CLI binaries directly via execFileSync, avoiding complex programmatic configurations.

- Added `getGitTarget()` to check for the base branch in order of priority: `origin/<base>` -> `<base>` -> `HEAD~1`. This prevents linter failures in shallow-cloned or detached-HEAD CI environments (like GitHub Actions) and avoids comparing against outdated local branches.
- Configured ESLint to run Prettier formatting checks directly as a lint rule by extending `plugin:prettier/recommended` in the root .eslintrc.json.
- Installed `eslint-plugin-import` and `eslint-plugin-promise` to enforce best practices.
- Configured the rules as blocking errors in .eslintrc.json (only checked against touched files in the PR) `import/no-extraneous-dependencies`, `promise/catch-or-return`, and `promise/always-return`
- Implements `checkTypeSafety()` which runs `tsc --noEmit` on parent packages using `node_modules/typescript/bin/tsc`
- Any compilation or type safety failures in modified packages are blocking and will fail the linter execution.
- Added early exit logging `No TypeScript files changed. Skipping checks.` to provide immediate feedback when no .ts files are touched.


bin/linter.mjs
```diff
@@ -0,0 +1,88 @@
...
+
+// Define the base branch to compare against
+const baseBranch = process.env.GITHUB_BASE_REF || 'main';
+
+// Extensions to check with Prettier
+const targetExtensions = new Set([
+  '.js',
+  '.mjs',
+  '.cjs',
+  '.ts',
```
pearigee: I think we should exclusively test TS files. The .cjs, .mjs, .js are all either compiled or unpublished scripts.

What do you think? It should also improve runtime performance?

shivanee-p: You're so right. There's no need to check the other file types. This will drastically improve performance. Thanks!

bin/linter.mjs
```diff
@@ -0,0 +1,138 @@
...
+// Extensions to check for Prettier and ESLint/GTS
+const targetExtensions = new Set(['.ts']);
+
+function getChangedFiles() {
+  try {
+    const output = execFileSync(
+      'git',
+      ['diff', '--name-only', '--diff-filter=ACMRT', baseBranch],
+      {encoding: 'utf8'},
+    );
```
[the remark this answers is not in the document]
shivanee-p: Ack.

bin/linter.mjs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+(async () => {
+  const changedTsFiles = getChangedFiles();
+
+  // Run Prettier and ESLint concurrently
+  await Promise.all([
+    checkPrettierFormatting(changedTsFiles),
+    checkEslint(changedTsFiles),
+  ]);
```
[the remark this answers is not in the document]
shivanee-p: Ack.

bin/linter.mjs
```diff
@@ -0,0 +1,163 @@
...
+    );
+    return output
+      .split('\n')
+      .map(f => f.trim())
+      .filter(
+        f =>
+          f.length > 0 &&
+          targetExtensions.has(path.extname(f).toLowerCase()) &&
+          existsSync(f),
+      );
```
pearigee: Could we replace this extension checking filter by just modifying the git command above?

```
git diff --name-only --diff-filter=ACMRT origin/main -- '*.ts'
```

or 

```
const output = execFileSync(
      'git',
      ['diff', '--name-only', '--diff-filter=ACMRT', targetRef, '--', '*.ts'],
      {encoding: 'utf8'},
    );
```

bin/linter.mjs
```diff
@@ -0,0 +1,163 @@
...
+  if (filesToCheck.length === 0) {
+    return true;
+  }
+
+  // Map files to their package directories
+  const packagesToCheck = new Set();
+  for (const file of filesToCheck) {
+    const parts = file.split(path.sep);
+    // e.g. packages/google-cloud-storage/src/index.ts -> packages/google-cloud-storage
+    if (parts[0] === 'packages' && parts.length > 1) {
```
pearigee: Hmm, do we need to limit this to the packages directory alone? I think the same script could also work in core/handwritten no?

Could we just walk up the directory chain checking for a tsconfig.json file?

I took a quick stab in Antigravity to see what this might look like:

```
// Dynamically resolve closest package roots with a tsconfig.json
  for (const file of filesToCheck) {
    let dir = path.dirname(file);
    while (dir !== '.' && dir !== path.sep) {
      if (existsSync(path.join(dir, 'tsconfig.json'))) {
        packagesToCheck.add(dir);
        break;
      }
      dir = path.dirname(dir);
    }
  }
```

bin/linter.mjs
shivanee-p: Ah, yes we can. Thanks for the callout!

bin/linter.mjs
shivanee-p: Ack.

bin/linter.mjs
```diff
@@ -0,0 +1,235 @@
...
+    }
+    dir = path.dirname(dir);
+  }
+
+  const result = (dir === '.' && existsSync(path.join('.', 'tsconfig.json'))) ? '.' : null;
+  for (const v of visited) {
+    tsconfigCache.set(v, result);
+  }
+  return result;
+}
```
pearigee, line 163: It looks like the TypeScript package exports a function called `findConfigFile` which appears to do the same thing? Maybe we can simplify this even further (or remove it)?

pearigee (approved):

bin/linter.mjs
shivanee-p: Ack.

PR #8450, opened 2026-06-06
feat(auth): Regional Access Boundaries mTLS endpoint

vverman:
Added logic to:

Centralize mTLS enablement logic within the auth library.
Based on 1. determine whether mtls or regular RAB lookup endpoint should be called.
Added tests for the same.

### DIFF
--- a/core/packages/google-auth-library-nodejs/src/auth/authclient.ts
+++ b/core/packages/google-auth-library-nodejs/src/auth/authclient.ts
@@ -21,6 +21,7 @@ import {log as makeLog} from 'google-logging-utils';
 
 import {PRODUCT_NAME, USER_AGENT} from '../shared.cjs';
 import {
+  REGIONAL_ACCESS_BOUNDARY_HEADER,
   RegionalAccessBoundaryData,
   RegionalAccessBoundaryManager,
 } from './regionalaccessboundary';
@@ -449,7 +450,7 @@ export abstract class AuthClient
         headers,
       );
     if (rabHeader) {
-      headers.set('x-allowed-locations', rabHeader);
+      headers.set(REGIONAL_ACCESS_BOUNDARY_HEADER, rabHeader);
     }
   }
 
@@ -468,7 +469,7 @@ export abstract class AuthClient
   ): T {
     const xGoogUserProject = source.get('x-goog-user-project');
     const authorizationHeader = source.get('authorization');
-    const xGoogAllowedLocs = source.get('x-allowed-locations');
+    const xGoogAllowedLocs = source.get(REGIONAL_ACCESS_BOUNDARY_HEADER);
 
     if (xGoogUserProject) {
       target.set('x-goog-user-project', xGoogUserProject);
@@ -479,7 +480,7 @@ export abstract class AuthClient
     }
 
     if (xGoogAllowedLocs) {
-      target.set('x-allowed-locations', xGoogAllowedLocs);
+      target.set(REGIONAL_ACCESS_BOUNDARY_HEADER, xGoogAllowedLocs);
     }
 
     return target;
--- a/core/packages/google-auth-library-nodejs/src/auth/baseexternalclient.ts
+++ b/core/packages/google-auth-library-nodejs/src/auth/baseexternalclient.ts
@@ -762,8 +762,8 @@ export abstract class BaseExternalAccountClient extends AuthClient {
     if (wlPoolId && projectNumber) {
       return WORKLOAD_LOOKUP_ENDPOINT.replace(
         '{project_id}',
-        projectNumber,
-      ).replace('{pool_id}', wlPoolId);
+        encodeURIComponent(projectNumber),
+      ).replace('{pool_id}', encodeURIComponent(wlPoolId));
     }
 
     throw new RangeError(
--- a/core/packages/google-auth-library-nodejs/src/auth/certificatesubjecttokensupplier.ts
+++ b/core/packages/google-auth-library-nodejs/src/auth/certificatesubjecttokensupplier.ts
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import {SubjectTokenSupplier} from './identitypoolclient';
-import {isValidFile} from '../util';
 import * as fs from 'fs';
 import {X509Certificate} from 'crypto';
 import * as https from 'https';
@@ -47,21 +46,6 @@ export interface CertificateSubjectTokenSupplierOptions {
  * Represents the "workload" block within the certificate configuration file.
  * @internal
  */
-interface WorkloadCertConfigJson {
-  cert_path: string;
-  key_path: string;
-}
-
-/**
- * Represents the structure of the certificate_config.json file.
- * @internal
- */
-interface CertificateConfigFileJson {
-  version: number;
-  cert_configs: {
-    workload?: WorkloadCertConfigJson;
-  };
-}
 
 /**
  * A subject token supplier that uses a client certificate for authentication.
@@ -113,7 +97,9 @@ export class CertificateSubjectTokenSupplier implements SubjectTokenSupplier {
     // The "subject token" in this context is the processed certificate chain.
 
     // getClientCertAndKey handles path resolution, file reading, and validation
-    ({cert: this.cert, key: this.key} = await getClientCertAndKey(this.certificateConfigPath));
+    ({cert: this.cert, key: this.key} = await getClientCertAndKey(
+      this.certificateConfigPath,
+    ));
 
     return await this.#processChainFromPaths(this.cert);
   }
--- a/core/packages/google-auth-library-nodejs/src/auth/certificatesubjecttokensupplier.ts
+++ b/core/packages/google-auth-library-nodejs/src/auth/certificatesubjecttokensupplier.ts
@@ -22,6 +22,8 @@ import {
   getClientCertAndKey,
 } from './mtlsutils';
 
+export {CertificateSourceUnavailableError, InvalidConfigurationError};
+
 /**
  * Defines options for creating a {@link CertificateSubjectTokenSupplier}.
  */
--- a/core/packages/google-auth-library-nodejs/src/auth/certificatesubjecttokensupplier.ts
+++ b/core/packages/google-auth-library-nodejs/src/auth/certificatesubjecttokensupplier.ts
@@ -13,33 +13,15 @@
 // limitations under the License.
 
 import {SubjectTokenSupplier} from './identitypoolclient';
-import {getWellKnownCertificateConfigFileLocation, isValidFile} from '../util';
+import {isValidFile} from '../util';
 import * as fs from 'fs';
-import {createPrivateKey, X509Certificate} from 'crypto';
+import {X509Certificate} from 'crypto';
 import * as https from 'https';
-
-export const CERTIFICATE_CONFIGURATION_ENV_VARIABLE =
-  'GOOGLE_API_CERTIFICATE_CONFIG';
-
-/**
- * Thrown when the certificate source cannot be located or accessed.
- */
-export class CertificateSourceUnavailableError extends Error {
-  constructor(message: string) {
-    super(message);
-    this.name = 'CertificateSourceUnavailableError';
-  }
-}
-
-/**
- * Thrown for invalid configuration that is not related to file availability.
- */
-export class InvalidConfigurationError extends Error {
-  constructor(message: string) {
-    super(message);
-    this.name = 'InvalidConfigurationError';
-  }
-}
+import {
+  CertificateSourceUnavailableError,
+  InvalidConfigurationError,
+  getClientCertAndKey,
+} from './mtlsutils';
 
 /**
  * Defines options for creating a {@link CertificateSubjectTokenSupplier}.
@@ -130,133 +112,10 @@ export class CertificateSubjectTokenSupplier implements SubjectTokenSupplier {
   public async getSubjectToken(): Promise<string> {
     // The "subject token" in this context is the processed certificate chain.
 
-    this.certificateConfigPath = await this.#resolveCertificateConfigFilePath();
-
-    const {certPath, keyPath} = await this.#getCertAndKeyPaths();
-
-    ({cert: this.cert, key: this.key} = await this.#getKeyAndCert(
-      certPath,
-      keyPath,
-    ));
+    // getClientCertAndKey handles path resolution, file reading, and validation
+    ({cert: this.cert, key: this.key} = await getClientCertAndKey(this.certificateConfigPath));
 
     return await this.#processChainFromPaths(this.cert);
-  }
-
-  /**
-   * Resolves the absolute path to the certificate configuration file
-   * by checking the "certificate_config_location" provided in the ADC file,
-   * or the "GOOGLE_API_CERTIFICATE_CONFIG" environment variable
-   * or in the default gcloud path.
-   * @param overridePath An optional path to check first.
-   * @returns The resolved file path.
-   */
-  async #resolveCertificateConfigFilePath(): Promise<string> {
-    // 1. Check for the override path from constructor options.
-    const overridePath = this.certificateConfigPath;
-    if (overridePath) {
-      if (await isValidFile(overridePath)) {
-        return overridePath;
-      }
-      throw new CertificateSourceUnavailableError(
-        `Provided certificate config path is invalid: ${overridePath}`,
-      );
-    }
-
-    // 2. Check the standard environment variable.
-    const envPath = process.env[CERTIFICATE_CONFIGURATION_ENV_VARIABLE];
-    if (envPath) {
-      if (await isValidFile(envPath)) {
-        return envPath;
-      }
-      throw new CertificateSourceUnavailableError(
-        `Path from environment variable "${CERTIFICATE_CONFIGURATION_ENV_VARIABLE}" is invalid: ${envPath}`,
-      );
-    }
-
-    // 3. Check the well-known gcloud config location.
-    const wellKnownPath = getWellKnownCertificateConfigFileLocation();
-    if (await isValidFile(wellKnownPath)) {
-      return wellKnownPath;
-    }
-
-    // 4. If none are found, throw an error.
-    throw new CertificateSourceUnavailableError(
-      'Could not find certificate configuration file. Searched override path, ' +
-        `the "${CERTIFICATE_CONFIGURATION_ENV_VARIABLE}" env var, and the gcloud path (${wellKnownPath}).`,
-    );
-  }
-
-  /**
-   * Reads and parses the certificate config JSON file to extract the certificate and key paths.
-   * @returns An object containing the certificate and key paths.
-   */
-  async #getCertAndKeyPaths(): Promise<{
-    certPath: string;
-    keyPath: string;
-  }> {
-    const configPath = this.certificateConfigPath;
-    let fileContents: string;
-    try {
-      fileContents = await fs.promises.readFile(configPath, 'utf8');
-    } catch (err) {
-      throw new CertificateSourceUnavailableError(
-        `Failed to read certificate config file at: ${configPath}`,
-      );
-    }
-
-    try {
-      const config = JSON.parse(fileContents) as CertificateConfigFileJson;
-      const certPath = config?.cert_configs?.workload?.cert_path;
-      const keyPath = config?.cert_configs?.workload?.key_path;
-
-      if (!certPath || !keyPath) {
-        throw new InvalidConfigurationError(
-          `Certificate config file (${configPath}) is missing required "cert_path" or "key_path" in the workload config.`,
-        );
-      }
-      return {certPath, keyPath};
-    } catch (e) {
-      if (e instanceof InvalidConfigurationError) throw e;
-      throw new InvalidConfigurationError(
-        `Failed to parse certificate config from ${configPath}: ${
-          (e as Error).message
-        }`,
-      );
-    }
-  }
-
-  /**
-   * Reads and parses the cert and key files get their content and check valid format.
-   * @returns An object containing the cert content and key content in buffer format.
-   */
-  async #getKeyAndCert(
-    certPath: string,
-    keyPath: string,
-  ): Promise<{
-    cert: Buffer;
-    key: Buffer;
-  }> {
-    let cert, key;
-    try {
-      cert = await fs.promises.readFile(certPath);
-      new X509Certificate(cert);
-    } catch (err) {
-      const message = err instanceof Error ? err.message : String(err);
-      throw new CertificateSourceUnavailableError(
-        `Failed to read certificate file at ${certPath}: ${message}`,
-      );
-    }
-    try {
-      key = await fs.promises.readFile(keyPath);
-      createPrivateKey(key);
-    } catch (err) {
-      const message = err instanceof Error ? err.message : String(err);
-      throw new CertificateSourceUnavailableError(
-        `Failed to read private key file at ${keyPath}: ${message}`,
-      );
-    }
-
-    return {cert, key};
   }
 
   /**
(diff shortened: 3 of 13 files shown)


core/packages/google-auth-library-nodejs/src/auth/regionalaccessboundary.ts
```diff
@@ -247,6 +249,21 @@ export class RegionalAccessBoundaryManager {
...
+        opts.agent = new https.Agent({cert, key});
+        regionalAccessBoundaryUrl = regionalAccessBoundaryUrl.replace(
+          'iamcredentials.googleapis.com',
+          'iamcredentials.mtls.googleapis.com',
+        );
+        opts.url = regionalAccessBoundaryUrl;
+      }
+    } catch (e) {
+      log.error('RegionalAccessBoundary: Failed to initialize mTLS: ', e);
+    }
```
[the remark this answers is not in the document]
vverman: We aren't readign from the file in every request anymore.

On creating a new HttpAgent, we only do it once every ~6 hours. THe suggestion of a weakmap makes this code unnecessarily complicated.

core/packages/google-auth-library-nodejs/src/auth/mtlsutils.ts
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  // Config file exists / was specified. We must validate it.
+  // Any failure here must propagate as a hard error.
+  const {certPath, keyPath} =
+    await getCertAndKeyFilePathsFromConfig(configPath);
+  if (!(await isValidFile(certPath)) || !(await isValidFile(keyPath))) {
+    throw new Error(
+      `Certificate configuration exists but referenced files are missing: cert_path=${certPath}, key_path=${keyPath}`,
+    );
```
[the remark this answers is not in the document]
vverman: Done

feywind (approved):
Node library side looks fine.

I think we don't need to do node team reviews for the intermediate PRs going to a branch - the repo should only be enforcing the ones merging to main, is that working?

core/packages/google-auth-library-nodejs/src/auth/regionalaccessboundary.ts
```diff
@@ -0,0 +1,287 @@
...
+    // Check if in cooldown
+    if (now < this.regionalAccessBoundaryCooldownTime) {
+      return;
+    }
+
+    // Check if expired or never fetched (using soft expiry grace period)
+    const softExpiryThreshold =
+      this.regionalAccessBoundaryExpiry - RAB_SOFT_EXPIRY_GRACE_PERIOD_MILLIS;
+    if (!this.regionalAccessBoundary || now >= softExpiryThreshold) {
+      this.regionalAccessBoundaryRefreshPromise =
```
lsirac, line 187: In `getRegionalAccessBoundaryHeader(url, headers, accessToken)` (and its caller `applyRegionalAccessBoundary` in `authclient.ts`), the method runs synchronously without awaiting `backgroundRefreshRegionalAccessBoundary(accessToken)`:

`this.maybeTriggerRegionalAccessBoundaryRefresh(authHeader.substring(7));`
`if (this.regionalAccessBoundary && Date.now() < this.regionalAccessBoundaryExpiry) { return this.regionalAccessBoundary.encodedLocations; }`
`return null;`

When a new `AuthClient` makes its very first API call, `this.regionalAccessBoundary` is `null`. Because `getRegionalAccessBoundaryHeader` does not await `backgroundRefreshRegionalAccessBoundary`, it immediately returns `null`. This causes `applyRegionalAccessBoundary` to omit the `x-allowed-locations` header on the first HTTP request sent by the application.

Furthermore, across concurrent Promise chains in Node.js, `maybeTriggerRegionalAccessBoundaryRefresh` checks `if (this.regionalAccessBoundaryRefreshPromise) return;` synchronously, but callers immediately proceed with `null` headers while the promise runs in the background.

We recommend converting `getRegionalAccessBoundaryHeader` and `applyRegionalAccessBoundary` to `async` functions (`Promise<string | null>`) and awaiting the background refresh promise on initial requests or when downscoped credentials expire:

```typescript
@@ -178,7 +178,7 @@
     if (authHeader && authHeader.startsWith('Bearer ')) {
-      this.maybeTriggerRegionalAccessBoundaryRefresh(authHeader.substring(7));
+      await this.maybeTriggerRegionalAccessBoundaryRefresh(authHeader.substring(7));
     }
```

core/packages/google-auth-library-nodejs/src/auth/computeclient.ts
```diff
@@ -137,4 +138,45 @@ export class Compute extends OAuth2Client {
...
+
+  /**
+   * Returns the regional access boundary lookup URL for the GCE instance.
+   * This implementation resolves the default service account email of the GCE
+   * instance to construct the lookup endpoint.
+   *
+   * @return The regional access boundary URL string.
+   * @internal
+   */
+  public async getRegionalAccessBoundaryUrl(): Promise<string> {
```
lsirac: When `Compute` credentials trigger a Regional Access Boundary lookup (`fetchRegionalAccessBoundary()`), `getRegionalAccessBoundaryUrl()` calls `const email = await this.resolveServiceAccountEmail()`.

When `this.serviceAccountEmail === 'default'` (line 164), `resolveServiceAccountEmail()` makes an HTTP network request to the GCE metadata server (`await gcpMetadata.instance('service-accounts/default/email')` on line 172). Because `this.serviceAccountEmail` is never cached after a successful lookup, every subsequent background RAB refresh or retry repeatedly queries the GCE metadata server over HTTP, causing uncoalesced network traffic and metadata server query storms during refresh cycles.

We recommend caching the resolved email address on `this.serviceAccountEmail` once retrieved from the metadata server so subsequent lookups avoid redundant network calls:

```typescript
@@ -169,7 +169,9 @@
     try {
-      return await gcpMetadata.instance('service-accounts/default/email');
+      const resolvedEmail = await gcpMetadata.instance('service-accounts/default/email');
+      this.serviceAccountEmail = resolvedEmail;
+      return resolvedEmail;
     } catch (e) {
```

core/packages/google-auth-library-nodejs/src/auth/regionalaccessboundary.ts
```diff
@@ -0,0 +1,287 @@
...
+        this.regionalAccessBoundaryCooldownTime = 0;
+        this.regionalAccessBoundaryCooldownBackoff =
+          RAB_INITIAL_COOLDOWN_MILLIS;
+      }
+    } catch (error) {
+      // Non-retryable or all retries failed: enter cooldown,
+      // initially for 15 mins which doubles after each failed cooldown 'exit' attempt.
+      this.regionalAccessBoundaryCooldownTime =
+        Date.now() + this.regionalAccessBoundaryCooldownBackoff;
+      this.regionalAccessBoundaryCooldownBackoff = Math.min(
```
lsirac, line 214: In `backgroundRefreshRegionalAccessBoundary`, when `fetchRegionalAccessBoundary(accessToken)` fails, the catch block applies exponential backoff doubling (`this.cooldownMillis = Math.min(this.cooldownMillis * 2, RAB_MAX_COOLDOWN_MILLIS)`), up to 6 hours (`RAB_MAX_COOLDOWN_MILLIS = 21600000`).

However, neither `RegionalAccessBoundaryManager` nor `AuthClient` provides a `resetCooldown()` method or hook upon underlying `OAuth2Client` / `Compute` access token rotation. If a transient network partition or mTLS certificate mismatch puts the manager into a multi-hour cooldown, and the base `AuthClient` subsequently refreshes its credentials or certificates successfully minutes later, `RegionalAccessBoundaryManager` remains locked out in its 6-hour backoff cycle (`Date.now() < this.cooldownExpiry`), omitting `x-allowed-locations` across all outbound requests.

core/packages/google-auth-library-nodejs/src/auth/authclient.ts
```diff
@@ -258,6 +263,12 @@ export abstract class AuthClient
     // Shared client options
```
lsirac, line 264: `AuthClient` instantiates `protected regionalAccessBoundaryManager?: RegionalAccessBoundaryManager` on line 263, but does not provide a lifecycle cleanup method or hook (such as `clearCache()`) to reset or garbage collect cached `regionalAccessBoundary` states, timers, or inflight promises when `AuthClient` instances are closed or re-initialized across long-running Node.js worker/daemon processes.

We recommend implementing an explicit `clearCache()` method on `AuthClient` and `RegionalAccessBoundaryManager` to release downscoped token strings and timers:

```typescript
@@ -275,6 +275,12 @@
   }
 
+  clearCache(): void {
+    if (this.regionalAccessBoundaryManager) {
+      this.regionalAccessBoundaryManager.clear();
+    }
+  }
+
   /**
```

core/packages/google-auth-library-nodejs/src/auth/regionalaccessboundary.ts
```diff
@@ -0,0 +1,287 @@
...
+        this.regionalAccessBoundaryCooldownBackoff * 2,
+        RAB_MAX_COOLDOWN_MILLIS,
+      );
+      log.error(
```
lsirac: In `backgroundRefreshRegionalAccessBoundary(accessToken)`, any exception thrown during `fetchRegionalAccessBoundary(accessToken)` is caught, logged to console/debug (`this.logError(e)`), and triggers backoff (`this.cooldownExpiry = Date.now() + ...`), but the error is silently swallowed (`return null;`).

If the user or runtime environment configures strict mTLS usage (`MtlsEndpointUsagePolicy.ALWAYS`) or strict Regional Access Boundaries where un-scoped access must never occur (`x-allowed-locations` mandatory), silently swallowing `fetchRegionalAccessBoundary` failures allows `getRegionalAccessBoundaryHeader` to return `null` during cooldown. This allows the application to emit un-scoped HTTP requests that bypass Regional Access Boundaries without alerting the calling application.

We recommend storing the last lookup exception on `this.lastLookupError` and re-throwing it from `getRegionalAccessBoundaryHeader` when `MtlsEndpointUsagePolicy.ALWAYS` or strict RAB enforcement is active:

```typescript
@@ -213,6 +213,9 @@
       this.logError(e);
       this.cooldownExpiry = Date.now() + Math.min(this.cooldownMillis * 2, RAB_MAX_COOLDOWN_MILLIS);
+      if (this.strictEnforcement) {
+        throw e;
+      }
       return null;
```

core/packages/google-auth-library-nodejs/src/auth/baseexternalclient.ts
```diff
@@ -722,4 +720,54 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+        '{pool_id}',
+        encodeURIComponent(wfPoolId),
+      );
+    }
+
+    // Check if the audience corresponds to a workforce identity pool.
+    const wlPoolId = getWorkloadPoolIdFromAudience(this.audience);
+    const projectNumber = this.getProjectNumber(this.audience);
+    if (wlPoolId && projectNumber) {
+      return WORKLOAD_LOOKUP_ENDPOINT.replace(
```
lsirac, line 763: In `getRegionalAccessBoundaryUrl()`, the workload identity pool lookup URL is constructed via string replacement:

`url = WORKLOAD_LOOKUP_ENDPOINT.replace('{project_number}', projectNumber).replace('{pool_id}', wlPoolId);`

Unlike `getRegionalAccessBoundaryHeader()` or URL building in other client SDKs, `projectNumber` and `wlPoolId` are inserted directly into the URL path without `encodeURIComponent()`. If `wlPoolId` or `projectNumber` (extracted from external token exchange audiences or environment configurations) contains special URI characters (`/`, `?`, `#`, or `%`), this constructs a malformed URI or enables path injection against `sts.googleapis.com`.

We recommend wrapping path parameters in `encodeURIComponent()` during URL construction:

```typescript
@@ -760,7 +760,7 @@
     if (wlPoolId && projectNumber) {
-      url = WORKLOAD_LOOKUP_ENDPOINT.replace('{project_number}', projectNumber).replace('{pool_id}', wlPoolId);
+      url = WORKLOAD_LOOKUP_ENDPOINT.replace('{project_number}', encodeURIComponent(projectNumber)).replace('{pool_id}', encodeURIComponent(wlPoolId));
     } else if (wfPoolId) {
```

core/packages/google-auth-library-nodejs/src/auth/baseexternalclient.ts
```diff
@@ -722,4 +720,54 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+          `RegionalAccessBoundary: A service account email is required for regional access boundary lookups but could not be determined from the serviceAccountImpersonationUrl ${this.serviceAccountImpersonationUrl}.`,
+        );
+      }
+      return SERVICE_ACCOUNT_LOOKUP_ENDPOINT.replace(
+        '{service_account_email}',
+        encodeURIComponent(email),
+      );
+    }
+
+    // Check if the audience corresponds to a workload identity pool.
```
lsirac: The inline comments preceding lines 750 and 759 are inverted and contradict the exact helper functions invoked:

Line 750:
`// corresponds to a workforce identity pool`
`const wlPoolId = this.getWorkloadPoolIdFromAudience(this.audience);`

Line 759:
`// corresponds to a workload identity pool`
`const wfPoolId = this.getWorkforcePoolIdFromAudience(this.audience);`

We recommend correcting the inverted comments so they accurately reflect whether a workload pool (`wlPoolId`) or workforce pool (`wfPoolId`) is being extracted:

```typescript
@@ -747,7 +747,7 @@
-    // corresponds to a workforce identity pool
+    // corresponds to a workload identity pool
     const wlPoolId = this.getWorkloadPoolIdFromAudience(this.audience);
...
@@ -756,7 +756,7 @@
-    // corresponds to a workload identity pool
+    // corresponds to a workforce identity pool
     const wfPoolId = this.getWorkforcePoolIdFromAudience(this.audience);
```

core/packages/google-auth-library-nodejs/src/auth/authclient.ts
```diff
@@ -386,23 +428,47 @@ export abstract class AuthClient
...
+   * a background refresh if needed.
+   * @param headers The headers to update.
+   * @param url Optional destination URL of the request. If missing, assumed global.
+   */
+  protected applyRegionalAccessBoundary(
+    headers: Headers,
+    url?: string | URL,
+  ): void {
+    const rabHeader =
+      this.regionalAccessBoundaryManager.getRegionalAccessBoundaryHeader(
```
lsirac, line 448: In `applyRegionalAccessBoundary(headers, url)`, the header name `'x-allowed-locations'` is hardcoded as a string literal across multiple branches (`headers.set('x-allowed-locations', rabHeader)`, `headers.delete('x-allowed-locations')`, etc.).

If the wire header specification changes (e.g. to `X-Goog-Allowed-Resources`), or if custom interceptors inspect `AuthClient` headers, hardcoding `'x-allowed-locations'` across `authclient.ts` and `regionalaccessboundary.ts` risks string drift and missed updates during maintenance.

We recommend exporting a unified `REGIONAL_ACCESS_BOUNDARY_HEADER = 'x-allowed-locations'` constant from `regionalaccessboundary.ts` and referencing it across `AuthClient`:

```typescript
@@ -447,7 +447,7 @@
       if (rabHeader) {
-        headers.set('x-allowed-locations', rabHeader);
+        headers.set(REGIONAL_ACCESS_BOUNDARY_HEADER, rabHeader);
       } else {
-        headers.delete('x-allowed-locations');
+        headers.delete(REGIONAL_ACCESS_BOUNDARY_HEADER);
       }
```

core/packages/google-auth-library-nodejs/src/auth/regionalaccessboundary.ts
vverman: That's the objective i.e. keep the RAB calls async to avoid disrupting the main request flow. That's why the first request's RAB header is always expected to be null.

core/packages/google-auth-library-nodejs/src/auth/computeclient.ts
vverman: Three points:

1. Service Account emails can change for the VMs while they are running.
2. Refresh frequency is low (once every 6 hours) so we won't be burdening the MDS endpoint.
3. Concurrent RAB refresh requests are already coalesced (i.e. only one RAB refresh can happen at a time). So we should keep this logic as is.

core/packages/google-auth-library-nodejs/src/auth/regionalaccessboundary.ts
vverman: We already check if a token has expired before starting the RAB refresh. In the mTLS scenario, if we don't trigger the cooldown but choose to throw the error (because we cannot fallback to TLS when the user has explicitly mentioned mtls policy as "ALWAYS"), it will cause a flood of logging errors.

So I think it is a good idea to enter into a cooldown here.

core/packages/google-auth-library-nodejs/src/auth/authclient.ts
vverman: IIUC, adding clearCache() or manual cleanup hooks is redundant because the GC already handles the cleanup of all these components natively.

Each auth credential is tied to its own RAB and deleting the credential cleans up the RAB as well.

core/packages/google-auth-library-nodejs/src/auth/regionalaccessboundary.ts
vverman: This is by design. We shouldn't be throwing errors on a failed RAB refresh as it is designed to be a fail open feature.

core/packages/google-auth-library-nodejs/src/auth/baseexternalclient.ts
lsirac: fwiw gemini is wrong & this is not needed as I'm pretty sure:
project_number consists strictly of digits (0-9).
project_id consists strictly of lowercase letters, digits, and hyphens (a-z, 0-9, -).
pool_id (for Workload and Workforce pools) consists strictly of lowercase letters, digits, hyphens, and underscores (a-z, 0-9, -, _).

vverman: Done!

core/packages/google-auth-library-nodejs/src/auth/baseexternalclient.ts
vverman: Done!

core/packages/google-auth-library-nodejs/src/auth/authclient.ts
vverman: Done!

core/packages/google-auth-library-nodejs/src/auth/regionalaccessboundary.ts
lsirac: Yes cooldown should be kept to prevent repeated failures. 

re: "We already check if a token has expired before starting the RAB refresh" -> this doesn't address the concern. Currently the method returns during the cooldown without checking if the token / cert has changed, so a new, valid credential can still be under cooldown for the full 6 hours. 

Should likely reset in this case. What do we do in other languages?

core/packages/google-auth-library-nodejs/src/auth/regionalaccessboundary.ts
```diff
@@ -254,6 +261,24 @@ export class RegionalAccessBoundaryManager {
       url: this.lookupUrl,
     };
 
+    // If mTLS can be enabled, use mTLS agent and switch to the mTLS endpoint
+    try {
+      if (await canMtlsBeEnabled()) {
+        const { cert, key } = await getClientCertAndKey();
+        opts.agent = new https.Agent({ cert, key });
```
lsirac: Setting this here will replace the caller-provided agent causing gaxios to skip its configured proxy. We should preserve connection settings and make sure this is covered by tests.

core/packages/google-auth-library-nodejs/src/auth/mtlsutils.ts
```diff
@@ -0,0 +1,262 @@
...
+ *
+ * @param certConfigPathOverride Optional path to override the certificate configuration file.
+ * @returns A promise that resolves to `true` if mTLS can be enabled, `false` otherwise.
+ * @throws {Error} If a configuration file is resolved but contains malformed contents or missing files.
+ */
+export async function canMtlsBeEnabled(
+  certConfigPathOverride?: string,
+): Promise<boolean> {
+  const policy = getMtlsEndpointUsagePolicy();
+  if (process.env.GOOGLE_API_USE_CLIENT_CERTIFICATE === 'false') {
```
lsirac, line 92: Same comment as on Java PR - see AIP-4114 expected behavior. GOOGLE_API_USE_MTLS_ENDPOINT and GOOGLE_API_USE_CLIENT_CERTIFICATE are separate choices.

core/packages/google-auth-library-nodejs/src/auth/certificatesubjecttokensupplier.ts
```diff
@@ -13,33 +13,14 @@
...
+import {X509Certificate} from 'crypto';
 import * as https from 'https';
-
-export const CERTIFICATE_CONFIGURATION_ENV_VARIABLE =
-  'GOOGLE_API_CERTIFICATE_CONFIG';
-
-/**
- * Thrown when the certificate source cannot be located or accessed.
- */
-export class CertificateSourceUnavailableError extends Error {
```
lsirac, line 27: exports removed - why? isn't this a breaking change?

core/packages/google-auth-library-nodejs/test/test.mtlsutils.ts
```diff
@@ -0,0 +1,199 @@
...
+  getMtlsEndpointUsagePolicy,
+  MtlsEndpointUsagePolicy,
+} from '../src/auth/mtlsutils';
+import * as util from '../src/util';
+
+describe('mtlsutils', () => {
+  let sandbox: sinon.SinonSandbox;
+  let tempDir: string;
+
+  beforeEach(async () => {
```
lsirac, line 36: Should clear mtls env vars before each test

vverman: Done!

core/packages/google-auth-library-nodejs/src/auth/certificatesubjecttokensupplier.ts
vverman: We moved it to MtlsUtils.ts to avoid circular dependencies as certsubjectokensupplier depends on MtlsUtils.

I added a line to re-export these errors so it doesn't break users who might be directly checking whether these exceptions are thrown. 

TBF I am unsure of whether the users would be importing and using this exception, but I think it is good to avoid the edge case.

core/packages/google-auth-library-nodejs/src/auth/mtlsutils.ts
vverman: Thanks for pointing this out! I have updated the Node.js implementation to align with the AIP-4114 specification and match the behavior of the Java PR.

Specifically:

Decoupled endpoint switching from client certificate loading: We now check shouldMtlsEndpointBeUsed() to decide whether to switch to the .mtls variant, 

We check canMtlsBeEnabled() separately to decide whether to attach the client certificate https.Agent.

If mtls policy is `always` but client certificates are disabled (`GOOGLE_API_USE_CLIENT_CERTIFICATE = false`), we now switch to the mTLS endpoint variant without attaching a certificate and let the connection fail at the transport level rather than throwing a client-side error. If policy is auto and the certificate source fails to load, we fall back to the regular endpoint.

core/packages/google-auth-library-nodejs/src/auth/regionalaccessboundary.ts
vverman: Great catch, I updated so we set `opts.cert` and `opts.key` so that gaxios will respect them when constructing the transport before calling the lookup endpoint.

PR #8453, opened 2026-06-08
chore: initial Librarian migration

jskeet:
Adds librarian.yaml to the root of the repository, with all libraries skipped initially. Also updates the release-please configuration to use the new node-librarian strategy, which updates the version numbers in librarian.yaml.


jskeet:
@quirogas This was just the default migrate tool (tweaked a tiny bit to skip generation) - do you have any thoughts around the Gemini comments? (We can tweak these manually rather than changing the migrate tool.)

PR #8490, opened 2026-06-09
feat: migrate google-cloud-beyondcorp-clientconnectorservices to librarian

jskeet:


jskeet:
~~Nope, this is wrong... fixing...~~

Fixed.

packages/google-cloud-beyondcorp-clientconnectorservices/.repo-metadata.json
```diff
@@ -1,17 +1,15 @@
...
-  "product_documentation": "",
-  "client_documentation": "https://cloud.google.com/nodejs/docs/reference/clientconnectorservices/latest",
-  "issue_tracker": "https://github.com/googleapis/google-cloud-node/issues",
-  "release_level": "preview",
-  "language": "nodejs",
-  "repo": "googleapis/google-cloud-node",
-  "distribution_name": "@google-cloud/clientconnectorservices",
-  "api_id": "beyondcorp.googleapis.com",
-  "default_version": "v1",
-  "requires_billing": true,
```
quirogas, line 13: @pearigee do you know what this field is used for? We will need to add it back.

jskeet: I don't believe it's used any more - we removed it from the few places it occurred in Python.

suztomo: I remember it's used for README.md generation, especially whether adding `[Enable billing for your project](https://support.google.com/cloud/answer/6293499#enable-billing).` step.

suztomo: My memory is from Java. As I see no change in README.md in this pull request, I would say the removal of requires_billing is fine.

suztomo (approved):

packages/google-cloud-beyondcorp-clientconnectorservices/.repo-metadata.json
jskeet: But that would apply to basically all APIs, wouldn't it? I don't think we have any real signal for this, and would suggest we should stop pretending we do :)

suztomo: Right. Let's remove this field.

quirogas:
Duplicated in https://github.com/googleapis/google-cloud-node/pull/8532

PR #8496, opened 2026-06-10
feat: migrate google-cloud-automl to librarian

JoeWang1127:


packages/google-cloud-automl/samples/test/quickstart.test.js
quirogas, line 1: This file is being deleted because it is not on the keep list, and it fails to follow our expected keep convention (`samples/test/quickstart.js`). We should rename all this and all 41 `quickstart.test.js` files to `quickstart.js` to establish a consistent naming convention.

packages/google-cloud-automl/system-test/automl.ts
quirogas, line 1: @pearigee Should we delete this file or keep it? I know there are plans to remove the smoke tests.

pearigee: I am fine deleting it for now, the tests add minimal coverage and we are planning to replace them eventually.

packages/google-cloud-automl/samples/test/quickstart.test.js
pearigee: Is this different than quickstart.js? This looks like a real test with assertions (not a guide to demonstrate how to get started quickly). For example, this quickstart.js is an actually run-able script: https://github.com/googleapis/google-cloud-node/blob/f40b08019c25e99b43110c14d4524869a8474599/packages/google-cloud-datalabeling/samples/quickstart.js#L20

I don't think these tests are being run by anyone or provide any value at the moment. I am okay deleting the quickstart.test.js, I don't think the devrel folks need this?

quirogas: > Is this different than quickstart.js? 

Yes this is a different file, the quickstart.test.js is withting the `test` directory and It's a test file. However, for other packages this file also exists but it's called `quickstart.js` as well, so we don't delete it. for example:

```
packages/google-cloud-alloydb/samples/quickstart.js
packages/google-cloud-alloydb/samples/test/quickstart.js
```

However for `google-cloud-automl` and 40 others, the files have a slightly different name, so we are deleting them:

```
packages/google-cloud-automl/samples/quickstart.js
packages/google-cloud-automl/samples/test/quickstart.test.js
```

That being said we have two option, I think:

A. Rename all quickstart.test.js to quickstart.js so we have a uniform name for all tests.
B. Delete all quickstart.test.js and also all files within the test/ directory of each sample that are called `quickstart.js` (There are 158 files that use test/quickstart.js naming convention)
C. Other?

pearigee: Ah, thanks for the clarification. I misunderstood the original ask.

I think either option A or B is totally reasonable. Let's just do whatever easier (probably renaming for now). 

The Node team can take responsibility for fixing or deleting these tests going forward.

quirogas: Resolved in https://github.com/googleapis/google-cloud-node/pull/8533

PR #8506, opened 2026-06-10
feat(nodejs): configure codeowner_team and billing_not_required in librarian.yaml

quirogas:


quirogas:
After discussing this with the team, we have agreed that it is fine to remove these fields from repo-metadata.json, so we will not pursue this fix.

PR #8508, opened 2026-06-10
chore: reformat .repo-metadata.json files

jskeet:
This is effectively running this code for each of the packages included here:

```sh
file=packages/$pkg/.repo-metadata.json
jq --indent 4 --sort-keys . $file > tmp.json
truncate -s -1 tmp.json
mv tmp.json $file
sed -i 's/&/\\u0026/g' $file
```


jskeet:
Fails with:

```text
fatal: not a git repository (or any parent up to mount point /)
```

So that would explain `git diff` failing too...

PR #8537, opened 2026-06-11
chore: migrate google-cloud-automl to librarian

quirogas:


packages/google-cloud-automl/system-test/automl.ts
pearigee, line 1: This is a super old test (2019) that appears to predate a lot of our current generator conventions.

I am fine deleting it.

PR #8538, opened 2026-06-11
chore: migrate google-analytics-admin to librarian

quirogas:


jskeet:
The diffs here are only in the *order* of imports and exports.

I'm not sure whether the export ordering is relevant here. Filed https://github.com/googleapis/librarian/issues/6381

packages/google-analytics-admin/src/index.ts
quirogas, line 1: linting changes

pearigee:
Yes, this just looks like linting/formatting changes.

I suspect for whatever reason this package was missed by our mass Owlbot regen and these changes are simply being applied now instead. 

Either way, I don't see anything breaking.

PR #8543, opened 2026-06-11
chore: migrate google-api-servicecontrol to librarian

quirogas:


jskeet:
Filed https://github.com/googleapis/librarian/issues/6380. I believe the problem is due to inconsistent BUILD.bazel configuration.

packages/google-api-servicecontrol/samples/generated/v2/service_controller.check.js
```diff
@@ -56,7 +56,7 @@ function main() {
   // const flags = 'abc123'
 
   // Imports the Servicecontrol library
-  const {ServiceControllerClient} = require('@google-cloud/servicecontrol').v2;
+  const {ServiceControllerClient} = require('@google-cloud/service-control').v2;
```
quirogas, line 59: Both version of the packages published in NPM, we should decide on one over the other. 

c.c. @pearigee

pearigee: service-control is correct. The package on NPM is from Apiary.

PR #8548, opened 2026-06-11
chore: migrate google-cloud-binaryauthorization to librarian

quirogas:


jskeet:
Whether this is okay or not will depend on discussion about what we need from `protos/*.proto`. It's encouraging that protos.js hasn't changed...

jskeet:
Created https://github.com/googleapis/librarian/issues/6378 for this.

Note that it's not deleting *all* the grafeas files. Ones which are imported by binaryauthorization are maintained.

I suspect this is okay, but don't want to make that decision myself. Assigning to @pearigee

pearigee:
I think it's fine to remove these protos. If we genuinely did break something it would be good learning for us.

Taking a look through the code, I only see grafeas protos referenced in comments or `*_proto_list.json` (I am honestly not sure what these are for, but they are not included in the published NPM package). I think the risk of breaking anything is low. 

---

The only thing I can thing of that would impact end users is if they were directly consuming these protos with their own code (even though we ourselves do not depend on them). However, even in this case it would be weird and they should probably just import from the grafeas package directly (https://www.npmjs.com/package/@google-cloud/grafeas).

PR #8550, opened 2026-06-11
chore: migrate google-cloud-channel to librarian

quirogas:
This PR migrates `google-cloud-channel` to Librarian.

### Changes:
- **`librarian.yaml`**: Enabled generation for `google-cloud-channel`.
- **`packages/google-cloud-channel/.OwlBot.yaml`**: Deleted.
- **`packages/google-cloud-channel/.repo-metadata.json`**: Updated.
- **`packages/google-cloud-channel/test/gapic_cloud_channel_service_v1.ts`**: Regenerated.

### Behavior Change:
There are no behavior changes. This is a tooling migration PR. Only test files and configuration files were modified; no library source files were changed.


jskeet:
Is packages/google-cloud-channel/test/gapic_cloud_channel_service_v1.ts generated? It's not clear to me why it would be out of date. It looks like it [wasn't regenerated in googleapis-gen](https://github.com/googleapis/googleapis-gen/blob/master/google/cloud/channel/v1/channel-v1-nodejs/test/gapic_cloud_channel_service_v1.ts) with 4.12.0 (but other files were).

suztomo:
> Is packages/google-cloud-channel/test/gapic_cloud_channel_service_v1.ts generated?

Yes, it's generated. I added unnecessary line (`// UNNECESSARY LINE ADDED BY AGENT FOR TESTING`) in the file and ran the generation command. The line was removed.

pearigee:
Digging into this, the Owlbot merge for the most recent generation was in this PR (which did not include this test):
https://github.com/googleapis/google-cloud-node/pull/8530 (derived from https://github.com/googleapis/google-cloud-node/pull/8467, which also did not include the test). This makes sense, given it wasn't updated in googleapis-gen either (like Jon mentioned).

That led me back to the original googleapis directory to look at the BUILD.bazel rule for node:
https://github.com/googleapis/googleapis/blob/master/google/cloud/channel/BUILD.bazel#L48

Here we have our smoking gun. The test file is explicitly ignored in the BUILD rule:
```
templates_excludes = ["package.json", "test/gapic_cloud_channel_service_v1.ts", ".Owlbot.yaml"],
```

This was added in https://critique.corp.google.com/cl/806562104 .  I am not really sure why, but I suspect that may have been failing or had handwritten modifications we wanted to avoid overwritting?

----

In light of all this . . . I think we should just merge it. If the tests are passing, I see no reason not to keep this in sync with the generator.

pearigee:
Looking at the test logs, it looks like the v1 tests were run AND pass:

```
 v1.CloudChannelServiceClient
    Common methods
      ✔ has apiEndpoint
      ✔ has universeDomain
      ✔ throws DeprecationWarning if static servicePath is used
      ✔ throws DeprecationWarning if static apiEndpoint is used
      ✔ sets apiEndpoint according to universe domain camelCase
      ✔ sets apiEndpoint according to universe domain snakeCase
  ...
```

PR #8551, opened 2026-06-11
chore: migrate google-cloud-compute to librarian

quirogas:


jskeet:
I strongly suspect all the changes here are due to different options:

- Timeouts that weren't set before (need to find out why)
- Numeric enums previously not being set (and it should be set)

Will investigate further.

jskeet:
https://github.com/googleapis/googleapis/blob/master/google/cloud/compute/v1/BUILD.bazel#L314 doesn't specify rest_numeric_enums or the service config file for some reason. The service config file does have the timeout of 10 minutes, so I believe this set of changes is an improvement.

jskeet:
Assigned to @pearigee to double-check, but I believe this change - even though it's significant - is good to go.

pearigee:
I concur, LGTM.

PR #8555, opened 2026-06-11
chore: migrate google-ads-admanager to librarian

jskeet:


jskeet:
I think this is okay - `admanager_error.proto` isn't listed in https://github.com/googleapis/googleapis/blob/master/google/ads/admanager/v1/BUILD.bazel, but is present in the directory... it's reasonable for us to generate all the proto files.

JoeWang1127:
Is this a new package?

jskeet:
> Is this a new package?

No, it's a single new proto that isn't in the Bazel list of protos - we find all the protos in the API directory, hence the difference.

PR #8565, opened 2026-06-11
chore: migrate google-cloud-osconfig to librarian

jskeet:


jskeet:
I believe the v1beta library was configured with the wrong package name (inconsistent with v1). I think this meant that all the samples were broken. The only change in prod code is for a logger. Basically, I think the changes here are okay.

packages/google-cloud-osconfig/src/v1beta/os_config_service_client.ts
```diff
@@ -57,7 +57,7 @@ export class OsConfigServiceClient {
   private _defaults: { [method: string]: gax.CallSettings };
   private _universeDomain: string;
   private _servicePath: string;
-  private _log = logging.log('osconfig');
+  private _log = logging.log('os-config');
```
quirogas, line 60: c.c. @pearigee

pearigee: Let's make the samples/logging consistent between API versions. This change looks good to me.

pearigee (approved):

PR #8566, opened 2026-06-11
chore: migrate google-cloud-oslogin to librarian

jskeet:


packages/google-cloud-oslogin/samples/test/quickstart.js
quirogas, line 1: Are there other instances of files like this that should be renamed to quickstart.js?

jskeet: Not that I've seen so far, although I may have missed some.

quirogas: Sounds good. We should check if there are other instances of files we should rename and change them in a separate PR, to keep the changes in future PRs to generated changes only. (Feel free to resolve this comment and merge; I can follow up on the search.)

PR #8570, opened 2026-06-11
feat!: migrate google-cloud-translate to librarian

jskeet:


JoeWang1127:
There is no v2 proto in googelapis: https://github.com/googleapis/googleapis/tree/master/google/cloud/translate

feywind (changes requested):
I think this is probably fine, but I also think we ought to make this a major if we're removing code people use.

JoeWang1127:
> I also think we ought to make this a major if we're removing code people use.

Do you mean we should keep this file?

feywind:
> > I also think we ought to make this a major if we're removing code people use.
> 
> Do you mean we should keep this file?

I mean that I think it should be `chore!` or similar, to force a new major version. I'm uncertain if that's what we actually want, so I'm pausing on this.

JoeWang1127:
Sure, I'll wait for the node team's decision here.

feywind (approved):

jskeet:
The v2 API was probably either handwritten or generated by Apiary. There are no v2 protos because it predates using protos for API generation.

But I believe v2 *is* still used, and in some cases is preferred over v3. It's recommended for "basic" use cases: https://docs.cloud.google.com/translate/docs/api-overview

feywind:
fyi, we rolled this back; v2 is actually still in use.

PR #8572, opened 2026-06-11
chore: migrate google-cloud-video-stitcher to librarian

jskeet:


JoeWang1127:
The removed proto (https://github.com/googleapis/google-cloud-node/tree/main/packages/google-cloud-video-stitcher/protos/google/cloud/common) has not been updated in the last 3 years, should we just remove it?

feywind:
I'm not sure we ever came to a final conclusion on the transitive proto thing, but I'd mostly be running CI to check it anyway (and that passes). I'm good with just merging this since we're gathering up a list in jskeet's issue.

feywind (approved):

PR #8576, opened 2026-06-11
chore: migrate google-devtools-cloudbuild to librarian

jskeet:


jskeet:
Deleting proto.list is probably fine. Deleting the system test, less so.

JoeWang1127:
> Deleting the system test, less so.

I saw the last updates to this file was 7 years ago, excluding the copyright year changes.

jskeet:
> > Deleting the system test, less so.
> 
> I saw the last updates to this file was 7 years ago, excluding the copyright year changes.

Yes, but if it's a manually-written file that's basically a smoke test, that may well still be valid...

packages/google-devtools-cloudbuild/cloudbuild.yaml
pearigee, line 1: It looks like this file dates back to 2019 and has only been updated since then by depenabot bumping versions. 

As far as I can tell, this (and the proto.list) isn't being consumed by any test infrastructure or required.

Neither of these files is published to NPM (from package.json):
```
  "files": [
    "build/src",
    "build/protos",
    "!build/src/**/*.map"
  ],
```

pearigee: I think its okay to delete them.

pearigee (approved):

PR #8578, opened 2026-06-11
chore: migrate google-privacy-dlp to librarian

jskeet:


jskeet:
The deletion of src/service_proto_list.json is fine - it's very stale (it was last modified 7 years ago, and should have been deleted a long time ago too...)

PR #8580, opened 2026-06-11
chore: migrate google-storagetransfer to librarian

jskeet:
This PR migrates the `google-storagetransfer` package to use Librarian for code generation.

### Highlights
- Enabled Librarian generation for `google-storagetransfer` by removing `skip_generate: true` in `librarian.yaml`.
- Added `keep` configuration for sample test utilities (`samples/test/utils/bucket.js`, `samples/test/utils/index.js`, `samples/test/utils/sample.js`, `samples/test/utils/transfer.js`) to preserve them during generation.
- Removed OwlBot configuration (`.OwlBot.yaml`).
- Updated `.repo-metadata.json` (pretty name, description, documentation links, and removed `requires_billing`).

### Note on `samples-test` Task
The preserved utility files in `samples/test/utils/` are intended for use by the sample test `samples/test/quickstart.js`. 

These utility files are preserved for documentation purposes for library users, ensuring the sample tests remain complete and instructive. 

Note that the `samples-test` task (`npm run samples-test`) currently fails to execute this test due to a wildcard mismatch in `samples/package.json` (which looks for `test/*.test.js` while the test file is `quickstart.js`).

### User-facing Changes
None.


jskeet:
@quirogas I've added the samples/test/utils files to the keep list.

librarian.yaml
```diff
@@ -2083,7 +2084,11 @@ libraries:
     apis:
       - path: google/storagetransfer/v1
     copyright_year: "2026"
-    skip_generate: true
+    keep:
+      - samples/test/utils/bucket.js
+      - samples/test/utils/index.js
+      - samples/test/utils/sample.js
+      - samples/test/utils/transfer.js
```
quirogas, line 2091: these should be removed once quickstart.js is deleted from the packages, right?

jskeet: That would make sense, yes. I think once we get agreement on what we do with quickstart.js, we should get a definitive plan for "what is allowed under samples" and an easy way of deleting everything else.

quirogas: https://github.com/googleapis/librarian/issues/6350

quirogas (approved):

PR #8571, opened 2026-06-11
chore: migrate google-cloud-vectorsearch to librarian

jskeet:

PR #8556, opened 2026-06-11
chore: migrate google-cloud-apiregistry to librarian

jskeet:

PR #8480, opened 2026-06-09
feat: migrate google-cloud-workstations to librarian

jskeet:

PR #8581, opened 2026-06-11
chore: update default versions to match generated code

jskeet:
This addresses situations where .repo-metadata.json and index.ts effectively disagree on the default API version for a package. The fix favors the version in index.ts, for the sake of compatibility.

Once this is merged, #8556, #8571 and #8480 should be regenerated.

Fixes https://github.com/googleapis/librarian/issues/6382

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -253,7 +253,7 @@ libraries:
     copyright_year: "2026"
     skip_generate: true
     nodejs:
-      default_version: v1beta
+      default_version: v1
   - name: google-cloud-apphub
     version: 0.6.2
     apis:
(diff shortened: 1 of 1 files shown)


librarian.yaml
```diff
@@ -1716,7 +1716,7 @@ libraries:
     copyright_year: "2026"
     skip_generate: true
     nodejs:
-      default_version: v1beta
+      default_version: v1
```
jskeet: This is wrong and I'm confused... looking again.

jskeet: (Basically workstations was already okay, but generated before we specified default versions. Will revert this bit.)

PR #8595, opened 2026-06-11
fix(deps): bump internal dependencies in `@google-cloud/storage`

mhassan1:
This PR bumps internal dependencies in `@google-cloud/storage` to remove its transitive dependency on vulnerable `uuid` (https://github.com/advisories/GHSA-w5hq-g745-h8pq).


handwritten/storage/package.json
```diff
@@ -81,13 +81,13 @@
     "async-retry": "^1.3.3",
     "duplexify": "^4.1.3",
     "fast-xml-parser": "^5.3.4",
-    "gaxios": "^6.0.2",
-    "google-auth-library": "^9.6.3",
+    "gaxios": "^7.1.5",
```
mhassan1, line 84: @shivanee-p @pearigee one potential problem with this approach is that the target versions of these dependencies have `engines.node >= 18`, while `@google-cloud/storage@7` has `engines.node >= 14`; what do you think?

mhassan1:
superseded by https://github.com/googleapis/google-cloud-node/pull/9289

PR #8598, opened 2026-06-11
chore: migrate google-cloud-documentai to librarian

quirogas:


packages/google-cloud-documentai/samples/resources/invoice.pdf
quirogas, line 1: Should we add the sample to the keep file?

pearigee: This file is referenced by the tests: https://github.com/googleapis/google-cloud-node/blob/main/packages/google-cloud-documentai/samples/test/quickstart.js

If it's easy to keep, let's keep it.

However, noone is running these tests right now and nothing would break if we deleted it. We could always recover it from git on the slim chance anyone cares to run these tests again. I would also be fine deleting it, if keeping it is at all inconvenient.

pearigee: I've changed my mind. None of these packages should have handwritten tests (and otherwise be entirely generated). Let's just get rid of it and delete the custom sample test as well. If we need sample tests, let's instead find a way to generate them.

PR #8599, opened 2026-06-11
chore: migrate google-cloud-eventarc-publishing to librarian

quirogas:


packages/google-cloud-eventarc-publishing/protos/spec.json
quirogas, line 1: what is this file for? do we need it?

pearigee: This is another adhoc test file added manually 4 years ago. The spec.json is required by the sample tests . . . which noone is running.

In this case, I am happy to delete it. The sample tests will break, but IMO shouldn't be relying on statically generated protobufjs artifacts. If the source proto drifted, this artifact would be out of date.

pearigee: For reference, added here: https://github.com/googleapis/google-cloud-node/commit/cedece439acd85856088fef2fbaf40fc6b593ee9#diff-7259aad9396d77a093fac7e28bdeef0e9b98979a7129def6c8a13889f41e0de6

pearigee: I think we should also delete the handwritten sample test in this case.

quirogas: https://github.com/googleapis/google-cloud-node/pull/8627

PR #8600, opened 2026-06-11
chore: migrate google-cloud-functions to librarian

quirogas:


packages/google-cloud-functions/src/v2/function_service_client.ts
```diff
@@ -27,8 +27,6 @@ import type {
   LROperation,
   PaginationCallback,
   GaxCall,
-  IamClient,
-  IamProtos,
```
pearigee, line 31: I am not sure what is causing this. I see that these lines are still included in googleapis-gen:
https://github.com/googleapis/googleapis-gen/blob/master/google/cloud/functions/v2/functions-v2-nodejs/src/v2/function_service_client.ts#L30

However, it appears to have been last updated a month ago with our prior generator rollout: [chore: update generator version to newest release v4.11.14](https://github.com/googleapis/googleapis-gen/commit/7cbddad1ef2968fc24d9aa1fe68aa5c6cc6b1adf). It could be stale?

pearigee: It looks like this imports are added by enabling the IamPolicy Mixin:
https://github.com/googleapis/google-cloud-node/blob/main/core/generator/gapic-generator-typescript/templates/cjs/typescript_gapic/src/%24version/%24service_client.ts.njk#L40
```
{%- if service.IAMPolicyMixin > 0 %}, IamClient, IamProtos{%- endif -%} ({%- if service.IAMPolicyMixin > 0 %}, IamClient, IamProtos{%- endif -%})
```

pearigee: The service yaml for v2 DOES include IAM: https://github.com/googleapis/googleapis/blob/master/google/cloud/functions/v2/cloudfunctions_v2.yaml#L9

This could be genuine issue. If I understand the generator correctly, a service yaml like this should trigger the inclusion of these lines.

packages/google-cloud-functions/src/v1/cloud_functions_service_client.ts
```diff
@@ -2235,6 +2240,92 @@ export class CloudFunctionsServiceClient {
...
+   *   Optional parameters. You can override the default settings for this call, e.g, timeout,
+   *   retries, paginations, etc. See {@link https://googleapis.github.io/gax-nodejs/interfaces/CallOptions.html | gax.CallOptions} for the details.
+   * @param {function(?Error, ?Object)} [callback]
+   *   The function which will be called with the result of the API call.
+   *
+   *   The second parameter to the callback is an object representing {@link google.iam.v1.Policy | Policy}.
+   * @returns {Promise} - The promise which resolves to an array.
+   *   The first element of the array is an object representing {@link google.iam.v1.Policy | Policy}.
+   *   The promise has a method named "cancel" which cancels the ongoing API call.
+   */
```
suztomo, line 2267: (As discussed yesterday) This doc comment does not have the function definition after that. This seems wrong.

suztomo: ### Root Cause
The issue is a bug in the code generator templates inside [gapic-generator-typescript](https://github.com/googleapis/google-cloud-node/tree/main/core/generator/gapic-generator-typescript):
*   When a package migration enables Librarian and configures `google.iam.v1.IAMPolicy` in the `mixins` list, the generator enables the IAM mixin.
*   For versioned APIs that **already define IAM methods natively** in their protos (like `google-cloud-functions` `v1`), the generator correctly decides **not** to generate the duplicate mixin methods to avoid conflicts. This logic is handled in [typescript/src/schema/proto.ts#L1061-L1066](https://github.com/googleapis/google-cloud-node/blob/main/core/generator/gapic-generator-typescript/typescript/src/schema/proto.ts#L1061-L1066).
*   However, in the generator's templates:
    *   [templates/cjs/typescript_gapic/_iam.njk](https://github.com/googleapis/google-cloud-node/blob/main/core/generator/gapic-generator-typescript/templates/cjs/typescript_gapic/_iam.njk)
    *   [templates/esm/typescript_gapic/_iam.njk](https://github.com/googleapis/google-cloud-node/blob/main/core/generator/gapic-generator-typescript/templates/esm/typescript_gapic/_iam.njk)
    the JSDoc comments for these mixin methods (e.g., `getIamPolicy`, `setIamPolicy`, `testIamPermissions`) were placed **outside** the conditional checks (e.g., `{% if service.iamPolicyMixinFlags.getIamPolicy %}`).
*   This caused the JSDoc comments to be generated anyway, but without the corresponding method definitions, leaving orphaned JSDoc comments at the end of the class.

### Functions API versions and Mixins

Here is the table showing the support for each mixin API across the different versions of the `google-cloud-functions` library:

| API Version | LRO (`google.longrunning.Operations`) | Locations (`google.cloud.location.Locations`) | IAM (`google.iam.v1.IAMPolicy`) |
| :--- | :--- | :--- | :--- |
| **`v1`** | Mixin | Mixin | **not mixin** (natively supported) |
| **`v2`** | Mixin | Mixin | Mixin |
| **`v2alpha`** | Mixin | Mixin | Mixin |
| **`v2beta`** | Mixin | Mixin | Mixin |

### Notes:
*   **"Mixin"** means the API is enabled via the `mixins` configuration and the calls are delegated to a separate helper client (e.g. `IamClient`).
*   **"not mixin"** for `v1` IAM means the `v1` protos natively define `GetIamPolicy`, `SetIamPolicy`, and `TestIamPermissions` as part of `CloudFunctionsService`, so they are generated as native client methods.

quirogas: https://github.com/googleapis/librarian/issues/6453

PR #8607, opened 2026-06-12
chore: migrate google-cloud-dialogflow to librarian

jskeet:


jskeet:
The only oddity here is the deletion of a `.cloud-repo-tools.json` file, which was last updated 8 years ago, and is the only example of its kind in the repo. I find it hard to believe it's still needed.

pearigee:
I concur, I can't see how deleting this would break anything. It doesn't appear to be important.

pearigee (approved):

PR #8609, opened 2026-06-12
chore: migrate google-cloud-contentwarehouse to librarian

jskeet:


jskeet:
documentai protos removed.

pearigee:
I do see this protos still exist in googleapis-gen: https://github.com/googleapis/googleapis-gen/tree/master/google/cloud/contentwarehouse/google-cloud-contentwarehouse-nodejs/protos/google/cloud/documentai/v1

These protos are marked as a dependency in the library BUILD rules here:
https://github.com/googleapis/googleapis/blob/master/google/cloud/contentwarehouse/v1/BUILD.bazel#L312

Added in this CL: https://critique.corp.google.com/cl/582493526

pearigee:
The content warehouse protos depend on the documentai ones here: https://github.com/googleapis/googleapis-gen/blob/0338218fe3694d49671198b459cc50a8a7bb542f/google/cloud/contentwarehouse/google-cloud-contentwarehouse-nodejs/protos/google/cloud/contentwarehouse/v1/document.proto#L21

It looks like this was added to enable compilation of the protos in the legacy generator infrastructure.

If Librarian doesn't need them (or doesn't have this problem), I think we are safe to delete them.

pearigee:
Passing back to Jon to confirm.

jskeet:
Agreed, if @pearigee is okay with the proto files being deleted, I think this is fine.

I also suspect that this isn't deleting any protos that contentwarehouse directly depends on anyway. It's only deleting other protos from documentai (I think).

jskeet:
@pearigee As this is my PR, I can't approve, but it sounds like we're good to go.

pearigee (approved):

PR #8610, opened 2026-06-12
chore: migrate google-cloud-config to librarian

jskeet:


jskeet:
Regenerated, and it became a "simple" API. I think I messed up when initially running my script. This should be an easy one to review.

quirogas (approved):

PR #8617, opened 2026-06-12
feat: regenerate google-cloud-workflows

jskeet:
Packages:

- google-cloud-workflows


jskeet:
(The other packages I'd *expected* to be in this batch had already been done.)

jskeet:
I suspect this is because we *want* to have both `workflows/v1` and `workflows/executions/v1` in the same library. This may need a bit more work in the combine-libraries code.

feywind:
Yeah, I don't think this one can go right now. Is this something that needs to be done on the librarian side, or ours? (We have a library combiner too, dunno if you all are using it.)

packages/google-cloud-workflows/src/index.ts
```diff
@@ -16,13 +16,15 @@
 // ** https://github.com/googleapis/synthtool **
 // ** All changes to this file may be overwritten. **
 
+import * as v1 from './v1';
```
suztomo, line 19: This is the same as line 21. This addition of line 19 and 20 seems unnecessary.

jskeet:
We need to first work out what we want to do with https://github.com/googleapis/librarian/issues/6278 and then work backwards to a suitable OwlBot PR.

PR #8619, opened 2026-06-12
chore: fix OwlBot configs

jskeet:
The paths listed before didn't exist, so we haven't been getting OwlBot updates for those packages.


JoeWang1127 (approved):

jskeet:
While this is probably safe, I'd like to wait until someone from the Node team can double-check it before merging.

pearigee (approved):

PR #8623, opened 2026-06-12
feat(storage): add IpFilter support to bucket metadata

thiyaguk09:


handwritten/storage/system-test/storage.ts
```diff
@@ -1289,6 +1290,79 @@ describe('storage', function () {
...
+      before(async () => {
+        ipFilterBucket = storage.bucket(generateName());
+        await ipFilterBucket.create();
+      });
+
+      after(async () => {
+        await ipFilterBucket.delete().catch(() => { });
+      });
+
+      it('should enable and set ipFilter', async () => {
```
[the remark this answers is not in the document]
thiyaguk09: test name was updated

handwritten/storage/system-test/storage.ts
```diff
@@ -1289,6 +1290,79 @@ describe('storage', function () {
...
+              allowedIpCidrRanges: [],
+            },
+            allowAllServiceAgentAccess: false,
+          },
+        };
+        const [meta] = await ipFilterBucket.setMetadata(metadata);
+        assert.strictEqual(meta.ipFilter?.mode, 'Disabled');
+        assert.strictEqual(meta.ipFilter?.publicNetworkSource?.allowedIpCidrRanges, undefined);
+        assert.strictEqual(meta.ipFilter?.allowAllServiceAgentAccess, false);
+      });
```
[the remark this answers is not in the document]
thiyaguk09: test name was changed as `should clear allowedIpCidrRanges`. but a new test suggested `should delete and clear ipFilter` was not added. currently api not support `ipFilter: null,`

handwritten/storage/system-test/storage.ts
```diff
@@ -4590,7 +4664,7 @@ describe('storage', function () {
       setTimeout(resolve, RETENTION_DURATION_SECONDS * 1000),
     );
     return Promise.all(
-      buckets.map(bucket => limit(() => deleteBucketAsync(bucket))),
+      buckets.map(bucket => limit(() => deleteBucketAsync(bucket).catch(() => {}))),
```
kalragauri: Catching and ignoring all errors is an anti-pattern. If this was added to handle buckets that were already deleted, you should explicitly check the error code instead.

handwritten/storage/src/bucket.ts
```diff
@@ -328,6 +328,47 @@ export interface EncryptionEnforcementConfig {
...
+ * bucket.setMetadata(metadata, (err, apiResponse) => {
+ *   if (err) {
+ *     console.error(err);
+ *   } else {
+ *     console.log('IP filter updated successfully.');
+ *   }
+ * });
+ * ```
+ */
+export interface IpFilter {
```
kalragauri, line 339: Do we also need to export this in handwritten/storage/src/index.ts?

handwritten/storage/test/bucket.ts
```diff
@@ -3692,5 +3692,132 @@ describe('Bucket', () => {
...
+          ],
+          allowAllServiceAgentAccess: true,
+          allowCrossOrgVpcs: true,
+        };
+
+        bucket.getMetadata = () => {
+          return Promise.resolve([{ipFilter}]);
+        };
+
+        bucket.getMetadata().then((data: any) => {
```
kalragauri: Can we use a more specific type here? We should avoid "any"  types wherever possible. Pls fix here and elsewhere in this PR.

handwritten/storage/system-test/storage.ts
```diff
@@ -1294,6 +1295,79 @@ describe('storage', function () {
...
+        assert.deepStrictEqual(
+          getMeta.ipFilter?.publicNetworkSource?.allowedIpCidrRanges,
+          ['203.0.113.0/24']
+        );
+
+        const metadata: BucketMetadata = {
+          ipFilter: {
+            mode: 'Disabled',
+            publicNetworkSource: {
+              allowedIpCidrRanges: [],
```
kalragauri, line 1381: nit: pls add a comment explaining that the API omits the field when the array is cleared.

handwritten/storage/test/index.ts
```diff
@@ -1258,6 +1258,48 @@ describe('Storage', () => {
...
+            location: 'US'
+          }
+        }
+      ];
+
+      storage.getBuckets = () => {
+        return Promise.resolve([bucketsResponse]);
+      };
+
+      storage.getBuckets().then((data: any) => {
```
kalragauri: In this test, `storage.getBuckets` is being overwritten (above) with a mock function before being called. Because the method itself is replaced, the actual library logic (which transforms raw API response items into Bucket instances and attaches bucket.metadata) is completely bypassed.

handwritten/storage/test/bucket.ts
```diff
@@ -3692,5 +3692,132 @@ describe('Bucket', () => {
...
+              network: 'projects/my-project/global/networks/my-vpc',
+              allowedIpCidrRanges: ['10.0.0.0/8'],
+            },
+          ],
+          allowAllServiceAgentAccess: true,
+          allowCrossOrgVpcs: true,
+        };
+
+        bucket.getMetadata = () => {
+          return Promise.resolve([{ipFilter}]);
```
kalragauri: Similar to the feedback in test/index.ts, it looks like bucket.getMetadata is being directly overwritten with a mock before it's called. By doing this, we skip the actual implementation of getMetadata and are only validating the mock itself.

handwritten/storage/system-test/storage.ts
thiyaguk09: done. I added this comment.

handwritten/storage/test/bucket.ts
thiyaguk09: done!  replaced `any` with the appropriate specific types.

handwritten/storage/system-test/storage.ts
thiyaguk09: done. I have  updated the catch block to explicitly check for a 404.

handwritten/storage/test/index.ts
thiyaguk09: I've updated the tests to mock storage.request instead of directly replacing the `storage.getBuckets` method.

handwritten/storage/src/bucket.ts
thiyaguk09: Yes, since it's part of the public metadata interface, it has been added.

handwritten/storage/test/bucket.ts
thiyaguk09: Done. I've refactored these tests to mock `bucket.parent.request` rather than overwriting "getMetadata" and "setMetadata" directly, ensuring the actual Bucket implementations are covered.

handwritten/storage/src/index.ts
```diff
@@ -167,6 +167,7 @@ export {
   UploadCallback,
   UploadOptions,
   UploadResponse,
+  IpFilter,
```
kalragauri: These exports are organized alphabetically. IpFilter is currently placed at the bottom.

kalragauri (approved):
Please add a test that verifies creation of a bucket with `ipFilter`.

handwritten/storage/src/index.ts
thiyaguk09: I've updated the export list to correctly alphabetize IpFilter.

thiyaguk09:
> Please add a test that verifies creation of a bucket with `ipFilter`.

Create a bucket with an ipFilter test added to both the unit and system tests.

PR #8626, opened 2026-06-12
feat: regenerated google-cloud-sql

jskeet:


jskeet:
We expect the tests to fail on this one - it's really only to allow the Node team to have a look. This is a "locally regenerated OwlBot PR, just for sql".

packages/google-cloud-sql/src/v1/index.ts
shivanee-p, line 1: None of the added files seem to have been added in this PR and it's causing [build failures](https://github.com/googleapis/google-cloud-node/actions/runs/27431336010/job/81082188307?pr=8626)

PR #8635, opened 2026-06-12
chore: migrate google-cloud-speech to librarian

jskeet:
This PR migrates the `google-cloud-speech` package to use Librarian for code generation.

### Highlights
- Enabled Librarian generation for `google-cloud-speech` by removing `skip_generate: true` in `librarian.yaml`.
- Removed OwlBot configuration (`.OwlBot.yaml`).
- Updated `.repo-metadata.json` (updated description, documentation links).
- Removed obsolete test/source files:
  - `packages/google-cloud-speech/src/helpers.ts`: The custom `streamingRecognize` helper.
  - `packages/google-cloud-speech/test/helpers.test.ts`: Tests for the custom helper.
  - `packages/google-cloud-speech/system-test/*`: Legacy system tests.
  <details>
  <summary>Why are these helper and test files obsolete?</summary>
  The custom <code>streamingRecognize</code> helper in <code>helpers.ts</code> was originally added to provide a better user experience for streaming (e.g., accepting raw audio and wrapping it). However, we are moving towards fully auto-generated clients to ensure consistency and reduce maintenance overhead. The associated tests (<code>helpers.test.ts</code> and legacy system tests) are removed as they target the deleted helper.
  
  Prior to this migration, OwlBot copied files from the Bazel-generated assembly package [speech-v1-nodejs](https://github.com/googleapis/googleapis/blob/05c4b6d37258ab81054529451cdf804f2918441f/google/cloud/speech/v1/BUILD.bazel#L273-L280) which included all dependencies. Librarian is more precise and only copies/keeps the protos defined for the API or transitively imported by them (resolved via `*_proto_list.json` generated during GAPIC generation), and filters out common protos that are already provided by `google-gax`. This is handled by [copyMissingProtos](https://github.com/googleapis/librarian/blob/v0.20.0/internal/librarian/nodejs/generate.go#L501-L552) in Librarian.
  </details>

<details>
<summary>Detail of breaking changes and migration path</summary>

#### What changed?
Previously, the custom helper was mixed into `SpeechClient.prototype.streamingRecognize` at runtime. It wrapped the raw gRPC stream to provide a more convenient API:
1. It accepted `streamingConfig` as the first argument to `streamingRecognize(config)`.
2. The returned stream accepted raw audio buffers (via `write()` or `pipe()`) and automatically wrapped them into `StreamingRecognizeRequest` objects (`{audioContent: buffer}`).

With the removal of the helper, users now interact directly with the raw generated gRPC stream.

#### Migration Path
**Before (with helper):**
```javascript
const stream = client.streamingRecognize(config).on('data', (response) => {
  // Handle response
});
// You could pipe a raw audio stream directly:
fs.createReadStream('audio.raw').pipe(stream);
// Or write buffers directly:
stream.write(audioBuffer);
```

**After (without helper):**
You must now write `StreamingRecognizeRequest` objects. The first write must contain the configuration, and subsequent writes must contain the audio content. You can no longer pipe raw audio streams directly without a transform.
```javascript
const stream = client.streamingRecognize().on('data', (response) => {
  // Handle response
});

// 1. Write the config first
stream.write({streamingConfig: config});

// 2. Write audio content wrapped in an object
stream.write({audioContent: audioBuffer});

// If you want to pipe a file, you must transform it first:
const {Transform} = require('stream');
const chunker = new Transform({
  objectMode: true,
  transform(chunk, encoding, callback) {
    callback(null, {audioContent: chunk});
  }
});
// Write config first, then pipe the chunked audio
stream.write({streamingConfig: config});
fs.createReadStream('audio.raw').pipe(chunker).pipe(stream);
```
</details>

### Fixes needed (for reviewers/author)
- **Delete `packages/google-cloud-speech/librarian.js`**: This file is still present and is trying to post-process the generated files by importing `./helpers` (which was deleted). This is causing compilation/test failures.
- **Regenerate code**: After deleting `librarian.js`, the code should be regenerated to remove the appended helper imports from:
  - `packages/google-cloud-speech/src/v1/speech_client.ts`
  - `packages/google-cloud-speech/src/v1p1beta1/speech_client.ts`
  - `packages/google-cloud-speech/src/v2/speech_client.ts`


shivanee-p (approved):

shivanee-p:
@pearigee and I were debugging the test failures, and we found that this code is being post-processed in by [librarian.js](https://github.com/googleapis/google-cloud-node/blob/566bab5c2f42e1ca01425fec4f9fe7ea6efa6ec3/packages/google-cloud-speech/librarian.js).

Additionally, the [imports in the index file](https://github.com/googleapis/google-cloud-node/blob/566bab5c2f42e1ca01425fec4f9fe7ea6efa6ec3/packages/google-cloud-speech/src/index.ts#L15
) need to be updated to account for the deleted files.

suztomo:
I enriched the pull request description. Let me see a way to avoid user-facing changes.

suztomo (changes requested):
Adding "Request changes" while I investigate this.

suztomo:
With updated keep list: https://github.com/googleapis/google-cloud-node/pull/8648

PR #8639, opened 2026-06-15
chore: migrate google-cloud-translate to librarian

jskeet:


jskeet:
Note that this has a keep list with three entries:

- src/v2/index.ts is preserved as the v2 client
- test/index.ts is preserved as a handwritten test for the v2 client
- src/index.ts is preserved as there is no way currently of generating it in combine-libraries; we want to effectively have a default version of v2, but that isn't a GAPIC client. In the BazelBot/OwlBot pipeline, this is implemented by skipping the file in https://github.com/googleapis/googleapis/blob/master/google/cloud/translate/BUILD.bazel:

```text
nodejs_gapic_combined_pkg(
    name = "google-cloud-translate-nodejs",
    default_version = "v3",
    release_level = "stable",
    templates_excludes = ["package.json", "src/index.ts", ".OwlBot.yaml"],
    deps = [
        "//google/cloud/translate/v3:translation-v3-nodejs",
        "//google/cloud/translate/v3beta1:translation-v3beta1-nodejs"
    ],
)
```

PR #8645, opened 2026-06-15
test(auth): add test for retry exhaustion on transient errors

westarle:
Add retry exhaustion test to verify client request rejects with 500 error after all transporter retries are exhausted.

This brings Node.js test coverage in line with other resilient implementations (like Python, Go, Rust, and Swift) which also perform internal credential-level retries with exponential backoff on the token endpoint before ultimately exhausting and bubbling the error up. (Contrast this with Java, C++, and PHP which lack internal credential retries entirely and fail immediately on transient auth errors).

### DIFF
--- a/core/packages/google-auth-library-nodejs/test/test.oauth2.ts
+++ b/core/packages/google-auth-library-nodejs/test/test.oauth2.ts
@@ -1026,6 +1026,28 @@ describe('oauth2', () => {
       assert.strictEqual('abc123', client.credentials.access_token);
     });
 
+    it('should throw an error after retries are exhausted on transient errors', async () => {
+      // Mock 4 failures (1 initial + 3 retries)
+      const scopes = [
+        nock(baseUrl, {
+          reqheaders: {
+            'content-type': 'application/x-www-form-urlencoded;charset=UTF-8',
+          },
+        })
+          .post('/token')
+          .times(4)
+          .reply(500, 'Internal Server Error'),
+      ];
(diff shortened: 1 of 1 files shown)


core/packages/google-auth-library-nodejs/test/test.oauth2.ts
```diff
@@ -1026,6 +1026,28 @@ describe('oauth2', () => {
...
+    it('should throw an error after retries are exhausted on transient errors', async () => {
+      // Mock 4 failures (1 initial + 3 retries)
+      const scopes = [
+        nock(baseUrl, {
+          reqheaders: {
+            'content-type': 'application/x-www-form-urlencoded;charset=UTF-8',
+          },
+        })
+          .post('/token')
+          .times(4)
```
pearigee, line 1038: Should this be a constant or explicitly configured for this test? I think we are implicitly falling back to the default 3 retries in gaxios: https://github.com/googleapis/google-cloud-node/blob/550635c136027d01e32fcbfa7b0550d0f91d7cda/core/packages/gaxios/src/common.ts#L476

It looks like this library stores its config in this static method: https://github.com/googleapis/google-cloud-node/blob/550635c136027d01e32fcbfa7b0550d0f91d7cda/core/packages/google-auth-library-nodejs/src/auth/oauth2common.ts#L213

We could reference the config explictly like this:
```
.times(1 + OAuthClientAuthHandler.RETRY_CONFIG().retryConfig?.retry ?? 3)
```

PR #8648, opened 2026-06-15
feat: migrate google-cloud-speech to librarian (keep helpers)

suztomo:
This PR migrates the `google-cloud-speech` package to use Librarian for code generation, while preserving the custom `streamingRecognize` helper and its tests to avoid breaking changes and preserve test coverage.

This is a follow-up to the initial attempt in #8635, which removed the helper and tests.

### Highlights
- Enabled Librarian generation for `google-cloud-speech` by removing `skip_generate: true` in `librarian.yaml`.
- Added `keep` configuration in `librarian.yaml` to preserve:
  - `src/helpers.ts` (custom helper)
  - `src/index.ts` (mixin logic)
  - `test/helpers.test.ts` (unit tests)
  - `system-test/fixtures/streamingRecognize/input.wav` (test fixture)
  - `system-test/speech_system_test.js` etc. (legacy system tests)
- Removed OwlBot configuration (`.OwlBot.yaml`) and other obsolete configs.

### Detailed Rationale for `keep` List

<details>
<summary>Why is <code>src/helpers.ts</code> kept?</summary>

`src/helpers.ts` contains the custom `streamingRecognize` helper. In the initial attempt (#8635), this helper was removed to transition to a fully auto-generated client. However, this helper is exposed to users and removing it introduces a major breaking change.

**The Breaking Change:**
The custom helper wraps the raw gRPC stream. Without it:
1. Users can no longer pass `streamingConfig` directly as the first argument to `streamingRecognize(config)`.
2. The returned stream no longer accepts raw audio buffers directly (e.g. via `stream.write(audioBuffer)` or `.pipe()`). Instead, users must write `StreamingRecognizeRequest` objects directly, wrapping the audio in `{audioContent: buffer}`.

Keeping `src/helpers.ts` preserves this helper and avoids these breaking changes.
</details>

<details>
<summary>Why is <code>src/index.ts</code> kept?</summary>

`src/index.ts` contains the runtime mixin logic that injects the helper into the generated clients:

```typescript
import {ImprovedStreamingClient} from './helpers';
// ...
Object.defineProperty(
  v1.SpeechClient.prototype,
  'streamingRecognize',
  Object.getOwnPropertyDescriptor(
    ImprovedStreamingClient.prototype,
    'streamingRecognize'
  )!
);
```

If `src/index.ts` is not in the `keep` list, Librarian will overwrite it with a default generated index file that only contains standard exports. 
Without the mixin logic in `src/index.ts`, the custom helper in `helpers.ts` is never bound to the `SpeechClient` prototype at runtime. Calling `client.streamingRecognize()` would fail (as the generated method is renamed to `_streamingRecognize` by `librarian.js` to avoid conflicts, leaving `streamingRecognize` undefined).
</details>

<details>
<summary>Why are the system tests and <code>input.wav</code> kept?</summary>

We preserve the legacy system tests (`system-test/speech_system_test.js`, `system-test/speech_typescript_system_test.ts`, etc.) and the associated test fixture (`system-test/fixtures/streamingRecognize/input.wav`) to maintain test coverage for the preserved helper. These tests verify the `streamingRecognize` helper functionality at a system level.

Note: We may delete these legacy system tests later (see [Librarian #6377](https://github.com/googleapis/librarian/issues/6377)).
</details>

### Detailed Rationale for Deleted Files

<details>
<summary>Why was <code>prettier.config.js</code> deleted?</summary>

`prettier.config.js` was deleted because it is redundant. 

It previously contained:
```javascript
module.exports = {
  singleQuote: true,
  trailingComma: 'es5',
};
```

This is identical to the standard Google TypeScript Style (GTS) config. The monorepo has a root `.prettierrc.cjs` that extends `gts/.prettierrc.json`. With this package-level config removed, Prettier will correctly fall back to the root config, ensuring consistent formatting across the entire repository without maintaining redundant local configuration files.
</details>

### User-facing Changes
None. The custom `streamingRecognize` helper and its mixin logic are preserved, so behavior is identical to pre-migration.


suztomo:
All checks have passed.

suztomo:
Superseding https://github.com/googleapis/google-cloud-node/pull/8635

shivanee-p (approved):

PR #8650, opened 2026-06-15
chore: migrate google-storage-control

### LINKED ISSUE #5955: nodejs: Investigate Offline NPM Install Timeout in google-storage-control
Generation fails for `google-storage-control` inside containerized or isolated build environments because its custom post-processor triggers a blocking `npm install` over the network.

We need to investigate why `librarian.js` triggers `npm install` in this package, and see if there is a better way to resolve what `librarian.js` is trying to accomplish (e.g., loading dependencies gracefully or checking system registry states without blocking offline container executions).

### Reproduction steps

Set your npm config registry to strict offline registry mapping or disconnect internet access, then trigger generation on storage-control:
```bash
librarian generate google-storage-control
```

### LINKED ISSUE #6359: nodejs: google-storage-control fails to generate in Docker
```text
2026/06/10 10:00:48 librarian: generate library "google-storage-control" (nodejs): failed to run post processor: librarian.js failed: /root/.nvm/versions/node/current/bin/node /repo/packages/google-storage-control/librarian.js: npm error code EACCES
npm error syscall mkdir
npm error path /.npm
npm error errno EACCES
npm error
npm error Your cache folder contains root-owned files, due to a bug in
npm error previous versions of npm which has since been addressed.
npm error
npm error To permanently fix this problem, please run:
npm error   sudo chown -R 80098:89939 "/.npm"
npm notice
npm notice New major version of npm available! 10.8.2 -> 11.16.0
npm notice Changelog: https://github.com/npm/cli/releases/tag/v11.16.0
npm notice To update run: npm install -g npm@11.16.0
npm notice
npm error Log files were not written due to an error writing to the directory: /.npm/_logs
npm error You can rerun the command with `--loglevel=verbose` to see the logs in your terminal
Error: Command failed: npm install
    at checkExecSyncError (node:child_process:890:11)
    at execSync (node:child_process:962:15)
    at Object.<anonymous> (/repo/packages/google-storage-control/librarian.js:23:3)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Module._load (node:internal/modules/cjs/loader:1019:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:128:12)
    at node:internal/main/run_main_module:28:49 {
  status: 1,
  signal: null,
  output: [ null, null, null ],
  pid: 97,
  stdout: null,
  stderr: null
}
```

JoeWang1127:
Migration google-storage-control to librarian generate.

Fixes https://github.com/googleapis/librarian/issues/6359
Fixes https://github.com/googleapis/librarian/issues/5955


packages/google-storage-control/src/v2/storage_client.ts
```diff
@@ -80,7 +80,7 @@ export class StorageClient {
   private _defaults: { [method: string]: gax.CallSettings };
   private _universeDomain: string;
   private _servicePath: string;
-  private _log = logging.log('storage');
+  private _log = logging.log('storage-control');
```
danieljbruce, line 83: nit: Do we need to change this log? What improvement does it offer?

packages/google-storage-control/LICENSE
```diff
@@ -1,202 +0,0 @@
-
```
danieljbruce, line 1: What is the purpose of removing these files?

danieljbruce (approved):
Just a few nits, but nothing seems particularly risky about this PR to me so approving to unblock librarian migration.

packages/google-storage-control/LICENSE
JoeWang1127: This is one of the file we removed during migration because the monorepo should have one license.

packages/google-storage-control/src/v2/storage_client.ts
JoeWang1127: Looks like the previous name is not correct since this is storage control client.

PR #8653, opened 2026-06-16
chore: migrate google-cloud-visionai to librarian

jskeet:


jskeet:
This supercedes https://github.com/googleapis/google-cloud-node/pull/8634 - the significant change is removing v1alpha from librarian.yaml (not sure why it was included by the migration tool) which makes the diffs much smaller.

feywind (approved):

PR #8667, opened 2026-06-17
chore: migrate google-cloud-dataproc to librarian

jskeet:


jskeet:
Note: this deletes a smoke test, but given the age of the package, I don't think that smoke test is particularly useful any more. I can add it to a keep list if necessary.

danieljbruce (approved):

PR #8685, opened 2026-06-18
fix(generator): add long dependency to CJS template and update baselines

suztomo:
This PR adds the missing "long" dependency to the CommonJS package.json template in gapic-generator-typescript, and updates the baselines to match.

### Why `long` is in `devDependencies`
`long` is only required during the compilation phase of the library itself to resolve type definitions in the generated `protos/protos.d.ts` (which has `import Long = require("long");`). It is not directly imported in the runtime JavaScript code. At runtime, `long` is transitively resolved via `google-gax` -> `protobufjs`, so it does not need to be a production dependency of the generated library.

### GAX
It also updates the `google-gax` dependency version to `^5.0.0` in the template to match the recently introduced `google-maps-mapmanagement` package:
https://github.com/googleapis/google-cloud-node/blob/6f9710c504c5272dfb59a77c7da98235ed9cf3ff/packages/google-maps-mapmanagement/package.json. The latest (non-RC) GAX version is 5.0.7: https://www.npmjs.com/package/google-gax?activeTab=versions.

Last week, I ran generate -all in https://github.com/googleapis/google-cloud-node/pull/8672 and package.json was not updated. Package.json file in the keep list https://github.com/googleapis/google-cloud-node/blob/ed701c6bfbe626f95f94159533140a1f921caf3c/librarian.yaml#L24. The effect is only to upcoming newly generated libraries.


quirogas (commented):
Reverting to 5.0.0 could introduce a significant number of changes to the generated code, as `google-gax` provides a transitive dependency used by `protobufjs` during client generation. Let's talk about this change before we  merge it.

suztomo:
@quirogas For google-gax version I found that the latest version 5.0.x versions are descendants (having the changes) of  5.1.1-rc.1. See b/527420531#comment3. Would you confirm this address your concern and remove do-not-merge label?

quirogas:
> @quirogas For google-gax version I found that the latest version 5.0.x versions are descendants (having the changes) of 5.1.1-rc.1. See b/527420531#comment3. Would you confirm this address your concern and remove do-not-merge label?

Thank you so much for all your work! Yes, I agree that we should be good to remove the labels. However, could you perform one final sanity check before merging the PR and review the generated changes to ensure they won't cause any issues? Feel free to send me the generated branch, and I can help look them over. 

@pearigee WDYT?

suztomo:
@quirogas  With `core/generator/gapic-generator-typescript/templates/cjs/typescript_gapic/package.json`, this pull request only changes the package.json in a new library. This does not change the dependencies of gapic-generator-typescript.

feywind (approved):
Thanks for throwing in the gax version fix too.

feywind:
> @quirogas For google-gax version I found that the latest version 5.0.x versions are descendants (having the changes) of 5.1.1-rc.1. See b/527420531#comment3. Would you confirm this address your concern and remove do-not-merge label?

I had a PR to fix that, a while back, but it didn't end up getting merged for unrelated reasons, so I'm glad to see it in here. The rc is much older than the 5.0.x releases. Our plan is eventually to just bump gax to 5.2.x so we can avoid future weirdness with the rc.

PR #8707, opened 2026-06-22
test: run the spanner system tests in google cloud build

danieljbruce:
## Description

This pull request adds a yaml file to instruct the Spanner system test CI check to work with the new Cloud Build trigger thereby making the new CI check effectively run our system tests. The new test environment caused tests to fail so we will skip those tests and address them later in [this bug](https://b.corp.google.com/issues/530911768).

## Impact

Leverages the strengths of running system tests in GCB rather than relying on kokoro for system tests.

## Testing

This pull request tells the tests how to work with the new check to the continuous integration pipeline for the Spanner system tests thus improves the effectiveness of that test.

## Next Steps

1. Merge a pull request that removes the system tests from kokoro. This should be done in a separate pull request to separate the concerns of removing the kokoro system tests from adding the Google Cloud Build system tests.


danieljbruce (commented):
LGTM. I would prefer we skip each individual test, but just skipping the Backups test is fine too.

alkatrivedi (approved):

surbhigarg92 (approved):

PR #8752, opened 2026-06-25
fix(storage): destroy local read stream on upload write failure to prevent resource leaks

### LINKED ISSUE #7325: File descriptor not closed on remote bucket.upload() errors

N/A, the bug is in the client code.

https://gist.github.com/MartinKolarik/04bcb1ac4209d3afee62a31667f666d7

1) Create a large local file.  
2) Start `bucket.upload`.  
3) Force an early failure (e.g., upload to a bucket without write permissions).  
4) Delete the local file while the process is alive.  
5) Check `/proc/<pid>/fd` and `df/du`.  

The file descriptor open in `bucket.upload` remains allocated until the process exists. This causes a small memory leak and prevents the local file from being fully deleted. The relevant code is: https://github.com/googleapis/nodejs-storage/blob/c7004daae0b506eeefc5f4f241d9e3df3589b0e8/src/bucket.ts#L4364-L4379

The file descriptor should always be closed when the function finishes.

thiyaguk09:

Fixes #7325

### DIFF
--- a/handwritten/storage/src/bucket.ts
+++ b/handwritten/storage/src/bucket.ts
@@ -4517,7 +4517,7 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
                 readStream.destroy();
                 if (
                   this.storage.retryOptions.autoRetry &&
-                  this.storage.retryOptions.retryableErrorFn!(err as any)
+                  this.storage.retryOptions.retryableErrorFn!(err as ApiError)
                 ) {
                   return reject(err);
                 } else {
--- a/handwritten/storage/src/bucket.ts
+++ b/handwritten/storage/src/bucket.ts
@@ -4501,17 +4501,22 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
             ) {
               newFile.storage.retryOptions.autoRetry = false;
             }
+            const readStream = fs.createReadStream(pathString);
             const writable = newFile.createWriteStream(options);
             if (options.onUploadProgress) {
               writable.on('progress', options.onUploadProgress);
             }
-            fs.createReadStream(pathString)
-              .on('error', bail)
+            readStream
+              .on('error', err => {
+                readStream.destroy();
+                bail(err);
+              })
               .pipe(writable)
               .on('error', err => {
+                readStream.destroy();
                 if (
                   this.storage.retryOptions.autoRetry &&
-                  this.storage.retryOptions.retryableErrorFn!(err)
+                  this.storage.retryOptions.retryableErrorFn!(err as any)
                 ) {
                   return reject(err);
                 } else {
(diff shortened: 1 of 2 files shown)


handwritten/storage/src/bucket.ts
```diff
@@ -4501,17 +4501,23 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
...
               newFile.storage.retryOptions.autoRetry = false;
             }
+            const readStream = fs.createReadStream(pathString);
             const writable = newFile.createWriteStream(options);
             if (options.onUploadProgress) {
               writable.on('progress', options.onUploadProgress);
             }
-            fs.createReadStream(pathString)
-              .on('error', bail)
+            readStream
```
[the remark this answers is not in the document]
thiyaguk09: Reordered the readStream initialization.

handwritten/storage/src/bucket.ts
```diff
@@ -4505,13 +4505,19 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
...
+                writable.destroy();
+                bail(err);
+              })
               .pipe(writable)
               .on('error', err => {
+                readStream.destroy();
                 if (
                   this.storage.retryOptions.autoRetry &&
-                  this.storage.retryOptions.retryableErrorFn!(err)
+                  this.storage.retryOptions.retryableErrorFn!(err as any)
```
Dhriti07: Using as any bypasses TypeScript's type-checking guarantees, leaving the compiler unaware of potential type mismatches. Instead of casting to any, explicitly cast err to the expected error type

thiyaguk09: I’ve updated the error handler block to explicitly cast err to `ApiError`  instead of bypassing type checks with as any.

handwritten/storage/test/bucket.ts
```diff
@@ -100,11 +100,18 @@ class FakeNotification {
 }
 
 let fsStatOverride: Function | null;
+let fsCreateReadStreamOverride: Function | null;
 const fakeFs = {
   ...fs,
   stat: (filePath: string, callback: Function) => {
     return (fsStatOverride || fs.stat)(filePath, callback);
   },
+  createReadStream: (filePath: string, options?: any) => {
```
Dhriti07: Nit: Avoid use of any. It would be better to type this parameter using Node's built-in types, such as fs.ReadStreamOptions | BufferEncoding | string | undefined, or simply omit the explicit any if the compiler can infer it from the fs module's types.

Dhriti07 (commented):
LGTM, Will add approval once the linter issue is fixed

handwritten/storage/test/bucket.ts
thiyaguk09: Updated to use Parameters<typeof fs.createReadStream>[1] to leverage Node's built-in types cleanly without any

Dhriti07 (approved):

PR #8771, opened 2026-06-29
feat(storage): support copying CSEK-encrypted files to non-CSEK destinations

### LINKED ISSUE #7351: Ability to copy a file encrypted with customer-supplied key to a file without customer-supplied key

Not possible, as the "Try this method" GUI does not allow setting the necessary optional extension headers.

While using the `file.copy()` method, I would like a way to copy a file encrypted with a customer-supplied encryption key (CSEK) to a file that is not encrypted with a CSEK (e.g., using Google default encryption for destination file). This is not currently possible, as the source `file`'s key is copied to the destination `file` object and reused. I believe this is the result of the library's code and not the API.

In more detail, it is currently possible to copy a file and change the CSEK, as follows:
```typescript
const srcFile = bucket.file('my-file1').setEncryptionKey('a'.repeat(32));

await bucket.upload(filePath, {
	destination: srcFile,
});

// Metadata will indicate CSEK with customerEncryption
const [srcMeta] = await srcFile.getMetadata(); 

// Note we are using a different key here:
const dstFile = bucket.file('my-file-2').setEncryptionKey('b'.repeat(32));

await srcFile.copy(dstFile);

// Metadata will indicate a CSEK with customerEncryption, and a different customerEncryption.keySha256 value.
const [dstMeta] = await dstFile.getMetadata();
```

However, I don't think it's currently possible to avoid setting a CSEK on the `dstFile`. Consider the behavior with this:
```typescript
const srcFile = bucket.file('my-file1').setEncryptionKey('a'.repeat(32));

await bucket.upload(filePath, {
	destination: srcFile,
});

// Metadata will indicate CSEK with `customerEncryption`
const [srcMeta] = await srcFile.getMetadata(); 

// Note we are purposely not setting a key for the dst file
const dstFile = bucket.file('my-file-2');

await srcFile.copy(dstFile);

// Metadata will indicate a CSEK with customerEncryption, and the same customerEncryption.keySha256 as the source (key is reused).
const [dstMeta] = await dstFile.getMetadata();
```

I've tried to explicitly set a null or empty-string key on the `dstFile`, but that gives a "Missing an encryption key, or it is not base64 encoded, or it does not meet the required length of the encryption algorithm." error.

Perhaps as a solution, the library could keep the current key-copying / reusing behavior if `setEncryptionKey()` is not called on the destination file, but allow explicitly setting a `null` key for the dstFile to avoid the key copying.

### Additional context/notes

_No response_

thiyaguk09:

Fixes #7351

### DIFF
--- a/handwritten/storage/src/file.ts
+++ b/handwritten/storage/src/file.ts
@@ -1461,20 +1461,18 @@ class File extends ServiceObject<File, FileMetadata> {
         this.encryptionKeyHash;
     }
 
-    let copiedKey = false;
     if (
       this.encryptionKey !== undefined &&
       this.encryptionKey !== null &&
       newFile.encryptionKey === undefined
     ) {
-      newFile.encryptionKey = this.encryptionKey;
-      newFile.encryptionKeyBase64 = this.encryptionKeyBase64;
-      newFile.encryptionKeyHash = this.encryptionKeyHash;
-      copiedKey = true;
+      newFile.setEncryptionKey(this.encryptionKey);
     }
 
-    if (newFile.encryptionKey !== undefined && !copiedKey) {
-      this.setEncryptionKey(newFile.encryptionKey!);
+    if (newFile.encryptionKey !== undefined && newFile.encryptionKey !== null) {
+      headers['x-goog-encryption-algorithm'] = 'AES256';
+      headers['x-goog-encryption-key'] = newFile.encryptionKeyBase64;
+      headers['x-goog-encryption-key-sha256'] = newFile.encryptionKeyHash;
     } else if (options.destinationKmsKeyName !== undefined) {
       query.destinationKmsKeyName = options.destinationKmsKeyName;
       delete options.destinationKmsKeyName;
@@ -1505,6 +1503,14 @@ class File extends ServiceObject<File, FileMetadata> {
       query.ifGenerationMatch = options.preconditionOpts?.ifGenerationMatch;
       delete options.preconditionOpts;
     }
+
+    const originalGetRequestInterceptors = this.getRequestInterceptors;
+    this.getRequestInterceptors = () => {
+      return originalGetRequestInterceptors.call(this).filter(
+        interceptorFn =>
+          interceptorFn !== this.encryptionKeyInterceptor?.request
+      );
+    };
 
     this.request(
       {
@@ -1543,6 +1549,8 @@ class File extends ServiceObject<File, FileMetadata> {
         callback!(null, newFile, resp);
       },
     );
+
+    this.getRequestInterceptors = originalGetRequestInterceptors;
   }
 
   /**
--- a/handwritten/storage/src/file.ts
+++ b/handwritten/storage/src/file.ts
@@ -1469,9 +1469,7 @@ class File extends ServiceObject<File, FileMetadata> {
     }
 
     const destinationKmsKeyName =
-      options.destinationKmsKeyName ||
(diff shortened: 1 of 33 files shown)


handwritten/storage/src/file.ts
```diff
@@ -1456,6 +1474,8 @@ class File extends ServiceObject<File, FileMetadata> {
         callback!(null, newFile, resp);
       },
     );
+
+    this.getRequestInterceptors = originalGetRequestInterceptors;
```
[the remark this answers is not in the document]
thiyaguk09: removed

handwritten/storage/src/file.ts
```diff
@@ -1419,6 +1429,14 @@ class File extends ServiceObject<File, FileMetadata> {
       delete options.preconditionOpts;
     }
 
+    const originalGetRequestInterceptors = this.getRequestInterceptors;
+    this.getRequestInterceptors = () => {
+      return originalGetRequestInterceptors.call(this).filter(
+        interceptorFn =>
+          interceptorFn !== this.encryptionKeyInterceptor?.request
+      );
+    };
```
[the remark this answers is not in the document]
thiyaguk09: there is a possible for race condition risk, but passing `requestOpts.interceptors` won't work because `@google-cloud/common` (`ServiceObject`) unconditionally appends local instance-level interceptors to every request.

To route the request through the parent bucket (`this.bucket.request`) is the best fix here. This bypasses the file-level encryption interceptors entirely without any  race conditions."

handwritten/storage/src/file.ts
```diff
@@ -1452,15 +1452,25 @@ class File extends ServiceObject<File, FileMetadata> {
...
     }
 
-    if (newFile.encryptionKey !== undefined) {
-      this.setEncryptionKey(newFile.encryptionKey!);
+    if (
+      this.encryptionKey !== undefined &&
+      this.encryptionKey !== null &&
+      newFile.encryptionKey === undefined
+    ) {
+      newFile.setEncryptionKey(this.encryptionKey);
```
Dhriti07, line 1479: If a user is trying to copy the file to switch to a KMS key, will this block will still accidentally copy the old CSEK key and skip the KMS logic below ?

handwritten/storage/test/file.ts
```diff
@@ -274,6 +290,19 @@ describe('File', () => {
     specialCharsFile = new File(BUCKET, "special/azAZ!*'()*%/file.jpg");
     specialCharsFile.request = util.noop;
 
+    activeFile = null;
```
Dhriti07: File.copy was updated to use this.bucket.request instead of this.request. Overriding BUCKET.request was used to make the old tests pass without rewriting them. Modifying File.prototype.copy globally like this can cause flaky tests and randomly break other tests. It also makes the tests a bit confusing to read, since they still check file.request even though the real code no longer uses it.

Would it be simpler to just remove that workaround and update the tests to assert against file.bucket.request directly?

thiyaguk09: Removed the workaround and updated the tests to assert directly on file.bucket.request.

handwritten/storage/src/file.ts
thiyaguk09: The new !destinationKmsKeyName check added to the if condition prevents this. It safely skips the CSEK block and falls through to the KMS logic below.

Dhriti07 (approved):

PR #8794, opened 2026-07-03
chore: migrate google-cloud-workflows

### LINKED ISSUE #6278: chore: onboard google-cloud-workflows to librarian
Currently, the migrated configuration for google-cloud-workflows is:

```yaml
  - name: google-cloud-workflows
    version: 5.1.1
    apis:
      - path: google/cloud/workflows/v1
      - path: google/cloud/workflows/v1beta
```

This omits google/cloud/workflows/executions/v1 (and the v1beta)

Ironically, it looks like what's in https://github.com/googleapis/google-cloud-node/tree/main/packages/google-cloud-workflows/src/v1 is *only* `google/cloud/workflows/executions/v1` (and the v1beta). Looking in googleapis-gen, there are two different "top level" directories which I suspect the OwlBot config isn't handling properly:

- https://github.com/googleapis/googleapis-gen/tree/master/google/cloud/workflows/executions/google-cloud-workflows-executions-nodejs
- https://github.com/googleapis/googleapis-gen/tree/master/google/cloud/workflows/google-cloud-workflows-nodejs

After manually changing the librarian.yaml file to generate workflows/executions (but not workflows itself) and regenerating, we end up with a diff which is:

- Deletions of workflows *samples* (which I suspect won't work at the moment)
- Deletions of workflows *protos* (which I suspect were being pulled in tranisitively)
- Various modifications (I haven't gone into these in detail)

If instead we change librarian.yaml to have all 4 APIs, like this:

```yaml
  - name: google-cloud-workflows
    version: 5.1.1
    apis:
      - path: google/cloud/workflows/v1
      - path: google/cloud/workflows/v1beta
      - path: google/cloud/workflows/executions/v1
      - path: google/cloud/workflows/executions/v1beta
    copyright_year: "2026"
```

... then we *add* a bunch of source code under src/ and test/ (but not samples/)

So basically I believe:

- The current state in google-cloud-node is somewhat broken (due to not including workflows prod code)
- Including all four APIs is the right way forward
- It would be simpler to either modify librarian.yaml by hand after a one-off migrate, or hard-code the change into the migrate tool

JoeWang1127:
Migrate google-cloud-workflows to librarian generate.

Fixes https://github.com/googleapis/librarian/issues/6278


packages/google-cloud-workflows/protos/google/cloud/workflows/v1/workflows.proto
JoeWang1127, line 1: We've decided to generate `google-cloud-workflows` as-is, protos from `google/cloud/workflows` are not included in package anymore.

pearigee (approved):

PR #8811, opened 2026-07-06
test: remove kokoro system tests for cloud-profiler

danieljbruce:
# Description

https://github.com/googleapis/google-cloud-node/pull/8698 added the CI check to run cloud-profiler system tests in GCB so now we don't need to run the system tests in kokoro anymore. This pull request removes the CI check to run system tests from the kokoro environment because it is now redundant.

## Impact

Removes redundant testing to save resources.

## Files changed

- Removes the system-test.cfg files so that kokoro doesn't try to run system-test.sh from continuous or presubmit environments.
- Removes the system-test.sh file as it will not be used anymore now that the config files are gone
- trampoline_v2.sh example refers to test.sh instead since system-test.sh doesn't exist anymore


aabmass (approved):
Great thanks!

PR #8827, opened 2026-07-07
fix!: Update storage to use newer versions of gaxios, teeny-request, and google-auth-library

pearigee:
These newer versions required code changes to be compatible, specifically changes to how headers are handled.


rexxars:
Could this PR also include a bump to retry-request@^8.0.4? As it stands, this will still pull retry-request@7 -> teeny-request@9 -> node-fetch@2 -> whatwg-url, which causes deprecation warning as it reaches into `punycode`

PR #8833, opened 2026-07-08
feat: integrate monorepo-wide linter into CI

shivanee-p:
This pull request centralizes the ESLint and GTS checking architecture across the entire google-cloud-node monorepo. It removes redundant package-level configurations, integrates type-aware linting rules, and adds the unified linter runner (bin/linter.mjs) to the CI pipeline in a non-blocking mode to ensure a smooth transition.

- Extended plugin:@typescript-eslint/recommended-requiring-type-checking with "project": true to catch subtle async bug patterns (e.g., floating promises, unawaited thenables).
- Re-enabled default Google TypeScript Style (GTS) Prettier rules while preserving necessary overrides for monorepo import/promise hygiene.
- Added a dedicated lint job to the presubmit workflow that executes npm run lint (node ./bin/linter.mjs) across PR diffs that is non-blocking  for PR submission. Existing formatting or linting violations will be reported as warning badges without blocking PR merging or releasing.


.github/workflows/presubmit.yaml
```diff
@@ -28,3 +28,18 @@ jobs:
...
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          fetch-depth: 300
+          persist-credentials: false
+      - name: Use Node.js 18
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
+        with:
+          node-version: 18
```
pearigee: Can we please use a supported Node? Maybe Node 24?

bin/linter.mjs
```diff
@@ -45,9 +45,7 @@ async function run() {
   } catch (err) {
     console.error('\nLinter failed:', err.message);
     // Setting exit code 1 to indicate failure. In the CI pipeline,
-    // a non-zero exit code will cause the check to fail and block the PR.
-    // Note: TypeScript (tsc) compile failures and ESLint errors (severity 2)
-    // are blocking.
+    // continue-on-error is used to ensure this does not block PRs.
     process.exitCode = 1;
```
pearigee, line 49: Is this blocking or not? For now can we me make it non blocking?

I am not sure what this comment about "continue-on-error" means in this context?

pearigee:
The linter CI emitted the following errors. Is this intended? I know there were not modified TS files? Do these messages indicate a failure (as "fatal" implies) or it is just stating these branches were not found?

```
fatal: bad revision 'origin/main'
fatal: bad revision 'main'
```

bin/linter.mjs
shivanee-p: This is not blocking! `continue-on-error: true` is set in the presubmit.yaml. The presubmit check will show that it failed if there's an error, but the failure will not be blocking.

.github/workflows/presubmit.yaml
shivanee-p: Ack

shivanee-p:
This is not intended - thanks for calling that out. The root of the issue is how a PR is created. When origin/main or main aren't locally available, it'll show this error message. I've updated the code to check for upstream/main or FETCH_HEAD if those are available.

pearigee (approved):

PR #8839, opened 2026-07-09
feat(google/pubsub/v1): onboard a new library

quirogas:
Onboard the lightweight Pub/Sub API client library (@google-cloud/pubsub-api) for google/pubsub/v1.

Add README disclaimer directing end users to the handwritten @google-cloud/pubsub package. Verify generated client structure, proto definitions, and test suite.

For b/475341222 and b/512488166


librarian.yaml
```diff
@@ -1347,6 +1347,12 @@ libraries:
     copyright_year: "2026"
     nodejs:
       default_version: v1
+  - name: google-cloud-pubsub-api
+    version: 0.1.0
```
feywind, line 1351: We should be able to make this 1.0.0, or did you want to be cautious about it?

feywind (approved):
This looks great in principle, good starting point. (I say because we will no doubt find stuff after we get started...)

librarian.yaml
feywind: Actually, come to think of it, maybe it's better to start with 0.1.0 in case we need to break things to get it into place.

quirogas: yes, just for the POC at least. once we go over it, we could upgrade it.

PR #8858, opened 2026-07-10
feat(google/devtools/clouderrorreporting/v1beta1): onboard a new library

quirogas:


feywind:
Do we need this one? Or was it an accident vs google.logging.v2?

quirogas:
> Do we need this one? Or was it an accident vs google.logging.v2?

Yes :(

PR #8861, opened 2026-07-10
feat(google/cloud/bigquery/storage/v1): onboard a new library

quirogas:


packages/google-cloud-bigquery-storage-api/README.md
```diff
@@ -0,0 +1,108 @@
+[//]: # "This README.md file is auto-generated, all changes to this file will be lost."
+[//]: # "The comments you see below are used to generate those parts of the template in later states."
+<img src="https://avatars2.githubusercontent.com/u/2810941?v=3&s=96" alt="Google Cloud Platform logo" title="Google Cloud Platform" align="right" height="96" width="96"/>
+
+# [BigQuery Storage API: Nodejs Client][homepage]
```
feywind, line 5: This is missing the warning too.

feywind (commented):
The original package has a bunch of alpha and beta gapics, but I don't see where they're being used or exported. So I think those are vestigial, and we can just use v1 here.

quirogas:
> The original package has a bunch of alpha and beta gapics, but I don't see where they're being used or exported. So I think those are vestigial, and we can just use v1 here.

Yes,  currently only `v1` is used, the other versions where only part of the index.ts exports.

feywind (approved):

PR #8862, opened 2026-07-10
feat(google/cloud/bigquery/v2): onboard a new library

quirogas:


feywind:
We're going to close this one for now, because the BQ library is still using apiary rather than gapic. BQ storage is separate.

PR #8877, opened 2026-07-14
fix(pubsub-api): use the release version of gax

feywind:
This was using the old template, so it got the false beta version of gax.


quirogas (approved):
Does this have to change in the template as well?

feywind:
> Does this have to change in the template as well?

Ah, I'm pretty sure Tomo changed it there, to 5.0.0, which should have the same effect.

PR #8891, opened 2026-07-16
test(gax): Add an integration test for testing post quantum encryption

danieljbruce:
## Description

Adds an integration test for the Gapic clients that only passes for Node versions 22.20 and up as those versions of Node are the only ones that use a lattice based algorithm that is quantum safe. The integration test checks that the algorithm used for network traffic encryption is safe from quantum based decryption methods. It does this by setting up the showcase server with a certificate to tell the server to use the quantum safe algorithm and using the grpc echo client to call the server much like the other test-application tests do.

As an added bonus to using the test in test-application, the test also is run in a new jest framework that we plan on integrating into our clients.

## Impact

Ensures Gapic library traffic is safe from decryption after further advances in quantum computing technology.

## Testing

You can run the test in isolation from the `test-application` directory with `npm run test`.

The test fails for Node versions below 22.20 and passes for 22.20+ versions.

You can also run `npm run test-application` from the gax directory to run the tests.


core/packages/gax/test/test-application/src/pqc-test.ts
```diff
@@ -0,0 +1,195 @@
...
+    // --- 1. gRPC PQC Test ---
+    currentTestType = 'grpc';
+    let negotiatedGroupGrpc: string | undefined;
+
+    const interceptor = (options: any, nextCall: any) => {
+      return new grpc.InterceptingCall(nextCall(options), {
+        start: (metadata: any, listener: any, next: any) => {
+          next(metadata, {
+            onReceiveMetadata: (receivedMetadata: any, nextListener: any) => {
+              const group = receivedMetadata.get('x-showcase-tls-group');
```
hongalex: In addition to this, could you also verify that the client included `X25519MLKEM768` hybrid KEM in `x-showcase-tls-client-supported-groups`?

core/packages/gax/test/test-application/src/pqc-test.ts
```diff
@@ -0,0 +1,195 @@
...
+            interceptors: [interceptor],
+          },
+        },
+      }
+    );
+
+    assert.strictEqual(responseGrpc.content, 'grpc-pqc-test');
+    assert.ok(grpcSocket, 'Expected to intercept gRPC TLS socket');
+    assert.strictEqual(grpcSocket.getProtocol(), 'TLSv1.3');
+    assert.strictEqual(grpcSocket.getCipher()?.name, 'TLS_AES_128_GCM_SHA256', 'Expected specific negotiated cipher');
```
westarle: nit: are these assertions necessary @hongalex @suzmue ?  I don't think these symmetric ciphers / hashes are relevant (mainly we are deciding if the KEM support is there?)

core/packages/gax/test/test-application/src/pqc-test.ts
```diff
@@ -0,0 +1,195 @@
...
+    // Restore the original tls.connect
+    (tls as any).connect = originalTlsConnect;
+  }
+}
+
+/**
+ * Spins up a TLS-enabled showcase server and runs the PQC compliance tests.
+ * Cleans up the generated certificate file after the tests complete.
+ */
+export async function runPqcComplianceTests() {
```
westarle, line 167: Rather than a test that fails on specific Node versions, would it make sense to skip this test conditionally (i.e. on `semver.satisfies(process.version, '>=22.20.0')`)?

It could print a "skipping" message.

This way it could be added to our CI more permanently.

core/packages/gax/test/test-application/src/pqc-test.ts
```diff
@@ -0,0 +1,195 @@
...
+  process.env['SHOWCASE_VERSION'] = '0.41.1';
+  const showcaseServerTls = new ShowcaseServer();
+  const tlsPort = 7443;
+  const caCertOutputFile = 'showcase.pem';
+  const pemPath = path.join(process.cwd(), caCertOutputFile);
```
westarle: Are there ways we could increase test isolation and reduce potential side-effects? --

* by replacing the env variable when it completes.
* use something like `fs.mkdtempSync(path.join(os.tmpdir(), 'pqc-test-'))` for the PEM file.
* reducing our need for patching `connect` globally (less sure about this one)

core/packages/gax/test/test-application/src/pqc-test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  const originalTlsConnect = tls.connect;
+  let grpcSocket: tls.TLSSocket | undefined;
+  let restSocket: tls.TLSSocket | undefined;
+
+  let currentTestType: 'grpc' | 'rest' | 'none' = 'none';
+
+  (tls as any).connect = function (...args: any[]) {
+    const socket = originalTlsConnect.apply(this, args as any);
+    socket.on('secureConnect', () => {
```
westarle: you could test in here that the `socket.remoteport == port` (to make sure this doesn't affect other tests.)

core/packages/gax/test/test-application/src/pqc-test.ts
danieljbruce: I added a commit to check for that metadata just now.

core/packages/gax/test/test-application/src/pqc-test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * Tests Post Quantum Cryptography (PQC) using the specified CA cert and port.
+ * It verifies both gRPC and HTTP/REST clients by inspecting the negotiated TLS group.
+ * @param pemPath Path to the generated CA certificate file.
+ * @param port The port the TLS showcase server is listening on.
+ */
+async function testPqc(pemPath: string, port: number) {
+
+  // Verify the CA certificate file exists
```
westarle: Would it make sense for this to live in showcase-server.ts start, maybe having it return a Promise that we can wait on here?

core/packages/gax/test/test-application/jest.config.js
westarle, line 1: Personally, I'd rather separate changes like this from adding new tests.  Is there something in jest that is required for the PQC test to work, or are we rolling out jest according to some plan?  cc @pearigee

westarle (commented):
This is a really exciting test, thanks for adding it.  Interesting about the .20 support.  Please take my feedback with a grain of salt, I'm new in this area.

core/packages/gax/test/test-application/src/pqc-test.ts
hongalex: Yeah I think we can remove this and the line above.

core/packages/gax/test/test-application/src/pqc-test.ts
danieljbruce: start already returns a promise because it is async, but we could move this code there and then have the promise return the pem information.

start is used by quite a few tests so we just have to be careful to run this code only when a certificate is provided.

core/packages/gax/test/test-application/jest.config.js
danieljbruce: In retrospect I think we should add jest in a separate PR. Isolating these into 2 changes will make it much easier to debug/rollback later on if necessary.

core/packages/gax/test/test-application/src/pqc-test.ts
danieljbruce: I pushed changes for the first two. The problem with the last bullet point is that the TLS version is only available at a really low level at the socket and we need need to know it for the assertion check that TLS v1.3 was used. That said, checking for X25519MLKEM768 already ensures the right TLS version so I pushed a change to remove the patching as well.

core/packages/gax/test/test-application/src/pqc-test.ts
danieljbruce: pushed a change for this

core/packages/gax/test/test-application/src/pqc-test.ts
danieljbruce: assertion checks for the ciphers have been removed

core/packages/gax/test/test-application/src/pqc-test.ts
danieljbruce: I guess this is no longer relevant since we removed patching the socket

core/packages/gax/test/showcase-server/src/index.ts
```diff
@@ -93,6 +93,25 @@ export class ShowcaseServer {
     );
 
     this.server = childProcess;
+
+    if (resolvedCaCertPath) {
+      let pemExists = false;
+      for (let i = 0; i < 15; i++) {
```
westarle: is it possible in Node to reliably 'watch' for the file somehow and avoid this loop?

core/packages/gax/test/test-application/src/pqc-test.ts
```diff
@@ -22,56 +22,20 @@ import {grpc, GoogleAuth, googleAuthLibrary} from 'google-gax';
...
-        restSocket = socket as tls.TLSSocket;
-      }
-    });
-    return socket;
-  };
-
-  try {
-    // --- 1. gRPC PQC Test ---
-    currentTestType = 'grpc';
+  // --- 1. gRPC PQC Test ---
```
westarle: nit: when labeling things like this, I wonder if it's better to create a new function (named `grpcPqcTest` or something).

core/packages/gax/test/test-application/src/pqc-test.ts
```diff
@@ -137,6 +104,10 @@ async function testPqc(pemPath: string, port: number) {
       if (group) {
         negotiatedGroupRest = group;
       }
+      const supportedGroups = typeof res.headers.get === 'function' ? res.headers.get('x-showcase-tls-client-supported-groups') : (res.headers as any)['x-showcase-tls-client-supported-groups'];
```
westarle: nit: consider a comment to describe when this will be a function and when it is an object.

core/packages/gax/test/test-application/src/pqc-test.ts
danieljbruce: Move this into start

westarle (changes requested):
I think it looks good, I'd just like to see if you can firm up watching for the file without adding flakiness or a ton of complexity.

core/packages/gax/test/showcase-server/src/index.ts
danieljbruce: I didn't like this loop either and I thought about just having this code wait until the certificate is created. But we should definitely have a timeout because a test that just waits here if there is a problem with creating this certificate will be impossible to debug. Pushed changes to add a timeout.

core/packages/gax/test/showcase-server/src/index.ts
```diff
@@ -95,23 +95,35 @@ export class ShowcaseServer {
...
         }
-        await sleep(1000);
-      }
+
+        const timeoutId = setTimeout(() => {
+          watcher.close();
+          resolve(false);
+        }, 15000);
+
+        const watcher = fs.watch(dir, (eventType, triggeredFilename) => {
```
westarle, line 111: nit: if this triggers early (before showcase has written) it might read a partial file.  Consider a few ms wait here with a comment to give it some time.

core/packages/gax/test/showcase-server/src/index.ts
westarle: good idea to add a timeout!

westarle (approved):

westarle:
will this get executed by CI?  If not, we should consider adding it (in another PR is ok!)

core/packages/gax/test/test-application/src/pqc-test.ts
danieljbruce: Done

core/packages/gax/test/test-application/src/pqc-test.ts
danieljbruce: Done

core/packages/gax/test/showcase-server/src/index.ts
danieljbruce: Done

danieljbruce:
> will this get executed by CI? If not, we should consider adding it (in another PR is ok!)

I have had to verify this manually with `npm run test-application` each time. I don't think our gax testing infrastructure got copied over to the monorepo and other tests might have been missed too.

PR #8898, opened 2026-07-17
chore(tools): check in pnpm lockfiles and enforce frozen installs for generator tools

quirogas:
Running librarian install nodejs previously installed global tools (gapic-tools, gapic-node-processing) via loose pnpm add -g without lockfile enforcement. Because published npm packages resolve floating semver ranges (^, ~) dynamically at install time, fresh installations across different dates or CI runners produced non-deterministic transitive dependencies. Additionally, core/packages/tools was missing an explicit dependency on @babel/template.

This change adds @babel/template and @types/babel__template to core/packages/tools/package.json, checks in pnpm-lock.yaml files for core/packages/tools and core/packages/gapic-node-processing, and updates librarian.yaml build steps to execute pnpm install --frozen-lockfile.

For https://github.com/googleapis/librarian/issues/6638


librarian.yaml
```diff
@@ -20,16 +20,12 @@ sources:
     sha256: 503314c3f49e4de70138d35d5b9b12cca0c74d7c5ba34380895440b83aa45b94
 tools:
   pnpm:
-    - name: pnpm
```
quirogas, line 23: this is no longer needed (just in case).

danieljbruce (approved):

PR #8928, opened 2026-07-21
feat(firestore): decouple handwritten sdk wrapper from embedded gapic clients

quirogas:
Decouple the handwritten Firestore SDK wrapper from embedded GAPIC client implementations by transitioning to the standalone `@google-cloud/firestore-api` package. Standalone `-api` packages are now generated and versioned independently.

### Summary of Changes
- **GAPIC Dependency**: Adds runtime dependency on standalone `@google-cloud/firestore-api` (`^0.2.0`).
- **Runtime Proto Delegation**: Replaces static compiled protobuf runtime files in `dev/protos/*.js` (`firestore_v1`, `firestore_admin_v1`, `firestore_v1beta1`) with dynamic forwarding exports directly to `@google-cloud/firestore-api/build/protos/protos`, removing **76,572 lines of static JavaScript**.
- **Getter Variable Caching**: Applies local variable caching around lazy getters in `dev/src/index.ts` to ensure referential equality (`===`) and avoid redundant module require and `Object.assign` execution on repeated accesses.
- **Deprecation Guidance**: Adds explicit JSDoc `@deprecated` annotations directing consumers to import `@google-cloud/firestore-api` directly instead of relying on re-exported client properties (`Firestore.v1`).
- **Build Script Cleanup**: Removes legacy OwlBot configuration (`owlbot.py`, `.OwlBot.yaml`) and proto compilation scripts (`dev/protos/update.sh`).

### Why Local Types Remain in This Commit
Historically, `dev/protos/update.sh` compiled static protobuf TypeScript interfaces (`firestore_v1_proto_api.d.ts`) using custom post-processing: passing `--force-enum-string` to force protobuf enums into string literal unions (`'ASCENDING' | 'DESCENDING'`), and executing a Perl regex to replace 64-bit integer `Long` references with `string`. Modern GAPIC packages generate standard numeric enums and `number | Long`.

In this initial pull request, we intentionally retain `firestore_v1_proto_api.d.ts` while delegating `.js` runtime files to `@google-cloud/firestore-api`. This is done specifically in PR #1 to separate the mechanical work of separating the library dependency from the deeper call-site refactoring required to transition from string literals to native GAPIC enums. Once the entire stacked series merges, the final state will be complete reliance on `@google-cloud/firestore-api` as our single source of truth for all types and runtime models.

Internal: b/531788771

---
### 📚 Stack Navigation Index
1. **Layer 1 (Base Disentanglement & Runtime Forwarding)**: #8928
2. **Layer 2 (Complete Decoupling & Dead Code Cleanup)**: #9074
3. **Layer 3 (Automated Presubmit Bundle Size Guard)**: #9075
4. **Layer 4 (Call-Site Modernization & Eliminate Local Protos)**: #9076


handwritten/firestore/dev/src/index.ts
```diff
@@ -2043,17 +2048,15 @@ Object.defineProperty(module.exports, 'v1beta1', {
 Object.defineProperty(module.exports, 'v1', {
   // The v1 module is very large. To avoid pulling it in from static
   // scope, we lazy-load  the module.
-  get: () => require('./v1'),
+  get: () => {
+    const api = require('@google-cloud/firestore-api').v1;
```
MarkDuckworth: Will `@google-cloud/firestore-api` be published to NPM? If so, it may be preferred to deprecate this `v1` property.

handwritten/firestore/dev/src/index.ts
```diff
@@ -2028,8 +2027,14 @@ module.exports = Object.assign(module.exports, existingExports);
  */
 Object.defineProperty(module.exports, 'v1beta1', {
   // The v1beta1 module is very large. To avoid pulling it in from static
-  // scope, we lazy-load the module.
-  get: () => require('./v1beta1'),
+  // scope, we lazy-load  the module.
+  get: () => {
+    const api = require('@google-cloud/firestore-api').v1beta1;
```
MarkDuckworth: Will `@google-cloud/firestore-api` be published to NPM? If so, it may be preferred to deprecate this `v1beta1` property.

handwritten/firestore/dev/src/index.ts
```diff
@@ -2043,17 +2048,15 @@ Object.defineProperty(module.exports, 'v1beta1', {
...
 });
 
-/**
- * {@link Status} factory function.
- *
- * @private
- * @internal
- * @name Firestore.GrpcStatus
- * @type {function}
- */
```
MarkDuckworth, line 2056: Was this removal intentional? If so, can you clarify why?

handwritten/firestore/owlbot.py
```diff
@@ -1,246 +0,0 @@
...
-        "../protos"
-    )
-    s.replace(
-        "handwritten/firestore/types/v1beta1/firestore_client.d.ts",
-        r"../../protos",
-        "../protos"
-    )
-    s.replace(
-        "handwritten/firestore/types/protos/firestore_admin_v1_proto_api.d.ts",
-        'import Long = require\("long"\);',
```
MarkDuckworth, line 213: TODO we need to understand why this replacement was made and determine if we need a new workaround

handwritten/firestore/owlbot.py
```diff
@@ -1,246 +0,0 @@
...
-    )
-    s.replace(
-       "handwritten/firestore/dev/test/gapic_firestore_v1beta1.ts",
-       r"\.\./protos/protos.json",
-       "../protos/v1beta1.json"
-    )
-
-    # Mark v1beta1 as deprecated
-    s.replace(
-      "handwritten/firestore/dev/src/v1beta1/firestore_client.ts",
```
MarkDuckworth, line 161: We may need to migrate this deprecation to the new implmementation, or simply mark the v1beta1 property deprecated in index.ts?

handwritten/firestore/owlbot.py
```diff
@@ -1,246 +0,0 @@
...
-    # skip index, protos, package.json, and README.md
-    s.copy(v1_admin_library, "handwritten/firestore/dev", excludes=["package.json", "README.md", "src/index.ts", "src/v1/index.ts",
-        "tsconfig.json", "linkinator.config.json", "webpack.config.js"])
-    s.copy(v1beta1_library, "handwritten/firestore/dev", excludes=["package.json", "README.md", "src/index.ts", "src/v1beta1/index.ts",
-        "tsconfig.json", "linkinator.config.json", "webpack.config.js"])
-    s.copy(v1_library, "handwritten/firestore/dev", excludes=["package.json", "README.md", "src/index.ts", "src/v1/index.ts",
-        "tsconfig.json", "linkinator.config.json", "webpack.config.js"])
-
-    # Fix dropping of google-cloud-resource-header
-    # See: https://github.com/googleapis/nodejs-firestore/pull/375
```
MarkDuckworth, line 35: TODO we should verify that this is not a regression in the new code

handwritten/firestore/owlbot.py
```diff
@@ -1,246 +0,0 @@
...
-    s.replace(".jsdoc.js", r"protos", "build/protos", 1)
-
-    # Remove auto-generated packaging tests
-    os.system('rm -rf handwritten/firestore/dev/system-test/fixtures handwritten/firestore/dev/system-test/install.ts')
-
-    node.compile_protos_hermetic(relative_dir="handwritten/firestore")
-    os.unlink('handwritten/firestore/dev/protos.js')
-    os.unlink('handwritten/firestore/dev/protos.d.ts')
-    os.unlink('handwritten/firestore/dev/protos.json')
-    subprocess.run('handwritten/firestore/dev/protos/update.sh', shell=True)
```
MarkDuckworth, line 188: I'm assuming we should also remove this script update.sh which is compiling the protos?

handwritten/firestore/dev/src/index.ts
```diff
@@ -2028,8 +2027,14 @@ module.exports = Object.assign(module.exports, existingExports);
...
-  // scope, we lazy-load the module.
-  get: () => require('./v1beta1'),
+  // scope, we lazy-load  the module.
+  get: () => {
+    const api = require('@google-cloud/firestore-api').v1beta1;
+    const fn = function(this: any, ...args: any[]) {
+      return new (api.FirestoreClient as any)(...args);
+    };
+    return Object.assign(fn, api);
+  },
```
MarkDuckworth, line 2042: Wondering if we also need to update the related manually generated type exports in .types/firestore.d.ts lines 2990 - 3005

handwritten/firestore/dev/src/index.ts
```diff
@@ -2043,17 +2048,15 @@ Object.defineProperty(module.exports, 'v1beta1', {
...
   // The v1 module is very large. To avoid pulling it in from static
   // scope, we lazy-load  the module.
-  get: () => require('./v1'),
+  get: () => {
+    const api = require('@google-cloud/firestore-api').v1;
+    const fn = function(this: any, ...args: any[]) {
+      return new (api.FirestoreClient as any)(...args);
+    };
+    return Object.assign(fn, api);
+  },
```
MarkDuckworth, line 2067: Wondering if we also need to update the related manually generated type exports in .types/firestore.d.ts lines 2990 - 3005

handwritten/firestore/dev/src/index.ts
MarkDuckworth: Or simply remove this and the related type in ./types/...

handwritten/firestore/dev/src/index.ts
MarkDuckworth: Or simply remove this and the related type in ./types/...

handwritten/firestore/owlbot.py
MarkDuckworth: Perhaps even remove all of handwritten/firestore/dev/protos/... and handwritten/firestore/dev/protos.(js|d.ts|json)

handwritten/firestore/owlbot.py
```diff
@@ -1,246 +0,0 @@
...
-
-    # Copy types into types/
-    logger.debug("Running compile...")
-    shell.run(["npm", "run", "compile"], cwd="handwritten/firestore", hide_output=True)
-    s.copy("handwritten/firestore/build/src/v1/firestore*.d.ts", "handwritten/firestore/types/v1")
-    s.copy("handwritten/firestore/build/src/v1beta1/firestore_client.d.ts", "handwritten/firestore/types/v1beta1")
-    s.copy("handwritten/firestore/build/protos/firestore*.d.ts", "handwritten/firestore/types/protos")
```
MarkDuckworth, line 195: possibly also remove handwritten/firestore/types/v1, handwritten/firestore/types/v1beta1, handwritten/firestore/types/protos

handwritten/firestore/owlbot.py
```diff
@@ -1,246 +0,0 @@
...
-    shutil.rmtree(staging)
-
-
-# Copy template files
-common_templates = gcp.CommonTemplates()
-templates = common_templates.node_mono_repo_library(relative_dir="handwritten/firestore",
-    source_location="build/src", test_project="node-gcloud-ci"
-)
-
-s.copy(templates, destination="handwritten/firestore", excludes=[".eslintrc.json", ".kokoro/**/*", ".github/CODEOWNERS", "README.md"])
```
MarkDuckworth, line 238: Are the CODEOWNERS and README.md still getting generated with Librarian? Or are they even still useful with the new repo structure?

handwritten/firestore/dev/src/index.ts
quirogas: Yes, @google-cloud/firestore-api was published on NPM at https://www.npmjs.com/package/@google-cloud/firestore-api. We can add @deprecated JSDoc annotations to the v1 and v1beta1 properties to send users to the new API package, that way we keep the package the same, which is what we are trying to do for now.

handwritten/firestore/dev/src/index.ts
quirogas: Yes, good catch they do, thank you!

handwritten/firestore/dev/src/index.ts
quirogas: Nope it wasn't, restored.

handwritten/firestore/owlbot.py
quirogas: Yes, `gapic-generator-typescript` now handles call options natively, so this replacement is not needed. For instance, [the new generated client](https://github.com/googleapis/google-cloud-node/blob/main/packages/google-cloud-firestore-api/src/v1/firestore_client.ts#L1841-L1870) already returns (null, options), which aligns with how the [old client](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/firestore/dev/src/v1/firestore_client.ts#L1857-L1886) looks currently.

handwritten/firestore/owlbot.py
```diff
@@ -1,246 +0,0 @@
...
-import subprocess
-from pathlib import Path
-from synthtool import _tracked_paths
-import shutil
-from synthtool import shell
-from synthtool.log import logger
-
-logging.basicConfig(level=logging.DEBUG)
-
-staging = Path("owl-bot-staging")
```
quirogas, line 15: We no longer need to copy the gapic files into dev, as they are now live in their own separate library: https://www.npmjs.com/package/@google-cloud/firestore-api.

handwritten/firestore/owlbot.py
```diff
@@ -1,246 +0,0 @@
...
-    s.replace(
-       "handwritten/firestore/dev/test/gapic_firestore_admin_v1.ts",
-       r"\.\./protos/protos.json",
-       "../protos/admin_v1.json"
-    )
-    s.replace(
-       "handwritten/firestore/dev/test/gapic_firestore_v1beta1.ts",
-       r"\.\./protos/protos.json",
-       "../protos/v1beta1.json"
-    )
```
quirogas, line 157: This imports are handled in the generated package so we don't need to do this changes here anymore, for example https://github.com/googleapis/google-cloud-node/blob/182bda2c8e92b27c93cd36e586d490d84a793e2d/packages/google-cloud-firestore-api/src/v1/firestore_client.ts#L32

handwritten/firestore/owlbot.py
quirogas: Yes, we have added @deprecated annotations in index.ts and types/firestore.d.ts.

handwritten/firestore/owlbot.py
```diff
@@ -1,246 +0,0 @@
...
-      "// tslint:disable deprecation\n\nconst version",
-      1
-    )
-
-    os.rename("handwritten/firestore/dev/.gitignore", "handwritten/firestore/.gitignore")
-    os.rename("handwritten/firestore/dev/.eslintignore", "handwritten/firestore/.eslintignore")
-    os.rename("handwritten/firestore/dev/.mocharc.js", "handwritten/firestore/.mocharc.js")
-    os.rename("handwritten/firestore/dev/.jsdoc.js", "handwritten/firestore/.jsdoc.js")
-    os.rename("handwritten/firestore/dev/.prettierrc.js", "handwritten/firestore/.prettierrc.js")
-    os.unlink("handwritten/firestore/dev/.eslintrc.json")
```
quirogas, line 177: The package configuration is now static; therefore, any changes should be applied directly there rather than through the generator.

handwritten/firestore/owlbot.py
quirogas: yes to both.

handwritten/firestore/owlbot.py
quirogas: yes!

handwritten/firestore/owlbot.py
quirogas: No, they are not. CODEOWNERS will be managed by the Node team, and the README should be owned and managed by the Firestore team.

handwritten/firestore/owlbot.py
```diff
@@ -1,246 +0,0 @@
...
-
-    s.replace(".jsdoc.js", r"protos", "build/protos", 1)
-
-    # Remove auto-generated packaging tests
-    os.system('rm -rf handwritten/firestore/dev/system-test/fixtures handwritten/firestore/dev/system-test/install.ts')
-
```
quirogas, line 183: This is no longer needed as we don't need to do any more synthtool moving and deleting.

handwritten/firestore/dev/src/index.ts
quirogas: same as the comment for v1beta1.

handwritten/firestore/dev/src/index.ts
quirogas: Yes, we do. We now re-export the types by importing them from the fully generated client.

handwritten/firestore/dev/src/index.ts
```diff
@@ -2044,6 +2045,7 @@ Object.defineProperty(module.exports, 'v1beta1', {
  * @internal
  * @name Firestore.v1
  * @type {function}
+ * @deprecated Use the top-level Firestore client instead.
```
MarkDuckworth: For users of Firestore.v1 or Firestore.v1beta1, their use case likely requires and API only exposed by the GAPIC. I'd recommend this deprecation message be:

> * `@deprecated Prefer the top-level Firestore client instead of the Firestore.v1 client. If your use case requires the Firestore.v1 client, this is now being published as a separate package at @google-cloud/firestore-api.`

handwritten/firestore/types/firestore.d.ts
```diff
@@ -2988,21 +2988,17 @@ declare namespace FirebaseFirestore {
...
    * @deprecated Use v1 instead.
    */
-  export const v1beta1: {
-    FirestoreClient: typeof import('./v1beta1/firestore_client').FirestoreClient;
-  };
+  export const v1beta1: typeof import('@google-cloud/firestore-api').v1beta1;
   /**
    * The v1 Veneer clients. These clients provide access to the Firestore Admin
    * API and the underlying Firestore v1 RPCs.
+   * @deprecated Use the top-level Firestore client instead.
```
MarkDuckworth: Same comment regarding the deprecation note.

handwritten/firestore/owlbot.py
MarkDuckworth: Not sure if you wanted to remove handwritten/firestore/dev/protos/, but it doesn't look like it was removed in this PR.

handwritten/firestore/owlbot.py
MarkDuckworth: same, doesn't look like this was removed

handwritten/firestore/owlbot.py
MarkDuckworth: I think we need to add 'long' as a dependency to packages/google-cloud-firestore-api. Reason: https://github.com/googleapis/nodejs-firestore/issues/2322. There may be a more elegant solution to this.

quirogas: Done! we've addressed this directly at the package level by adding long to the dependencies in package.json ([see](https://github.com/googleapis/google-cloud-node/blob/d5b66ff18167a4771f27f0d3e83528b6daec1933/packages/google-cloud-firestore-api/package.json#L79))

handwritten/firestore/owlbot.py
quirogas: We are keeping dev/protos/*.d.ts in this PR and we are removing it in [PR4](https://github.com/googleapis/google-cloud-node/pull/9076) because we had to do some additional work to cleanly remove them. Deleting the local declarations immediately triggers 184 compilation errors because update.sh was using --force-enum-string which and @google-cloud/firestore-api generates standard numeric enum objects which broke internal query dictionaries that passed string literals.

handwritten/firestore/owlbot.py
quirogas: same as https://github.com/googleapis/google-cloud-node/pull/8928#discussion_r3660003397

handwritten/firestore/dev/src/index.ts
quirogas: ack

MarkDuckworth (approved):
Looks good in context of the stacked PRs

quirogas:
Labeling as "do not merge" to make sure we merge all stacked prs before merging this one.

PR #8933, opened 2026-07-22
feat(Spanner): modify default maxConcurrentStreamsLowWatermark value …

alkatrivedi:
This PR modifies the default value of `maxConcurrentStreamsLowWatermark` from 100 to 25. This is to make the value consistent with Java client. As part of this change one channel can have at max 25 concurrent request at time, once the threshold will reach the channels will get scale up.


handwritten/spanner/package.json
```diff
@@ -139,12 +139,14 @@
   },
   "overrides": {
     "gcp-metadata": "^7.0.1",
-    "@sinonjs/fake-timers": "15.2.1"
+    "@sinonjs/fake-timers": "15.2.1",
+    "@eslint-community/eslint-utils": "<4.10.0"
```
alkatrivedi, line 143: need this change since, the package `@eslint-community/eslint-utils` (which is pulled in automatically by `eslint` via `gts@6.0.2`) recently released `version 4.10.0`.
In `version 4.10.0`, they updated their engines field in package.json to require Node.js `^20.19.0` || `^22.13.0` || `>=24`, effectively dropping support for `Node 18`.
Our GitHub Actions presubmit job is running Node `v18.20.8`. Because `pnpm install` is being executed with the `--engine-strict` flag, `pnpm` strictly enforces these Node engine requirements and immediately fails the installation.

rahul2393 (approved):

PR #8934, opened 2026-07-22
feat(bigtable): split gapic and proto usage from the handwritten library into bigtable-api

feywind:
This is in the effort to remove OwlBot from the standard repo path. A new package (bigtable-api) takes care of what OwlBot used to do (shuffle in new protos and generated gapic classes).

Internal: b/531788771


handwritten/bigtable/src/admin/table.ts
```diff
@@ -15,9 +15,19 @@
...
+import {protos} from '@google-cloud/bigtable-api';
+import google = protos.google;
+
+const PROTO_DIR = path.join(
+  path.dirname(require.resolve('@google-cloud/bigtable-api')),
+  '..',
+  'protos',
+);
+const JSON_PATH = path.join(PROTO_DIR, 'protos.json');
+const JSON_PROTOS = fs.readFileSync(JSON_PATH, 'utf-8');
```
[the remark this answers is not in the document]
feywind: Hah, that `require(JSON_PATH)` explicitly does not work; it has to be a const string.

feywind:
The PR might be a bit gross until I debug the resource leaks.

handwritten/bigtable/src/admin/v2/index.ts
quirogas, line 1: Can we delete this file or directory altogether and add the export logic to the `index.ts` file in the parent directory?

handwritten/bigtable/src/v2/index.ts
quirogas, line 1: Same as  admin/v2/index.ts. Could we move this export to the top-level file instead?

quirogas (commented):
PR looks good to me, Let's open follow up clean up bugs for:

- removing proto files and addressing test_proxy.proto
- cleanup tests, specially tests that depend on proto files
- transition to the .json definition of the service file as opposed to using the proto files.
- remove logs
- split test changes into a separate PR.
- remove utility script from publishing (if not already)
 
Feel free to reference this commnet to create the issues initially or create new ones altogether, up to you.

quirogas (approved):

handwritten/bigtable/src/v2/index.ts
feywind: I'm going to get this one merged and then open up some tweak PRs.

PR #8937, opened 2026-07-22
test(firestore): kokoro to gcb

MarkDuckworth:
Fixes issues in https://github.com/googleapis/google-cloud-node/pull/8700. 

Because Datastore SDK tests also run against Firestore database instances (in Datastore mode) within long-door-651, naming collisions occur between the database instances used for Firestore and Datastore tests. To resolve this, Firestore CI will utilize uniquely named databases: firestore-standard and firestore-enterprise

We can merge this PR, or merge this branch to `kokoro-to-gcb-firestore`, and then merge https://github.com/googleapis/google-cloud-node/pull/8700.


danieljbruce:
This is looking good, but I would also suggest that we just skip the small number of tests that seem to be failing shown in https://paste.googleplex.com/4859917847625728. The GCB check will be a lot more reliable if we can expect it to pass under normal circumstances.

danieljbruce (approved):

PR #8953, opened 2026-07-23
fix(gapic-node-processing): preserve index.ts exports and merge clients across multi-api staging subdirectories

quirogas:
Combine duplicate version barrel export files (src/v*/index.ts) using TypeScript AST parsing to prevent sequential gapic-generator passes from overwriting exports. Aggregate client class names across all staging subdirectories per version to ensure top-level index.ts re-exports all generated clients.

For https://github.com/googleapis/librarian/issues/6969

### DIFF
--- a/core/packages/gapic-node-processing/src/library.ts
+++ b/core/packages/gapic-node-processing/src/library.ts
@@ -248,12 +248,9 @@ export class LibraryConfig {
 }
 
 function alphaOrBetaPrecedence(preRelease: string): '' | 'beta' | 'alpha' {
-  console.log(preRelease);
   if (preRelease.startsWith('beta')) {
-    console.log('beta');
     return 'beta';
   } else if (preRelease.startsWith('alpha')) {
-    console.log('alpha');
     return 'alpha';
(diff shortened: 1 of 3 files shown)


core/packages/gapic-node-processing/src/library.ts
```diff
@@ -148,23 +142,30 @@ export class LibraryConfig {
...
-          });
+          if (!versionMap.has(version)) {
+            versionMap.set(version, new Set());
+          }
+          const set = versionMap.get(version)!;
+          for (const match of clientsRegexMatch) {
+            set.add(match[1]);
+          }
+        } catch (err) {
+          // ignore if file does not exist
```
feywind: Is there a worry here about catching an error besides something out of `fs.readFile`?

feywind (approved):
One quick double-check comment.

core/packages/gapic-node-processing/src/library.ts
quirogas: Better safe than sorry -- good catch.

PR #8955, opened 2026-07-23
fix(ci): fix discovery.yaml by installing root dependencies

### LINKED ISSUE #8954: ci(bigquery): discovery.yaml workflow fails during npm run fix due to missing root dependencies

### Library Name

google-cloud-node (CI/CD Infrastructure) / @google-cloud/bigquery

N/A - This is a bug with the repository's GitHub Actions CI/CD pipeline (discovery.yaml), not an API issue.

https://github.com/googleapis/google-cloud-node/actions/workflows/discovery.yaml (Example of a failing run: https://github.com/googleapis/google-cloud-node/actions/runs/29830371243)

1. Navigate to the Actions tab in the google-cloud-node repository.
2. Select the Update Discovery Generated Types (discovery.yaml) workflow.
3. Observe that every scheduled run has failed for the past several months.
4. Click into any recent failed run (e.g., the run from July 21st).
5. Expand the Run cd handwritten/bigquery && npm run fix step.
6. Observe the ESLint error: ESLint couldn't find the config "./node_modules/gts" to extend from.

The discovery.yaml GitHub Action is failing consistently during the npm run fix step. It throws an ESLint configuration error because it cannot find the gts config in ./node_modules/gts. This happens because the workflow only runs npm install inside the handwritten/bigquery directory, leaving the repository root node_modules empty. When ESLint runs, it reads the root .eslintrc.json, tries to resolve ./node_modules/gts, and crashes.

I expected the discovery.yaml workflow to successfully run npm run fix, generate the updated Discovery types, and open a pull request titled chore: update types from Discovery.

Proposed Fix: Add a step to run npm install at the repository root before running the fix command in the bigquery directory

The purpose of the discovery.yaml workflow is to automate the generation of Discovery types for handwritten libraries (specifically BigQuery) so that OwlBot is no longer needed for this task. Because this workflow is broken, the types are not being updated by the workflow.

This issue is currently blocking PR #8936 (which attempts to remove OwlBot config for BigQuery). Removing OwlBot while this Action is broken would completely halt type updates for BigQuery, as OwlBot running as a post-processor on release PRs is currently the only mechanism successfully updating these types.

westarle:
This fixes an issue where ESLint could not find the gts config because the root node_modules was empty, and ignores system-test/fixtures which are not included in the tsconfig.

Fixes #8954  🦕

### DIFF
--- a/.github/workflows/discovery.yaml
+++ b/.github/workflows/discovery.yaml
@@ -19,8 +19,7 @@ jobs:
       - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
         with:
           version: ^10.0.0
-      # Install root deps for ESLint
-      - run: pnpm install
+
       # Install all deps, including dev dependencies.
       - run: cd handwritten/bigquery && pnpm install
       # Generate types
--- a/.github/workflows/discovery.yaml
+++ b/.github/workflows/discovery.yaml
@@ -15,15 +15,18 @@ jobs:
           persist-credentials: false
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 16
+          node-version: 22
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
+        with:
+          version: ^10.0.0
       # Install root deps for ESLint
-      - run: npm install
+      - run: pnpm install
       # Install all deps, including dev dependencies.
-      - run: cd handwritten/bigquery && npm install
+      - run: cd handwritten/bigquery && pnpm install
       # Generate types
-      - run: cd handwritten/bigquery && npm run types
+      - run: cd handwritten/bigquery && pnpm run types
       # Fix formatting
-      - run: cd handwritten/bigquery && npm run fix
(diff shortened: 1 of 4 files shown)


.github/workflows/discovery.yaml
```diff
@@ -16,6 +16,8 @@ jobs:
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: 16
```
pearigee: Nit/Optional: May be worth bumping this to 22 or 24 while we are here.

pearigee (approved):

.github/workflows/discovery.yaml
```diff
@@ -16,6 +16,8 @@ jobs:
...
+      # Install root deps for ESLint
+      - run: npm install
```
pearigee: In the future, we will likely want to use PNPM to ensure our global settings/lockfile are respected.

westarle: done -- also pnpm was strict and didn't like clashing deps here and in root so I updated the bq eslintrc to be its own root for now.

.github/workflows/discovery.yaml
westarle: done

PR #8966, opened 2026-07-24
chore: fix test breakage in error-reporting

pearigee:
## Why I am addressing this

I am trying to fix all of the current outstanding test breakages blocking the merge of our new Bun CI (https://github.com/googleapis/google-cloud-node/pull/8904). To verify this PR, we are running ALL unit tests in the monorepo. This has revealed a handful of existing breakages, including this one.

---

## The problem

In `@hapi/hapi` (and modern `@types/node` / HTTP header type definitions), `req.headers` elements are typed as `unknown` (e.g. Record<string, unknown>). Passing header values like `req.headers['user-agent']`, `req.headers.referrer`, or `extractRemoteAddressFromRequest(req)` directly into `getSingleHeader(val: string | string[] | undefined)` caused `TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string | string[] | undefined'.`

---

## The fix

Update the internal implementation to correctly reflect the real types and add a new test to verify functionality.

I don't love the use of `as` casting in the tests, but mocking a type-compliant `hapi.Request` object is surprisingly difficult (it appears to require a running server to inject dependencies).  I spoke with Gemini about this and it recommended casting a bare JS object (like the existing tests). An alternative would be to update the implementation to have a less strict type, but I would prefer to keep the minimum diff.

### DIFF
--- a/handwritten/error-reporting/src/request-extractors/hapi.ts
+++ b/handwritten/error-reporting/src/request-extractors/hapi.ts
@@ -26,10 +26,12 @@ import * as hapi from '@hapi/hapi';
  * @returns {Number} - Either an HTTP status code or -1 in absence of an
  *  extractable status code.
  */
-function attemptToExtractStatusCode(req: hapi.Request) {
+function attemptToExtractStatusCode(
+  req: hapi.Request | Record<string, unknown>,
+) {
   // TODO: Handle the cases where `req.response` and `req.response.output` are
   // `null` in this function
-  if (typeof req.response === 'object') {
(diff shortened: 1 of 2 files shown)


pearigee:
Note, in response to the Gemini feedback, the lines it's reporting on are not modified by the PR. My preference here is to maintain the minimum possible diff to resolve the test failure.

shivanee-p (approved):
`it appears to require a running server to inject dependencies` <-- that's wild for a mock 👀 

Ty for updating the unit tests to reflect this

PR #8968, opened 2026-07-24
fix(ci): improve changed file detection in linter script

shivanee-p:
Fixes an issue where `./bin/linter.mjs` skipped TypeScript checks in CI pull request workflows because it's incorrectly detecting no TS file changes. Updates linter logic to be hermetic.

It looks like the CI was comparing "head" to "head". It was running `git diff FETCH_HEAD -- *.ts`, but we need to fetch from origin before doing that. Previously, the linter was going in order of the `refsToTry` array to check which reference existed in order to diff against it:
1. `main`, `upstream/main`, and `origin/main` failed
2. `'FETCH_HEAD'`, existed and pointed to the checked-out PR commit (`HEAD`).
3. Running `git diff --name-only FETCH_HEAD -- *.ts` compared `HEAD` against itself, which succeeded with an exit code of `0` and returned an empty file list (`""`)

Fixes:
- Removed git fetch calls so the script relies solely on local Git history provided by actions/checkout.
- Added `STRICT` mode to ./bin/linter.mjs, configured in presubmit.yaml
- Uses git merge-base to accurately identify changed files when both main and the branch have moved.
- Fails explicitly in CI if GITHUB_BASE_REF is missing or invalid, or if the GIT_DIFF_ARG is missing in STRICT mode
- Streamlined reference resolution for local development and added logging for the chosen base reference.

### DIFF
--- a/bin/linter.mjs
+++ b/bin/linter.mjs
@@ -94,7 +94,6 @@ function getDiffFiles(ref) {
 
 /**
  * Returns a list of changed TypeScript files comparing against target branches/references.
- * Fully hermetic (uses local git references and merge-base with no network calls).
  */
 function getChangedFiles() {
   const isCI = Boolean(process.env.CI || process.env.GITHUB_ACTIONS || process.env.GITHUB_BASE_REF);
@@ -117,7 +116,7 @@ function getChangedFiles() {
       // Continue with fallback refs if branch detection fails
     }
 
-    refsToTry = currentBranch === 'main'
+    refsToTry = (currentBranch === 'main')
       ? ['origin/main', 'upstream/main', 'HEAD~1']
       : ['upstream/main', 'origin/main', 'main'];
   }
--- a/bin/linter.mjs
+++ b/bin/linter.mjs
@@ -112,7 +112,6 @@ function getChangedFilesStrict() {
  */
 function getChangedFiles() {
   const base = process.env.GITHUB_BASE_REF || 'main';
-
   const refsToTry = [
     `origin/${base}...HEAD`,
     `${base}...HEAD`,
--- a/bin/linter.mjs
+++ b/bin/linter.mjs
@@ -64,72 +64,83 @@ function runGit(args, options = {}) {
 }
 
 /**
+ * Helper to get modified/added TypeScript files against a given git ref.
+ * Uses merge-base when possible to compare against the common ancestor (e.g. when base and branch have both moved).
+ */
+function getDiffFiles(ref) {
+  let diffTarget = ref;
+  try {
+    const mergeBase = runGit(['merge-base', ref, 'HEAD']).trim();
+    if (mergeBase) {
+      diffTarget = mergeBase;
+    }
+  } catch {
+    // If merge-base fails (e.g. shallow clone or invalid ref), fall back to using ref directly
+  }
+
+  const output = runGit([
+    'diff',
+    '--name-only',
+    '--diff-filter=ACMRT',
+    diffTarget,
+    '--',
+    '*.ts',
+  ]);
+  return output
+    .split('\n')
+    .map(f => f.trim())
+    .filter(f => f.length > 0 && existsSync(f));
+}
+
+/**
  * Returns a list of changed TypeScript files comparing against target branches/references.
  */
 function getChangedFiles() {
-  const base = process.env.GITHUB_BASE_REF || 'main';
+  const isCI = Boolean(process.env.CI || process.env.GITHUB_ACTIONS || process.env.GITHUB_BASE_REF);
 
-  // Attempt to fetch the base branch and set origin/${base} so it doesn't compare against itself
-  if (process.env.GITHUB_BASE_REF) {
+  if (isCI) {
+    const baseRef = process.env.GITHUB_BASE_REF;
+    if (!baseRef) {
+      throw new Error('Running in CI but GITHUB_BASE_REF environment variable is not set.');
+    }
     try {
-      runGit([
-        'fetch',
-        'origin',
-        `+refs/heads/${base}:refs/remotes/origin/${base}`,
-        '--depth=1',
-      ]);
+      const files = getDiffFiles(baseRef);
+      console.log(`Comparing against base reference: ${baseRef}`);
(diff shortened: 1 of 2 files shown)


bin/linter.mjs
```diff
@@ -68,11 +68,28 @@ function runGit(args, options = {}) {
  */
 function getChangedFiles() {
   const base = process.env.GITHUB_BASE_REF || 'main';
+
+  // Attempt to fetch the base branch and set origin/${base} so it doesn't compare against itself
```
westarle: I'd prefer to make the script hermetic and have our CI configuration produce the right environment for it to be effective.  Can we avoid mutation in the script and have the workflow use `actions/checkout`

bin/linter.mjs
```diff
@@ -68,11 +68,28 @@ function runGit(args, options = {}) {
  */
 function getChangedFiles() {
   const base = process.env.GITHUB_BASE_REF || 'main';
```
westarle, line 114: I think if the script is running in CI, it should fail here if GITHUB_BASE_REF isn't set.

bin/linter.mjs
```diff
@@ -68,11 +68,28 @@ function runGit(args, options = {}) {
...
+        'origin',
+        `+refs/heads/${base}:refs/remotes/origin/${base}`,
+        '--depth=1',
+      ]);
+    } catch {
+      // Continue if network fetch fails or remote does not exist
+    }
+  }
+
   const refsToTry = [
```
westarle, line 115: It might be simpler to say:
 
1. if it's in CI then use exactly process.env.GITHUB_BASE_REF or fail
2. if the local branch is "main" then use HEAD~1
3. for other local branches, use upstream/main, origin/main and then main and print a line that says what we chose.

bin/linter.mjs
```diff
@@ -68,11 +68,28 @@ function runGit(args, options = {}) {
...
   const base = process.env.GITHUB_BASE_REF || 'main';
+
+  // Attempt to fetch the base branch and set origin/${base} so it doesn't compare against itself
+  if (process.env.GITHUB_BASE_REF) {
+    try {
+      runGit([
+        'fetch',
+        'origin',
+        `+refs/heads/${base}:refs/remotes/origin/${base}`,
+        '--depth=1',
```
westarle: if `main` has moved and the branch has moved, the common ancestor might be higher than 1; worth considering here or as `fetch-depth` in `actions/checkout`.

bin/linter.mjs
shivanee-p: Ack. Thanks for the callout. I've updated it and the script doesn't make any network calls anymore.

bin/linter.mjs
shivanee-p: Good point. I've updated getDiffFiles() to resolve HEAD using `merge-base`

bin/linter.mjs
```diff
@@ -63,56 +63,144 @@ function runGit(args, options = {}) {
...
+    '*.ts',
+  ]);
+  return output
+    .split('\n')
+    .map(f => f.trim())
+    .filter(f => f.length > 0 && existsSync(f));
+}
+
+/**
+ * Retrieves the GIT_DIFF_ARG from CLI flags or environment variable.
```
westarle: I would just choose one or the other (flag or env), I think @bshaffer went with / reused the env variable so maybe let's just use that

bin/linter.mjs
westarle: I would introduce the fork here to avoid refactoring `getChangedFiles()`

You would populate `isStrict` and then:

```js
if (isStrict) {
  gitDiffArg = verifyDiffArg();
  getChangedFilesStrict(gitDiffArg);
} else {
  getChangedFiles();
}
```

and leave all the broken behavior in `getChangedFiles()` alone (don't try to make it work for CI, e.g.)

westarle (changes requested):

bin/linter.mjs
shivanee-p: Ah sorry, I was looking at an earlier commit of his where he used the flag. Just changed it to match!

bin/linter.mjs
```diff
@@ -26,7 +26,13 @@ const tsconfigCache = new Map();
 // --- Main Runner (Entry Point) ---
 async function run() {
   try {
-    const changedTsFiles = getChangedFiles();
+    const isStrict = Boolean(process.env.STRICT || process.argv.includes('--strict'));
```
westarle: nit: I would just choose one of these (env or arg) and go with it -- same for gitDiffArg.

Seems like the env is easier for the git diff arg, and the boolean flag arg is easier for strict.

bin/linter.mjs
```diff
@@ -63,56 +69,96 @@ function runGit(args, options = {}) {
...
+      'Please set the GIT_DIFF_ARG environment variable or provide --git-diff-arg <arg>.'
+    );
+  }
+
+  console.log(`Strict mode enabled. Comparing using GIT_DIFF_ARG: ${gitDiffArg}`);
+
+  const args = gitDiffArg.trim().split(/\s+/);
+
+  try {
+    runGit(['diff', '--quiet', ...args]);
```
westarle: maybe we don't need this if we immediately call 'diff' below -- maybe you can wrap the actual diff in a `try` and avoid extra work?

bin/linter.mjs
```diff
@@ -63,56 +69,96 @@ function runGit(args, options = {}) {
...
+    .split('\n')
+    .map(f => f.trim())
+    .filter(f => f.length > 0 && existsSync(f));
+}
+
 /**
- * Returns a list of changed TypeScript files comparing against target branches/references.
+ * Returns a list of changed TypeScript files comparing against target branches/references when not in strict mode.
  */
 function getChangedFiles() {
```
westarle, line 113: nit: I would remove all the edits to `getChangedFiles()` in this PR.

westarle (approved):

westarle:
please clean up params (args or env) and submit.

bin/linter.mjs
shivanee-p: Ack

PR #8974, opened 2026-07-27
docs: update README to provide users with guidance to turn on PQC encryption

danieljbruce:
## Description

Users need to know that they must upgrade their Node version to 22.20 or higher to ensure their network traffic is secure against post-quantum decryption algorithms like Shor's algorithm. So this PR updates the README to ensure customers are well informed.

## Impact

Informs users so that they can better encrypt their network traffic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,9 @@ The dist-tags follow the naming convention `legacy-(version)`.
 For example, `{{ metadata['lib_install_cmd'] }}@legacy-10` installs client libraries
 for versions compatible with Node.js 10.
 
+Network traffic is only secure against post-quantum decryption algorithms for
+users running Node version 22.20+.
+
 ## Versioning
 
 Our libraries follow [Semantic Versioning][semver].
--- a/README.md
+++ b/README.md
@@ -292,9 +292,8 @@ for versions compatible with Node.js 10.
 ## Post-quantum Cryptography
 
 A minimum Node version of >=22.20 is required for post-quantum cryptography to
-be used. When running an application with a supported node version and making
-requests with a client library, network traffic will automatically be secure
-against post-quantum decryption methods.
+be used. When running an application with a supported node version, the client
+library will attempt to negotiate and use post-quantum cryptography for all requests.
 
 ## Versioning
 
--- a/README.md
+++ b/README.md
@@ -289,8 +289,12 @@ The dist-tags follow the naming convention `legacy-(version)`.
 For example, `{{ metadata['lib_install_cmd'] }}@legacy-10` installs client libraries
 for versions compatible with Node.js 10.
 
-Network traffic is only secure against post-quantum decryption algorithms for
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -289,6 +289,9 @@ The dist-tags follow the naming convention `legacy-(version)`.
 For example, `{{ metadata['lib_install_cmd'] }}@legacy-10` installs client libraries
 for versions compatible with Node.js 10.
 
+Network traffic is only secure against post-quantum decryption algorithms for
+users running Node version 22.20+.
```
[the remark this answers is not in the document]
danieljbruce: This statement doesn't appear to be true. I tried this version of Node out on https://github.com/googleapis/google-cloud-node/compare/main...pqc-version-check-again and the test fails so I think 22.20 is the right version.

image

pearigee:
Should we give it its on heading?

I think the key things people need to know:

* What is the minimum node version?
* Is it enabled automatically? If not, how do I turn it on?
* If automatic, what triggers it?

Suggestion (feel free to ignore/rewrite accordingly):
```
## Post-quantum Cryptography

A minimum Node version of >=22.20 is required for post-quantum cryptography to be used. If using a supported runtime, traffic will be automatically upgraded when [condition].
```

danieljbruce:
> Should we give it its on heading?
> 
> I think the key things people need to know:
> 
> * What is the minimum node version?
> * Is it enabled automatically? If not, how do I turn it on?
> * If automatic, what triggers it?
> 
> Suggestion (feel free to ignore/rewrite accordingly):
> 
> ```
> ## Post-quantum Cryptography
> 
> A minimum Node version of >=22.20 is required for post-quantum cryptography to be used. If using a supported runtime, traffic will be automatically upgraded when [condition].
> ```

I modified this slightly just to capture the fact that it is the client library requests specifically that are secure when the user is importing a client library.

README.md
```diff
@@ -289,6 +289,13 @@ The dist-tags follow the naming convention `legacy-(version)`.
 For example, `{{ metadata['lib_install_cmd'] }}@legacy-10` installs client libraries
 for versions compatible with Node.js 10.
 
+## Post-quantum Cryptography
+
+A minimum Node version of >=22.20 is required for post-quantum cryptography to
+be used. When running an application with a supported node version and making
+requests with a client library, network traffic will automatically be secure
+against post-quantum decryption methods.
```
westarle: I think we should adjust the langauge.

```suggestion
be used. When running an application with a supported node version, the client
library will attempt to negotiate and use post-quantum cryptography for all requests.
```

I think it would be helpful to get a review and maybe a link from @viacheslav-rostovtsev

westarle: actually, better a review from @suzmue

westarle (approved):
I think we can submit and then change it if we get feedback from @suzmue -- thanks!

PR #8982, opened 2026-07-28
fix(storage): pass signingEndpoint to URLSigner in file.getSignedUrl

### LINKED ISSUE #8829: Config option `signingEndpoint` is not used when passed to `getSignedUrl`
This started off as a [discussion](https://github.com/googleapis/google-cloud-node/discussions/8795) but I've now made a well-formatted issue since I believe this is an actual bug.

### Library Name

@google-cloud/storage

I'm implementing an integration test so trying the api doesn't really make sense and I couldn't find the corresponding endpoints to try.

https://gist.github.com/MisanthropicBit/0c2283cb9e8de5e195361a75793d075f

Setting up a full reproducible example would require me to set up authentication that would most likely not work for the tester's environment. I think perhaps running one of the tests in https://github.com/googleapis/google-cloud-node/blob/main/handwritten/storage/test/file.ts might make more sense, modifying a test to temporarily call into the code that I have outlined below.

Otherwise, there is a small gist you can run provided you've set up the authentication appropriate for your system.

1. Patch the lines [here](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/storage/src/signer.ts#L361-L364) with a `console.log(config.signingEndpoint)` to see that the value is `undefined`.
2. Run the script with `ts-node get-signed-url.ts`.

I've confirmed the issue in my tests (where authentication is done through test servers so the code will be allowed to eventually run the code in question).

I'm implementing an integration test for some code that uses signed urls (via `file.getSignedUrl`) and I'm testing with a bunch of test servers. In order to test signed urls I'm passing a `signingEndpoint` that points to a local test server that is started up when the test starts.

The problem is that the `file.getSignedUrl` seems to have a bug where that config option is not passed down through the code ([here](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/storage/src/file.ts#L3230-L3239)). If you follow the code down through [`URLSigner.getSignedUrl`](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/storage/src/signer.ts#L158-L223) to [`URLSigner.getSignedUrlV4`](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/storage/src/signer.ts#L269-L379), you'll eventually get to [this code](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/storage/src/signer.ts#L361-L364) which looks for the `signingEndpoint` in the config but never finds it. Effectively, the config option is ignored. The `URLSigner.getSignedUrlV2` suffers from the same issue although I'm only using version 4.

I think this is a bug because the [`GetSignedUrlConfig`](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/storage/src/file.ts#L152-L167) interface includes a `signingEndpoint` option which is never used.

I expected the provided `signingEndpoint` config option to be used in place of a default one.

A simple solution would be the following diff:

```diff
    const signConfig: SignerGetSignedUrlConfig = {
      method,
      expires: cfg.expires,
      accessibleAt: cfg.accessibleAt,
      extensionHeaders,
      queryParams,
      contentMd5: cfg.contentMd5,
      contentType: cfg.contentType,
      host: cfg.host,
+     signingEndpoint: cfg.signingEndpoint,
    };
```

If this is not a bug, what is the point of `signingEndpoint`? Can I redirect the signing process in another fashion? My code will run in a GCE environment so this is only an issue for testing and if someone else needs to use another signing endpoint in the future.

Not relevant. I didn't expect this to happen at all.

thiyaguk09:

Fixes #8829

### DIFF
--- a/handwritten/storage/src/file.ts
+++ b/handwritten/storage/src/file.ts
@@ -3236,6 +3236,7 @@ class File extends ServiceObject<File, FileMetadata> {
       contentMd5: cfg.contentMd5,
       contentType: cfg.contentType,
       host: cfg.host,
+      signingEndpoint: cfg.signingEndpoint,
     };
 
     if (cfg.cname) {
--- a/handwritten/storage/test/file.ts
+++ b/handwritten/storage/test/file.ts
@@ -3794,7 +3794,26 @@ describe('File', () => {
           contentType: config.contentType,
           cname: CNAME,
           virtualHostedStyle: true,
+          signingEndpoint: undefined,
         });
+        done();
+      });
+    });
+
+    it('should pass signingEndpoint to URLSigner', done => {
+      const signingEndpoint = 'https://my-endpoint.com';
+      const config = {
+        signingEndpoint,
+        ...SIGNED_URL_CONFIG,
+      };
+
+      file.getSignedUrl(config, (err: Error | null) => {
+        assert.ifError(err);
+        const getSignedUrlArgs = signerGetSignedUrlStub.getCall(0).args;
+        assert.strictEqual(
+          getSignedUrlArgs[0]['signingEndpoint'],
+          signingEndpoint
+        );
         done();
       });
     });
--- a/handwritten/storage/test/file.ts
+++ b/handwritten/storage/test/file.ts
@@ -3803,8 +3803,8 @@ describe('File', () => {
     it('should pass signingEndpoint to URLSigner', done => {
       const signingEndpoint = 'https://my-endpoint.com';
       const config = {
-        signingEndpoint,
         ...SIGNED_URL_CONFIG,
+        signingEndpoint,
       };
 
       file.getSignedUrl(config, (err: Error | null) => {
--- a/handwritten/storage/src/signer.ts
+++ b/handwritten/storage/src/signer.ts
@@ -25,7 +25,7 @@ type GoogleAuthLike = Pick<GoogleAuth, 'getCredentials' | 'sign'>;
  * @deprecated Use {@link GoogleAuth} instead
  */
 export interface AuthClient {
-  sign(blobToSign: string): Promise<string>;
+  sign(blobToSign: string, signingEndpoint?: string): Promise<string>;
   getCredentials(): Promise<{
     client_email?: string;
   }>;


handwritten/storage/test/file.ts
```diff
@@ -3794,11 +3794,30 @@ describe('File', () => {
...
         done();
       });
     });
 
+    it('should pass signingEndpoint to URLSigner', done => {
+      const signingEndpoint = 'https://my-endpoint.com';
+      const config = {
+        signingEndpoint,
+        ...SIGNED_URL_CONFIG,
+      };
```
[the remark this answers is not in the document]
thiyaguk09: updated

handwritten/storage/src/file.ts
```diff
@@ -3236,6 +3236,7 @@ class File extends ServiceObject<File, FileMetadata> {
       contentMd5: cfg.contentMd5,
       contentType: cfg.contentType,
       host: cfg.host,
+      signingEndpoint: cfg.signingEndpoint,
```
rajeevpodar, line 3239: When URLSigner receives this parameter, it [passes it to auth.sign() as a second argument](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/storage/src/signer.ts#L248-L251). The issue is that the [public AuthClient interface](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/storage/src/signer.ts#L27-L32) only defines one parameter (sign(blobToSign: string)).

If a developer provides a custom AuthClient implementation based on our interface, Javascript will silently drop this new signingEndpoint argument at runtime. The user will receive a signed URL generated from the default endpoint, and they'll have no idea their custom config was ignored.

Suggestion: Could we update the AuthClient interface in src/signer.ts to explicitly include the optional parameter? sign(blobToSign: string, signingEndpoint?: string): Promise<string>;

This will ensure custom implementers are aware they need to support it.

rajeevpodar (commented):
Verify the custom auth support code flow.

handwritten/storage/src/file.ts
thiyaguk09: I've updated the AuthClient interface in src/signer.ts, it is now ​`sign(blobToSign: string, signingEndpoint?: string): Promise<string>;`​

Dhriti07 (approved):

Dhriti07 (changes requested):
Unapproving for now to await fixes for the failing presubmit checks

thiyaguk09:
> Unapproving for now to await fixes for the failing presubmit checks

@Dhriti07 The presubmit checks are now passing. Could you please approve and merge this PR?

Dhriti07 (approved):

PR #8983, opened 2026-07-28
feat(spanner-driver): Add support for DSN resolution and Error State

surbhigarg92:
Introduces DSN connection resolution (`resolveDsn`) and SQLSTATE error code mapping (`enrichPgError`, `DatabaseError`) for `@google-cloud/spanner-pg`.
- Implements `resolveDsn` to parse standard connection strings, configuration objects
- Implements `DatabaseError` class and `enrichPgError` to map Spanner gRPC status errors to PostgreSQL SQLSTATE codes 
- Adds unit tests in `test/unit/config_test.ts` and `test/unit/errors_test.ts`.


handwritten/spanner-driver/src/lib/errors.ts
```diff
@@ -0,0 +1,95 @@
...
+  const match = msg.match(/\[SQLSTATE\s+([0-9A-Z]{5})\]/i);
+  if (match) {
+    pgErr.code = match[1].toUpperCase();
+    return pgErr;
+  }
+
+  // Unconditionally overwrite non-SQLSTATE code with standard PostgreSQL internal error XX000
+  pgErr.code = 'XX000';
+
+  return pgErr;
```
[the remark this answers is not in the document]
surbhigarg92: This mapping is done in Go shared driver, hence we dont need it here

handwritten/spanner-driver/src/lib/errors.ts
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @param err - Thrown error object or message string to enrich.
+ * @returns DatabaseError instance enriched with PostgreSQL SQLSTATE code and severity.
+ */
+export function enrichPgError(err: unknown): PgError {
+  let pgErr: PgError;
+  if (!err) {
+    pgErr = new DatabaseError('Unknown error');
+  } else if (err instanceof DatabaseError) {
+    pgErr = err;
```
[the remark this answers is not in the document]
surbhigarg92: not valid. if error is already an instance of DatabaseError it make sense to return it directly

handwritten/spanner-driver/src/lib/config.ts
```diff
@@ -0,0 +1,126 @@
...
+      if (
+        !cfg?.database &&
+        (!database ||
+          database === 'postgres' ||
+          (process.env.PGDATABASE && database === process.env.PGDATABASE))
+      ) {
+        database = parts[5].split('?')[0];
+      }
+    }
+  }
```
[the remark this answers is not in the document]
surbhigarg92: done

handwritten/spanner-driver/src/lib/config.ts
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Resolves a Spanner DSN connection string from a connection string, configuration object, or environment variables.
+ *
+ * Priority order:
+ * 1. String connection DSN or postgresql:// URL passed directly.
+ * 2. `config.connectionString` or explicit `config.database` resource path (`projects/...`).
+ * 3. `config` object parts (`project`, `instance`, `database`, `host`, `port`). Explicit `config.database` overrides environment variables.
+ * 4. Environment variable fallbacks (`DATABASE_URL`, `PGCONNECTSTRING`, `SPANNER_PROJECT_ID`, `SPANNER_INSTANCE_ID`, `PGDATABASE`).
+ * 5. Preserves existing DSN query parameters and appends `auto_config_emulator=true` if `SPANNER_EMULATOR_HOST` is set.
```
olavloite: This is slightly different from how `SPANNER_EMULATOR_HOST` normally works. `auto_config_emulator=true` will not only connect the driver to the emulator, but also automatically create the instance and database on the emulator. Are we sure that we want this driver to behave differently with regards to this, compared to all other drivers and clients?

handwritten/spanner-driver/src/lib/config.ts
```diff
@@ -0,0 +1,144 @@
...
+  if (typeof config === 'string') {
+    return config;
+  }
+  const cfg =
+    typeof config === 'object' && config !== null && !Array.isArray(config)
+      ? config
+      : undefined;
+  if (cfg?.connectionString) {
+    return cfg.connectionString;
+  }
```
olavloite, line 59: Note that these early returns will make the statement in the documentation above regarding what happens when SPANNER_EMULATOR_HOST has been set to be incorrect.

handwritten/spanner-driver/src/lib/config.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  if (cfg?.database && cfg.database.startsWith('projects/')) {
+    return cfg.database;
```
olavloite: This also does nothing with SPANNER_EMULATOR_HOST. (But see also my comment above: I am not sure we even want that special handling of SPANNER_EMULATOR_HOST)

handwritten/spanner-driver/src/lib/config.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/**
+ * Resolves a Spanner DSN connection string from a connection string, configuration object, or environment variables.
+ *
+ * Priority order:
+ * 1. String connection DSN or postgresql:// URL passed directly.
+ * 2. `config.connectionString` or explicit `config.database` resource path (`projects/...`).
+ * 3. `config` object parts (`project`, `instance`, `database`, `host`, `port`). Explicit `config.database` overrides environment variables.
+ * 4. Environment variable fallbacks (`DATABASE_URL`, `PGCONNECTSTRING`, `SPANNER_PROJECT_ID`, `SPANNER_INSTANCE_ID`, `PGDATABASE`).
```
olavloite: If we are supporting PGCONNECTSTRING and PGDATABASE, then I think that it would also make sense to support PGHOST and PGPORT.

handwritten/spanner-driver/src/index.ts
```diff
@@ -12,8 +12,28 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {DatabaseError} from './lib/errors.js';
+import {ClientConfig, resolveDsn} from './lib/config.js';
+
+export {DatabaseError, ClientConfig, resolveDsn};
+
+/**
+ * Escapes a PostgreSQL SQL identifier (such as a table name or column name)
```
olavloite: This comment also 'locks' this API to a PostgreSQL database. Would it make sense to put it in a file or folder that better indicates that it is a PostgreSQL-specific function?

handwritten/spanner-driver/src/index.ts
```diff
@@ -12,8 +12,28 @@
...
+ * by enclosing it in double quotes and doubling internal double-quote characters.
+ *
+ * @param str - The identifier string to escape.
+ * @returns Safely double-quoted identifier string.
+ */
 export const escapeIdentifier = (str: string) => `"${str.replace(/"/g, '""')}"`;
 
+/**
+ * Escapes a PostgreSQL string literal value by enclosing it in single quotes,
```
olavloite: Same here as above: We need to think through a bit how we make sure that we do not 'lock' this driver in to only PostgreSQL.

handwritten/spanner-driver/src/lib/errors.ts
```diff
@@ -0,0 +1,95 @@
...
+  } else if (typeof err === 'object' && err !== null && !Array.isArray(err)) {
+    const errorObj = err as Record<string, unknown>;
+    pgErr = new DatabaseError(String(errorObj.message || 'Unknown error'));
+    Object.assign(pgErr, errorObj);
+    pgErr.name = 'DatabaseError';
+    if (typeof errorObj.stack === 'string') {
+      pgErr.stack = errorObj.stack;
+    }
```
olavloite: Using `Object.assign(pgErr, errorObj)` on Error objects has a couple of subtle issues in Node:

* Non-enumerable properties are skipped: In Node/V8, standard Error properties and prototype getters are non-enumerable, so Object.assign won't copy them (which is why stack had to be manually assigned on line 65).
* Type safety: If errorObj is a gRPC error where code is a number (e.g. 5), pgErr.code gets assigned a number instead of string | undefined.

Instead of `Object.assign`, it's safer to extract and convert the properties explicitly:

```typescript
const message = typeof errorObj.message === 'string' ? errorObj.message : String(err);
pgErr = new DatabaseError(message);

if (typeof errorObj.stack === 'string') pgErr.stack = errorObj.stack;
if (errorObj.code !== undefined) pgErr.code = String(errorObj.code);
if (typeof errorObj.severity === 'string') pgErr.severity = errorObj.severity;
if (typeof errorObj.detail === 'string') pgErr.detail = errorObj.detail;
if (typeof errorObj.hint === 'string') pgErr.hint = errorObj.hint;
```

handwritten/spanner-driver/src/lib/errors.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  // If code is already a 5-char SQLSTATE string, preserve it
+  if (typeof pgErr.code === 'string' && /^[0-9A-Z]{5}$/i.test(pgErr.code)) {
+    return pgErr;
+  }
+
+  const msg = (pgErr.message || '').toString();
+
+  // Extract SQLSTATE code enriched by the underlying Go SQL driver (go-sql-spanner)
+  const match = msg.match(/\[SQLSTATE\s+([0-9A-Z]{5})\]/i);
```
olavloite: `/^[0-9A-Z]{5}$/i` is a too broad and also matches non-SQLSTATE errors.

Node system error codes like EPERM, EBUSY, and EPIPE (or arbitrary strings like 'ERROR') would for example also match this regex.

I think that it would be better to match against known PG SQLSTATE prefixes (e.g., 00-58, XX, HV, P0). See https://www.postgresql.org/docs/current/errcodes-appendix.html for a complete list.

handwritten/spanner-driver/src/lib/pg/errors.ts
```diff
@@ -0,0 +1,99 @@
...
+    pgErr = new DatabaseError('Unknown error');
+  } else if (err instanceof DatabaseError) {
+    pgErr = err;
+  } else if (typeof err === 'object' && err !== null && !Array.isArray(err)) {
+    const errorObj = err as Record<string, unknown>;
+    const message =
+      typeof errorObj.message === 'string'
+        ? errorObj.message
+        : String(errorObj.message || 'Unknown error');
+    pgErr = new DatabaseError(message);
```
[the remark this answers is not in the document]
surbhigarg92: In @google-cloud/spanner-driver, underlying errors from the Go driver (go-sql-spanner) cross the CGO / N-API C++ boundary as stringified exception messages (Napi::Error::New(env, message)).

Because CGO converts Go errors into standard JavaScript Error instances, there are no nested sub-error objects or gRPC metadata structs passed across the native C boundary. Attaching { cause: err } would simply point to the same generic exception string without adding extra context.

handwritten/spanner-driver/src/lib/pg/errors.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+    if (
+      typeof errorObj.position === 'string' ||
+      typeof errorObj.position === 'number'
+    ) {
+      pgErr.position = errorObj.position;
+    }
+  } else {
+    pgErr = new DatabaseError(String(err));
+  }
```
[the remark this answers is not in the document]
surbhigarg92: In @google-cloud/spanner-driver, underlying errors from the Go driver (go-sql-spanner) cross the CGO / N-API C++ boundary as stringified exception messages (Napi::Error::New(env, message)).

Because CGO converts Go errors into standard JavaScript Error instances, there are no nested sub-error objects or gRPC metadata structs passed across the native C boundary. Attaching { cause: err } would simply point to the same generic exception string without adding extra context.

handwritten/spanner-driver/src/lib/errors.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @returns DatabaseError instance enriched with dialect-specific error codes.
+ */
+export function enrichError(
+  err: unknown,
+  dialect: Dialect = DEFAULT_DIALECT,
+): DatabaseError {
+  if (dialect === 'pg') {
+    return enrichPgError(err);
+  }
+  return err instanceof DatabaseError ? err : new DatabaseError(String(err));
```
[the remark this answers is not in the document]
surbhigarg92: In @google-cloud/spanner-driver, underlying errors from the Go driver (go-sql-spanner) cross the CGO / N-API C++ boundary as stringified exception messages (Napi::Error::New(env, message)).

Because CGO converts Go errors into standard JavaScript Error instances, there are no nested sub-error objects or gRPC metadata structs passed across the native C boundary. Attaching { cause: err } would simply point to the same generic exception string without adding extra context.

handwritten/spanner-driver/src/lib/errors.ts
```diff
@@ -0,0 +1,60 @@
...
+  hint?: string;
+  /** Position index of the syntax error within the SQL query string. */
+  position?: string | number;
+
+  constructor(message: string) {
+    super(message);
+    this.name = 'DatabaseError';
+    Object.setPrototypeOf(this, DatabaseError.prototype);
+  }
+}
```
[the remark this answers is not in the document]
surbhigarg92: In @google-cloud/spanner-driver, underlying errors from the Go driver (go-sql-spanner) cross the CGO / N-API C++ boundary as stringified exception messages (Napi::Error::New(env, message)).

Because CGO converts Go errors into standard JavaScript Error instances, there are no nested sub-error objects or gRPC metadata structs passed across the native C boundary. Attaching { cause: err } would simply point to the same generic exception string without adding extra context.

handwritten/spanner-driver/src/lib/config.ts
surbhigarg92: Have removed this

handwritten/spanner-driver/src/lib/config.ts
surbhigarg92: Removed

handwritten/spanner-driver/src/lib/config.ts
surbhigarg92: Have removed the support for PG ENVs for the time being, can add later if required.

handwritten/spanner-driver/src/index.ts
surbhigarg92: Moved to separate "pg" directory

handwritten/spanner-driver/src/lib/errors.ts
surbhigarg92: Done

handwritten/spanner-driver/src/lib/errors.ts
surbhigarg92: Done

handwritten/spanner-driver/src/lib/config.ts
```diff
@@ -0,0 +1,75 @@
...
+    typeof config === 'object' && config !== null && !Array.isArray(config)
+      ? config
+      : undefined;
+  if (cfg?.connectionString) {
+    return cfg.connectionString;
+  }
+
+  const project = cfg?.project || process.env.GOOGLE_CLOUD_PROJECT;
+  const instance = cfg?.instance;
+  const database = cfg?.database;
```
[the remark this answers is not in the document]
surbhigarg92: Updated the PR description

handwritten/spanner-driver/test/unit/config_test.ts
```diff
@@ -0,0 +1,119 @@
...
+  const origEnv = {...process.env};
+
+  beforeEach(() => {
+    delete process.env.GOOGLE_CLOUD_PROJECT;
+  });
+
+  afterEach(() => {
+    delete process.env.GOOGLE_CLOUD_PROJECT;
+    Object.assign(process.env, origEnv);
+  });
```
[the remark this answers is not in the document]
surbhigarg92: Its already done

handwritten/spanner-driver/test/unit/config_test.ts
```diff
@@ -0,0 +1,119 @@
...
+    const origEnv = {...process.env};
+
+    beforeEach(() => {
+      delete process.env.GOOGLE_CLOUD_PROJECT;
+    });
+
+    afterEach(() => {
+      delete process.env.GOOGLE_CLOUD_PROJECT;
+      Object.assign(process.env, origEnv);
+    });
```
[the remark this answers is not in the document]
surbhigarg92: Will keep these blocks for future if required

handwritten/spanner-driver/src/lib/pg/errors.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @param err - Thrown error object or message string to enrich.
+ * @returns DatabaseError instance enriched with PostgreSQL SQLSTATE code and severity.
+ */
+export function enrichPgError(err: unknown): PgError {
+  let pgErr: PgError;
+  if (!err) {
+    pgErr = new DatabaseError('Unknown error');
+  } else if (err instanceof DatabaseError) {
+    pgErr = err;
```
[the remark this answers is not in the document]
surbhigarg92: Not required if err is already an instance of "DatabaseError"

handwritten/spanner-driver/src/lib/constants.ts
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+/**
+ * Supported SQL dialects for Google Cloud Spanner.
+ * Note: Currently 'pg' (PostgreSQL) dialect is supported.
+ */
+export type Dialect = 'pg' | 'googlesql';
+
+/** Default SQL dialect used by the Spanner driver. */
+export const DEFAULT_DIALECT: Dialect = 'pg';
```
olavloite, line 22: I do worry a bit about this, at least long-term. We should preferably gradually move towards a setup that automatically detects the dialect that is being used for a connection by querying the underlying database, and use that dialect for the dialect-aware functions. If we keep this as an exported constant when this is released, then making any changes to this after that will be a breaking change.

Note: This is fine for now, but I do think that we need to make sure that we do not 'lock our selves in' before we release the driver.

olavloite (approved):

handwritten/spanner-driver/src/lib/constants.ts
surbhigarg92: We are not exporting DEFAULT_DIALECT and Dialect for customers, these are not exported from "index.ts" file. This will be only available for internal files. 

Our `package.json` only allow exports from `index` file and blocks deep imports

```
"exports": {
    ".": {
      "import": {
        "types": "./build/esm/src/index.d.ts",
        "default": "./build/esm/src/index.js"
 ```

PR #8993, opened 2026-07-28
test: remove testing on node 18 from matrix

danieljbruce:
# Description

Right now we are upgrading Node on client libraries to only support versions 22+ so open PR tests are naturally failing when trying to run the tests on Node version 18. While we could remove testing for Node 18 only after we officially drop support for Node 18 to catch specific Node 18 issues in the short term, we create a situation in our CI where our unit tests will always fail on Node 18 and unit tests for other Node versions don't run like in https://github.com/googleapis/google-cloud-node/pull/8988 after the upgrade. Therefore, we want to remove Node 18 testing right now in order to keep the tests running for other versions and avoid having a Node 18 unit test that we expect to fail.

The github actions also needed to be updated for the zizmor check.

## Impact

Ensure testing infrastructure is still useful as we do the Node upgrade.

### DIFF
--- a/.github/workflows/continuous.yaml
+++ b/.github/workflows/continuous.yaml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         node: [22, 24, 26]
     steps:
-      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           persist-credentials: false
       - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
--- a/.github/workflows/continuous.yaml
+++ b/.github/workflows/continuous.yaml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         node: [22, 24, 26]
     steps:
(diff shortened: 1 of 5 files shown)


danieljbruce:
Could I get approval to remove the units (18) check?

.github/workflows/continuous.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
       matrix:
         node: [22, 24, 26]
     steps:
-      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
```
pearigee, line 16: Are we really going from v6 to v4? Or were these just mis-labled before?

We wouldn't want to downgrade our actions right? Do the newer versions just need to have their hashes corrected?

pearigee (approved):

.github/workflows/continuous.yaml
danieljbruce: They were not mislabelled before, but I looked into this further and it seems the problem is we were using floating tags ie. v6 which were trying to match the sha in the yaml to the latest v6 sha and the mismatch was producing a zizmor error. Therefore, I pinned the versions after the `#` so that we don't see zizmor matches again.

PR #8973, opened 2026-07-27
feat: PNPM workspace

bshaffer:
### Description

Migrate repository to a **PNPM workspace** with **Turborepo caching** and **CI test sharding**.

This modernization speeds up dependency installation, reduces disk space usage, enforces strict dependency boundaries across monorepo packages, and drastically cuts CI turnaround time.

---

### Key Highlights & Changes

1. **PNPM Workspace & Turborepo (`feat(workspace)`)**:
   - Added `pnpm-workspace.yaml` and root `pnpm-lock.yaml` to manage packages across `core/*`, `handwritten/*`, and `packages/*`.
   - Configured Turborepo (`turbo.json`) to cache `compile` tasks across all packages with proper input tracking (`protos/**`, `src/**`, `test/**`), saving ~20 minutes on CI build steps.
   - Updated root `package.json` to enforce `pnpm` package management and set compilation concurrency.
   - Configured `.npmrc` with `link-workspace-packages=false` to mirror production registry resolution.
   - Updated root tooling (`bin/linter.mjs`, `.mocharc.cjs`, `.gitignore`).

2. **Core Tools & Packages Compatibility (`fix(core)`)**:
   - Added standalone cross-platform proto copying (`core/packages/gax/scripts/copy-protos.js`) and compilation scripts in `gax` and `tools`.
   - Updated `pack-n-play` test fixtures and `gapic-generator-typescript` baselines.

3. **Handwritten Package & Build Fixes (`fix(handwritten)`)**:
   - Resolved test timeouts, mock servers, and Node.js compatibility across `storage`, `bigtable`, `spanner`, and `cloud-profiler`.
   - Regenerated `google-storage-control` protos.

4. **Kokoro & Cloud Build Modernization (`ci(scripts)`)**:
   - Updated all Kokoro (`.kokoro/*.sh`) and `cloudbuild.yaml` scripts across handwritten and core packages to use `pnpm`.

5. **GitHub Actions CI Workflows & Matrix Sharding (`ci(github)`)**:
   - Configured 15-shard matrix unit test execution across Node 22, 24, and 26 for Linux, Windows, and Bun.
   - Pinned GitHub action hashes for security (Zizmor compliance).

---

> **Note**: Non-critical workspace documentation and config additions have been separated into follow-up PR #9144.

PR #9021, opened 2026-07-29
feat(ci): implement sharding for unit test performance

### LINKED ISSUE #7540: reenable test-all when CI scripts change
For the Node monorepo migration project we will temporarily be turning off the feature of CI to test all of packages when we make changes to CI. These types of change, which happen during the migration, cause tests runs of 7+ hours which make it hard to land changes. Once the monorepo migration is complete this bug should be triaged and resolved independently

bshaffer:
fixes #7540

- **Conditional Test Sharding**: Dynamically shards tests when a large number of packages are modified. Adds a job that rolls up shard results to satisfy branch protection. Uses "composite" actions to deduplicate workflow YAML between unit and windows unit tests. For what this looks like with just ONE package modified, see https://github.com/googleapis/google-cloud-node/pull/9027

- **Run the full unit test suite**: Restores triggering the full unit test suite across all packages whenever `ci/` scripts change, as sharding now allows running all tests without timing out.

- **Clean CI Logs**: Switches Mocha to use `dot` reporter and removes accidental `console.log` from the Firestore conformance tests.

**NOTE**: These tests run in about ~12min per shard (as compared to 25-30) once we migrate to a PNPM Workspace #8973 

 - [x] Depends on #9061


ci/run_conditional_tests.sh
```diff
@@ -61,11 +61,11 @@ set -e
...
-    echo "skipping trigger of tests for now: tracking in #7540"
-    # echo "change detected in ci, we should test everything"
-    # echo "result of git diff ${GIT_DIFF_ARG} ci:"
-    # git diff ${GIT_DIFF_ARG} ci
-    # GIT_DIFF_ARG=""
+    # echo "skipping trigger of tests for now: tracking in #7540"
+    echo "change detected in ci, we should test everything"
+    echo "result of git diff ${GIT_DIFF_ARG} ci:"
+    git diff ${GIT_DIFF_ARG} ci
+    GIT_DIFF_ARG=""
```
bshaffer: As tests now take ~30 minutes (down from 2 hours), we can reenable this and close https://github.com/googleapis/google-cloud-node/issues/7540.

Once we add PNPM workspaces and turbo caching, this number goes down to ~15!!

bshaffer:
@gemini-code-assist review

ci/run_conditional_tests.sh
```diff
@@ -204,21 +209,51 @@ for subdir in ${subdirs[@]}; do
...
+    exit 0
+fi
+
+# If SHARD_TOTAL and SHARD_INDEX are provided, we will only run a subset of the tests.
+for i in "${!test_dirs[@]}"; do
+    d="${test_dirs[$i]}"
+
+    if [[ -n "${SHARD_TOTAL}" && -n "${SHARD_INDEX}" ]]; then
+        if (( SHARD_TOTAL > 0 && i % SHARD_TOTAL != SHARD_INDEX )); then
+            continue
```
westarle, line 287: We should just keep an eye on whether there are differences between the shard durations (in case multiple "heavy" targets land in a single shard consistently.)

ci/run_conditional_tests.sh
```diff
@@ -31,25 +31,27 @@ fi
...
+if [ -z "${GIT_DIFF_ARG}" ]; then
+    if [ "${BUILD_TYPE}" == "presubmit" ]; then
+        # For presubmit build, we want to know the difference from the
+        # common commit in origin/main.
+        GIT_DIFF_ARG="origin/main..."
 
-    # Then fetch enough history for finding the common commit.
-    git fetch origin main --deepen=300
+        # Then fetch enough history for finding the common commit.
+        git fetch origin main --deepen=300
```
westarle: Can we have the action fetch the right level of detail?  I think we do 2 levels in setup then immediately fill in 300 more.

handwritten/bigtable/protos/protos.js
westarle, line 1: could this be in another PR?

.github/workflows/presubmit.yaml
```diff
@@ -43,3 +68,37 @@ jobs:
...
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
+        with:
+          script: |
+            const { data: { jobs } } = await github.rest.actions.listJobsForWorkflowRun({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              run_id: context.runId,
+            });
+            const nodeVersion = '${{ matrix.node-version }}';
+            const shardJobs = jobs.filter(j => j.name.includes(`units (Node ${nodeVersion},`));
```
westarle: can we reuse test_dirs from setup?

core/dev-packages/pack-n-play/test/test.ts
```diff
@@ -18,7 +18,11 @@ import execa = require('execa');
 import {describe, it} from 'mocha';
 
 describe('pack-n-play', () => {
-  it('should run tests', async () => {
+  it('should run tests', async function () {
+    this.timeout(120000); // 2 minutes
+    if (process.platform === 'win32') {
```
westarle: I'd be more comfortable doing this after we set up a postsubmit for the coverage.  It seems like Windows would cause more problems in packaging...

ci/run_single_test.sh
```diff
@@ -47,8 +49,8 @@ if command -v cygpath >/dev/null 2>&1; then
     PNPMFILE_PATH=$(cygpath -m "${PNPMFILE_PATH}")
 fi
 
-echo "pnpm install --ignore-scripts --engine-strict --prod --pnpmfile \"${PNPMFILE_PATH}\"; pnpm install --pnpmfile \"${PNPMFILE_PATH}\""
-pnpm install --ignore-scripts --engine-strict --prod --pnpmfile "${PNPMFILE_PATH}"; pnpm install --pnpmfile "${PNPMFILE_PATH}"
+echo "pnpm install --reporter=silent --engine-strict --pnpmfile \"${PNPMFILE_PATH}\""
+pnpm install --reporter=silent --engine-strict --pnpmfile "${PNPMFILE_PATH}"
```
westarle: Are errors still reported with --reporter=silent?

westarle (commented):
A few comments, I think this is a really great idea!

core/dev-packages/pack-n-play/test/test.ts
bshaffer: after testing, this did not affect performance that much, and windows STILL took 30-40 minutes, so I think removing it is the right call

ci/run_conditional_tests.sh
bshaffer: My understanding is that 2 is sufficient for github actions, but 300 might be needed for kokoro (which is why the conditional here exists). The 300 is just supposed to be a sufficiently large number to ensure a common commit ancestor is found.

So I guess Im not sure what your suggestion is here.

ci/run_conditional_tests.sh
bshaffer: Yes, I've already seen this happen that Shard 0 is a little heavier, probably because of the `bigtable` and `pack-n-play` tests. If a big discrepancy emerges (right now, it seems like just a few minutes, so I am not sure it's a big concern), we could take the known longer tests and manually separate them? What are your thoughts on the best way to accomplish this?

handwritten/bigtable/protos/protos.js
bshaffer: The bigtable unit tests were failing, so I added these changes to fix them... however I'm not 100% sure THIS specific file update is needed (I suspect it is). I can verify!

westarle (commented):
Also, would you be OK with breaking the test fixes out to their own little PRs?

ci/run_conditional_tests.sh
westarle: Thanks for handling the newbie questions Brent.  TIL about actions/checkout synthetic commits (and also I missed the `if [ -z...` at the top of this diff).

I'd like all these scripts to be hermetic, and have things work like your actions work (operator fetches the right depth and passes the GIT_DIFF_ARG).  Today, if I set `GIT_DIFF_ARG` but fail to fetch it fails, and if I don't set `GIT_DIFF_ARG` it can waste some time if I already fetched what I know I need.

Here's a practical fix for now:
- Add a new "strict" flag or variable `--strict`/`STRICT` to the script and pass it from the action
- If "strict" is set, `GIT_DIFF_ARG` must be set and `git diff --quiet ${GIT_DIFF_ARG}` must return 0 or 1.
- The test on line 34 becomes "if not strict".

Then we can visit GCB builds later to the model you're introducing for actions here.

ci/run_conditional_tests.sh
westarle: > we could take the known longer tests and manually separate them

This it what I'd do.

> right now, it seems like just a few minutes, so I am not sure it's a big concern

+1

ci/run_conditional_tests.sh
bshaffer: This is implemented in 2446ce6928af6e7bcb14d395b64232c426503a91

ci/run_single_test.sh
bshaffer: Yes, errors are still reported! `--reporter=silent` suppresses stdout, but errors are still sent to stderr.

.github/workflows/presubmit.yaml
bshaffer: I went with this way because:

 - Computing `test_dirs` with `git diff` takes ~1 second. Passing it via GitHub Actions adds ~5-10 seconds of network transfer and disk I/O per shard.
 - Because we have to calculate on both Windows vs. Linux, this way allows each OS runner to apply platform-specific handling (e.g., path separators and windows_exempt_tests) natively without needing serialization/translation across platforms.
 - It allows `ci/run_conditional_tests.sh` to be executed standalone

handwritten/bigtable/protos/protos.js
bshaffer: aparently not because every time I revert it owlbot commits it 😆 

We do not have owlbot set up like that for PHP... that is quite baffling to me...

bshaffer: I've separated out https://github.com/googleapis/google-cloud-node/pull/9057 so it can be merged separately

.github/actions/check-shard-status/action.yaml
```diff
@@ -0,0 +1,31 @@
...
+            repo: context.repo.repo,
+            run_id: context.runId,
+          });
+          const nodeVersion = '${{ inputs.node-version }}';
+          const shardJobs = jobs.filter(j => j.name.includes(`units (Node ${nodeVersion},`));
+          if (shardJobs.length === 0) {
+            core.setFailed(`No shard jobs found for Node ${nodeVersion}`);
+            return;
+          }
+          const failed = shardJobs.filter(j => j.conclusion === 'failure' || j.conclusion === 'cancelled');
```
westarle: can we test for 'success' to make sure we are catching 'skips' (when setup fails)

.pnpmfile.cjs
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+  hooks: {
+    readPackage(pkg, context) {
+      // Override yargs for Node >= 24 to prevent ESM scope require errors
```
westarle, line 4: Is there an issue we can track to remove this?

.github/workflows/presubmit.yaml
```diff
@@ -43,3 +51,22 @@ jobs:
       - run: npm install
       - run: npm run lint
         name: Run monorepo linter
+
+  # Dummy jobs to satisfy branch protection requirements for each node version
```
westarle: Did you consider updating the branch protection requirements to match our new tests?

ci/run_conditional_tests.sh
```diff
@@ -209,21 +236,51 @@ for subdir in ${subdirs[@]}; do
...
-                # Since there are so many APIs, we should exit early if there's an error
-                exit ${RETVAL}
-            fi
-            popd
+            test_dirs+=("${d}")
         fi
     done
 done
+# If DRY_RUN_SHARDS is set, output dynamic matrix values to GitHub Actions and exit
+if [[ "${DRY_RUN_SHARDS}" == "true" ]]; then
```
westarle: It might be cleaner to change the API to something like:
RUN_TESTS_MODE={CALCULATE_SHARD_MATRIX | RUN_UNIT_TESTS}
and validate it is set in --strict mode.

Then you can also validate that SHARD_TOTAL/SHARD_INDEX are set with RUN_UNIT_TESTS.

westarle: ideally we could replace this with multiple js or go scripts with tests...

.github/workflows/presubmit.yaml
bshaffer: With this system you won't have to - the "dummy jobs" represent the current branch protection rules, and they aggregate the result of the shards

.github/actions/check-shard-status/action.yaml
bshaffer: I do not believe setup failures result in a skip, but I'm happy to check for success anyway

.github/workflows/presubmit.yaml
westarle: Maybe the comment can say

> consolidate shards into jobs representing aggregate status to simplify branch protection requirements

(i.e. the branch protection depends on these jobs rather than the other way around)

ci/run_conditional_tests.sh
bshaffer: I'm not sure what is meant by the multiple scripts comment, but I'm happy to change the script to use RUN_TESTS_MODE instead of DRY_RUN_SHARDS

.github/actions/check-shard-status/action.yaml
bshaffer:
```suggestion
          const failed = shardJobs.filter(j => j.conclusion !== 'success');
```

.github/workflows/presubmit.yaml
bshaffer:
```suggestion
  # Consolidate shards into jobs representing aggregate status to simplify branch protection requirements
```

.pnpmfile.cjs
bshaffer: Yes, I have created issue #9130 to track removing this override once we transition to PNPM workspaces (or once all sub-packages are updated to compatible `yargs` versions). Since the upcoming PNPM workspace PR (#8973) deletes `.pnpmfile.cjs` entirely, this specific commented line will be cleaned up automatically during that merge. However, the global override will still exist under `"pnpm.overrides"` in the root `package.json`, which is what issue #9130 is tracking for ultimate removal.

I've also updated this PR to add a `TODO` pointing to the issue directly on `.pnpmfile.cjs`.

westarle (approved):

PR #9026, opened 2026-07-30
feat!: upgrade minimum Node.js version to 22

danieljbruce:
## Description

Upgrade all generated packages to Node 22 along with their dependencies.

For all `package.json` files inside `packages/**` folder we set "node": ">=22" for `engines`. We also make sure that for all of these `package.json` files the dependencies that match the core libraries are all on the latest version so that all core library dependencies correspond to node v22.

Note that a branch update is likely required to get the zizmor checks to pass.

Also note that upgrading the auth library dependencies causes compiler errors so we opt out of upgrading the auth library.

## Impact

Encourages customers to adopt new Node version which avoids issues due to lack of Node 18 support.

## Next Steps

Upgrade the auth dependencies that we chose not to upgrade this time due to the compiler error. Tracked [here](https://b.corp.google.com/issues/541318013).


westarle (commented):
Can you please share the script or commands used to make this change in the PR description?

danieljbruce:
> Can you please share the script or commands used to make this change in the PR description?

There was a lot of back and forth in gemini, but I think I landed on [script](https://paste.googleplex.com/5213911568744448) and then gemini applied `bdc816a1a9fa41cfa5d625d8fd048d06cf7f0f5a` commit to drop the auth changes.

westarle (approved):

PR #9029, opened 2026-07-30
feat!: upgrade minimum Node.js version to 22 for the gapic-generator-typescript

danieljbruce:
## Description

Upgrade the gapic-generator-typescript to node version 22. 

## Impact

Ensures users upgrade their runtime to Node 22 and don't encounter problems due to legacy Node 18 issues.

## Changes

Node was upgraded to version 22 in two package.json files and then `npm run baseline` was run.

Just the following files were updated:
core/generator/gapic-generator-typescript/templates/cjs/typescript_gapic/package.json
core/generator/gapic-generator-typescript/templates/esm/typescript_gapic/package.json

Then in core/generator/gapic-generator-typescript we ran the following command:
`npm run baseline`

## Next steps

Next we are going to upgrade all the packages in the root `packages` directory to separate concerns about the packages and the gapic-typescript-generator.


westarle:
can you please add some (step-by-step) details about how you generated this PR (which package.json were changed, for example)

danieljbruce:
> can you please add some (step-by-step) details about how you generated this PR (which package.json were changed, for example)

Done

westarle (approved):

core/generator/gapic-generator-typescript/baselines/videointelligence-esm/package.json
westarle: do we need to update this too (better do all the breaking changes at once?)

westarle (changes requested):

core/generator/gapic-generator-typescript/baselines/videointelligence-esm/package.json
danieljbruce: Yeah. Probably a good idea. Though it will still work if the dependency has a lesser constraint (ie. >=18), but we should probably stick to the plan for this one anyway.

danieljbruce: Done

westarle: oh right, thanks for explaining.  I would like to keep the dev dependencies up to date as well, if it's convenient.  Up to you how / when you want to update those.

westarle (approved):

PR #9037, opened 2026-07-31
chore: update CODEOWNERS file

shivanee-p:
Updated team_members: comment definition in `.github/CODEOWNERS`

### DIFF
--- a/.github/workflows/assign-reviewers.yml
+++ b/.github/workflows/assign-reviewers.yml
@@ -20,17 +20,14 @@ jobs:
         uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
         with:
           persist-credentials: false
-          fetch-depth: 0
-
-      - name: Fetch base branch for diff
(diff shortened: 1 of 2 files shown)


.github/workflows/assign-reviewers.yml
```diff
@@ -16,94 +16,128 @@ jobs:
...
-              console.log(`PR already has route-specific team reviewer requested: ${requestedTeams.map(t => t.slug).join(', ')}.`);
-              shouldExit = true;
-            } else if (hasReviews) {
-              shouldExit = true;
-            }
+            // 3. Obtain changed files hermetically from local git checkout
+            let files = [];
+            const baseRef = pr.base.ref || 'main';
+
+            const isStrict = Boolean(process.env.STRICT);
```
westarle: Because this script is embedded in a GitHub workflow, this type of thing isn't necessary -- you can make more assumptions about the process.

I suggest we drop this PR for now (except for `CODEOWNERS`), unless there's a bug or something you observed in the action that this is addressing.

danieljbruce (approved):

PR #9039, opened 2026-07-31
ci(workflows): add librarian generation check presubmit

### LINKED ISSUE #5352: node: create a presubmit check to verify `librarian generate -all` in google-cloud-node
Create a presubmit check to verify `librarian generate -all` in google-cloud-node, similar to https://github.com/googleapis/librarian/blob/main/.github/workflows/golang.yaml

quirogas:
Modifications to librarian configuration and generator scripts require downstream generated code to be kept in sync with upstream changes.

Adds .github/workflows/generation_check.yaml to run on pull requests that modify librarian.yaml, .readme-partials.yaml, librarian.js, or librarian.ts, as well as on main branch pushes and workflow_dispatch events. The workflow sets up Node.js 22, installs Librarian dependencies, executes librarian generate --all across configured packages, and asserts zero uncommitted git diffs exist.

Closes googleapis/librarian#5352

### DIFF
--- a/.github/workflows/generation_check.yaml
+++ b/.github/workflows/generation_check.yaml
@@ -1,5 +1,6 @@
 ---
 name: generation-check
+
 on:
   push:
     branches:
@@ -7,12 +8,22 @@ on:
   pull_request:
   # Allows manual triggering for debugging purpose.
   workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
-  issues: write
+
 jobs:
   regeneration:
+    name: Regenerate and verify clean git tree
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      # zizmor: ignore[undocumented-permissions] - Required for creating issue on main branch failure
+      issues: write
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/generation_check.yaml
```diff
@@ -0,0 +1,65 @@
...
+          persist-credentials: false
+      - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
+        id: changes
+        with:
+          filters: |
+            librarian:
+              - 'librarian.yaml'
+      # Version of librarian is pulled from librarian.yaml,
+      # action is pinned to commit SHA to satisfy blanket security policy.
+      - uses: googleapis/librarian@35997441eafc2b02716804f9baba1e3f04f8a44f # v0.31.1
```
westarle, line 38: Can we move this definition to librarian.yaml?

JoeWang1127: This is an action defined in librarian repo, it is not the librarian version used in librarian generation.

Could you explain why do we want to add this definition to librarian.yaml?

westarle: does this need to be in sync with the version in librarian.yaml?

quirogas: No it does not, the only time this action needs to changes is when the action logic itself is updated.

westarle: OK, I was confused because there is a librarian actoin that has the same versioning scheme as the librarian tool, and everything is "librarian" -- I'd appreciate if this had a more descriptive name.  Something with a verb in it like "run_librarian".

westarle (approved):

PR #9044, opened 2026-07-31
chore: update baseline generation

quirogas:
Ran the following commands on librarian version 0.31.1:

```
go install github.com/googleapis/librarian/cmd/librarian@0.31..1
librarian install nodejs
librarian generate --all 
```


packages/google-cloud-agentidentity/protos/protos.json
```diff
@@ -4481,4 +4481,4 @@
       }
     }
   }
-}
+}
```
westarle, line 4484: I guess this whitespace noise here and in the repo metadata is the difference between "onboarding" and "updating" -- can you file an issue to close it out?

westarle (approved):

packages/google-cloud-agentidentity/protos/protos.json
quirogas: created https://github.com/googleapis/librarian/issues/7147 to track this.

PR #9057, opened 2026-08-02
chore: update bigtable from OwlBot post-processor

bshaffer:
Separate the owlbot changes in https://github.com/googleapis/google-cloud-node/pull/9021 in a separate PR

PR #9058, opened 2026-08-03
chore(ci): misc ci fixes, workflow renames, and system test concurrency

bshaffer:
Misc test improvements and fixes, namely:
- Renames `conformance-test` and `conformance` to `bigtable-conformance` and `storage-conformance` respectively.
- Increases Windows `pack-n-play` test timeouts to 10 minutes and adds `--no-audit --no-fund` to speed up installs.
- Scopes `--no-experimental-require-module` strictly to Node 22 in CI scripts to prevent ESM crashes on Node 24 and 26.
- Prevents Bigtable's `reapInstances` cleanup script from deleting active test instances missing a `time_created` label.
- Fixes Bigtable OpenTelemetry timestamp clamping (`endTime = startTime + 1ns`) to prevent Cloud Monitoring `INVALID_ARGUMENT` export errors.
- Fixes Firestore GCB system test contention by running standard and enterprise databases in two concurrent streams instead of four.
  
We will be able to comment out triggering all unit tests on `ci/` script once the sharding PR (#9021) lands.


bshaffer:
@gemini-code-assist please review!

ci/run_conditional_tests.sh
```diff
@@ -87,18 +113,21 @@ subdirs=(
 )
 
 RETVAL=0
-# These following APIs need an explicit credential file to run properly (or oAuth2, which we don't support in this repo). 
+# These following APIs need an explicit credential file to run properly (or oAuth2, which we don't support in this repo).
```
westarle:
```suggestion
# These following APIs need an explicit credential file to run properly (or OAuth2, which we don't support in this repo).
```

ci/run_conditional_tests.sh
```diff
@@ -209,21 +238,51 @@ for subdir in ${subdirs[@]}; do
...
-                RETVAL=${ret}
-                # Since there are so many APIs, we should exit early if there's an error
-                exit ${RETVAL}
-            fi
-            popd
+            test_dirs+=("${d}")
         fi
     done
 done
+# If DRY_RUN_SHARDS is set, output dynamic matrix values to GitHub Actions and exit
```
westarle: I think we should remove all the sharding stuff from this PR.

handwritten/bigtable/protos/protos.js
westarle, line 1: Can you move this to an independent PR?  And include the command you used to generate it?

bshaffer: I didn't generate it, owlbot generates it and automatically adds it to PRs which don't have it but touch files in the directory (because it's out of date)... 

I actually did create a separate PR for it, but since tests fail on it without THIS PR, I thought it was appropriate to put it in this one. Here it is (I closed it, but just reopened it if you'd rather merge it first): https://github.com/googleapis/google-cloud-node/pull/9057

ci/run_single_test.sh
```diff
@@ -47,8 +49,8 @@ if command -v cygpath >/dev/null 2>&1; then
     PNPMFILE_PATH=$(cygpath -m "${PNPMFILE_PATH}")
 fi
 
-echo "pnpm install --ignore-scripts --engine-strict --prod --pnpmfile \"${PNPMFILE_PATH}\"; pnpm install --pnpmfile \"${PNPMFILE_PATH}\""
-pnpm install --ignore-scripts --engine-strict --prod --pnpmfile "${PNPMFILE_PATH}"; pnpm install --pnpmfile "${PNPMFILE_PATH}"
+echo "pnpm install --reporter=silent --engine-strict --pnpmfile \"${PNPMFILE_PATH}\""
```
westarle: Heads up I tried it out and `--reporter=silent` will eat `pnpm` errors.

ci/run_conditional_tests.sh
bshaffer: 100%! I am surprised I didn't catch this ☹️ 

I've removed it

handwritten/bigtable/protos/protos.js
westarle: Thanks for explaining, I don't love Owlbot creating commits.

For now, ideally you could push a commit to #9057 to just breaks caused by owlbot and submit that first.

bshaffer: I also think it's very strange that owlbot commits directly to branches, and I don't like it at all.

As for #9057, we can just ignore the failing `bigtable-system-tests` knowing that they'll be fixed by this PR, and merge it.

Note: I think we should disable `bigtable-system-tests` anyway because they take 90 minutes to run and are flakey. Until they are faster and more reliable they are only slowing us down...

ci/run_single_test.sh
bshaffer: I have reverted this change, and will also revert it in the Shard PR

bshaffer:
I have broken out this PR into smaller, component-isolated PRs as requested:
- #9059: Bigtable CI fixes (`exporter.ts`, `bigtable.ts`, `common.ts`, `gcp-metrics-handler.ts`)
- #9060: Firestore CI fixes (`runner.ts`, `firestore.ts`, `package.json`)
- #9061: Core library & `pack-n-play` CI fixes (`gcp-metadata`, `nodejs-googleapis-common`, `pack-n-play`)
- #9062: GitHub Actions workflow renames (`bigtable-conformance.yaml`, `storage-conformance.yaml`)
- https://github.com/googleapis/google-cloud-node/pull/9057 (owlbot changes for Bigtable)

PR #9059, opened 2026-08-03
fix(bigtable): test resource leaks, timestamp clamping, and metrics exporter

bshaffer:
Separated from #9058 to isolate Bigtable CI fixes:
- Fixes resource leaks in Bigtable system tests by preventing reapInstances from deleting active test instances
- Fixes OpenTelemetry timestamp clamping in exporter
- Overrides OpenTelemetry metrics periodic exporter interval to 1s in tests to prevent timeouts


handwritten/bigtable/src/client-side-metrics/exporter.ts
```diff
@@ -52,14 +52,28 @@ function getInterval(
...
 ) {
+  let endSec = dataPoint.endTime[0];
+  let endNanos = dataPoint.endTime[1];
+  const startSec = dataPoint.startTime[0];
+  const startNanos = dataPoint.startTime[1];
+
+  if (startSec > endSec || (startSec === endSec && startNanos >= endNanos)) {
+    endSec = startSec;
+    endNanos = startNanos + 1;
+    if (endNanos >= 1000000000) {
```
westarle: nit: a comment can help here, like "check for overflow at 1B nanos"

handwritten/bigtable/system-test/common.ts
```diff
@@ -42,7 +42,7 @@ export async function reapBackups(instance: Instance) {
 
 export async function reapInstances(
   bigtable: Bigtable,
-  maxAgeMs = 15 * 60 * 1000,
+  maxAgeMs = 120 * 60 * 1000, // 2 hours
```
westarle, line 45: 😅

handwritten/bigtable/src/client-side-metrics/exporter.ts
westarle: can you please add a test for this change (both the overshoot and the 1B-1 boundary?

westarle (approved):
it's a bit hard to review (e.g. the change in `exporter.ts` without knowing what the failure mode was.

bshaffer:
> it's a bit hard to review (e.g. the change in `exporter.ts` without knowing what the failure mode was.

When OpenTelemetry records very fast operations (or during system tests), it can emit metric data points where `startTime >= endTime`. Google Cloud Monitoring's `CreateTimeSeries` API was rejecting these with `INVALID_ARGUMENT: startTime must be before endTime`. So this change just ensures that `endTime` is at least 1 nanosecond after `startTime` to prevent those API rejections.

handwritten/bigtable/src/client-side-metrics/exporter.ts
bshaffer: I've added that comment and also comment that explains the whole section

handwritten/bigtable/src/client-side-metrics/exporter.ts
bshaffer: done!

PR #9061, opened 2026-08-03
fix(core): discovery url handling, error code review helper, and pack-n-play test timeouts

bshaffer:
Separated from #9058 to isolate Core library and pack-n-play CI fixes:
- Resolves url.parse deprecations and Node 22+ URL discovery handling in nodejs-googleapis-common
- Adds getErrorCode helper to gcp-metadata to properly unwrap AggregateError / cause hierarchies
- Increases Windows CI timeout to 120s and adds --no-audit --no-fund flags in pack-n-play test fixtures

### DIFF
--- a/core/dev-packages/pack-n-play/test/test.ts
+++ b/core/dev-packages/pack-n-play/test/test.ts
@@ -18,7 +18,8 @@ import execa = require('execa');
 import {describe, it} from 'mocha';
 
 describe('pack-n-play', () => {
-  it('should run tests', async () => {
+  it('should run tests', async function () {
+    this.timeout(600000); // 10 minutes
     const fixturesPath = path.resolve('./test/fixtures');
     const dirs = fs
       .readdirSync(fixturesPath)
@@ -29,7 +30,7 @@ describe('pack-n-play', () => {
         stdio: 'inherit',
         cwd: dir,
       };
-      await execa('npm', ['install'], opts);
+      await execa('npm', ['install', '--no-audit', '--no-fund'], opts);
       await execa('npm', ['link', '../../../'], opts);
       await execa('npm', ['test'], opts);
     }
--- a/core/dev-packages/pack-n-play/test/test.ts
+++ b/core/dev-packages/pack-n-play/test/test.ts
@@ -18,8 +18,7 @@ import execa = require('execa');
 import {describe, it} from 'mocha';
 
 describe('pack-n-play', () => {
-  it('should run tests', async function () {
-    this.timeout(600000); // 10 minutes
+  it('should run tests', async () => {
     const fixturesPath = path.resolve('./test/fixtures');
     const dirs = fs
       .readdirSync(fixturesPath)
@@ -30,7 +29,7 @@ describe('pack-n-play', () => {
         stdio: 'inherit',
         cwd: dir,
       };
-      await execa('npm', ['install', '--no-audit', '--no-fund'], opts);
+      await execa('npm', ['install'], opts);
       await execa('npm', ['link', '../../../'], opts);
       await execa('npm', ['test'], opts);
     }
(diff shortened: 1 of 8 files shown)


core/dev-packages/pack-n-play/test/test.ts
```diff
@@ -29,7 +30,7 @@ describe('pack-n-play', () => {
         stdio: 'inherit',
         cwd: dir,
       };
-      await execa('npm', ['install'], opts);
+      await execa('npm', ['install', '--no-audit', '--no-fund'], opts);
```
westarle, line 33: I guess we should always do this for non-interactive npm installs, since we don't do anything with this info?

core/packages/nodejs-googleapis-common/test/test.discovery.ts
westarle, line 34: I think `discoverAPI` used to handle URLs without ports, so we should keep that part of the test (even if we need to pass the port to `nock`.)

Unless the test never passed?!

core/packages/nodejs-googleapis-common/test/test.discovery.ts
```diff
@@ -37,7 +37,7 @@ describe(__filename, () => {
     scope.done();
   });
   it('should discover an API through second weird path', async () => {
-    const discoUrl = 'http://test.local';
+    const discoUrl = 'http://test.local:80';
```
westarle: see above.

core/packages/gcp-metadata/src/index.ts
```diff
@@ -383,12 +383,22 @@ export async function isAvailable() {
...
+              if (err.code) {
+                return [err.code.toString()];
+              }
+              if (err.cause) {
+                return getErrorCodes(err.cause);
+              }
+              return ['UNKNOWN'];
+            };
+
+            const codes = getErrorCodes(errObj);
```
westarle, line 401: can you add tests to make sure this handles error.cause and AggregateError like the old code?

core/dev-packages/pack-n-play/test/test.ts
bshaffer: Yes i believe so, it's a small optimization

core/packages/nodejs-googleapis-common/test/test.discovery.ts
bshaffer: the reason we added the port is we switched to using `new URL()` API from the deprecated `url.parse()`, and the `new URL()` API requires a port.

core/packages/nodejs-googleapis-common/test/test.discovery.ts
bshaffer: Same as above, the reason we added the port is we switched to using new URL() API from the deprecated url.parse(), and the new URL() API requires a port.

core/packages/gcp-metadata/src/index.ts
bshaffer: done in ac1f4781c3

core/packages/nodejs-googleapis-common/test/test.discovery.ts
westarle: I'm not sure new URL() requires a port: https://nodejs.org/api/url.html#new-urlinput-base

If it did, I think we would need to update `discoverAPI` to infer a port based on the scheme (or modify the callers to make sure they pass a port)

I do think `nock` might need one.  So it might be more correct to pass a port to `nock` and skip the port for `discoverAPI`.

bshaffer: You're right, it must have been a hallucination 😦 I've reverted it.

core/packages/nodejs-googleapis-common/test/test.discovery.ts
bshaffer: See above, I've reverted it.

westarle (approved):

PR #9070, opened 2026-08-04
feat: add Otel Tracing helpers and tests

shivanee-p:
Introduces OpenTelemetry tracing helpers, trace context definitions, and a test harness in google-gax to support API call tracing across Google Cloud Node.js client libraries.

- Added @opentelemetry/api to dependencies.
- Wraps API call attempts in active spans, populates static and dynamic GCP trace attributes, and records exceptions and error metadata
- Defined DynamicTraceContext and StaticTraceContext
- Implemented @opentelemetry/api interfaces for unit testing
- Added unit tests verifying span creation, attribute population, error recording, and transport handling (grpc and http).

### DIFF
--- a/core/packages/gax/src/observability/TracerHelper.ts
+++ b/core/packages/gax/src/observability/TracerHelper.ts
@@ -1,0 +1,50 @@
+import { Span, trace, Tracer } from '@opentelemetry/api';
+
+export interface StaticTraceContext {
+  gcpClientService?: string;
+  gcpVersion?: string;
+  gcpRepo?: string;
+  gcpArtifact?: string;
+}
+
+export interface DynamicTraceContext {
+  clientName: string;
+  methodName: string;
+  rpcType: 'grpc' | 'http';
+}
+
+export function getGaxTracer(): Tracer {
+  return trace.getTracer('google-gax');
+}
+
+export async function traceAttempt(dynamicArgs: DynamicTraceContext, staticArgs: StaticTraceContext, fn: () => Promise<any>): Promise<any> {
+  const spanName = `${dynamicArgs.clientName}.${dynamicArgs.methodName}`;
+  return getGaxTracer().startActiveSpan(spanName, {}, async (span: Span) => {
+    span.setAttributes({
+      'gcp.client.service': staticArgs.gcpClientService,
+      'gcp.client.version': staticArgs.gcpVersion,
+      'gcp.repo': staticArgs.gcpRepo,
+      'gcp.artifact': staticArgs.gcpArtifact,
+      'gcp.method.name': dynamicArgs.methodName,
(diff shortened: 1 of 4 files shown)


core/packages/gax/src/observability/TracerHelper.ts
```diff
@@ -0,0 +1,70 @@
...
+  clientName: string;
+  methodName: string;
+  rpcType: 'grpc' | 'http';
+}
+
+export function getGaxTracer(): Tracer {
+  return trace.getTracer('google-gax');
+}
+
+export async function traceAttempt<T>(
```
danieljbruce: nit: We might be able to use generic constraints or a more specific type for T in this method.

core/packages/gax/src/observability/TracerHelper.ts
```diff
@@ -0,0 +1,70 @@
...
+        'error.message': err.message,
+        'error.type': err.constructor?.name ?? err.name,
+      });
+      span.recordException(err);
+      if (err.name) {
+        span.setAttribute('exception.type', err.name);
+      }
+      throw e;
+    } finally {
+      span.end();
```
danieljbruce, line 111: I'd love to see how this span records trace data in the context of one of the handwritten client libraries. That is likely out of scope for this PR, but some tests to show this would be really useful.

core/packages/gax/test/unit/otelHarness.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  getSpans(tracerName?: string): ReadableSpan[] {
+    const spans = this.exporter.getFinishedSpans();
+    if (tracerName) {
+      return spans.filter(
+        span =>
+          span.instrumentationScope?.name?.startsWith(tracerName) ||
+          (
+            span as unknown as {instrumentationLibrary?: {name?: string}}
```
danieljbruce: Not blocking, but I think we want to avoid the as unknown as x pattern if possible

core/packages/gax/src/observability/TracerHelper.ts
danieljbruce: nit: Maybe a JS doc comment about what traceAttempt does would be good.

danieljbruce (approved):
LGTM. A few nits, but feel free to just merge the PR and look at them later because they don't need to block.

core/packages/gax/src/observability/TracerHelper.ts
shivanee-p: Ack. Handwritten libraries are out of scope rn, but I'll be sure to add this within S2.

core/packages/gax/src/observability/TracerHelper.ts
shivanee-p: Ack

PR #9073, opened 2026-08-04
feat(firestore-api): implement subpath exports, declare long dependency, and discontinue v1beta1 

quirogas:
Parallel architectural update to `@google-cloud/firestore-api` supporting the handwritten Firestore SDK disentanglement (#8928).

### Summary of Changes
- **Runtime `long` Dependency**: Moves `"long"` from `devDependencies` to runtime `"dependencies"`. Generated declaration files (`protos.d.ts`) explicitly import `long`. In environments using strict package managers with node_modules isolation (such as `pnpm`), omitting `long` from runtime dependencies causes TypeScript compilation to fail with `error TS2307: Cannot find module 'long'` (per structural context in `googleapis/nodejs-firestore#2322`).
- **Subpath Exports**: Configures conditional subpath exports (`./v1`, `./client_v1`, `./admin_v1`, `./protos`, and configuration files) with the `"types"` condition placed before `"default"`. This ensures declaration files resolve correctly for consumers using modern TypeScript module resolution (`"moduleResolution": "nodenext"`).
- **Discontinue `v1beta1`**: Removed `v1beta1` from `librarian.yaml` and regenerated the library via `librarian` to stop generating beta client code.

Internal: b/531788771


packages/google-cloud-firestore-api/package.json
```diff
@@ -11,6 +11,46 @@
...
+      "default": "./build/src/index.js"
+    },
+    "./v1": {
+      "types": "./build/src/v1/index.d.ts",
+      "default": "./build/src/v1/index.js"
+    },
+    "./v1/*": {
+      "types": "./build/src/v1/*.d.ts",
+      "default": "./build/src/v1/*"
+    },
```
MarkDuckworth: Consider removing these exports with asterisks, unless we have a specific reason to include them? These can lead to breaking changes if an internal file is imported by a customer and that internal file is renamed, refactored, or deleted.

packages/google-cloud-firestore-api/package.json
```diff
@@ -11,6 +11,46 @@
...
+      "default": "./build/src/v1/firestore_admin_client.js"
+    },
+    "./protos": {
+      "types": "./build/protos/protos.d.ts",
+      "default": "./build/protos/protos.js"
+    },
+    "./protos/*": {
+      "types": "./build/protos/*.d.ts",
+      "default": "./build/protos/*"
+    },
```
MarkDuckworth: Consider removing these exports with asterisks, unless we have a specific reason to include them? These can lead to breaking changes if an internal file is imported by a customer and that internal file is renamed, refactored, or deleted.

packages/google-cloud-firestore-api/package.json
```diff
@@ -11,6 +11,46 @@
...
+    "./build/protos/protos": {
+      "types": "./build/protos/protos.d.ts",
+      "default": "./build/protos/protos.js"
+    },
+    "./build/src/v1": {
+      "types": "./build/src/v1/index.d.ts",
+      "default": "./build/src/v1/index.js"
+    },
+    "./build/*": "./build/*",
```
MarkDuckworth: Consider removing these exports with asterisks, unless we have a specific reason to include them? These can lead to breaking changes if an internal file is imported by a customer and that internal file is renamed, refactored, or deleted.

MarkDuckworth (approved):
Approved, but I'm strongly in favor of removing the exports with wildcards from package.json, unless these are required.

PR #9101, opened 2026-08-06
fix: prefer native fetch over node-fetch in Node.js

HelloWNW:
## Problem

`Gaxios.#getFetch()` only checks for a browser `window` before falling back to
`node-fetch`, so Node.js always uses `node-fetch` even though native `fetch` has
been available since Node 18 — this package's minimum supported version.

`node-fetch` intermittently rejects with `Premature close` on some networks. This
reproduces as a consistent `clasp login` failure, whose OAuth2 token exchange runs
through google-auth-library and gaxios:

```
Invalid response body while trying to fetch https://oauth2.googleapis.com/token: Premature close
```

The same POST succeeds via curl and via Node's native `fetch` on the same machine,
so the failure is specific to `node-fetch`.

## Change

Prefer native `fetch` in Node.js, keeping `node-fetch` as a fallback when no global
`fetch` exists. Browser behavior is unchanged.

Two implementation differences are normalized to preserve existing behavior:

- `responseType: 'stream'` converts native `fetch`'s `ReadableStream` back to a
  `stream.Readable`. A caller-supplied `fetchImplementation` still returns its own
  body type, as covered by the existing test.
- An `AbortSignal` reason that is not an `Error` is propagated rather than replaced
  with a generic message. The existing test for this case carried a comment
  anticipating the change, which is now applied.

## Testing

`npm test` — 129 passing.

### DIFF
--- a/core/packages/gaxios/src/gaxios.ts
+++ b/core/packages/gaxios/src/gaxios.ts
@@ -283,16 +283,29 @@ export class Gaxios implements FetchCompliance {
     }
 
     switch (opts.responseType) {
-      case 'stream':
-        if (usingInternalFetch && res.body && !(res.body instanceof Readable)) {
-          // Native `fetch` resolves a `ReadableStream`, so convert it to retain
-          // the `stream.Readable` contract. A caller-provided
-          // `fetchImplementation` keeps its own body type.
+      case 'stream': {
+        // Native `fetch` resolves a `ReadableStream`, so convert it to retain
+        // the `stream.Readable` contract. Browser bundles stub or polyfill
+        // `stream` without `fromWeb` and have always resolved a
+        // `ReadableStream`, so the conversion is skipped there. A
+        // caller-provided `fetchImplementation` keeps its own body type.
+        const body = res.body as
+          | (ReadableStream & {getReader?: unknown})
+          | null;
+        const isWebStream = typeof body?.getReader === 'function';
+
+        if (
+          usingInternalFetch &&
+          isWebStream &&
+          typeof Readable?.fromWeb === 'function'
+        ) {
           return Readable.fromWeb(
-            res.body as unknown as import('stream/web').ReadableStream,
+            body as unknown as import('stream/web').ReadableStream,
           );
         }
+
         return res.body;
+      }
       case 'json': {
         const data = await res.text();
         try {
--- a/core/packages/gaxios/src/gaxios.ts
+++ b/core/packages/gaxios/src/gaxios.ts
@@ -147,6 +147,8 @@ export class Gaxios implements FetchCompliance {
   private async _defaultAdapter<T>(
     config: GaxiosOptionsPrepared,
   ): Promise<GaxiosResponse<T>> {
+    const usingInternalFetch =
+      !config.fetchImplementation && !this.defaults.fetchImplementation;
     const fetchImpl =
       config.fetchImplementation ||
       this.defaults.fetchImplementation ||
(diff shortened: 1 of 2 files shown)


HelloWNW:
Addressed in e8ec31e — the conversion now checks that fromWeb is available before using it, and no longer references instanceofReadable.

 This is my first contribution to a project of this size, so apologies for any review overhead. Thanks for taking a look.

feywind:
@HelloWNW Thanks for the PR! My main comment would honestly be, if the built-in fetch is overall better, maybe we should just use it (and not worry about the `usingInternalFetch`). With the supply chain issues going around, we've been trying to move to more built-ins anyhow.

ScottMansfield:
Adding a data point for @feywind's "we've been trying to move to more built-ins anyhow" — the dependency-weight case for this change is stronger than it looks, and I don't think it's been quantified on either this PR or #8107.

**Removing the `node-fetch` fallback drops 8.8 MB and 6 packages from every install.**

Measured on a clean `npm i @google/genai@2.9.0` (Node v26.7.0), which reaches gaxios via `google-auth-library`:

| package | installed |
|---|---:|
| `web-streams-polyfill` | **8.8 MB** |
| `node-domexception` | 160 KB |
| `node-fetch` | 148 KB |
| `formdata-polyfill` | 52 KB |
| `fetch-blob` | 48 KB |
| `data-uri-to-buffer` | 24 KB |

That's 28% of a 32 MB install tree, for a transitive chain rooted entirely in `#getFetch()`'s `await import('node-fetch')`.

**The 8.8 MB is never even loaded.** `web-streams-polyfill` arrives via `fetch-blob`, whose `streams.cjs` guards it:

```js
if (!globalThis.ReadableStream) {
  try   { Object.assign(globalThis, require('node:stream/web')) }        // Node 16.5+
  catch { Object.assign(globalThis, require('web-streams-polyfill/…')) } // fallback
}
```

On any Node ≥ 16.5 the branch is skipped entirely. Traced with a `Module._resolveFilename` hook, `require('fetch-blob')` on Node 20+ resolves `web-streams-polyfill` **zero times**. It is pure install-time weight on every supported runtime.

It is also 8.8 MB for a ~200 KB library because `web-streams-polyfill@3.3.3` ships twelve build variants (`polyfill`/`ponyfill` × `es5`/`es6`/`es2018` × `.js`/`.mjs`/`.min.js`), each with a ~400 KB sourcemap. Its own v4 is 441 KB, but `fetch-blob` pins `^3.0.3`, so consumers can't fix it downstream without an override.

**Scale.** gaxios is at ~77.7M weekly downloads and sits under `google-auth-library` (~69.6M/wk), `googleapis`, and the `@google-cloud/*` clients, so this is one dependency edge in front of most of the Node GCP surface.

**Independently verified.** Overriding `node-fetch` with a nine-line shim that re-exports `globalThis.fetch`/`Headers`/`Request`/`Response`/`FormData`/`Blob`/`File`:

```
gaxios status: 200 | body bytes: 559
GoogleAuth constructs OK: true
```

Install tree goes 34 MB → 25 MB with no code changes. That's obviously not a substitute for doing it properly in `#getFetch()`, but it does suggest the native path is drop-in for the common request shapes.

On the `usingInternalFetch` question — agreed with @feywind and with @Sainan's note on #8107 that, given `engines.node: ">=18"`, the conditional may not be needed at all. Dropping `node-fetch` outright rather than keeping it as a fallback is what actually removes the 8.8 MB; a conditional that still lists `node-fetch` in `dependencies` keeps the install cost for everyone and only saves the runtime import.

PR #9103, opened 2026-08-06
feat(firestore)!: update minimum Node version to 22

MarkDuckworth:


quirogas (approved):

quirogas:
Labeling as "do not merge" because we should ensure that this PR and https://github.com/googleapis/google-cloud-node/pull/8928 are included in the same release, as they both represent the same breaking change.

PR #9104, opened 2026-08-06
feat: update API sources and regenerate

sofisl:


sofisl:
we should merge https://github.com/googleapis/google-cloud-node/pull/9109 first

packages/google-ads-datamanager/protos/google/ads/datamanager/v1/ad_event.proto
```diff
@@ -101,9 +101,9 @@ message AdEvent {
   // found.
   UserData user_data = 7 [(google.api.field_behavior) = OPTIONAL];
 
-  // Optional. Information gathered about the device being used when the ad
+  // Required. Information gathered about the device being used when the ad
   // event happened.
-  DeviceInfo device_info = 8 [(google.api.field_behavior) = OPTIONAL];
+  DeviceInfo device_info = 8 [(google.api.field_behavior) = REQUIRED];
```
feywind, line 106: Is that going to cause a breaking change?

feywind (approved):
I know my comment is not about your stuff, but the protos themselves... just wanted to make sure!

packages/google-ads-datamanager/protos/google/ads/datamanager/v1/ad_event.proto
sofisl: this does look breaking to me. Technically I don't think ads has the same restrictions as cloud for breaking changes, but it doesn't mean we shouldn't specify it.

Let me bring it up with the Librarian team next week.

quirogas: Yes, technically, this is backwards incompatible -- [AIP 203]( https://google.aip.dev/203#backwards-compatibility) describes this verbatim in the examples for backwards incompatible changes -- However there is an exception tracking this change b/538653921 as backwards compatible.

PR #9122, opened 2026-08-11
fix(firestore): observe rejection of lazily-started transaction ID promise

rockwotj:
## Problem

When the first read of a transaction fails, the client can emit an `unhandledRejection`. Under Node's default `--unhandled-rejections=throw` this terminates the process, even though the application handled the error that `runTransaction()` returned.

`withLazyStartedTransaction()` derives a second promise from the first read:

```js
const resultPromise = resultFn.call(this, param, opts);

this._transactionIdPromise = resultPromise.then(r => { ... });

return resultPromise.then(r => r.result);
```

The caller handles the returned promise. Nothing attaches a handler to `_transactionIdPromise` at creation time. Its only handler comes later, from a subsequent read, `commit()`, or `rollback()`. Two cases leave it unobserved:

1. **Read-only transactions.** `rollback()` returns early when `this._writeBatch` is unset, so it never awaits `_transactionIdPromise`. No handler is ever attached.
2. **Read-write transactions where the callback continues after a failed read.** `rollback()` does await the promise, but only after the callback returns. If the callback yields to the event loop in between, Node has already reported the rejection. Node also logs `PromiseRejectionHandledWarning` here, which confirms the handler arrived too late.

Case 1 is deterministic: for a read-only transaction, no handler is ever attached, so any failure of the first read terminates the process.

We hit case 1 in production on `@google-cloud/firestore@8.7.1`. The first read of a read-only transaction, a `BatchGetDocuments`, stalled for 293 seconds and then failed with `16 UNAUTHENTICATED`. The application caught the error from `runTransaction()`, and the process still died on the orphaned rejection. Read-only transactions set `_maxAttempts` to 1, so no retry masked it.

## Reproduction

Point the client at a port where nothing listens, then run a read-only transaction whose first read fails:

```js
const firestore = new Firestore({
  projectId: 'repro-project',
  host: 'localhost:1',
  ssl: false,
});

process.on('unhandledRejection', err => console.log('unhandled:', err.code));

try {
  await firestore.runTransaction(
    async txn => {
      await txn.get(firestore.doc('things/one'));
    },
    {readOnly: true},
  );
} catch {
  // The caller handles the error, and the process still dies without the
  // `process.on` above.
}
```

Before this change the listener fires with code 14. After it, it does not. The same script reproduces case 2 with a read-write transaction whose callback swallows the read error and then awaits a timer.

## Fix

Attach a no-op handler to `_transactionIdPromise` when it is created. This marks the rejection observed. Later awaiters in `commit()` and `rollback()` still reject exactly as before, so error propagation is unchanged.

## Test

Adds a regression test to `handwritten/firestore/dev/test/transaction.ts`. It records `unhandledRejection` events while a read-only transaction's first read fails, and asserts none are emitted.

Verified both directions against the `firestore` package:

- Without the source change: `1 failing` — `expected [ Error: Test Error, …(1) ] to be empty`.
- With it: the new test passes and the full `transaction` suite is green at `37 passing`.

### DIFF
--- a/handwritten/firestore/dev/src/transaction.ts
+++ b/handwritten/firestore/dev/src/transaction.ts
@@ -791,7 +791,7 @@ export class Transaction implements firestore.Transaction {
         // macrotask boundary. Node then reports an unhandled rejection, which
         // terminates the process under the default `--unhandled-rejections=throw`.
         // Observe the rejection here; awaiters still see it.
-        this._transactionIdPromise.catch(() => {});
+        void this._transactionIdPromise.catch(() => {});
 
         return resultPromise.then(r => r.result);
       }
--- a/handwritten/firestore/dev/src/transaction.ts
+++ b/handwritten/firestore/dev/src/transaction.ts
@@ -784,6 +784,15 @@ export class Transaction implements firestore.Transaction {
           return r.transaction;
         });
 
+        // Nothing awaits `_transactionIdPromise` until a subsequent read, a
+        // commit, or a rollback. When the first read is the operation that
+        // fails, `rollback()` returns early for read-only transactions and never
+        // awaits it, and a read-write transaction can leave it rejected across a
+        // macrotask boundary. Node then reports an unhandled rejection, which
+        // terminates the process under the default `--unhandled-rejections=throw`.
+        // Observe the rejection here; awaiters still see it.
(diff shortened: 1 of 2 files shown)


dlarocque (approved):

rockwotj:
@MarkDuckworth would you be willing to approve?

PR #9134, opened 2026-08-12
refactor(storage): Migrate conformance tests to StorageTransport and enhance retry logic

thiyaguk09:


handwritten/storage/src/storage-transport.ts
```diff
@@ -239,8 +243,14 @@ export class StorageTransport {
...
+          .then(resp => {
+            callback(null, decorateMetadata(resp), resp);
+            return resp;
+          })
+          .catch(err => {
+            callback(err, null, err.response);
+            throw err;
+          });
         return requestPromise;
       }
```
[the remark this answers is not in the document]
thiyaguk09: return was updated

handwritten/storage/conformance-test/conformanceCommon.ts
```diff
@@ -70,6 +67,23 @@ const TESTBENCH_HOST =
...
 const RETRY_MULTIPLIER_FOR_CONFORMANCE_TESTS = 0.01;
+const SERVICE_ACCOUNT = path.join(
+  getDirName(),
+  '../../../conformance-test/fixtures/signing-service-account.json',
+);
+
+const authClient = new GoogleAuth({
+  keyFilename: SERVICE_ACCOUNT,
+  scopes: ['https://www.googleapis.com/auth/devstorage.full_control'],
+}).fromJSON(require(SERVICE_ACCOUNT));
```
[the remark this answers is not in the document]
thiyaguk09: imported `fs` and replaced dynamic require() with fs.

handwritten/storage/conformance-test/libraryMethods.ts
```diff
@@ -51,100 +52,151 @@ export interface ConformanceTestOptions {
...
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      pageToken = (listResult as any)?.nextPageToken;
+    } while (pageToken);
+
+    const deleteBucketReq: StorageRequestOptions = {
+      method: 'DELETE',
+      url: `storage/v1/b/${bucketName}`,
+    };
+    await options.storageTransport.makeRequest(deleteBucketReq);
   }
```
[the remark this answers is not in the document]
thiyaguk09: `encodeURIComponent` was added.

handwritten/storage/src/storage.ts
```diff
@@ -1297,36 +1297,35 @@ export class Storage {
...
-          callback(
-            null,
-            hmacKey,
-            hmacKey.secret,
-            resp as unknown as HmacKeyResourceResponse,
-          );
-        },
-      )
-      .catch(err => callback!(err));
+    return (this.storageTransport as any).makeRequest(
```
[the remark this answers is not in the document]
thiyaguk09: removed return before `(this.storageTransport as any).makeRequest(...)`

handwritten/storage/conformance-test/conformanceCommon.ts
```diff
@@ -15,22 +15,19 @@
...
 import * as assert from 'assert';
 import {
   StorageRequestOptions,
   StorageTransport,
-} from '../src/storage-transport';
+  StorageTransportCallback,
+} from '../src/storage-transport.js';
+import {getDirName} from '../src/util.js';
+import path from 'path';
+import {GoogleAuth} from 'google-auth-library';
```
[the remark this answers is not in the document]
thiyaguk09: imported

Dhriti07 (commented):
Hi, 
One major concern I have with the changes in libraryMethods.ts is that we're replacing calls to the actual SDK methods with direct storageTransport.makeRequest() calls.
The main goal of the conformance test suite is to verify that our public SDK layer adheres to GCS contracts, retries, and preconditions.  

We lose test coverage for the SDK itself -any regression in say bucket.addLifecycleRule or file.save would go unnoticed since the tests wouldn't even run that code. 

Could we instead keep calling the high-level SDK methods on bucket/file, have the Storage instance use StorageTransport under the hood, and inject the `x-retry-test-id` header via a request interceptor?

thiyaguk09:
> Hi, One major concern I have with the changes in libraryMethods.ts is that we're replacing calls to the actual SDK methods with direct storageTransport.makeRequest() calls. The main goal of the conformance test suite is to verify that our public SDK layer adheres to GCS contracts, retries, and preconditions.
> 
> We lose test coverage for the SDK itself -any regression in say bucket.addLifecycleRule or file.save would go unnoticed since the tests wouldn't even run that code.
> 
> Could we instead keep calling the high-level SDK methods on bucket/file, have the Storage instance use StorageTransport under the hood, and inject the `x-retry-test-id` header via a request interceptor?

We restored all high-level SDK method calls in libraryMethods.ts and switched to injecting the x-retry-test-id header via a request interceptor on StorageTransport. All conformance and unit tests are  passing.

handwritten/storage/src/nodejs-common/service-object.ts
```diff
@@ -478,27 +490,32 @@ class ServiceObject<T, K extends BaseMetadata> extends EventEmitter {
...
+      queryParameters: {
+        ...methodConfig.reqOpts?.queryParameters,
+        ...query,
+      },
+    };
+
+    if (callback) {
+      this.storageTransport
+        .makeRequest<K>(req, (err, data, resp) => {
+          this.metadata = data!;
```
Dhriti07: Did this accidentally get dropped during the merge conflict? If the request fails, this.metadata = data! will wipe out any cached metadata with undefined. Should we keep the if (!err && data) guard here.

handwritten/storage/src/storage.ts
```diff
@@ -58,7 +58,7 @@ export interface ServiceAccount {
 export type GetServiceAccountResponse = [ServiceAccount, unknown];
 export interface GetServiceAccountCallback {
   (
-    err: Error | null,
+    err: GaxiosError | null,
```
Dhriti07: Is changing err from Error | null to GaxiosError | null intentional? All the other public callbacks in storage.ts still use Error | null, and narrowing this might be a breaking change for anyone explicitly typing their callback with Error ?

Dhriti07 (commented):
Hi, 
Thanks for fixing the previous changes, As I was reviewing this pr along with https://github.com/googleapis/google-cloud-node/pull/9133 I noticed both the branches include include a mix of core library updates alongside the respective test changes.

Extracting the code changes

handwritten/storage/src/nodejs-common/service-object.ts
thiyaguk09: Yes, that guard got dropped accidentally during the merge conflict resolution. Thanks, restored now.

handwritten/storage/src/storage.ts
thiyaguk09: Got it. Reverting this back to Error | null.

handwritten/storage/src/nodejs-common/service-object.ts
```diff
@@ -543,35 +562,40 @@ class ServiceObject<T, K extends BaseMetadata> extends EventEmitter {
...
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    };
+
+    if (callback) {
+      this.storageTransport
+        .makeRequest<K>(req, (err, data, resp) => {
+          if (err) {
+            callback(err);
```
Dhriti07: It looks like we're dropping the resp argument in the error callback. The original code used to pass (err, undefined, resp). We should probably restore the resp argument so we don't break backward compatibility for users who inspect the raw HTTP response on failures

handwritten/storage/conformance-test/conformanceCommon.ts
```diff
@@ -117,92 +138,120 @@ export function executeScenario(testCase: RetryTestCase) {
...
+                assert.strictEqual(testBenchResult.completed, true);
+              } else {
+                await assert.rejects(async () => {
+                  await storageMethodObject(methodParameters);
+                }, undefined);
+              }
+            } finally {
+              storage.interceptors = [];
+              storage.storageTransport.gaxiosInstance?.interceptors?.request?.clear();
+              defaultGaxios.instance?.interceptors?.request?.clear();
```
Dhriti07: Instead of calling .clear() on the default gaxios interceptors, should we capture the interceptor ID returned by .add(interceptor) and use .eject(id)? Since defaultGaxios.instance is global, calling .clear() might wipe out interceptors from other tests if test concurrency is ever introduced, which could lead to flaky tests.

handwritten/storage/conformance-test/conformanceCommon.ts
```diff
@@ -117,92 +138,120 @@ export function executeScenario(testCase: RetryTestCase) {
...
+            try {
+              if (testCase.expectSuccess) {
+                await storageMethodObject(methodParameters);
+                const testBenchResult = await getTestBenchRetryTest(
+                  creationResult.id,
+                  storageTransport,
+                );
+                assert.strictEqual(testBenchResult.completed, true);
+              } else {
+                await assert.rejects(async () => {
```
Dhriti07, line 238: Should we also be asserting assert.strictEqual(testBenchResult.completed, true); in this failure block like we do for the success case? If a bug causes the SDK to immediately fail without exhausting the required retries, assert.rejects will catch the error and the test will pass, but we wouldn't catch that the retry logic actually failed to execute.

handwritten/storage/src/nodejs-common/service-object.ts
thiyaguk09: Done. Updated to `callback(err, undefined, resp);` to preserve the response argument and maintain backward compatibility.

handwritten/storage/conformance-test/conformanceCommon.ts
thiyaguk09: Removed `.clear()` and updated the finally block to remove only the test interceptor via `.delete(interceptor)`.

handwritten/storage/conformance-test/conformanceCommon.ts
thiyaguk09: Added `getTestBenchRetryTest` and `assert.strictEqual(testBenchResult.completed, true);`.

handwritten/storage/src/storage.ts
```diff
@@ -918,8 +925,12 @@ export class Storage {
...
     this.retryOptions = config.retryOptions;
 
-    this.storageTransport = new StorageTransport({...config, ...options});
-    this.interceptors = [];
+    this.interceptors = options.interceptors_ || [];
+    this.storageTransport = new StorageTransport({
+      ...config,
+      ...options,
+      interceptors: this.interceptors,
+    });
```
dashpole: Spreading `...options` after `...config` overwrites `config.retryOptions` whenever a caller passes partial `retryOptions` (like `{retryDelayMultiplier: 0.01}`), dropping all the merged defaults (`maxRetries`, `maxRetryDelay`, `totalTimeout`, `autoRetry`, `idempotencyStrategy`). Passing `retryOptions: this.retryOptions` after `...options` preserves the defaults and keeps the object reference shared with `StorageTransport`:
```suggestion
    this.storageTransport = new StorageTransport({
      ...config,
      ...options,
      retryOptions: this.retryOptions,
      interceptors: this.interceptors,
    });
```

handwritten/storage/src/nodejs-common/service-object.ts
```diff
@@ -313,28 +313,40 @@ class ServiceObject<T, K extends BaseMetadata> extends EventEmitter {
...
             if (err.status === 404 && ignoreNotFound) {
               err = null;
             }
           }
           callback(err, resp);
-        },
-      )
-      .catch(err => callback!(err));
+        })
+        .catch(err => callback!(err));
```
dashpole: When `StorageTransport.makeRequest` fails, it invokes its callback argument and also returns a rejected `enrichedPromise`. Because of that, `.catch(err => callback!(err))` will call `callback` a second time on errors (and for a 404 with `ignoreNotFound: true`, the inner callback calls `callback(null, resp)` and then `.catch` immediately calls `callback(err)` with the 404 error). Consider guarding `callback` so it is only invoked once (and applying the same `ignoreNotFound` check in `.catch`, as well as guarding `.catch` in `getMetadata` and `setMetadata`).

handwritten/storage/conformance-test/conformanceCommon.ts
```diff
@@ -117,92 +133,124 @@ export function executeScenario(testCase: RetryTestCase) {
...
-              jsonMethod?.name.toString(),
-              storageTransport,
+            bucket = await createBucketForTest(
+              storage,
+              testCase.preconditionProvided &&
+                !storageMethodString.includes('combine'),
+              storageMethodString,
+            );
+            file = await createFileForTest(
+              testCase.preconditionProvided,
```
dashpole: Passing `testCase.preconditionProvided` for all methods attaches instance-level `preconditionOpts` to `bucket` and `file` even during non-`InstancePrecondition` tests, which are meant to verify method-level preconditions on plain instances. We should only set instance preconditions when `storageMethodString.includes('InstancePrecondition')`:
```suggestion
              Boolean(testCase.preconditionProvided) &&
                storageMethodString.includes('InstancePrecondition') &&
                !storageMethodString.includes('combine'),
              storageMethodString,
            );
            file = await createFileForTest(
              Boolean(testCase.preconditionProvided) &&
                storageMethodString.includes('InstancePrecondition'),
```

Dhriti07 (approved):

PR #9141, opened 2026-08-13
feat(spanner-driver): integrate native spannerlib-node wrapper and implement type system

surbhigarg92:
This PR integrates the native `spannerlib-node` wrapper into `@google-cloud/spanner-driver` and implements the complete PostgreSQL-compatible data type codec and parsing system following standard `node-postgres` (pg) conventions. It enables connecting to Cloud Spanner PostgreSQL-dialect databases, executing parameterized queries, decoding all Cloud Spanner scalar and array data types, streaming rows/events, managing transactions (BEGIN, COMMIT, ROLLBACK), and pooling connections.

### 1. Native Bridge Integration (spannerlib-node)
Connection & Pool Bridge: Integrated Connection, Pool, and Rows from `spannerlib-node` in `src/lib/native.ts`
Execution & Teardown Lifecycle: Implemented connection handshake, sequential query task queueing, and safe resource closure in Client and Pool.

### 2. Comprehensive Type System & Codecs

- Protobuf to PostgreSQL Catalog OID Mapping: Implemented `mapSpannerTypeToPgOid` in `src/lib/pg/types.ts` , mapping all Spanner TypeCodes and TypeAnnotationCodes to standard numeric PostgreSQL OIDs.
- Type Parsers: Built default text parsers converting Spanner raw values to native JavaScript types (string, number, boolean, Date, Buffer, object, Array).
- TypeOverrides Registry: Added scoped and hierarchical TypeOverrides support matching pg.types for custom parser overrides per client or query.
- ORM .toPostgres() Serialization: Added automatic parameter serialization for objects implementing .toPostgres() (e.g. Knex/ORMs) in `src/lib/pg/utilities.ts`

### DIFF
--- a/handwritten/spanner-driver/README.md
+++ b/handwritten/spanner-driver/README.md
@@ -131,10 +131,12 @@ console.log(`Waiting Requests:  ${pool.waitingCount}`); // Queued queries waitin
 
 ---
 
-### 3. Streaming Rows & Events
+### 3. Row & Query Event Listeners
+
+`Query` instances extend `EventEmitter`, allowing you to listen to `row`, `fields`, `end`, and `error` events:
 
 ```typescript
-import { Client, Query } from '@google-cloud/spanner-driver';
+import { Client } from '@google-cloud/spanner-driver';
 
 const client = new Client({
   project: 'my-gcp-project',
@@ -144,13 +146,15 @@ const client = new Client({
 
 await client.connect();
 
-// Stream rows as they arrive from Spanner gRPC stream
-client.query('SELECT * FROM large_table')
+// Listen to query events as rows are decoded
+client.query('SELECT id, name FROM users')
   .on('fields', fields => console.log('Column metadata:', fields))
   .on('row', row => console.log('Received Row:', row))
   .on('end', result => console.log('Query completed. Row count:', result.rowCount))
   .on('error', err => console.error('Query error:', err));
 ```
+
+> **Note on Memory**: Listening to `.on('row')` events does not bypass in-memory row buffering in `QueryResult.rows`. For large queries, use SQL pagination (`LIMIT` / `OFFSET`).
 
 ---
 
@@ -233,6 +237,11 @@ When working with Cloud Spanner's PostgreSQL dialect, keep the following behavio
    - SQL query projections and expressions can generate multidimensional arrays (e.g. `SELECT ARRAY[ARRAY[1, 2], ARRAY[3, 4]]`). The driver's array parser recursively decodes nested arrays into multidimensional JavaScript arrays (e.g. `[[1, 2], [3, 4]]`).
 4. **64-bit Integer Precision (`INT8` / `BIGINT`)**:
    - By default, `INT8` columns are returned as strings to prevent precision loss beyond JavaScript's 53-bit `Number.MAX_SAFE_INTEGER` ($9,007,199,254,740,991$). Use `types.setTypeParser(BuiltinOids.INT8, BigInt)` or `Number` based on your application's requirements.
+5. **Memory Usage & Result Set Buffering**:
+   - Like standard `node-postgres`, calling `client.query()` or `pool.query()` buffers all returned rows into the in-memory `QueryResult.rows` array before resolving the Promise or callback.
+   - For large datasets, use SQL query pagination (`LIMIT` / `OFFSET`) or selective filtering to avoid excessive memory consumption.
+6. **Wire Format & Parsers**:
+   - The driver communicates with Cloud Spanner via gRPC Protobuf rather than PostgreSQL's raw TCP wire protocol. Type parsers operate on text, JSON, and structured Protobuf representations; binary wire protocol parsers (`format: 'binary'`) are not applicable.
 
 ---
 
--- a/handwritten/spanner-driver/README.md
+++ b/handwritten/spanner-driver/README.md
@@ -1,6 +1,6 @@
 # Google Spanner Node.js Driver (`@google-cloud/spanner-driver`)
 
-The `@google-cloud/spanner-driver` package provides a high-performance, `node-postgres` (`pg`) compatible client and connection pool interface for Google Spanner. It bridges Node.js applications directly to Spanner using a native Go CGO engine, delivering full PostgreSQL dialect support.
+The `@google-cloud/spanner-driver` package provides a high-performance, `node-postgres` (`pg`) compatible client and connection pool interface for Cloud Spanner, delivering full PostgreSQL dialect compatibility with low latency and seamless ORM integration.
 
 [![npm version](https://img.shields.io/npm/v/@google-cloud/spanner-driver.svg)](https://www.npmjs.com/package/@google-cloud/spanner-driver)
 [![License](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](LICENSE)
@@ -10,10 +10,14 @@ The `@google-cloud/spanner-driver` package provides a high-performance, `node-po
 ## Key Features
 
 - **`node-postgres` Compatibility**: Drop-in compatible `Client` and `Pool` interfaces matching standard PostgreSQL drivers.
-- **Connection Pooling**: Full-featured connection pool (`Pool`) with idle eviction, connection recycling (`maxUses`, `maxLifetimeSeconds`), and backpressure wait queues.
-- **Dual ESM & CommonJS**: Full support for both `import` (ESM) and `require()` (CommonJS) modules.
-- **PostgreSQL Dialect Utilities**: Escaping tools (`escapeIdentifier`, `escapeLiteral`) and SQLSTATE error code enrichment (`DatabaseError`).
-- **Flexible Invocation Modes**: Supports Promises (`async`/`await`), Node callbacks, and streaming row event emitters.
+- **High Performance**: Optimized, low-overhead communication directly with Cloud Spanner.
+- **Transaction Support**: Explicit transaction management (`BEGIN`, `COMMIT`, `ROLLBACK`) with transaction readiness status tracking (`client.txStatus`).
+- **Robust Connection Pooling**: Full-featured connection pool (`Pool`) with idle eviction, connection recycling (`maxUses`, `maxLifetimeSeconds`), backpressure wait queues, and async `onConnect` initialization hooks.
+- **Custom Type System**: `pg.types`-compatible `TypeOverrides` registry allowing global, per-client, or per-query custom parsers for PostgreSQL OID types.
+- **Rich Parameter Serialization**: Automatic encoding of JavaScript primitives, `Date` objects, `Buffer` bytes, JSON, and ORM objects implementing `.toPostgres()`.
+- **Flexible Invocation Modes**: Full support for Promises (`async`/`await`), Node callbacks (`(err, res) => ...`), and streaming row event emitters (`.on('row')`, `.on('fields')`).
+- **Dual ESM & CommonJS**: Native exports supporting both modern `import` (ESM) and legacy `require()` (CommonJS).
+- **PostgreSQL Utilities**: Escaping helpers (`escapeIdentifier`, `escapeLiteral`) and SQLSTATE error code enrichment (`DatabaseError`).
 
 ---
 
@@ -39,15 +43,16 @@ const client = new Client({
   database: 'my-spanner-database',
 });
 
-// Option B: Connection DSN String or postgresql:// URL
+// Option B: Connection DSN Resource String
 // const client = new Client('projects/my-gcp-project/instances/my-spanner-instance/databases/my-spanner-database');
 
 async function main() {
+  // connect() returns Promise<Client>
   await client.connect();
 
   // Executing queries with positional parameters ($1, $2, etc.)
   const result = await client.query(
-    'SELECT user_id, email FROM users WHERE status = $1',
+    'SELECT user_id, email, created_at FROM users WHERE status = $1',
     ['ACTIVE']
   );
 
@@ -60,6 +65,8 @@ async function main() {
 
 main().catch(console.error);
 ```
+
+---
 
 ### 2. Connection Pooling via `Pool`
 
@@ -83,7 +90,6 @@ const pool = new Pool({
   },
 });
 
-// Pool Lifecycle Events (Fire-and-forget notifications)
 // Note: 'connect' event listeners do NOT wait for async functions; use onConnect option for async setup.
 pool.on('connect', client => console.log('New client connected to pool'));
 pool.on('acquire', client => console.log('Client checked out from pool'));
@@ -115,7 +121,7 @@ async function shutdown() {
 
 #### Monitoring Pool Metrics
 
-You can inspect real-time connection metrics on the `Pool` instance:
+You can inspect real-time connection metrics directly on the `Pool` instance:
 
 ```typescript
 console.log(`Total Connections: ${pool.totalCount}`);   // Total clients in pool (active + idle)
@@ -123,10 +129,12 @@ console.log(`Idle Connections:  ${pool.idleCount}`);    // Clients currently ava
 console.log(`Waiting Requests:  ${pool.waitingCount}`); // Queued queries waiting for an available client
 ```
 
-### 3. Streaming Rows & Callbacks
+---
+
+### 3. Streaming Rows & Events
 
 ```typescript
-import { Client } from '@google-cloud/spanner-driver';
+import { Client, Query } from '@google-cloud/spanner-driver';
 
 const client = new Client({
   project: 'my-gcp-project',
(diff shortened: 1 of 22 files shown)


handwritten/spanner-driver/src/lib/pg/types.ts
```diff
@@ -0,0 +1,808 @@
...
+  // Fallback to pass-through string
+  return parseString;
+}
+
+/**
+ * TypeOverrides provides a scoped type parser and encoder registry.
+ * Matches `pg-types.TypeOverrides` and allows overriding type parsers per query, client, or globally.
+ */
+export class TypeOverrides implements ITypeOverrides {
+  private readonly _parsers: Map<number, TypeParser> = new Map();
```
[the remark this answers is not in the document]
surbhigarg92: We dont support "GoogleSQL" dialect for now. So we need not implement this

handwritten/spanner-driver/src/lib/pg/types.ts
```diff
@@ -0,0 +1,808 @@
...
+    if (elementOid !== undefined) {
+      const elemParser = this.getTypeParser(elementOid, format);
+      const arrayParser = (val: unknown) => parsePgArray(val, elemParser);
+      this._arrayParserCache.set(numOid, arrayParser);
+      return arrayParser;
+    }
+    // 6. Fallback to global built-in defaults
+    return getDefaultTypeParser(numOid);
+  }
+
```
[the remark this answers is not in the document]
surbhigarg92: We dont support "GoogleSQL" dialect for now. So we need not implement this

handwritten/spanner-driver/src/lib/pg/types.ts
```diff
@@ -0,0 +1,808 @@
...
+    }
+    const parser = typeof formatOrFn === 'function' ? formatOrFn : fn!;
+    if (typeof parser !== 'function') {
+      throw new TypeError('Type parser must be a function');
+    }
+    this._parsers.set(numOid, parser);
+    // Invalidate synthesized array cache so any updated element parser is immediately picked up
+    this._arrayParserCache.clear();
+  }
+
```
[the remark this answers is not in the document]
surbhigarg92: We dont support "GoogleSQL" dialect for now. So we need not implement this

handwritten/spanner-driver/src/lib/pg/types.ts
```diff
@@ -0,0 +1,558 @@
...
+   * @param format - Formatting mode ('text' or 'binary'). Defaults to 'text'.
+   * @returns Parsing function converting wire string into JavaScript value.
+   */
+  public getTypeParser(oid: number | string, format = 'text'): TypeParser {
+    const numOid = typeof oid === 'string' ? Number(oid) : oid;
+    if (typeof numOid !== 'number' || Number.isNaN(numOid)) {
+      throw new TypeError(
+        `Invalid PostgreSQL OID: "${oid}". OID must be numeric.`,
+      );
+    }
```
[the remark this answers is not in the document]
surbhigarg92: Currently this driver does not support GoogleSQL dialect

handwritten/spanner-driver/src/lib/client.ts
```diff
@@ -111,27 +126,40 @@ export class Client extends EventEmitter {
...
-    }
     try {
       if (!this.dsn) {
         throw new Error(
           'Invalid Spanner connection configuration: project, instance, and database must be provided.',
         );
       }
-      // TODO(PR 4 - Native CGO Bridge): Instantiate native CGO Spanner connection handle via spannerlib-node
+      const pool = await Pool.create(this.dsn);
+      this.nativePool = pool;
```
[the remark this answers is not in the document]
surbhigarg92: The JS Pool / Client separation implements the standard `node-postgres` contract where each Client is an independent connection object. 

Will pick this up in separate PR if required

handwritten/spanner-driver/package.json
```diff
@@ -58,5 +59,8 @@
   },
   "engines": {
     "node": ">=18"
+  },
+  "optionalDependencies": {
+    "spannerlib-node": "file:../../../go-sql-spanner/spannerlib/wrappers/spannerlib-node/spannerlib-node-0.1.0.tgz"
   }
```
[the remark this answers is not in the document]
surbhigarg92: Will be fixed in next PR, once we release these packages

handwritten/spanner-driver/src/lib/client.ts
```diff
@@ -207,28 +247,126 @@ export class Client extends EventEmitter {
...
+                  const decoded = Codec.decodeRow<R>(
+                    rawRow,
+                    fields,
+                    parsers,
+                    query.rowMode,
+                  );
+                  currentRows.push(decoded);
+                  currentResult.rowCount = currentRows.length;
+                  query.emit('row', decoded, currentResult);
+                }
```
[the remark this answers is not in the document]
surbhigarg92: Standard `node-postgres` queries (client.query) buffer `result.rows` so that `await client.query()` resolves with the row array, while simultaneously emitting streaming 'row' events.

handwritten/spanner-driver/src/lib/codec.ts
```diff
@@ -0,0 +1,338 @@
...
+      };
+    }
+
+    if (val instanceof Date) {
+      // Serialized as ISO-8601 UTC string for Spanner TIMESTAMPTZ
+      return {
+        valueProto: {stringValue: val.toISOString()},
+        typeProto: {code: TypeCode.TIMESTAMP},
+      };
+    }
```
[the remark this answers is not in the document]
surbhigarg92: done

handwritten/spanner-driver/src/lib/pg/types.ts
```diff
@@ -0,0 +1,558 @@
...
+  }
+  if (typeof source === 'string') {
+    const trimmed = source.trim();
+    if (trimmed.startsWith('{') && trimmed.endsWith('}')) {
+      const inner = trimmed.slice(1, -1).trim();
+      if (!inner) return [];
+      return inner
+        .split(',')
+        .map(s => s.trim().replace(/^"(.*)"$/, '$1'))
+        .map(s => (s.toUpperCase() === 'NULL' ? null : elementParser(s)));
```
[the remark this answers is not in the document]
surbhigarg92: done

handwritten/spanner-driver/src/lib/client.ts
```diff
@@ -88,15 +100,18 @@ export class Client extends EventEmitter {
...
     if (this.isConnected) {
+      const err = enrichError(new Error('Client has already been connected'));
       if (callback) {
-        process.nextTick(() => callback(null));
+        process.nextTick(() => callback(err));
         return;
       }
-      return Promise.resolve();
+      return Promise.reject(err);
     }
```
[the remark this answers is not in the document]
surbhigarg92: done

handwritten/spanner-driver/src/lib/pg/types.ts
```diff
@@ -0,0 +1,456 @@
...
+        .split(/,(?=(?:(?:[^"]*"){2})*[^"]*$)/)
+        .map(s => s.trim().replace(/^"(.*)"$/, '$1'))
+        .map(s => (s.toUpperCase() === 'NULL' ? null : elementParser(s)));
+    }
+  }
+  return [];
+}
+
+// -----------------------------------------------------------------------------
+// Binary Format Parsers (matches pg-types/lib/binaryParsers.js)
```
olavloite: Technically, these will never be used in the Spanner driver. So if we keep them, that is only for code compatibility with node-pg. But I don't think any application code would ever use this, as these functions return the low-level PG wire-protocol binary representation of a value. That is not something I would expect an application to use. We can keep them here if that is better for compatibility, but we should preferably try to make sure that no-one tries to use them with the driver, as that would fail.

handwritten/spanner-driver/src/lib/pg/types.ts
```diff
@@ -0,0 +1,456 @@
...
+  [BuiltinOids.TIMESTAMPTZ]: parseBinaryTimestamp,
+  [BuiltinOids.INTERVAL]: parseBinaryText,
+  [BuiltinOids.NUMERIC]: parseBinaryText,
+  [BuiltinOids.UUID]: parseBinaryText,
+  [BuiltinOids.JSONB]: parseJson,
+};
+
+const defaultTypeParsers: Record<string, Record<number, TypeParser>> = {
+  text: textParsers,
+  binary: binaryParsers,
```
olavloite: I think that it would be better to let this just default to the textParsers. Anyone who tries to use this with this driver, will run into problems.

handwritten/spanner-driver/src/lib/codec.ts
```diff
@@ -0,0 +1,324 @@
...
+ */
+export class Codec {
+  /**
+   * Encodes a JavaScript value to Spanner Protobuf value and type format.
+   *
+   * @param val - JavaScript value.
+   * @returns EncodedParam containing `valueProto` and `typeProto`.
+   */
+  static encodeValue(val: unknown): EncodedParam {
+    if (val === null || val === undefined) {
```
olavloite, line 50: This should handle `.toPostgres()` first, as otherwise an array of `.toPostgres()` instances will be treated as an object:

```typescript
  // 1. Unwrap custom objects implementing .toPostgres() first (handles both top-level and nested array items)
  if (
    typeof val === 'object' &&
    val !== null &&
    typeof (val as {toPostgres?: unknown}).toPostgres === 'function'
  ) {
    return Codec.encodeValue(
      (val as {toPostgres: () => unknown}).toPostgres(),
    );
  }
```

Test case:

```typescript
it('should unwrap custom objects with .toPostgres() inside array parameters', () => {
  // Custom ORM / domain model wrappers
  const customId1 = { toPostgres: () => 101 };
  const customId2 = { toPostgres: () => 102 };
  // Pass array of custom objects to parameter $1
  const { fields } = Codec.encodeParams([[customId1, customId2]], 'pg');
  assert.deepStrictEqual(fields.p1, {
    listValue: {
      values: [
        { stringValue: '101' },
        { stringValue: '102' },
      ],
    },
  });
});
```

handwritten/spanner-driver/src/lib/client.ts
```diff
@@ -111,28 +125,50 @@ export class Client extends EventEmitter {
...
     try {
       if (!this.dsn) {
         throw new Error(
           'Invalid Spanner connection configuration: project, instance, and database must be provided.',
         );
       }
-      // TODO(PR 4 - Native CGO Bridge): Instantiate native CGO Spanner connection handle via spannerlib-node
-      if (!this.isEnded) {
+      const pool = await Pool.create(this.dsn);
+      this.nativePool = pool;
```
olavloite, line 156: We should not create a new `Pool` for each new `Client`. A `Pool` creates an actual connection to Spanner (e.g. a gRPC channel pool), and each pool can create many `Connection` instances. A `Connection` instance is lightweight in SpannerLib. `Pool` is not a lightweight object.

olavloite: Compare to the implementation of the .NET driver here: https://github.com/googleapis/dotnet-spanner-entity-framework/blob/e4dd7124a6b60c70f0f38db3f7507d2bb087d629/spanner-ado-net/spanner-ado-net/SpannerConnection.cs#L371

Each Pool is shared across every connection that uses the same connection string (DSN). So we should introduce something similar here, where clients that are created like this create or pick a pool based on the DSN.

olavloite: We can do this in a follow-up PR, but we should put a TODO here in that case.

handwritten/spanner-driver/src/lib/client.ts
olavloite, line 202: Can we split this function into smaller parts? It is getting very long and hard to read.

handwritten/spanner-driver/src/lib/client.ts
```diff
@@ -207,28 +255,122 @@ export class Client extends EventEmitter {
...
+                let listValue;
+                while ((listValue = await nativeRows.next()) !== null) {
+                  const rawRow = Codec.extractRawRow(listValue);
+                  const decoded = Codec.decodeRow<R>(
+                    rawRow,
+                    fields,
+                    parsers,
+                    query.rowMode,
+                  );
+                  currentRows.push(decoded);
```
olavloite: This will collect all rows in the query in memory. From what I understand, this is indeed also what node-pg does. But we should be clear (and document it accordingly) that this means that this function is not usable for large queries. And it means that we also need to implement/support `pg-cursor` and/or `pg-query-stream`.

handwritten/spanner-driver/src/lib/codec.ts
```diff
@@ -0,0 +1,324 @@
...
+    }
+    if (v.stringValue !== undefined && v.stringValue !== null) {
+      return v.stringValue;
+    }
+    if (v.numberValue !== undefined && v.numberValue !== null) {
+      return String(v.numberValue);
+    }
+    if (v.boolValue !== undefined && v.boolValue !== null) {
+      return v.boolValue ? 't' : 'f';
+    }
```
olavloite, line 305: This seems to 'emulate' the PostgreSQL wire-protocol. I don't think we need that (and it makes it less efficient). Now a boolean is converted from bool => string => bool. A number is converted from number => string => number.

handwritten/spanner-driver/src/lib/codec.ts
```diff
@@ -0,0 +1,324 @@
...
+    if (v.structValue) {
+      return JSON.stringify(v.structValue);
```
olavloite: This will return something like this:

```
{
  "fields": {
    "user_id": { "stringValue": "100" },
    "score": { "numberValue": 98.5 }
  }
}
```

Is that really what we would want here?

handwritten/spanner-driver/README.md
```diff
@@ -115,34 +121,121 @@ async function shutdown() {
...
 const client = new Client({
   project: 'my-gcp-project',
   instance: 'my-spanner-instance',
   database: 'my-spanner-database',
 });
 
-// Streaming row events
+await client.connect();
+
+// Stream rows as they arrive from Spanner gRPC stream
```
olavloite: This comment is a bit misleading. We are indeed streaming the rows as they come from the Spanner gRPC stream. But the driver is also collecting all of them in memory, so if the query returns a large number of rows, then you will get an OOM.

handwritten/spanner-driver/src/lib/pg/types.ts
```diff
@@ -0,0 +1,456 @@
...
+): unknown[] {
+  if (Array.isArray(source)) {
+    return source.map(item => {
+      if (item === null || item === undefined) return null;
+      return Array.isArray(item)
+        ? parsePgArray(item, elementParser)
+        : elementParser(item);
+    });
+  }
+  if (typeof source === 'string') {
```
olavloite: Spanner will never return this, so in that sense, we can remove everything from here in this function.

handwritten/spanner-driver/system-test/driver.ts
```diff
@@ -0,0 +1,940 @@
...
+          false,
+          '95000.00',
+          {genre: 'pop'},
+          ['pop', 'dance'],
+        ],
+      );
+
+      // Seed AllTypes table
+      await client.query('DELETE FROM AllTypes WHERE Id = 1');
+      await client.query(
```
olavloite, line 261: Can we also add rows here with:
1. NULL values (to verify that NULL parameter values work correctly)
2. Empty arrays (to verify that an empty array is correctly sent to Spanner)
3. Arrays with NULL elements in the array. So for example `[1, NULL, 2]`

olavloite: Can we also add a test for using `.toPostgres()`

handwritten/spanner-driver/src/lib/client.ts
surbhigarg92: Yes, I was planning to do this in a follow-up PR. Will add a todo

handwritten/spanner-driver/src/lib/pg/types.ts
surbhigarg92: Remove the binary parsers . Retained the same syntax for "setTypeParsers" as per `node-pg`, if someone tries to use format as `binary` , the driver will return error `Binary wire format is not supported in @google-cloud/spanner-driver because rows are transported via Spanner Protobuf.'`

handwritten/spanner-driver/src/lib/pg/types.ts
surbhigarg92: Remove the binary parsers . Retained the same syntax for "setTypeParsers" as per `node-pg`, if someone tries to use format as `binary` , the driver will return error `Binary wire format is not supported in @google-cloud/spanner-driver because rows are transported via Spanner Protobuf.'`

handwritten/spanner-driver/src/lib/codec.ts
surbhigarg92: We were already using "toPostgres" via `preparePgValue` method, but this was only called once during top level param not for nested objects. Fixed this

handwritten/spanner-driver/src/lib/client.ts
surbhigarg92: Added a TODO

handwritten/spanner-driver/src/lib/codec.ts
surbhigarg92: Removed this

handwritten/spanner-driver/src/lib/codec.ts
surbhigarg92: Fixed this. "struct" is not supported in PG anyway, but have added it for GoogleSQL dialect

handwritten/spanner-driver/README.md
surbhigarg92: Updated

handwritten/spanner-driver/src/lib/pg/types.ts
surbhigarg92: Done

handwritten/spanner-driver/system-test/driver.ts
surbhigarg92: Added and verified the test suite by running on local

handwritten/spanner-driver/src/lib/client.ts
surbhigarg92: Done

handwritten/spanner-driver/src/lib/client.ts
surbhigarg92: Updated the comment that rows will be buffered in memory. 

Need to think about how to implement support streaming, via `pg-cursor` and/or `pg-query-stream` or if there is a simpler way

handwritten/spanner-driver/src/lib/pg/types.ts
```diff
@@ -0,0 +1,370 @@
...
+      throw new Error(
+        'Binary wire format is not supported in @google-cloud/spanner-driver because rows are transported via Spanner Protobuf.',
+      );
+    }
+    const numOid = typeof oid === 'string' ? Number(oid) : oid;
+    if (typeof numOid !== 'number' || Number.isNaN(numOid)) {
+      throw new TypeError(
+        `Invalid PostgreSQL OID: "${oid}". OID must be numeric.`,
+      );
+    }
```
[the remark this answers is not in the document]
surbhigarg92: Currently we are only support PG dialect

handwritten/spanner-driver/src/lib/native.ts
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {Connection, Pool, Rows, SpannerLibError} from 'spannerlib-node';
+
+// 1. Target Production Platform & Architecture Package Loading:
+// In production, prebuilt native binaries will be distributed as optional dependencies:
+// import { createRequire } from 'module';
+// const require = createRequire(import.meta.url);
+// export const NativeBridge = require(`@google-cloud/spannerlib-node-${process.platform}-${process.arch}`);
```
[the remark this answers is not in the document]
surbhigarg92: Will be fixed in coming PR

handwritten/spanner-driver/package.json
```diff
@@ -58,5 +59,8 @@
   },
   "engines": {
     "node": ">=18"
+  },
+  "optionalDependencies": {
+    "spannerlib-node": "file:../../../go-sql-spanner/spannerlib/wrappers/spannerlib-node/spannerlib-node-0.1.0.tgz"
```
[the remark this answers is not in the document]
surbhigarg92: Will be fixed in coming PR

handwritten/spanner-driver/src/lib/codec.ts
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    if (typeof val === 'number') {
+      // Cloud Spanner represents INT64 as decimal strings in protobuf to avoid 64-bit IEEE 754 precision loss
+      if (Number.isInteger(val)) {
+        return {
+          valueProto: {stringValue: String(val)},
+          typeProto: {code: TypeCode.INT64},
+        };
+      }
```
olavloite, line 81: I worry about how this will work in real life. If an application tries to insert the value `1.0` into a FLOAT64 column, what will happen then? Do we have an end-to-end test for that?

handwritten/spanner-driver/src/lib/codec.ts
```diff
@@ -0,0 +1,338 @@
...
+      return {
+        valueProto: {numberValue: val},
+        typeProto: {code: TypeCode.FLOAT64},
+      };
```
olavloite, line 85: This is also risky. What if the application is trying to insert the value `1.2` into a `float32` column. Does Spanner accept this?

Also, we probably need to manually check if the value is `NaN`, `Inf` or `-Inf`. Those values should also be sent as strings to Spanner.

olavloite: Actually, my question about `float32` can probably be ignored. I see now that all data types are removed before the query parameters are sent to Spanner, so the `float64` vs `float32` problem should not be there.

olavloite (approved):

PR #9150, opened 2026-08-14
refactor: Improve readability of buildQueryStringComponents method by removing repeated code fragments

danieljbruce:
## Description

`request[key]` is used in many different places and it would be useful if it was referenced in one place to reduce clutter. To the reader it also eliminates the need to understand the role of the key variable except when used to build requestValue. It would also probably save time on lookup operations.

## Impact

Improve code performance and readability.

### DIFF
--- a/core/packages/gax/src/transcoding.ts
+++ b/core/packages/gax/src/transcoding.ts
@@ -124,23 +124,23 @@ export function buildQueryStringComponents(
 ): string[] {
   const resultList = [];
   for (const key in request) {
-    const value = request[key];
-    if (Array.isArray(value)) {
-      for (const val of value as JSONObject[]) {
+    const requestValue = request[key];
+    if (Array.isArray(requestValue)) {
+      for (const value of requestValue as JSONObject[]) {
         resultList.push(
           `${prefix}${encodeWithoutSlashes(key)}=${encodeWithoutSlashes(
-            val.toString(),
+            value.toString(),
           )}`,
         );
       }
-    } else if (typeof value === 'object' && value !== null) {
+    } else if (typeof requestValue === 'object' && requestValue !== null) {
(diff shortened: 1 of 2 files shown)


core/packages/gax/src/transcoding.ts
```diff
@@ -124,22 +124,23 @@ export function buildQueryStringComponents(
...
+    } else if (typeof requestValue === 'object' && requestValue !== null) {
       resultList.push(
-        ...buildQueryStringComponents(request[key] as JSONObject, `${key}.`),
+        ...buildQueryStringComponents(requestValue as JSONObject, `${key}.`),
       );
     } else {
       resultList.push(
         `${prefix}${encodeWithoutSlashes(key)}=${encodeWithoutSlashes(
-          request[key] === null ? 'null' : request[key]!.toString(),
+          requestValue === null || requestValue === undefined ? 'null' : requestValue.toString(),
```
westarle, line 143: should we add a test for this behavior change?

```typescript
it('should gracefully handle undefined property values without throwing', () => {
  const request = {
    definedField: 'value',
    undefinedField: undefined
  };

  // Prior to PR 9150, this threw: TypeError: Cannot read properties of undefined (reading 'toString')
  // With PR 9150, it gracefully serializes the undefined property to 'null'
  const result = buildQueryStringComponents(request as any);
  
  assert.deepStrictEqual(result, [
    'definedField=value', 
    'undefinedField=null'
  ]);
});
```

westarle (approved):
As a reviewer, I really appreciate these small PRs that just do one thing.

core/packages/gax/src/transcoding.ts
danieljbruce: done

PR #9152, opened 2026-08-14
fix(common): prevent path traversal and parameter injection in REST clients

danieljbruce:
Fixes path traversal and parameter injection vulnerabilities in Node.js REST clients by validating `.` and `..` path segments and percent-encoding URL-unsafe characters.

---
*PR created automatically by Jules for task [10569082496054868007](https://jules.google.com/task/10569082496054868007) started by @danieljbruce*


danieljbruce:
Since [the bug](https://b.corp.google.com/issues/529359609) affects only Gapic clients I think we can close this as out of scope, but it may come in handy later.

PR #9160, opened 2026-08-18
fix(teeny-request): destroy response stream if request stream is destroyed early

### LINKED ISSUE #8670: teeny-request throws when response handler destroys the request stream
### Library Name

`teeny-request` 10.1.3

### Try this API

Not applicable. This is a local Node.js stream lifecycle failure in `teeny-request`; the remote HTTP response succeeds.

### Minimal reproduction

https://gist.github.com/rockwotj/a629eea7b9ccbd7e354659b9f09fead3

Run:

```sh
npm install teeny-request@10.1.3 nock
node repro.js
```

On Node.js 24.16.0 this reports an unhandled rejection:

```text
Error [ERR_STREAM_UNABLE_TO_PIPE]: Cannot pipe to a closed or destroyed stream
```

### Steps and root cause

1. Start consuming the stream returned by `teenyRequest()`.
2. Destroy that stream from its `response` event handler. Retry wrappers do this when discarding a response before another attempt.
3. `teeny-request` then runs its later `response` listener and calls `pipeline(responseStream, requestStream)` with the destroyed destination.
4. Node rejects the pipeline setup with `ERR_STREAM_UNABLE_TO_PIPE`; the rejection is not handled, and the response body is not disposed.

The stream-mode implementation defers piping until both reading has begun and the fetch response exists, but it does not recheck whether the returned request stream was destroyed by an earlier `response` listener.

### Expected behavior

If the consumer destroys the returned stream before piping starts, `teeny-request` should skip pipeline construction and destroy the unused response body. No unhandled rejection should escape.

This is the same production stack previously reported in googleapis/teeny-request#351 and googleapis/teeny-request#352. Those reports were automatically closed for lacking a runnable reproduction.

westarle:
If the consumer destroys the returned stream before piping starts, teeny-request should skip pipeline construction and destroy the unused response body. This prevents unhandled rejections (e.g. ERR_STREAM_UNABLE_TO_PIPE) when trying to pipe to a closed/destroyed stream.

Fixes #8670

### DIFF
--- a/core/packages/teeny-request/src/index.ts
+++ b/core/packages/teeny-request/src/index.ts
@@ -278,10 +278,18 @@ function teenyRequest(
     let responseStream: any;
     requestStream.once('reading', () => {
       if (responseStream) {
-        pipeline(responseStream, requestStream, () => {});
+        if (!requestStream.destroyed) {
+          pipeline(responseStream, requestStream, () => {});
+        } else if (responseStream && typeof responseStream.destroy === 'function') {
+          responseStream.destroy();
+        }
       } else {
         requestStream.once('response', () => {
-          pipeline(responseStream, requestStream, () => {});
+          if (!requestStream.destroyed) {
+            pipeline(responseStream, requestStream, () => {});
+          } else if (responseStream && typeof responseStream.destroy === 'function') {
+            responseStream.destroy();
+          }
         });
       }
     });
--- a/core/packages/teeny-request/test/index.ts
+++ b/core/packages/teeny-request/test/index.ts
@@ -298,6 +298,25 @@ describe('teeny', () => {
     });
   });
 
+  it('should not pipe if request stream is destroyed early', done => {
+    const scope = mockJson();
+
+    const stream = teenyRequest({uri});
+    stream.on('error', done);
+
+    stream.once('response', response => {
+      response.body.once('close', () => {
+        scope.done();
+        done();
+      });
+
+      // Destroy the stream before piping can be fully set up
+      stream.destroy();
+    });
+
+    stream.resume();
+  });
+
   it('should expose TeenyStatistics instance', () => {
     assert.ok(teenyRequest.stats instanceof TeenyStatistics);
   });


dich-bandana:
We independently confirmed this race in a production Cloud Run service using `@google-cloud/storage@7.18.0` -> `teeny-request@9.0.0` on Node 24. The rejected promise reached the service's process-level `unhandledRejection` handler, which logged the failure and exited the instance:

```text
Error [ERR_STREAM_UNABLE_TO_PIPE]: Cannot pipe to a closed or destroyed stream
    at pipelineImpl (node:internal/streams/pipeline:264:15)
    at pipeline (node:internal/streams/pipeline:183:10)
    at PassThrough.<anonymous> (.../build/server.js:145022:37)
```

Our trace and validation:

- The bundled stack location maps directly to `teeny-request`'s deferred `pipeline(responseStream, requestStream)` call.
- A deterministic reproduction delays the HTTP response, destroys the returned request stream, and then releases the response.
- Using the unmodified `teeny-request@9.0.0`, Node 22.16.0 exits cleanly while Node 24.15.0 reproduces the exact `ERR_STREAM_UNABLE_TO_PIPE` uncaught/unhandled failure.
- Matching production errors predate the downstream application's recent stream changes, ruling those changes out as the introduction point.
- We prepared a narrow downstream patch with the same destroyed-stream guard and response-body cleanup: [workwisecare/expert-octo#6178](https://github.com/workwisecare/expert-octo/pull/6178).

We evaluated rolling the service back to Node 22, forcing `teeny-request@8.0.2`, and replacing the Storage client with custom authenticated HTTP calls. The first only masks the latent race, the second is outside Storage's supported dependency range and reverts intentional cleanup behavior, and the third is a much larger rewrite. This PR is the preferred upstream resolution; once it is released through `teeny-request@11` and consumed by `@google-cloud/storage@8`, we can upgrade and remove the downstream patch.

Thanks for picking this up. The implementation here matches the fix we independently validated, and the passing Node 22/24/26 regression coverage addresses the runtime boundary we observed.

dich-bandana (commented):
The implementation and regression test match the production race we reproduced and look good to me. `Fixes: 8670` is not linking the issue; change it to `Fixes #8670`. Is anything else keeping this in draft, or would help getting it ready for review be useful?

tgilino:
Another independent production confirmation, and some data that may help this along.

We encountered this on Node **24.15.0** via `@google-cloud/storage@7.22.0` → `teeny-request@9.0.0`. Background worker processes exit outright, because the throw is synchronous from inside the `once('response')` handler rather than a rejected promise, so nothing in the call chain catches it:

```
Error [ERR_STREAM_UNABLE_TO_PIPE]: Cannot pipe to a closed or destroyed stream
    at pipelineImpl (node:internal/streams/pipeline:264:15)
    at pipeline (node:internal/streams/pipeline:183:10)
    at PassThrough.<anonymous> (teeny-request/build/src/index.js:178)
    at Object.onceWrapper (node:events:631:26)
    at PassThrough.emit (node:events:521:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

This corroborates the Node-version detail in #8670: we see it on 24.15.0, and the same code path was quiet for us before.

One thing worth flagging for anyone triaging this — **there is currently no released version to upgrade to.** The unguarded call is byte-identical in `9.0.0`, `10.1.3` and `11.0.1` (latest), so every consumer of the storage client is exposed regardless of which transport version resolves.

A detail that made this hard to reproduce, in case it helps with a test: the `'reading'` event fires asynchronously after `resume()`, so the `once('response')` handler has to be armed *before* the destroy. A synchronous destroy-then-respond does not reproduce it. Only a non-first stream position throws — a destroyed source is fine, a destroyed destination is not.

We have applied the diff from this PR as a local package patch across both resolved versions. It resolves the crash, and destroying the orphaned response body matters as much as skipping the pipe — without it the guard trades a crash for a leaked socket. We also added a regression test that drives the real module against a server which responds late, and confirmed it fails without the patch.

Anything we can do to help move this along? Happy to supply more detail if it's useful.

PR #9166, opened 2026-08-18
fix: validate path parameters and prevent traversal/injection in REST transcoder - tests relocated

danieljbruce:
## Description

Reintroduces changes https://github.com/googleapis/google-cloud-node/pull/9151 only with the tests relocated not to conflict with generated changes.

## Impact

Resolves the exploit.


core/packages/gax/test/unit/transcoding_validation.ts
```diff
@@ -0,0 +1,132 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as assert from 'assert';
+import { describe, it } from 'mocha';
+const { v3 } = require('../../../../../../packages/google-cloud-dialogflow-cx');
```
westarle, line 17: This seems brittle to me; is there a place that has visibility to both packages?  If not, we might want to introduce one.  Can you file a ticket to track this (moving this to another directory where the dependencies can be a bit more natural -- maybe reuse your 'de-skip' issue)

westarle (approved):
If this is a clean rollforward of the previous PR with test change it LGTM.

core/packages/gax/test/unit/transcoding_validation.ts
danieljbruce: https://b.corp.google.com/issues/548652610

PR #9174, opened 2026-08-19
feat(generator): add internalTelemetryInfo configuration to nunjucks templates

shivanee-p:
Pass enableTelemetryTracing and internalTelemetryInfo into constructSettings() to pipe o11y information through to gax
- Updated constructSettings() across gax, grpc, and fallback to accept enableTelemetryTracing and internalTelemetryInfo
- Initializes CallSettings.enableTelemetryTracing.
- Updated templates to pass opts.enableTelemetryTracing and internalTelemetryInfo directly into this._gaxGrpc.constructSettings()
- Added unit tests in gax.ts and generator.ts

### DIFF
--- a/core/generator/gapic-generator-typescript/templates/cjs/typescript_gapic/src/$version/$service_client.ts.njk
+++ b/core/generator/gapic-generator-typescript/templates/cjs/typescript_gapic/src/$version/$service_client.ts.njk
@@ -398,6 +398,23 @@ export class {{ service.name }}Client {
        , 'x-goog-api-version': '{{ service.apiVersion }}'
 {%- endif -%}});
 
+    {%- if api.enableTelemetryTracing %}
+      if (opts.enableTelemetryTracing) {
+        // set the internal telemetry info using `otherArgs` attribute of `CallSettings`
+        const internalTelemetryInfo = {
+          gcpClientService: '{{ api.loggingName }}',
+          gcpVersion: '{{ api.naming.version }}',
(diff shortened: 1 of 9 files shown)


core/generator/gapic-generator-typescript/templates/cjs/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -398,6 +398,23 @@ export class {{ service.name }}Client {
        , 'x-goog-api-version': '{{ service.apiVersion }}'
 {%- endif -%}});
 
+    {%- if api.enableTelemetryTracing %}
```
westarle: can we add enableTelemetryTracing and internalTelemetryInfo to `constructSettings` so that it is applied once (vs. patching the object after it is constructed)?  This can help reduce the dependency of the generator on the settings object strcuture, and help other non-generator clients of CallSettings if any.

shivanee-p: This makes a lot of sense. I didn't consider the implications of this design of patching after construction. I've updated `constructSettings()` per your suggestion and now the template files will pass through the internal telemetry info when the expected flag is enabled.

shivanee-p:
Note: linter failures related to the generator and google-gax are addressed in [PR#9235](https://github.com/googleapis/google-cloud-node/pull/9235)

bshaffer (approved):

PR #9176, opened 2026-08-19
feat(gax): accept clientOptions in createApiCall for telemetry tracing

shivanee-p:
Adds clientOptions support and telemetry gating logic to createApiCall in google-gax to support OpenTelemetry tracing.

- Defined StaticTraceContext and added enableTelemetryTracing?: boolean and internalTelemetryInfo?: StaticTraceContext to ClientOptions.
- Accepted clientOptions in createApiCall and added conditional gating checking enableTelemetryTracing, internalTelemetryInfo, and GOOGLE_SDK_NODE_EXPERIMENTAL_O11Y_ENABLED.
- Updated test helpers to forward clientOptions and added unit tests for tracing enablement.

### DIFF
--- a/core/packages/gax/src/clientInterface.ts
+++ b/core/packages/gax/src/clientInterface.ts
@@ -28,10 +28,19 @@ import {
 import * as longrunning from './longRunningCalls/longrunning';
 import * as operationProtos from '../protos/operations';
 
+/**
+ * The static trace context is information about the Google Cloud client library that is
+ * used to generate telemetry tracing information.
+ */
+export interface StaticTraceContext {
+  gcpClientService?: string;
+  gcpVersion?: string;
+  gcpRepo?: string;
+  gcpArtifact?: string;
+}
+
 export interface ClientOptions
-  extends GrpcClientOptions,
-    GoogleAuthOptions,
-    ClientStubOptions {
+  extends GrpcClientOptions, GoogleAuthOptions, ClientStubOptions {
   libName?: string;
   libVersion?: string;
   clientConfig?: gax.ClientConfig;
@@ -42,6 +51,15 @@ export interface ClientOptions
   // No preference; exception will be thrown if both are set to different values.
   universeDomain?: string;
   universe_domain?: string;
+  /**
+   * Whether to enable telemetry tracing for the client.
+   */
+  enableTelemetryTracing?: boolean;
+  /**
+   * The static trace context is information about the Google Cloud client library that is
+   * used to generate telemetry tracing information.
+   */
+  internalTelemetryInfo?: StaticTraceContext;
 }
 
 export interface Descriptors {
--- a/core/packages/gax/src/clientInterface.ts
+++ b/core/packages/gax/src/clientInterface.ts
@@ -28,17 +28,6 @@ import {
 import * as longrunning from './longRunningCalls/longrunning';
 import * as operationProtos from '../protos/operations';
 
-/**
- * The static trace context is information about the Google Cloud client library that is
- * used to generate telemetry tracing information.
- */
-export interface StaticTraceContext {
-  gcpClientService?: string;
-  gcpVersion?: string;
-  gcpRepo?: string;
-  gcpArtifact?: string;
-}
-
 export interface ClientOptions
   extends GrpcClientOptions, GoogleAuthOptions, ClientStubOptions {
   libName?: string;
--- a/core/packages/gax/src/clientInterface.ts
+++ b/core/packages/gax/src/clientInterface.ts
@@ -55,11 +55,6 @@ export interface ClientOptions
    * Whether to enable telemetry tracing for the client.
    */
   enableTelemetryTracing?: boolean;
-  /**
-   * The static trace context is information about the Google Cloud client library that is
-   * used to generate telemetry tracing information.
-   */
-  internalTelemetryInfo?: StaticTraceContext;
 }
 
 export interface Descriptors {
--- a/core/packages/gax/src/createApiCall.ts
+++ b/core/packages/gax/src/createApiCall.ts
@@ -35,19 +35,6 @@ import {StreamingApiCaller} from './streamingCalls/streamingApiCaller';
 import {warn} from './warnings';
 
 /**
- * Checks if telemetry tracing is enabled
- * @param settings
- * @returns true if telemetry tracing is enabled, false otherwise
- */
-export function checkTelemetryEnabled(settings?: CallSettings): boolean {
-  const tracingEnabled =
-    Boolean(settings?.enableTelemetryTracing) &&
-    process.env.GOOGLE_SDK_NODE_EXPERIMENTAL_O11Y_ENABLED === 'true' &&
-    settings?.internalTelemetryInfo !== undefined;
-  return Boolean(tracingEnabled);
-}
-
-/**
  * Converts an rpc call into an API call governed by the settings.
  *
  * In typical usage, `func` will be a promise to a callable used to make an rpc
--- a/core/packages/gax/src/createApiCall.ts
+++ b/core/packages/gax/src/createApiCall.ts
@@ -35,6 +35,19 @@ import {StreamingApiCaller} from './streamingCalls/streamingApiCaller';
 import {warn} from './warnings';
 
 /**
+ * Checks if telemetry tracing is enabled
+ * @param settings
+ * @returns true if telemetry tracing is enabled, false otherwise
+ */
+export function checkTelemetryEnabled(settings?: CallSettings): boolean {
+  const tracingEnabled =
+    Boolean(settings?.enableTelemetryTracing) &&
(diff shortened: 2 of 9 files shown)


core/packages/gax/src/clientInterface.ts
```diff
@@ -28,6 +28,13 @@ import {
 import * as longrunning from './longRunningCalls/longrunning';
 import * as operationProtos from '../protos/operations';
 
+export interface StaticTraceContext {
+  gcpClientService?: string;
+  gcpVersion?: string;
+  gcpRepo?: string;
+  gcpArtifact?: string;
+}
```
westarle: please add docs to exported types.

core/packages/gax/src/createApiCall.ts
```diff
@@ -59,6 +60,7 @@ export function createApiCall(
   descriptor?: Descriptor,
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   _fallback?: boolean | 'proto' | 'rest', // unused here, used in fallback.ts implementation
+  clientOptions?: ClientOptions
```
westarle: add param docs please.

core/packages/gax/src/clientInterface.ts
```diff
@@ -42,6 +49,8 @@ export interface ClientOptions
   // No preference; exception will be thrown if both are set to different values.
   universeDomain?: string;
   universe_domain?: string;
+  enableTelemetryTracing?: boolean;
+  internalTelemetryInfo?: StaticTraceContext;
```
westarle: If ClientOptions is exposed to our customers, can we avoid exposing internalTelemetryTracing?  See below about use of CallSettings.

core/packages/gax/src/createApiCall.ts
```diff
@@ -164,4 +171,10 @@ export function createApiCall(
     // or to cancel the ongoing call.
     return currentApiCaller.result(ongoingCall);
   };
+
+  if (tracingEnabled) {
+    return invokeCall;
+  } else {
+    return invokeCall;
+  }
```
westarle: I don't think `invokeCall` or this branching is necessary?

core/packages/gax/src/createApiCall.ts
```diff
@@ -67,7 +69,12 @@ export function createApiCall(
   // the following apiCaller will be used for all calls of this function...
   const apiCaller = createAPICaller(settings, descriptor);
 
-  return (
+  // Check if telemetry tracing is enabled and also check if internal telemetry information has been passed through the templates
+  const tracingEnabled =
+    clientOptions?.enableTelemetryTracing &&
+    process.env.GOOGLE_SDK_NODE_EXPERIMENTAL_O11Y_ENABLED === 'true' &&
+    clientOptions?.internalTelemetryInfo !== undefined;
```
westarle: this is complicated enough to be its own function, with a test.

The signal for me is that it needs a comment to describe what it does -- easier to make that the name of a function.

core/packages/gax/src/createApiCall.ts
westarle: Did we consider extracting what we need from ClientOptions into CallSettings before calling this function?  That might prevent the requirement to change all our generated clients to add this param.  It also makes it less ambiguous for future clients which value to use where data is available in both ClientOptions and CallSettings (e.g. endpoint)

core/packages/gax/src/clientInterface.ts
```diff
@@ -42,6 +49,8 @@ export interface ClientOptions
   // No preference; exception will be thrown if both are set to different values.
   universeDomain?: string;
   universe_domain?: string;
+  enableTelemetryTracing?: boolean;
```
westarle, line 46: docs

westarle (dismissed):
I think we should consider CallSettings carefully.  I would take the change to

core/packages/gax/src/createApiCall.ts
shivanee-p: It's not in this context. It will be later on when I call the TracerHelper. I removed it for now.

core/packages/gax/src/clientInterface.ts
shivanee-p: Ack - this makes much more sense. I've updated the design accordingly.

core/packages/gax/src/createApiCall.ts
shivanee-p: Ack.

core/packages/gax/src/clientInterface.ts
```diff
@@ -42,6 +51,15 @@ export interface ClientOptions
...
   universeDomain?: string;
   universe_domain?: string;
+  /**
+   * Whether to enable telemetry tracing for the client.
+   */
+  enableTelemetryTracing?: boolean;
+  /**
+   * The static trace context is information about the Google Cloud client library that is
+   * used to generate telemetry tracing information.
+   */
```
danieljbruce: Thanks for adding these comments. They are very helpful.

core/packages/gax/src/createApiCall.ts
```diff
@@ -34,6 +34,19 @@ import {addTimeoutArg} from './normalCalls/timeout';
 import {StreamingApiCaller} from './streamingCalls/streamingApiCaller';
 import {warn} from './warnings';
 
+/**
+ * Checks if telemetry tracing is enabled
```
danieljbruce: If the intent is to export this so that it is available to users then maybe a comment would be good about how to do that. Otherwise if it is used for the tests then maybe just moving it there.

core/packages/gax/src/clientInterface.ts
```diff
@@ -42,6 +51,15 @@ export interface ClientOptions
   // No preference; exception will be thrown if both are set to different values.
   universeDomain?: string;
   universe_domain?: string;
+  /**
+   * Whether to enable telemetry tracing for the client.
+   */
+  enableTelemetryTracing?: boolean;
```
danieljbruce, line 46: These two things were the agreed upon change to the API surface? I'm not too familiar with the project so just making sure.

danieljbruce (commented):
Do you have any more information about the CallSettings comment from Wes? Seems like it is okay to expose the options here, I just don't have enough context.

core/packages/gax/src/clientInterface.ts
```diff
@@ -42,6 +51,15 @@ export interface ClientOptions
...
+  /**
+   * The static trace context is information about the Google Cloud client library that is
+   * used to generate telemetry tracing information.
+   */
+  internalTelemetryInfo?: StaticTraceContext;
```
westarle: I don't think we should add this to our public API if we can avoid it (e.g. by stuffing it directly into callsettings in the generated code)

westarle (commented):
everything else lgtm if you can get a review from the team.

core/packages/gax/src/clientInterface.ts
shivanee-p: Thanks for catching this. We only want to expose `enableTelemetryTracing` to the API surface.

core/packages/gax/src/createApiCall.ts
shivanee-p: Gotcha. I think it makes sense to move this function out of this file. I moved it to util.ts and I've updated the corresponding until tests. This function will eventually be called in `createApiCall()`, not just in testing.

core/packages/gax/src/gax.ts
```diff
@@ -170,6 +171,8 @@ export interface CallOptions {
   longrunning?: BackoffSettings;
   apiName?: string;
   retryRequestOptions?: RetryRequestOptions;
+  enableTelemetryTracing?: boolean;
+  internalTelemetryInfo?: StaticTraceContext;
```
danieljbruce: I think it is worth noting here that this interface is exposed to the user. See https://github.com/googleapis/google-cloud-node/blob/main/handwritten/bigtable/system-test/read-rows.ts#L127-L135 for example where users pass in CallOptions to specify how the client behaves.

If the intent is to allow users to set internalTelemetryInfo then this LGTM, but just thought I would highlight this detail. It looks like merge needs this option here anyway.

danieljbruce (approved):
I don't know enough to know why merge and the gax constructor will need these new options yet, but supporting the new options only in CallSettings would be better if it is possible and if the user does not need them. Maybe we want to expose these options to the user though so I'll add a LGTM, but let's make sure there is no other choice.

core/packages/gax/src/gax.ts
shivanee-p: The intent is not to allow users to set internalTelemetryInfo. Thank you for mentioning that. I've removed that field from the CallSettings and CallOptions, and I've updated the design at go/client-libraries:nodejs-o11y-t4-t3-gapic to reflect this change. There's an `otherArgs` attribute that I can utilize to pass this information from the template files. Thank you!

PR #9179, opened 2026-08-20
chore(spanner-driver): Use spanner-api for protos instead of spanner

surbhigarg92:


handwritten/spanner-driver/src/lib/native.ts
olavloite, line 28: Should we use `createRequire(import.meta.url)` instead? That should make this independent of the current working directory.

handwritten/spanner-driver/package.json
```diff
@@ -45,9 +45,10 @@
     "test:pg-suite": "node test/pg-suite/run_pg_suite.cjs"
   },
   "dependencies": {
-    "@google-cloud/spanner": "^8.7.1"
+    "@google-cloud/spanner-api": "^0.3.0"
   },
   "devDependencies": {
+    "@google-cloud/spanner": "^8.7.1",
```
olavloite, line 51: Do we really need this here?

olavloite (approved):

handwritten/spanner-driver/package.json
olavloite, line 43: Should we also add this here:

```
"pretest": "npm run compile"
```

handwritten/spanner-driver/package.json
surbhigarg92: Yes, I am using it in system tests for admin calls to create database for tests

handwritten/spanner-driver/src/lib/native.ts
surbhigarg92: `createRequire(import.meta.url)` cannot be used directly in dual ESM/CJS packages. 

CJS will throw error `SyntaxError: Cannot use 'import.meta' outside a module`

handwritten/spanner-driver/package.json
surbhigarg92: Will add it in next PR

PR #9180, opened 2026-08-20
chore(spanner-driver): Add pg test suite

surbhigarg92:


handwritten/spanner-driver/src/lib/client.ts
```diff
@@ -91,7 +130,7 @@ export class Client extends EventEmitter {
     this.config =
       typeof config === 'string' ? {connectionString: config} : config || {};
     this.dsn = this.config.connectionString || resolveDsn(this.config);
-    this.types = this.config.types;
+    this.types = this.config.types || new TypeOverrides(defaultTypes);
```
olavloite: nit:

```suggestion
    this.types = this.config.types || new TypeOverrides(getDefaultTypeOverrides(this.dialect));
```

handwritten/spanner-driver/src/lib/client.ts
```diff
@@ -69,14 +70,52 @@ export class Client extends EventEmitter {
...
+
   /** Internal task queue managing sequential query execution. */
   private queryQueue: QueryTask<unknown>[] = [];
 
   /** Boolean flag tracking active query execution state. */
   private isExecuting = false;
 
   /** Boolean flag tracking whether client has been explicitly closed via end(). */
-  private isEnded = false;
+  public isEnded = false;
```
olavloite: Can we expose this through a getter? Or in some other way make sure that an outside user cannot update this manually?

handwritten/spanner-driver/src/lib/config.ts
```diff
@@ -90,9 +90,13 @@ export function resolveDsn(config?: string | ClientConfig): string {
...
-  const project = cfg?.project || process.env.GOOGLE_CLOUD_PROJECT;
-  const instance = cfg?.instance;
-  const database = cfg?.database;
+  const project =
+    cfg?.project ||
+    process.env.GCLOUD_PROJECT ||
+    process.env.GOOGLE_CLOUD_PROJECT ||
+    process.env.SPANNER_PROJECT;
+  const instance = cfg?.instance || process.env.SPANNER_INSTANCE;
+  const database = cfg?.database || process.env.SPANNER_DATABASE;
```
olavloite, line 99: Can we make sure that all of this is covered in our documentation (the README)?

handwritten/spanner-driver/test/pg-suite/run_pg_suite.cjs
```diff
@@ -0,0 +1,229 @@
...
+
+    // Summary
+    console.log('\n================================================================');
+    console.log('                   PG-SUITE EXECUTION SUMMARY                    ');
+    console.log('================================================================');
+    console.log(`Total Test Files Executed: ${testFiles.length}`);
+    console.log(`Passed Files:              ${totalPassedFiles}`);
+    console.log(`Failed Files:              ${totalFailedFiles}`);
+    console.log('================================================================');
+
```
olavloite, line 225: Should we not add something like this here:

```javascript
if (totalFailedFiles > 0) {
  process.exit(1);
}
```

(Or throw an error so the main method will exit with a non-zero code?)

handwritten/spanner-driver/test/pg-suite/run_pg_suite.cjs
```diff
@@ -0,0 +1,229 @@
...
+    testFiles = testFiles.filter(f => f.name.includes(targetArg));
+  }
+  if (testFiles.length === 0) {
+    console.error(`[pg-suite] No test files matching "${targetArg}" found in ${integrationDir}`);
+    process.exit(1);
+  }
+}
+
+async function main() {
+  const projectId = process.env.GCLOUD_PROJECT;
```
olavloite: Is this intentionally only reading GCLOUD_PROJECT? Or should we support the other environment variables that we also support in the driver itself (e.g. GOOGLE_CLOUD_PROJECT)?

handwritten/spanner-driver/cloudbuild.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - |
+    npm install
+  dir: 'handwritten/spanner-driver'
+  id: 'install-dependencies'
+
+# 2. Configure environment variables for the tests and run system tests
+#    - GOOGLE_APPLICATION_CREDENTIALS: GCB steps run as a service account
+#      that is typically granted permissions directly. Explicitly setting
+#      GOOGLE_APPLICATION_CREDENTIALS might not be needed if the GCB service
+#      account has the right roles (e.g., Bigtable Admin, Bigtable User).
```
olavloite: nit: remove copy-pasted BigTable references

handwritten/spanner-driver/cloudbuild.yaml
```diff
@@ -0,0 +1,51 @@
...
+# If you use Secret Manager for credentials, uncomment and configure:
+# availableSecrets:
+#   secretManager:
+#   - versionName: projects/${PROJECT_ID}/secrets/YOUR_SERVICE_ACCOUNT_KEY_SECRET_NAME/versions/latest
+#     env: 'SA_KEY' # This env var will hold the secret value. Use it as GOOGLE_APPLICATION_CREDENTIALS in step 3 if needed.
+
+# Define a substitution variable for your project ID
+# Replace 'long-door-651' with the actual GCP Project ID your system tests should run against.
+substitutions:
+  _GCP_PROJECT_ID: 'long-door-651'
```
olavloite, line 49: Is it really intentional to hard-code a project ID here?

handwritten/spanner-driver/test/pg-suite/integration/client/sasl-scram-tests.cjs
```diff
@@ -0,0 +1,203 @@
...
+
+'use strict'
+const helper = require('../../test-helper.cjs')
+const pg = helper.pg
+const suite = new helper.Suite()
+const { native } = helper.args
+const assert = require('assert')
+
+/**
+ * This test only executes if the env variables SCRAM_TEST_PGUSER and
```
olavloite: Can we remove this file, as it is for tests that we will never use for this driver. It is part of the pg wire-protocol startup sequence.

handwritten/spanner-driver/test/pg-suite/integration/connection-pool/idle-timeout-tests.cjs
```diff
@@ -0,0 +1,29 @@
...
+const helper = require('../../test-helper.cjs')
+const assert = require('assert')
+
+new helper.Suite().test('idle timeout', function () {
+  const config = Object.assign({}, helper.config, { idleTimeoutMillis: 50 })
+  const pool = new helper.pg.Pool(config)
+  pool.connect(
+    assert.calls(function (err, client, done) {
+      assert(!err)
+      client.query('SELECT NOW()')
```
olavloite: Should we add `await` here? (Or is this a carry-over bug from PG?). Also, is the pool actually closed anywhere in this test? Should we do that?

handwritten/spanner-driver/test/pg-suite/SPANNER_PG_SUITE_ADAPTATIONS.md
```diff
@@ -0,0 +1,112 @@
...
+| **`ssl-tests.cjs`** | Tested raw PostgreSQL `SSLRequest` packet negotiation (`80877103`) over TCP sockets. Cloud Spanner uses Google Cloud IAM / gRPC TLS authentication. |
+| **`network-partition-tests.cjs`** | Simulated TCP socket drops via local Node `net.createServer()`. Spanner uses HTTP/2 gRPC channel pools. |
+| **`connection-pool/tls-tests.cjs`** | Tested PostgreSQL SSL/TLS client certificate files (`PGSSLROOTCERT`, `PGSSLCERT`, `PGSSLKEY`) over raw TCP sockets. Cloud Spanner uses Google Cloud IAM / gRPC TLS authentication. |
+| **`connection-pool/native-instance-tests.cjs`** | Tested `libpq` C-binding toggles (`pg.native`). `@google-cloud/spanner-driver` uses the native `spannerlib-node` Go CGO bridge. |
+
+*Note: Accompanying TCP packet mocking helpers (`test-buffers.cjs` and `buffer-list.cjs`) were also removed in favor of a single unified `test/pg-suite/test-helper.cjs`.*
+
+---
+
+## 4. Skipped Test Files & Future Roadmap
```
olavloite, line 64: I think that there are a few more tests that are skipped. Can we do a full check for that? The things that cam up for me were:

- appname-tests.cjs (skipped: application_name GUC session variables)
- api-tests.cjs (skipped: raw TCP socket query_timeout tests)
- error-handling-tests.cjs (skipped: pg_sleep() socket drop tests)

handwritten/spanner-driver/src/lib/pool.ts
olavloite, line 428: This method should probably have a 'double-entry' protection:

```javascript
  if (!this.allClients.has(client)) {
    return;
  }
```

The current implementation causes this method to be called multiple times for the same client, which again can have an impact on the order in which events are handled.


This test case shows the problem:

```javascript
it('should maintain strict FIFO acquisition order and prevent waitQueue displacement on client error', async () => {
  const pool = new Pool({
    project: 'test-project',
    instance: 'test-instance',
    database: 'test-database',
    max: 1,
  });
  // 1. Check out the only available connection slot
  const initialClient = await pool.connect();
  assert.strictEqual(pool.totalCount, 1);
  // 2. Queue Request A (index 0) and Request B (index 1) in FIFO order
  const resolutionOrder: string[] = [];
  const requestA = pool.connect().then(client => {
    resolutionOrder.push('A');
    return client;
  });
  const requestB = pool.connect().then(client => {
    resolutionOrder.push('B');
    return client;
  });
  assert.strictEqual(pool.waitingCount, 2, 'Both requests should be queued');
  // 3. initialClient encounters a fatal error and is destroyed.
  // Exactly ONE client slot is freed, so only Request A should be dequeued to get the replacement connection.
  const releasePromise = initialClient.release(new Error('connection reset'));
  // 4. Concurrently queue Request C while release is in flight
  const requestC = pool.connect().then(client => {
    resolutionOrder.push('C');
    return client;
  });
  await releasePromise;
  // Wait a short tick for asynchronous handlers to settle
  await new Promise(resolve => setTimeout(resolve, 50));
  // 5. Request A completes its work and returns the client to the pool
  const clientA = await requestA;
  await clientA.release();
  await new Promise(resolve => setTimeout(resolve, 50));
  // Request B was queued before Request C, so Request B MUST be resolved before Request C.
  // Without the fix, re-entrant removeClient() prematurely pops Request B from waitQueue,
  // causing Request C to jump ahead of Request B in the queue (resulting in ['A', 'C']).
  assert.deepStrictEqual(
    resolutionOrder,
    ['A', 'B'],
    `FIFO queue order was violated. Expected Request B before Request C, but got: [${resolutionOrder.join(', ')}]`,
  );
  // 6. Request B completes and frees the connection for Request C
  const clientB = await requestB;
  await clientB.release();
  await new Promise(resolve => setTimeout(resolve, 50));
  assert.deepStrictEqual(
    resolutionOrder,
    ['A', 'B', 'C'],
    'All requests should eventually resolve in strict FIFO order',
  );
  const clientC = await requestC;
  await clientC.release();
  await pool.end();
});
```

handwritten/spanner-driver/src/lib/client.ts
```diff
@@ -54,7 +55,7 @@ export class Client extends EventEmitter {
   readonly dialect: Dialect = DEFAULT_DIALECT;
 
   /** Type parser override registry configured on this client. */
-  readonly types?: ITypeOverrides;
+  public types: ITypeOverrides;
```
olavloite: Can we make this `public readonly`?

handwritten/spanner-driver/src/lib/client.ts
```diff
@@ -526,5 +576,7 @@ export class Client extends EventEmitter {
         task.cancel(closeError);
       }
     }
+
+    this.emit('end');
```
olavloite, line 599: Can we add a small unit test for this? I know that it is covered by the PG integration tests, but having a quick unit test will make sure that any regression is quickly caught in the future.

handwritten/spanner-driver/src/lib/pool.ts
```diff
@@ -146,7 +146,7 @@ export class Pool extends EventEmitter {
             void client.release(releaseErr);
           }),
         )
-        .catch(err => callback(err));
+        .catch(err => callback(err, undefined, () => {}));
```
olavloite, line 149: Here also: Can we add a quick unit test to prevent regressions in the future?

handwritten/spanner-driver/src/lib/pool.ts
```diff
@@ -197,6 +197,11 @@ export class Pool extends EventEmitter {
         }
       });
 
+      // Remove client if it is explicitly ended
+      client.on('end', () => {
+        this.removeClient(client);
+      });
+
```
olavloite, line 204: here also: can we add a unit test?

handwritten/spanner-driver/cloudbuild.yaml
surbhigarg92: I have taken this whole yaml file from "spanner" client library, there as well we have this hardcoded project, infact all GCB files have this hardcoded project

handwritten/spanner-driver/src/lib/client.ts
surbhigarg92: Done

handwritten/spanner-driver/src/lib/client.ts
surbhigarg92: Done

handwritten/spanner-driver/src/lib/config.ts
surbhigarg92: Done

handwritten/spanner-driver/test/pg-suite/run_pg_suite.cjs
surbhigarg92: Not intentional, was using only one env for testing so missed the other one

handwritten/spanner-driver/test/pg-suite/run_pg_suite.cjs
surbhigarg92: Done

handwritten/spanner-driver/cloudbuild.yaml
surbhigarg92: Done

handwritten/spanner-driver/test/pg-suite/integration/client/sasl-scram-tests.cjs
surbhigarg92: Done

handwritten/spanner-driver/test/pg-suite/integration/connection-pool/idle-timeout-tests.cjs
surbhigarg92: Done. Also did a scan for other tests, found 3 more tests with the same issue.

handwritten/spanner-driver/test/pg-suite/SPANNER_PG_SUITE_ADAPTATIONS.md
surbhigarg92: So this section only captured the files where all the tests were skipped, not the files where some tests are skipped
Updated this to capture all the skipped test case, also asked agent to do a scan and updated the below section for modified files as well

handwritten/spanner-driver/src/lib/pool.ts
surbhigarg92: Done

handwritten/spanner-driver/src/lib/client.ts
surbhigarg92: Done

handwritten/spanner-driver/src/lib/client.ts
surbhigarg92: Done

handwritten/spanner-driver/src/lib/pool.ts
surbhigarg92: Done. `should invoke connect callback with error and dummy done function when connection acquisition fails`

handwritten/spanner-driver/src/lib/pool.ts
surbhigarg92: Done. `should remove client from pool when client.end() is called directly on checked-out client`

handwritten/spanner-driver/src/lib/client.ts
surbhigarg92: `should emit end event when client.end() is called`

handwritten/spanner-driver/test/test_results.json
```diff
@@ -0,0 +1,40 @@
+{
```
olavloite: nit: remove this file

olavloite (approved):

PR #9194, opened 2026-08-23
fix(storage): reject uploadFileInChunks when the automatic abort succeeds

### LINKED ISSUE #9193: TransferManager uploadFileInChunks does not error when aborted

### Library Name

@google-cloud/storage

N/A

https://gist.github.com/tcrognon/3102b05ce3e0aa902e22decb5be0a788

1. Replace SOME_BUCKET with your GCP test bucket.
2. Replace SOME_FILE_PATH with a file to upload to SOME_BUCKET.
3. Run the code and somehow cause one of the PUT requests to fail, but don't just disable internet because the DELETE request (to abort the XML multipart upload) needs to succeed.

`uploadFileInChunks` performs an XML multipart upload. This is:
1) A POST to create the multipart upload.
2) A series of PUTs to upload the parts.
3) A POST to complete the multipart upload.
4) A DELETE to abort the multipart upload if 2) or 3) fails.

If 4) happens and is successful, `uploadFileInChunks` returns `undefined`. I expected it would throw. All other error throw.

Here is the offending `return;`: https://github.com/googleapis/google-cloud-node/blob/b5b5afb9e053744ff7c3f6334aa2a60f86466c84/handwritten/storage/src/transfer-manager.ts#L955

Returning `undefined` in this single case is very unexpected and the documentation heavily implies that if `uploadFileInChunks` doesn't throw, then the file was written successfully:
- https://docs.cloud.google.com/nodejs/docs/reference/storage/latest/storage/transfermanager#_google_cloud_storage_TransferManager_uploadFileInChunks_member_1_
- https://docs.cloud.google.com/storage/docs/multipart-uploads#storage-upload-object-chunks-nodejs

If auth problem, throws. If invalid bucket or file, throws. If initial POST fails, throws. If subsequent PUTs/POST fails and the abort fails, throws. But if the abort succeeds, returns undefined.

I experienced silent data loss due to this. Code continued after aborted uploads thinking they were successfully written.

Om-singhaI:
# fix(storage): reject uploadFileInChunks when the automatic abort succeeds

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-cloud-node/issues) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes #9193 🦕

## What was wrong

`TransferManager.uploadFileInChunks` drives an XML multipart upload: initiate, upload the parts, complete. When any of those steps fails after an `uploadId` exists, the method aborts the upload session by default (`autoAbortFailure` unset or true). The catch block then did this:

```ts
try {
  await mpuHelper.abortUpload();
  return;
} catch (e) {
  throw new MultiPartUploadError(...);
}
throw new MultiPartUploadError(...);
```

So the only outcome that did not throw was "the upload failed and the abort succeeded". The promise resolved with `undefined`, and callers that treat a resolved promise as a written object carried on as if the file existed. The reporter hit exactly this and lost data silently. The method's doc comment promises the opposite ("otherwise a error containing the message, uploadId, and parts map"), and every other failure path (initiate fails, abort fails, `autoAbortFailure: false`) already throws a `MultiPartUploadError`.

## The fix

Remove the early `return` so a successful abort falls through to the existing `throw new MultiPartUploadError(message, uploadId, partsMap)`. The error now carries the original failure message plus the `uploadId` and the parts map, which is the documented contract and matches what the abort failure path and the `autoAbortFailure: false` path already do. No other logic changes.

Note on behaviour: a case that previously resolved now rejects. That is the point of the fix and it is what the documentation and the sample on cloud.google.com already describe, but I am calling it out explicitly so nobody is surprised. I framed this as a patch level fix rather than a breaking change because the resolved `undefined` was never documented and could not be distinguished from success; if the maintainers would rather treat it as breaking, the commit message can be adjusted.

## Tests

`test/transfer-manager.ts`, `describe('uploadFileInChunks')`:

- `should call abortUpload when a failure occurs after an uploadID is established` previously used `assert.doesNotThrow` on a promise returning arrow function without awaiting anything, so it passed regardless of whether the promise resolved or rejected and effectively pinned the bug. It now `await`s `assert.rejects`, checks that the rejection is a `MultiPartUploadError` whose message, `uploadId` and parts map match what the fake helper reported, and checks that `abortUpload` was called once. The fake now establishes the `uploadId` in `initiateUpload`, records a part in the parts map in `uploadPart`, and fails in `completeUpload`, so the parts map assertion is meaningful (a non empty map is propagated).
- New: `should reject with the abort error when abortUpload also fails` covers the inner catch, which had no unit test before: the abort rejects, and the method rejects with a `MultiPartUploadError` carrying the abort error's message, the `uploadId` and the parts map.
- `should reject with an error with empty uploadId and partsMap` (the `autoAbortFailure: false` test) called `assert.rejects` without `await`, so its assertion could never fail the test. It now awaits it. No other change to that test. I am happy to drop this one word change if you would rather keep the PR to the single path from the issue.

Why the fakes fail in `completeUpload` rather than `uploadPart`: `completeUpload` is awaited directly by the method. A fake whose `uploadPart` rejects synchronously rejects while the read loop is still waiting on the next chunk from disk, and the wrapper promise from `p-limit` sits unhandled for one event loop turn until `Promise.all` attaches a handler. With the tests now awaiting the result that shows up as two `PromiseRejectionHandledWarning` lines in the mocha output. Failing in `completeUpload` keeps the tests deterministic and the output clean. That short unhandled window in the source is pre existing and separate from this issue; I mention it under "Out of scope" below.

## Verification

All runs were on `handwritten/storage` with `npm run compile:cjs -- --sourceMap` followed by `mocha build/cjs/test/transfer-manager.js` (the same file the package's `npm test` runs under c8, limited to the transfer manager suite; the unit tests use the sinon `FakeXMLHelper` harness and need no GCP credentials).

- Pristine `main` (48e0941): 44 passing. A standalone script that calls `uploadFileInChunks` with a helper whose `uploadPart` fails and whose `abortUpload` succeeds printed `abortUpload called and succeeded` followed by the promise resolving to `undefined`.
- Updated tests with the source change reverted to `main` (test file only changed): 44 passing, 1 failing. The failing test is `should call abortUpload when a failure occurs after an uploadID is established` with `AssertionError [ERR_ASSERTION]: Missing expected rejection.` The new abort failure test passes on `main` as well, as expected, because that path already threw.
- Updated tests with the fix: 45 passing, 0 failing, no `PromiseRejectionHandledWarning` output. The same standalone script now prints `REJECTED: MultiPartUploadError part upload failed uploadId= upload-123 partsMap= Map(1) { 1 => 'etag-1' }`.

Coverage: the only removed source line is the early `return`; the inner `catch` (abort failure) gains its first unit test, so line coverage of `src/transfer-manager.ts` goes up, not down.

## Lint

I ran the monorepo linter the way `presubmit.yaml` does (`node ./bin/linter.mjs`, which diffs `main...HEAD`, runs ESLint with the root config, and type checks the package with `tsc --noEmit`). Type checking passes. Two pre existing findings are unrelated to this change and are identical on `main`:

- `src/transfer-manager.ts` reports 32 `prettier/prettier` "Delete `,`" errors both before and after this change (only the line numbers shift by one). The file on `main` is formatted with `trailingComma: "all"` (the root gts 6 config) while the package's own `.prettierrc.cjs` resolves the package local gts 5 config with `trailingComma: "es5"`. My one line deletion neither adds nor removes any of these.
- `test/transfer-manager.ts` cannot be parsed by that linter at all because the package `tsconfig.json` does not include `test/`. Linting it with `tsconfig.cjs.json` as the project instead gives 0 prettier findings and three pre existing rule findings on lines this PR does not touch (duplicate `fs` imports at lines 41 and 42, and an explicit `any` at line 639). Before this PR that run also flagged `@typescript-eslint/no-floating-promises` at the un awaited `assert.rejects` in the `autoAbortFailure: false` test; that is the one word `await` included here.

`npx prettier --check test/transfer-manager.ts` with the package config passes.

## Out of scope, noted for maintainers

- While writing the standalone reproduction I noticed that a part upload that rejects before the read loop reaches `await Promise.all(promises)` (that is, while the loop is still waiting on the next chunk from disk) produces an unhandled rejection on the `p-limit` wrapper promise for one event loop turn. Under Node's default `--unhandled-rejections=throw` that terminates the process instead of reaching the catch block at all. Real part uploads take far longer than a disk read, so this is rare in practice, but it is a separate latent issue and I did not touch it here.
- Open PR #9158 and draft PR #8827 both edit `src/transfer-manager.ts` and `test/transfer-manager.ts`. Neither touches the catch block changed here. #9158 does reformat a large part of the test file, so whichever lands second will need a rebase of the test hunks; the one line source change is unaffected.


Om-singhaI:
Main moved under this one. #9212 landed Aug 27 and already added the `await` to `assert.rejects` in the `autoAbortFailure: false` test, so that hunk of mine was redundant. I've merged main and it's gone. What's left is the one source line in `transfer-manager.ts` and the two tests on the abort path. The `assert.doesNotThrow` in `should call abortUpload when a failure occurs after an uploadID is established` is still on main, and that's the assertion that was hiding this.

@Dhriti07 you merged #9198 and #9212 through #9214, so I'd rather ask you than leave it on the team queue. There's no reviewer on it at all right now and `multi-approvers` reads 0 of 2.

One thing worth settling first: a call that used to resolve now rejects. I filed it as a fix rather than breaking, since the resolved `undefined` wasn't documented and you couldn't tell it from success. Say if you'd rather it ship as breaking and I'll redo the commit message.

Issue #9186, opened 2026-08-21
Passing DocumentReference to Firestore.pipeline().documents() throws  "Error: INTERNAL ERROR: Client is not yet ready to issue requests."


dvoloshyn:
### Please make sure you have searched for information in the following guides.

- [x] Search the issues already opened: https://github.com/GoogleCloudPlatform/google-cloud-node/issues
- [x] Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js
- [x] Check our Troubleshooting guide: https://github.com/googleapis/google-cloud-node/blob/main/docs/troubleshooting.md
- [x] Check our FAQ: https://github.com/googleapis/google-cloud-node/blob/main/docs/faq.md
- [x] Check our libraries HOW-TO: https://github.com/googleapis/gax-nodejs/blob/main/client-libraries.md
- [x] Check out our authentication guide: https://github.com/googleapis/google-auth-library-nodejs
- [x] Check out handwritten samples for many of our APIs: https://github.com/GoogleCloudPlatform/nodejs-docs-samples
- [x] Check the API's issue tracker: https://cloud.google.com/support/docs/issue-trackers

### Library Name

@google-cloud/firestore 9.0.0

### A screenshot that you have tested with "Try this API".

Issue is with the library, not with the API

### Link to the code that reproduces this issue. A link to a **public** Github Repository or gist with a minimal reproduction.

https://gist.github.com/dvoloshyn/9a2923e7e5e43182731ba5c33e6027a0

### A step-by-step description of how to reproduce the issue, based on the linked reproduction.

See the linked gist, it is obvious

### A clear and concise description of what the bug is, and what you expected to happen.

When validating DocumentReference, the library accesses projectId without triggering lazy initialization which results in exception below. Passing string value works fine. I expect it to work for references as well
```
"Error: INTERNAL ERROR: Client is not yet ready to issue requests.
    at mT.get projectId (/node_modules/.pnpm/@google-cloud+firestore@9.0.0/node_modules/@google-cloud/firestore/build/src/index.js:686:19)
    at mT.get formattedName (/node_modules/.pnpm/@google-cloud+firestore@9.0.0/node_modules/@google-cloud/firestore/build/src/index.js:711:33)
    at W$._validateReference (/node_modules/.pnpm/@google-cloud+firestore@9.0.0/node_modules/@google-cloud/firestore/build/src/pipelines/pipelines.js:91:45)
[...]

PR #9197, opened 2026-08-24
fix(firestore): do not throw when wrapping an error with a read only stack

### LINKED ISSUE #9154: firestore: wrapError() throws TypeError when the wrapped error has a non-writable stack, surfacing as an uncaughtException
### Environment

- `@google-cloud/firestore`: **9.0.0** (also reproduced on 7.11.6)
- `google-auth-library`: 11.0.2 (also reproduced on 10.6.2)
- Node.js: 20.19.5
- OS: macOS (not platform specific)

### Summary

`wrapError()` — [`handwritten/firestore/dev/src/util.ts:236`](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/firestore/dev/src/util.ts#L236), shipped as `build/src/util.js` — decorates an error by appending to its `stack`:

```js
function wrapError(err, stack) {
    err.stack += '\nCaused by: ' + stack;
    return err;
}
```

This assumes `err.stack` is writable. That is not guaranteed. When the wrapped error has a **non-writable** `stack`, the compound assignment throws a `TypeError` and the original error is destroyed — replaced by an unrelated one that points at `util.js`.

Because `util.js` is emitted with `"use strict"`, the write throws rather than silently no-op'ing (in sloppy mode the same assignment is silently discarded).

This is reachable in practice, not just theoretically: `google-auth-library` produces exactly such errors. `getErrorFromOAuthErrorResponse()` copies properties onto a new `Error` with `writable: false`, and explicitly includes `stack` in that copy ([`oauth2common.ts`, the `keys.push('stack')` / `Object.defineProperty(..., { writable: false })` block](https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/oauth2common.ts)). So **every STS/OAuth error response** surfaced through an `external_account` (Workload Identity Federation) credential carries a non-writable `stack`.

I've filed the counterpart issue on that side too: googleapis/google-auth-library-nodejs — see cross-reference below. Fixing either end resolves the crash, but the guard here seems worth having regardless, since `wrapError` cannot know what kind of error it is handed.

### Why this is worse than a confusing message

Firestore calls `wrapError` from stream `'error'` handlers. The `TypeError` is therefore thrown inside an `EventEmitter` emit, so it escapes the surrounding promise chain and every enclosing `try`/`catch`, and lands as an **`uncaughtException`**. In a serverless runtime that takes down the invocation instead of producing a clean rejected promise.

Observed trace (paths genericised), repeated once per failing request:

```
TypeError: Cannot assign to read only property 'stack' of object 'Error: Error code invalid_grant: ID Token issued at <ts> is stale to sign-in.'
    at wrapError (/app/node_modules/@google-cloud/firestore/build/src/util.js:213:15)
    at Transform.<anonymous> (/app/node_modules/@google-cloud/firestore/build/src/index.js:...)
    at Transform.emit (node:events:524:28)
    at emitErrorNT (node:internal/streams/destroy:169:8)
    at emitErrorCloseNT (node:internal/streams/destroy:128:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```

The real failure was an expired federated credential. What the process reported instead was a `TypeError` about a read-only property, which points at this library rather than at the credential — the actual cause is only visible once you inspect the message string.

### Reproduction

```js
// npm i google-auth-library
const { getErrorFromOAuthErrorResponse } = require('google-auth-library/build/src/auth/oauth2common.js');

// Exactly what the STS/external_account path produces on an error response:
const authError = getErrorFromOAuthErrorResponse(
  { error: 'invalid_grant', error_description: 'ID Token is stale to sign-in.' },
  new Error('underlying transport failure'),
);

console.log(Object.getOwnPropertyDescriptor(authError, 'stack').writable);
// => false

// This is the assignment wrapError() performs, in a strict-mode module:
(function () { 'use strict'; authError.stack += '\nCaused by: <caller stack>'; })();
// => TypeError: Cannot assign to read only property 'stack' of object
//    'Error: Error code invalid_grant: ID Token is stale to sign-in.'
```

Calling `wrapError` directly requires reaching past the package `exports` map, so the snippet above models the one line it executes. Passing that same `authError` into the real `wrapError` throws identically at `util.js:216` on 9.0.0 (`util.js:213` on 7.11.6), and routing it through a stream `'error'` handler that calls `wrapError` reproduces the `uncaughtException` end to end.

### Suggested fix

Make the decoration non-fatal, so a hostile `stack` degrades the message rather than replacing the error:

```js
function wrapError(err, stack) {
    try {
        err.stack += '\nCaused by: ' + stack;
    } catch {
        // `stack` may be non-writable (e.g. errors from google-auth-library's
        // OAuth error path). Preserve the original error rather than throwing.
    }
    return err;
}
```

If keeping the appended context matters in that case, `Object.defineProperty(err, 'stack', { value: <combined>, writable: true, configurable: true })` works when the property is `configurable` — it is in the `google-auth-library` case (`configurable: true, writable: false`).

Om-singhaI:
# fix(firestore): do not throw when wrapping an error with a read only stack

Fixes #9154

## What breaks

`wrapError()` in `handwritten/firestore/dev/src/util.ts` appends the captured
callsite stack to the error it is handed:

```ts
export function wrapError(err: Error, stack: string): Error {
  err.stack += '\nCaused by: ' + stack;
  return err;
}
```

The compiled `build/src/util.js` begins with `"use strict"`, so this compound
assignment is a strict mode write. When the incoming error carries a non
writable own `stack` property the write does not silently do nothing. It throws
a `TypeError`, and that `TypeError` takes the place of the real error the caller
was supposed to receive. The user loses both the failure and any hint of where
it came from.

## Why it takes the process down

Where `wrapError()` is called decides how bad this is.

At `dev/src/index.ts:1384` the call sits inside a promise `.catch()`, so a throw
there only turns into a rejected promise. Unpleasant, but catchable.

At `dev/src/reference/query-util.ts:87` it sits inside a stream `'error'`
listener:

```ts
.on('error', err => {
  reject(wrapError(err, callsiteError.stack!));
})
```

That listener runs long after the `Promise` executor returned, so a throw inside
it is not funnelled into `reject`. It propagates out of `EventEmitter.emit`,
through `emitErrorNT`, and becomes an `uncaughtException`. Two further call
sites have the identical shape and the identical exposure:
`dev/src/reference/aggregate-query.ts:151` and
`dev/src/pipelines/pipeline-util.ts:106`.

Running a query whose backend stream fails with such an error, against `main`
before this change, produces:

```
  1) query interface
       handles stream exception whose stack is not writable:
     Uncaught TypeError: Cannot assign to read only property 'stack' of object 'Error: Expected error'
      at wrapError (dev/src/util.ts:237:12)
      at Transform.<anonymous> (dev/src/reference/query-util.ts:87:27)
      at Transform.emit (node:events:508:20)
      at emitErrorNT (node:internal/streams/destroy:170:8)
      at emitErrorCloseNT (node:internal/streams/destroy:129:3)
      at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
```

## The fix

```ts
export function wrapError(err: Error, stack: string): Error {
  const wrappedStack = err.stack + '\nCaused by: ' + stack;
  try {
    if (Object.getOwnPropertyDescriptor(err, 'stack')?.writable === false) {
      // Some libraries define `stack` as a read only property, which makes a
      // plain assignment throw. Redefining the property still preserves the
      // callsite as long as the property stayed configurable.
      Object.defineProperty(err, 'stack', {value: wrappedStack});
    } else {
      err.stack = wrappedStack;
    }
  } catch {
    // The stack cannot be modified at all. Recording the callsite is best
    // effort, so return the error we were asked to wrap rather than let the
    // resulting TypeError take its place.
  }
  return err;
}
```

Two things are going on, and they are separate on purpose.

**Redefining rather than assigning keeps the diagnostic.** The errors that
actually trigger this in the wild are still configurable, so the callsite can be
recorded. Simply swallowing the `TypeError` would make `wrapError()` a no op for
exactly the failures where an async stack trace is most useful.

**The `try` is the safety net, not the mechanism.** A `stack` that is both non
writable and non configurable cannot be redefined either, and neither can one on
a frozen error. In those cases the original error is returned untouched.
Recording the callsite is a diagnostic nicety; delivering the actual error is
not optional.

Behaviour for ordinary errors is unchanged: on a normal `Error`, `stack` is an
accessor property, `descriptor.writable` is `undefined`, and the plain
assignment path runs exactly as before.

## Relationship to the auth library

One producer of such errors is
`core/packages/google-auth-library-nodejs/src/auth/oauth2common.ts`, where
`getErrorFromOAuthErrorResponse()` copies `stack` onto a new error:

```ts
Object.defineProperty(newError, key, {
  value: (err as {} as {[index: string]: string})[key],
  writable: false,
  enumerable: true,
});
```

Worth noting, because it decides which fix is correct here: `configurable` is
not specified, and `stack` already exists on `newError`, so `defineProperty`
leaves the existing attribute alone. The resulting descriptor is
`writable: false, configurable: true`. Verified against the
`google-auth-library` that Firestore actually resolves:

```
descriptor: {"writable":false,"configurable":true}
wrapError returned: the ORIGINAL error
message: Error code invalid_grant: Bad Request
stack ends with Caused by: true
```

That producer is a separate defect in a separate package with a separate owner
(`.github/CODEOWNERS` assigns `/handwritten/firestore` to
`@googleapis/firestore-team` and `/core/packages/google-auth-library-nodejs` to
`@googleapis/aion-team`), and it is addressed on its own. Fixing the producer
does not fix this one. `wrapError()` still converts *any* error with a non
writable stack, from any source, into a `TypeError` or an `uncaughtException`,
and Firestore resolves the auth library transitively through a published
`google-gax` (`@google-cloud/firestore-api` to `google-gax` to
`google-auth-library`), so released Firestore keeps crashing until every layer
upgrades. This change makes the Firestore side robust on its own.

## Testing

Three regression tests, all written before the fix and confirmed failing against
unmodified source.

`dev/test/util.ts`, unit coverage of the root cause:

* `appends the callsite stack to the error stack` guards the existing behaviour.
* `appends the callsite stack when the error stack is not writable` uses the
  descriptor shape that `google-auth-library` produces, and asserts the callsite
  is still recorded.
* `returns the original error when its stack cannot be modified` uses a non
  configurable stack and asserts the original error comes back untouched.

`dev/test/query.ts`, coverage of the reported crash path:

* `handles stream exception whose stack is not writable` drives a real query
  through `query-util.ts` with a mock backend stream that fails with an error
  whose `stack` was defined the way `google-auth-library` defines it, and
  asserts the returned promise rejects with the original error.

### All three fail without the source change

With `dev/src/util.ts` restored to `origin/main` and the tests left in place:

```
  wrapError()
    ✔ appends the callsite stack to the error stack
    1) appends the callsite stack when the error stack is not writable
    2) returns the original error when its stack cannot be modified

  1 passing (6ms)
  2 failing

  1) wrapError()
       appends the callsite stack when the error stack is not writable:
     TypeError: Cannot assign to read only property 'stack' of object 'Error: Expected error'
      at wrapError (dev/src/util.ts:237:12)
      at Context.<anonymous> (dev/test/util.ts:129:30)
      at process.processImmediate (node:internal/timers:504:21)

  2) wrapError()
       returns the original error when its stack cannot be modified:
     TypeError: Cannot assign to read only property 'stack' of object 'Error: Expected error'
      at wrapError (dev/src/util.ts:237:12)
      at Context.<anonymous> (dev/test/util.ts:146:30)
      at process.processImmediate (node:internal/timers:504:21)
```

```
  query interface
    1) handles stream exception whose stack is not writable

  0 passing (56ms)
  1 failing

  1) query interface
       handles stream exception whose stack is not writable:
     Uncaught TypeError: Cannot assign to read only property 'stack' of object 'Error: Expected error'
      at wrapError (dev/src/util.ts:237:12)
      at Transform.<anonymous> (dev/src/reference/query-util.ts:87:27)
      at Transform.emit (node:events:508:20)
      at emitErrorNT (node:internal/streams/destroy:170:8)
      at emitErrorCloseNT (node:internal/streams/destroy:129:3)
      at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
```

### Suites with the change applied

Run from `handwritten/firestore` on Node v25.6.1.

```
npx mocha build/test
  696 passing (9s)
  47 pending
```

```
npx mocha build/conformance
  228 passing (245ms)
```

No failures. Changed files pass `prettier` and the `gts` base ESLint config.
Note that `npm run lint` at the package level currently cannot run from a
package only install, because the repository root `.eslintrc.json` extends
`./node_modules/gts` relative to the repository root; that is unrelated to this
change.


Om-singhaI:
Good catch, taken.

I had only checked `writable === false`, which misses a getter with no setter: `writable` is `undefined` there, so it falls through to the plain assignment and still throws. Switched to the try then redefine shape you suggested, which also drops the descriptor lookup off the happy path.

Added a test for that case. One thing worth knowing if anyone writes a similar test later: the `set: undefined` in it is load bearing. V8 installs `stack` as an accessor with both a getter and a setter, and a partial descriptor only overrides the attributes it names, so defining just `get` leaves the original setter in place and the assignment quietly succeeds. My first attempt at the test passed against the unfixed code for exactly that reason.

Reverting only `util.ts` makes the new test fail with the callsite missing, so it is covering the right thing. 673 passing in the firestore unit suite either way.

Om-singhaI:
@wu-hui you picked up #9154 and #9186 last Wednesday. Both already have a PR sitting on them: this one for #9154, and #9196 for #9186.

This one's small. `wrapError` assigns to `err.stack`, and `google-auth-library` builds errors where that property is read only, so the assignment throws a TypeError inside the stream error handler and surfaces as an uncaughtException. The fix tries the assignment, falls back to redefining, and gives up quietly if neither works.

If you're already partway into either, say so and I'll close mine. Otherwise could you take a look?

Om-singhaI:
Closing, since #9325 landed the same `wrapError` change and closed #9154.

PR #9221, opened 2026-08-28
perf: optimize stream pipeline by eliminating events-intercept

alkatrivedi:
Optimizes performance by removing the events-intercept library and its associated CPU penalty caused by monkey-patching EventEmitter.emit.

**Changes:**

- Replaced merge-stream with standard Node.js PassThrough streams.
- Replaced error interception hacks with direct error listeners (.on('error')) on the request stream for retry handling.
- Removed events-intercept and merge-stream dependencies from package.json.

**Impact:** Eliminates artificial CPU overhead on every stream event and simplifies pipeline architecture without breaking retry functionality.


handwritten/spanner/src/partial-result-stream.ts
```diff
@@ -16,8 +16,6 @@
 
 import {GrpcService} from './common-grpc/service';
 import * as checkpointStream from 'checkpoint-stream';
-import * as eventsIntercept from 'events-intercept';
-import mergeStream = require('merge-stream');
 import {common as p} from 'protobufjs';
 import {Readable, Transform} from 'stream';
```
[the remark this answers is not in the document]
alkatrivedi: done

handwritten/spanner/src/partial-result-stream.ts
```diff
@@ -578,17 +576,18 @@ export function partialResultStream(
...
+  // requestsStream allows multiple streams to be connected into one. This is good;
   // if we need to retry a request and pipe more data to the user's stream.
   // We also add an additional stream that can be used to flush any remaining
   // items in the checkpoint stream that have been received, and that did not
   // contain a resume token.
-  const requestsStream = mergeStream();
+  const requestsStream = new stream.PassThrough({objectMode: true});
   const flushStream = new stream.PassThrough({objectMode: true});
-  requestsStream.add(flushStream);
+  flushStream.pipe(requestsStream, {end: false});
```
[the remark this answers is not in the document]
alkatrivedi: done

handwritten/spanner/src/partial-result-stream.ts
```diff
@@ -626,7 +623,11 @@ export function partialResultStream(
     }
     lastRequestStream = requestFn(lastResumeToken);
     lastRequestStream.on('end', endListener);
-    requestsStream.add(lastRequestStream);
+    errorListener = (err: grpc.ServiceError) => {
+      setImmediate(() => retry(err));
```
olavloite, line 709: Deferring the retry here with `setImmediate(..)` means that other events could be handled before the retry. If for example `lastRequestStream` emits 'end' before the retry, the retry will fail. It could also cause multiple errors to trigger multiple retries.

Verification test cases:

```typescript
it('should successfully retry when the failed stream emits an error followed by end', done => {
  const fakeCheckpointStream = through.obj();
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  (fakeCheckpointStream as any).reset = () => {};
  sandbox.stub(checkpointStream, 'obj').returns(fakeCheckpointStream);

  const firstStream = through.obj();
  const secondStream = through.obj();

  const requestFnStub = sandbox.stub();

  // First request fails with UNAVAILABLE and immediately ends
  requestFnStub.onCall(0).callsFake(() => {
    setImmediate(() => {
      firstStream.emit('error', {
        code: grpc.status.UNAVAILABLE,
        message: 'Unavailable',
      } as grpc.ServiceError);
      firstStream.end();
    });
    return firstStream;
  });

  // Retried request succeeds and delivers data
  requestFnStub.onCall(1).callsFake(() => {
    setImmediate(() => {
      secondStream.push(RESULT_WITH_TOKEN);
      fakeCheckpointStream.emit('checkpoint', RESULT_WITH_TOKEN);
      secondStream.end();
    });
    return secondStream;
  });

  const receivedRows: Row[] = [];
  partialResultStream(requestFnStub)
    .on('data', row => receivedRows.push(row))
    .on('error', done)
    .on('end', () => {
      try {
        assert.strictEqual(requestFnStub.callCount, 2, 'Should have retried once');
        assert.strictEqual(receivedRows.length, 1, 'Should receive data from retried stream');
        done();
      } catch (e) {
        done(e);
      }
    });
});

it('should only spawn a single retry when multiple errors are emitted in rapid succession', done => {
  const fakeCheckpointStream = through.obj();
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  (fakeCheckpointStream as any).reset = () => {};
  sandbox.stub(checkpointStream, 'obj').returns(fakeCheckpointStream);

  const firstStream = through.obj();
  const secondStream = through.obj();

  const requestFnStub = sandbox.stub();

  // First request emits two error events synchronously
  requestFnStub.onCall(0).callsFake(() => {
    setImmediate(() => {
      const err = {
        code: grpc.status.UNAVAILABLE,
        message: 'Unavailable',
      } as grpc.ServiceError;
      firstStream.emit('error', err);
      firstStream.emit('error', err);
    });
    return firstStream;
  });

  // Second request succeeds
  requestFnStub.onCall(1).callsFake(() => {
    setImmediate(() => {
      secondStream.push(RESULT_WITH_TOKEN);
      fakeCheckpointStream.emit('checkpoint', RESULT_WITH_TOKEN);
      secondStream.end();
    });
    return secondStream;
  });

  partialResultStream(requestFnStub)
    .on('error', done)
    .pipe(
      concat(rows => {
        try {
          // Exactly 1 initial request + 1 retry request = 2 calls total
          assert.strictEqual(requestFnStub.callCount, 2, 'Should only trigger one retry request');
          assert.strictEqual(rows.length, 1);
          done();
        } catch (e) {
          done(e);
        }
      }),
    );
});
```

Suggested fix:

```
--- a/handwritten/spanner/src/partial-result-stream.ts
+++ b/handwritten/spanner/src/partial-result-stream.ts
@@ -618,17 +618,27 @@ export function partialResultStream(
       flushStream.push(null);
     });
   };
+
+  const destroyRequestStream = (): void => {
+    if (lastRequestStream) {
+      lastRequestStream.removeListener('end', endListener);
+      lastRequestStream.removeAllListeners('error');
+      lastRequestStream.on('error', () => {});
+      lastRequestStream.unpipe(requestsStream);
+      lastRequestStream.destroy();
+    }
+  };
+
   const makeRequest = (): void => {
     if (isDefined(lastResumeToken) && lastResumeToken.length > 0) {
       partialRSStream._resetPendingValues();
     }
     lastRequestStream = requestFn(lastResumeToken);
     lastRequestStream.on('end', endListener);
     errorListener = (err: grpc.ServiceError) => {
+      destroyRequestStream();
       setImmediate(() => retry(err));
     };
     lastRequestStream.on('error', errorListener);
     lastRequestStream.pipe(requestsStream, {end: false});
   };
```

handwritten/spanner/src/partial-result-stream.ts
```diff
@@ -659,6 +660,8 @@ export function partialResultStream(
 
     if (lastRequestStream) {
       lastRequestStream.removeListener('end', endListener);
+      lastRequestStream.removeAllListeners('error');
+      lastRequestStream.on('error', () => {}); // Prevent unhandled exception crash
       lastRequestStream.destroy();
     }
```
olavloite: This cleanup code is skipped if the early return above for non-retriable errors is used (the one on line 657). That causes the `lastRequestStream` open and with all event listeners attached.

Verification test:

```typescript
it('should destroy the request stream and detach listeners on non-retryable errors', done => {
  const fakeCheckpointStream = through.obj();
  sandbox.stub(checkpointStream, 'obj').returns(fakeCheckpointStream);

  const fakeStream = through.obj();
  const destroySpy = sandbox.spy(fakeStream, 'destroy');

  const requestFnStub = sandbox.stub().callsFake(() => {
    setImmediate(() => {
      fakeStream.emit('error', {
        code: grpc.status.INVALID_ARGUMENT,
        message: 'Invalid query argument.',
      } as grpc.ServiceError);
    });
    return fakeStream;
  });

  partialResultStream(requestFnStub)
    .on('data', () => {})
    .on('error', err => {
      try {
        assert.strictEqual(err.code, grpc.status.INVALID_ARGUMENT);
        assert.strictEqual(destroySpy.called, true, 'Request stream should be destroyed on non-retryable error');
        assert.strictEqual(fakeStream.listenerCount('end'), 0, 'endListener should be removed');
        assert.strictEqual(fakeStream.listenerCount('error'), 0, 'errorListener should be removed');
        done();
      } catch (e) {
        done(e);
      }
    });
});
```

Suggested fix:

```
--- a/handwritten/spanner/src/partial-result-stream.ts
+++ b/handwritten/spanner/src/partial-result-stream.ts
@@ -633,6 +633,7 @@ export function partialResultStream(
   };
 
   const retry = (err: grpc.ServiceError): void => {
+    destroyRequestStream();
     const elapsed = Date.now() - startTime;
     if (elapsed >= timeout) {
       // The timeout has reached so this will flush any rows the
@@ -659,13 +660,6 @@ export function partialResultStream(
       return;
     }
 
-    if (lastRequestStream) {
-      lastRequestStream.removeListener('end', endListener);
-      lastRequestStream.removeAllListeners('error');
-      lastRequestStream.on('error', () => {}); // Prevent unhandled exception crash
-      lastRequestStream.destroy();
-    }
     // Delay the retry until all the values that are already in the stream
     // pipeline have been handled. This ensures that the checkpoint stream is
```

handwritten/spanner/src/partial-result-stream.ts
```diff
@@ -674,15 +677,6 @@ export function partialResultStream(
   };
 
   userStream.once('reading', makeRequest);
-  eventsIntercept.patch(requestsStream);
```
olavloite, line 677: If the application calls `stream.destroy()`, then that is not propagated into 'our' stream. We should add a listener on the `userStream` for `close` and make sure that we clean up then the user stream is closed.

Verification test case:

```typescript
it('should destroy the underlying request stream when the user destroys the returned stream', done => {
  const fakeStream = through.obj();
  const destroySpy = sandbox.spy(fakeStream, 'destroy');

  const requestFnStub = sandbox.stub().returns(fakeStream);

  const stream = partialResultStream(requestFnStub);

  // Read first row and immediately destroy stream
  stream.on('data', () => {
    stream.destroy();
  });

  stream.on('close', () => {
    setImmediate(() => {
      try {
        assert.strictEqual(
          destroySpy.called,
          true,
          'Underlying request stream must be destroyed when user cancels the stream',
        );
        done();
      } catch (e) {
        done(e);
      }
    });
  });

  fakeStream.push(RESULT_WITH_TOKEN);
});
```

Suggested fix:

```
--- a/handwritten/spanner/src/partial-result-stream.ts
+++ b/handwritten/spanner/src/partial-result-stream.ts
@@ -677,6 +677,12 @@ export function partialResultStream(
   };
 
   userStream.once('reading', makeRequest);
+  userStream.once('close', () => {
+    destroyRequestStream();
+    requestsStream.destroy();
+    flushStream.destroy();
+    batchAndSplitOnTokenStream.destroy();
+  });
 
   return (
     requestsStream
```

handwritten/spanner/src/partial-result-stream.ts
```diff
@@ -16,18 +16,15 @@
 
 import {GrpcService} from './common-grpc/service';
 import * as checkpointStream from 'checkpoint-stream';
```
olavloite: This actually also imports `events-intercept` as a transitive dependency. So the removal of the import below does not actually remove it entirely from this file. And `checkpointStream` also monkey-patches the `emit` method. So while this PR gets rid of some of the monkey-patching of the query stream, it does not get rid of all of it.

Test:

```typescript
it('should not have events-intercept monkey-patching the stream pipeline', () => {
  const stream = checkpointStream.obj();
  // @ts-ignore
  const hasIntercept = typeof stream.intercept === 'function';
  assert.strictEqual(
    hasIntercept,
    false,
    'events-intercept is still monkey-patching stream pipeline via checkpoint-stream',
  );
});
```

We could instead implement our own specific 'checkpointStream' without monkey-patching and remove the entire dependency on `checkpointStream` here:

```typescript
class CheckpointStream extends Transform {
  private queue: google.spanner.v1.PartialResultSet[] = [];
  private maxQueued: number;
  private isCheckpointFn: (chunk: google.spanner.v1.PartialResultSet) => boolean;

  constructor(options: {
    maxQueued?: number;
    isCheckpointFn: (chunk: google.spanner.v1.PartialResultSet) => boolean;
  }) {
    super({objectMode: true});
    this.maxQueued = options.maxQueued ?? 10;
    this.isCheckpointFn = options.isCheckpointFn;
  }

  _transform(
    chunk: google.spanner.v1.PartialResultSet,
    enc: string,
    callback: () => void,
  ): void {
    this.queue.push(chunk);
    const isCheckpoint = this.isCheckpointFn(chunk);
    if (isCheckpoint) {
      this.emit('checkpoint', chunk);
      this._flushQueue();
    } else if (this.queue.length > this.maxQueued) {
      this._flushQueue();
    }
    callback();
  }

  private _flushQueue(): void {
    while (this.queue.length > 0) {
      this.push(this.queue.shift());
    }
  }

  reset(): void {
    this.queue = [];
  }

  _flush(callback: () => void): void {
    this._flushQueue();
    callback();
  }
}
```

olavloite (changes requested):

handwritten/spanner/src/partial-result-stream.ts
alkatrivedi: done

handwritten/spanner/src/partial-result-stream.ts
alkatrivedi: done

handwritten/spanner/src/partial-result-stream.ts
alkatrivedi: done

handwritten/spanner/src/partial-result-stream.ts
alkatrivedi: thanks for pointing this out! I have removed the `checkpoint-stream` package and have added our custom `CheckpointStream` class with some modification.

TL-DR:
The suggested class is synchronous and blocking. To make it asynchronous and non-blocking and to ensure it behaves like a production Node stream component I have done some modifications. I noticed few differences against production `checkpoint-stream`:

**1. Synchronous vs. Asynchronous Flushing (Blocking the Event Loop)**

**In production:** The original `checkpoint-stream` used a helper called `split-array-stream` which pushes items to the user asynchronously, one by one, using setImmediate(). This yielded control back to the event loop between each item.

**Custom class:** Our `CheckpointStream` flushes the queue synchronously in a while loop:

```
while (this.queue.length > 0) {
  this.push(this.queue.shift());
}
```
This could block the event loop until all items are pushed.

**Mitigation:** To mitigate, I have implemented an asynchronous recursive loop (`loop`) using `setImmediate()` which matches behavior identical to `split-array-stream`.

**2. Stream Flow Control and Backpressure**

**Custom class:** In `_transform`, the `callback()` was called synchronously at the very end, regardless of whether a flush was happening asynchronously (or even synchronously).

```
_transform(...) {
  // ... flush logic ...
  callback(); // Called immediately
}

```
**Mitigation:** With the modification, we tied the callback directly to the completion of the flush. The callback is passed to `_flushQueue` and is only called when the queue will be completely empty.

```
if (!shouldFlush) {
  return callback(); // Call immediately if not flushing
}
this._flushQueue(callback); // Pass callback to be called AFTER async flush
```

It will ensure that Node streams will be aware of when this transform step is `done` processing the current chunk, which will take care of flow control and backpressure.

**3. Potential Data Loss on Errors**
**Custom class:** Misses mechanism to handle shutdown on errors. If a non-retryable error occurred and standard `stream.destroy(err)` was called, any buffered data in `this.queue` would be dropped.

We added the `flushAndDestroy(err)` method:
```
flushAndDestroy(err: Error): void {
  this._flushQueue(() => {
    this.destroy(err);
  });
}
```
It will guarantees that if a fatal error occurs, we first push all buffered rows to the user asynchronously, and only destroy the stream once the queue is empty. It will prevent prevents data loss right before a stream failure.

handwritten/spanner/src/partial-result-stream.ts
alkatrivedi: since we are having:
```
const destroyRequestStream = (): void => {
           if (lastRequestStream) {
               lastRequestStream.removeListener('end', endListener);
               lastRequestStream.removeAllListeners('error');
               lastRequestStream.on('error', () => {});
               lastRequestStream.unpipe(requestsStream);
               lastRequestStream.destroy();
           }
};
```

where we are listening on the error event to prevent uncaught exceptions as per `lastRequestStream.on('error', () => {});`. Now keeping this code to prevent uncaught exceptions is resulting in registering of a new event hence, `assert.strictEqual(fakeStream.listenerCount('error'), 0, 'errorListener should be removed');` is failing since the error listener count is 1.

Currently, to keep the production safe I am asserting on value `1` in the test. But, can there be a better way?

handwritten/spanner/src/partial-result-stream.ts
```diff
@@ -554,6 +550,115 @@ export class PartialResultStream extends Transform implements ResultEvents {
...
+   */
+  private _flushQueue(callback: () => void): void {
+    const loop = () => {
+      if (this.destroyed) {
+        return callback();
+      }
+
+      if (this.queue.length > 0) {
+        this.push(this.queue.shift());
+        setImmediate(loop);
```
olavloite: We should not do this. It could lead to `PartialResultSet`s being lost during retries. Instead, we should do something like this:

```
--- a/handwritten/spanner/src/partial-result-stream.ts
+++ b/handwritten/spanner/src/partial-result-stream.ts
@@ -605,29 +605,17 @@ class CheckpointStream extends Transform {
       return callback();
     }
 
-    this._flushQueue(callback);
+    this._flushQueue();
+    callback();
   }
 
   /**
-   * Flushes queued chunks asynchronously using `setImmediate` to avoid blocking the event loop.
-   *
-   * @private
-   * @param {Function} callback Callback to call when all chunks are flushed.
+   * Flushes queued chunks synchronously to prevent state races on retry/reset.
    */
-  private _flushQueue(callback: () => void): void {
-    const loop = () => {
-      if (this.destroyed) {
-        return callback();
-      }
-
-      if (this.queue.length > 0) {
-        this.push(this.queue.shift());
-        setImmediate(loop);
-      } else {
-        callback();
-      }
-    };
-
-    loop();
+  private _flushQueue(): void {
+    while (this.queue.length > 0 && !this.destroyed) {
+      this.push(this.queue.shift());
+    }
   }
 
   /**
@@ -637,9 +625,8 @@ class CheckpointStream extends Transform {
    * @param {Error} err The error to destroy the stream with.
    */
   flushAndDestroy(err: Error): void {
-    this._flushQueue(() => {
-      this.destroy(err);
-    });
+    this._flushQueue();
+    this.destroy(err);
   }
 
   /**
@@ -655,6 +642,7 @@ class CheckpointStream extends Transform {
    * @param {Function} callback Callback to call when flushing is complete.
    */
   _flush(callback: () => void): void {
-    this._flushQueue(callback);
+    this._flushQueue();
+    callback();
   }
 }
```

And add this test:

```javascript
it('should not drop buffered checkpointed chunks when a retry occurs during flush', done => {
  const firstStream = through.obj();
  const secondStream = through.obj();
  const requestFnStub = sandbox.stub();

  const token1 = 'token1';
  // Chunks 1 to 3 have no token; Chunk 4 has token1
  const chunk1 = Object.assign({}, RESULT, {resumeToken: ''});
  const chunk2 = Object.assign({}, RESULT, {resumeToken: ''});
  const chunk3 = Object.assign({}, RESULT, {resumeToken: ''});
  const chunk4 = Object.assign({}, RESULT, {resumeToken: token1});
  // Chunk 5 is returned after retry
  const chunk5 = Object.assign({}, RESULT, {resumeToken: 'token2'});

  requestFnStub.onCall(0).callsFake(() => {
    setImmediate(() => {
      firstStream.push(chunk1);
      firstStream.push(chunk2);
      firstStream.push(chunk3);
      firstStream.push(chunk4); // Checkpoint hit: queue has 4 items
      // Simulate network blip immediately after sending chunk4
      firstStream.emit('error', {
        code: grpc.status.UNAVAILABLE,
        message: 'Unavailable',
      } as grpc.ServiceError);
    });
    return firstStream;
  });

  requestFnStub.onCall(1).callsFake(resumeToken => {
    try {
      assert.strictEqual(resumeToken, token1, 'Must resume from token1');
    } catch (e) {
      done(e);
    }
    setImmediate(() => {
      secondStream.push(chunk5);
      secondStream.end();
    });
    return secondStream;
  });

  const receivedRows: Row[] = [];
  partialResultStream(requestFnStub)
    .on('data', row => receivedRows.push(row))
    .on('error', done)
    .on('end', () => {
      try {
        // Must receive all 4 rows from the checkpointed batch + 1 from retry = 5 total
        assert.strictEqual(
          receivedRows.length,
          5,
          'All checkpointed rows must be delivered without being dropped by retry reset()',
        );
        done();
      } catch (e) {
        done(e);
      }
    });
});
```

alkatrivedi: done

handwritten/spanner/src/partial-result-stream.ts
```diff
@@ -554,6 +550,104 @@ export class PartialResultStream extends Transform implements ResultEvents {
...
+/**
+ * A custom Transform stream that buffers PartialResultSet chunks and flushes them
+ * asynchronously to prevent blocking the event loop.
+ *
+ * It holds chunks in a queue until a "checkpoint" is reached (as determined by
+ * `isCheckpointFn`) or until the queue exceeds `maxQueued` items.
+ *
+ * This matches the legacy behavior of buffering chunks and flushing them
+ * asynchronously using `setImmediate` to yield control back to the event loop,
+ * preventing long-running synchronous loops from blocking other processing.
```
olavloite: nit: I think that it would be good to just remove this section. In 6 months, no one will know what 'legacy behavior' is referring to.

handwritten/spanner/package.json
olavloite: nit: I think that we can remove this dependency as well now (or move it to dev-dependencies, as it is currently used in a test, but I don't think that test is relevant anymore due to all the other changes here).

olavloite (approved):
This looks great! I think that this will significantly reduce the overhead for running streaming queries, and also cleans up the code quite a bit.

handwritten/spanner/src/partial-result-stream.ts
alkatrivedi: done

handwritten/spanner/package.json
alkatrivedi: thanks for pointing this out! I have removed the dependency and have updated the test accordingly.

PR #9223, opened 2026-08-28
ci(generator): parallelize generator-test action

bshaffer:
The `generator-test` action takes over 60 minutes to run when generator files are modified. This attempts to run them in parallel, which should decrease execution time significantly

This is also done in #8973, but I split it out in case we'd like to review it separately!


shivanee-p (approved):

bshaffer:
We merged https://github.com/googleapis/google-cloud-node/pull/8973 instead, so this PR is redundant

PR #9229, opened 2026-08-31
feat: add new test-utils library

feywind:
Factors out the test resources class from the Pub/Sub library into its own library, so we can share it with other libraries.

b/554057686


feywind:
Same answer to all of the Gemini comments: this class is meant to be self-contained, not to adopt resources for existing tests. So I don't think we need to do anything here.

bshaffer (commented):
potentially naive question but, if this factors out utilities from the pubsub library, shouldn't we use it IN the pubsub library instead of what's there currently?

feywind:
> potentially naive question but, if this factors out utilities from the pubsub library, shouldn't we use it IN the pubsub library instead of what's there currently?

Yes. I have a separate PR for that, but this one has to be merged before it can work.

feywind:
Follow-up PR that factors out Pub/Sub pieces: https://github.com/googleapis/google-cloud-node/pull/9339

shivanee-p (approved):

PR #9230, opened 2026-09-01
refactor(storage): dynamically import mime and p-limit with module caching

### LINKED ISSUE #7350: Feature Request: Explore Dynamic Imports for mime and p-limit for improve performance
## Feature Request: Explore Dynamic Imports for `mime` and `p-limit`

**Description:**

This issue proposes exploring the adoption of dynamic `import()` for the `mime` and `p-limit` libraries within the `@google-cloud/storage` codebase. Currently, these libraries are likely imported statically, which means their code is loaded and potentially initialized even in scenarios where their functionality might not be immediately or frequently required.

**Motivation:**

Introducing dynamic imports for `mime` and `p-limit` could offer several potential benefits:

* **Improved Initial Load Time:** By deferring the loading of these libraries until they are actually needed, the initial load time of the `@google-cloud/storage` library could be reduced. This can be particularly beneficial for applications running in environments where startup performance is critical (e.g., serverless functions, client-side applications bundling the library).
* **Reduced Memory Footprint:** Delaying the loading of these dependencies can also lead to a smaller initial memory footprint, as the code for `mime` and `p-limit` would only be loaded into memory when their functionality is invoked.
* **Tree-Shaking Potential:** While modern bundlers are generally good at tree-shaking, dynamic imports can sometimes provide clearer boundaries for bundlers to identify and eliminate unused code within these dependencies more effectively.

**Specific Libraries to Consider:**

* **`mime`:** This library is used for determining the MIME type of files based on their extensions. Its usage might be concentrated in specific parts of the library, such as during upload or download operations where content type negotiation is involved.
* **`p-limit`:** This library is used for limiting the concurrency of asynchronous operations. Its usage is likely within internal mechanisms that handle multiple parallel requests or tasks.

**Proposed Approach:**

The investigation should involve:

1.  **Identifying Usage Patterns:** Analyze the `@google-cloud/storage` codebase to pinpoint exactly where and how frequently `mime` and `p-limit` are used.
2.  **Assessing Import Impact:** Measure the current impact of statically importing these libraries on initial load time and memory usage (if feasible within the testing framework).
3.  **Implementing Dynamic Imports:** Experiment with replacing static imports with dynamic `import()` calls within the relevant modules. This would likely involve refactoring the code to handle the asynchronous nature of dynamic imports (e.g., using `async/await`).
4.  **Evaluating Performance:** After implementing dynamic imports, re-measure the load time and memory usage in relevant scenarios to quantify the potential benefits.
5.  **Considering Trade-offs:** Evaluate any potential downsides of dynamic imports, such as the added complexity of asynchronous code and potential performance overhead if the libraries are used very frequently in quick succession after the initial load.
6.  **Ensuring Compatibility:** Verify that the changes are compatible with the supported Node.js versions and browser environments (if applicable).

**Potential Implementation Details:**

* For modules where `mime` or `p-limit` are used, the import could be done within the function or method where they are first needed.
* A local module-level variable could cache the imported module to avoid repeated dynamic imports in the same scope.

**Example (Conceptual):**

```javascript
// Current (static import)
import mime from 'mime/lite';

async function uploadFile(filename: string) {
  const contentType = mime.getType(filename);
  // ... use contentType ...
}

// Proposed (dynamic import)
async function uploadFile(filename: string) {
  const mime = await import('mime/lite');
  const contentType = mime.getType(filename);
  // ... use contentType ...
}

### PR DESCRIPTION
**Pull request by thiyaguk09**
Fixes #7350

### CODE REVIEW

File: handwritten/storage/src/file.ts [hunk 1]
```diff
@@ -2280,96 +2264,127 @@ class File extends ServiceObject<File, FileMetadata> {
       fileWriteStreamMetadataReceived = true;
     });
 
-    writeStream.once('writing', () => {
-      if (options.resumable === false) {
-        this.startSimpleUpload_(fileWriteStream, options);
-      } else {
-        this.startResumableUpload_(fileWriteStream, options);
-      }
+    writeStream.once('writing', async () => {
```
kalragauri, line 2239: gzip handling is split between the synchronous setup (`gzip === true` uses `push`) and the async `'writing'` callback (`gzip === 'auto'` uses `unshift`).

Could we consolidate MIME detection, gzip evaluation, and `transformStreams` construction inside the `'writing'` handler? 

Additionally, if `options.contentType` or `options.metadata.contentType` is already set (and not `'auto'`), we can skip calling `getMime()` entirely.

handwritten/storage/src/util.ts
```diff
@@ -318,3 +318,59 @@ export function handleContextValidation(
...
+
+let pLimitPromise: Promise<typeof import('p-limit')> | undefined;
+
+/**
+ * Lazily loads and returns the `p-limit` module.
+ * Caches the resolved module so dynamic import is evaluated only once.
+ *
+ * @internal
+ */
+export async function getPLimit(): Promise<typeof import('p-limit')> {
```
kalragauri: This method returns the callable limiter function directly (`modObj.default || modObj`), but its return type is declared as the module namespace `Promise<typeof import('p-limit')>`.

To align the type with runtime behavior and prevent typing issues with `pLimit(concurrency)`, consider refining the return type.

And similarly ensure getMime() reflects the resolved MIME interface with .getType().

handwritten/storage/src/util.ts
kalragauri, line 1: Could we add dedicated unit tests under test/util.ts for `getMime()` and `getPLimit()` covering:
   - Promise caching
   - Error recovery

handwritten/storage/test/bucket.ts
```diff
@@ -234,6 +234,11 @@ describe('Bucket', () => {
         ServiceObject: FakeServiceObject,
         util: fakeUtil,
       },
+      './util.js': {
+        ...require('../src/util.js'),
+        getPLimit: async () => fakePLimit,
```
kalragauri, line 238: Is this line required? `src/bucket.ts` imports `getPLimit` via `./util.js`

handwritten/storage/src/file.ts
thiyaguk09: Consolidated MIME detection, gzip evaluation, and `transformStreams` construction inside the `'writing'` callback, removing the early synchronous setup and the `unshift(zlib.createGzip())`.

Also updated the check so `getMime()` is completely bypassed when `contentType` or `metadata.contentType` is already provided (and not `'auto'`), with unit tests added to verify this behavior.

handwritten/storage/src/util.ts
thiyaguk09: Updated. 
`getPLimit()` now returns `Promise<PLimit>` where `PLimit = (concurrency: number) => Limit`.
`getMime()` now returns `Promise<Mime>` with `getType(path: string): string | null`.

Also removed `any` assertions.

handwritten/storage/test/bucket.ts
thiyaguk09: Yes, this is required.

In `test/bucket.ts`, two tests (`deleteFiles` and `makeAllFilesPublicPrivate_` under `"should process 10 files at a time"`) rely on `pLimitOverride` in `fakePLimit` to assert `limit === 10` and invoke `done()`. 

Since `src/bucket.ts` now calls `const pLimit = await getPLimit()` from `./util.js`, proxyquire must stub `getPLimit` on `'./util.js'` to route calls through `fakePLimit`. Without it, real `p-limit` is used, `pLimitOverride` is never called, and both tests time out.

*We removed the legacy top-level `'p-limit': fakePLimit` in proxyquire as that was indeed obsolete.

handwritten/storage/src/util.ts
thiyaguk09: I have added a dedicated new unit test file, `test/util.ts`, for both `getMime()` and `getPLimit()`.

kalragauri (approved):

PR #9235, opened 2026-09-02
chore: gax linter fixes

shivanee-p:
Resolves legacy ESLint promise violations in core/packages/gax/test/unit/grpc.ts and grpc-fallback.ts
- Wrapped callback-based RPC calls in Promise constructs
- added async() + await callbacks


danieljbruce (approved):
LGTM. The async tests are cleaner to read too!

PR #9250, opened 2026-09-04
chore(firestore): add `--exit` flag to `test-only` and `conformance`

dlarocque:
Pass `--exit` to mocha in test-only and conformance to ensure Node terminates once tests complete. This prevents test runner from hanging on open grpc streams and background timer handles in modern node versions.

This technically could hide bugs in code that cause our tests to hang.


MarkDuckworth (approved):
Are we seeing instances of these script hang at the end of testing?

PR #9251, opened 2026-09-04
chore(firestore): disable promise lint rules in test overrides

dlarocque:
Expand the Firestore test override glob to match nested test subdirectories and disable `promise/always-return` and `promise/catch-or-return`.


MarkDuckworth (approved):
This is fine. I was planning to eventually address all of these issues d

Issue #9185, opened 2026-08-21
storage 8.0.1: MaxListenersExceededWarning (11 error/close listeners on PassThrough) on every createReadStream()


glenn-jocher:
Package: `@google-cloud/storage` (handwritten/storage)

## Environment
- @google-cloud/storage 8.0.1 (teeny-request 11.0.1, retry-request 9.0.1)
- Node 26.5.1 (also on Vercel's Node runtime)
- Does **not** occur with 8.0.0 (teeny-request 9.0.0), same code — 8.0.1's change vs 8.0.0 is that dependency bump (teeny-request 11 moved to node-fetch 3).

## Steps to reproduce
```js
import { Storage } from "@google-cloud/storage";
const file = new Storage().bucket("<bucket>").file("<~3 MB object>");
let n = 0;
for await (const chunk of file.createReadStream()) n += chunk.length;
```
Run with `node --trace-warnings`.

## Actual
Every streamed download emits both:
```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added to [PassThrough]. MaxListeners is 10.
    at pipelineImpl (node:internal/streams/pipeline:286:14)
    at pipeline (node:internal/streams/pipeline:183:10)
    at PassThrough.<anonymous> (teeny-request/build/src/index.js:194:43)
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [PassThrough]. MaxListeners is 10.
    at eos (node:internal/streams/end-of-stream:302:10)
    at pipe (node:internal/streams/pipeline:447:3)
    at pipelineImpl (node:internal/streams/pipeline:374:25)
    at PassThrough.<anonymous> (teeny-request/build/src/index.js:194:43)
```
The emitter is the node-fetch 3 response body `PassThrough` (teeny-request's `responseStream` in stream mode). Its `error` listeners when the warning fires are node-fetch's `FetchError` wrapper, three `onError`/`eos` pairs from node-fetch's own internal `pipeline` calls, teeny-request's `(err) => requestStream.emit('error', err)` forwarder (`index.js:201`), and the listeners `stream.pipeline(responseStream, requestStream)` adds at `index.js:194` — the count is reached by library wiring over node-fetch 3's body, not by the caller. (teeny-request's own repo is archived, so reporting here.)

## Expected
No warning for a single stream consumer, as with 8.0.0.

## Impact
[...]

PR #9255, opened 2026-09-05
feat(teeny-request)!: replace node-fetch with undici

re-taro:
teeny-request wrapped node-fetch, whose response body stream carries
enough internal pipeline listeners that streamed downloads through
consumers like @google-cloud/storage warn with
MaxListenersExceededWarning on every request. Rather than raising the
listener limit, this change moves the transport to undici's request
API, which hands back a plain readable stream.

Stream mode keeps its contract of delivering the raw response bytes
without decompression, which checksum validation in
@google-cloud/storage depends on; the built-in fetch API cannot provide
this because it always decompresses gzip responses. Callback mode still
requests and transparently decompresses gzip, deflate, and brotli
responses unless gzip is set to false. Redirects are followed through a
redirect interceptor, up to 20 like node-fetch. Proxies use undici's
ProxyAgent, replacing http-proxy-agent and https-proxy-agent, with the
same proxy environment variable and NO_PROXY handling as before.
System error codes such as ECONNRESET that undici wraps in its own
error types are surfaced on err.code, which downstream retry logic
keys off. Destroying an unread stream now aborts the in-flight request,
releasing the socket, which previously leaked and forced consumers to
destroy the shared agent as a workaround.

BREAKING CHANGE: Response.request.agent is now always false, connection
pooling is handled by undici dispatchers, and network errors are undici
error types instead of node-fetch's FetchError. The request timeout
option now applies again as undici's headers and body timeouts; it had
been silently ignored since node-fetch 3 removed its timeout option.
Agent pool options other than maxSockets are ignored.

For #9185

### DIFF
--- a/core/packages/teeny-request/README.md
+++ b/core/packages/teeny-request/README.md
@@ -4,7 +4,7 @@
 
 # teeny-request 
 
-Like `request`, but much smaller - and with less options. Uses `node-fetch` under the hood. 
+Like `request`, but much smaller - and with less options. Uses `undici` under the hood. 
 Pop it in where you would use `request`. Improves load and parse time of modules. 
 
 ```js
--- a/core/packages/teeny-request/package.json
+++ b/core/packages/teeny-request/package.json
@@ -30,7 +30,7 @@
   },
   "keywords": [
     "request",
-    "node-fetch",
+    "undici",
     "fetch"
   ],
   "author": "fhinkel",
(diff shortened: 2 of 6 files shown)


re-taro:
Stacked on #9254 — this branch includes that commit, so the diff will shrink to the migration itself once #9254 merges. Review order: #9254 first.

Note for reviewers: this migration moves the transport off node-fetch, so downstream consumers whose unit tests intercept HTTP with nock (which cannot intercept undici) will need their tests adapted when they pick up the new major. teeny-request's own tests were rewritten against a local HTTP server in this PR.

PR #9258, opened 2026-09-08
fix(spanner): complete query stream on PartialResultSet.last without resumeToken

olavloite:
Ensures query streams flush and complete immediately upon receiving `PartialResultSet.last`:
- **Immediate flush**: Flushes and completes the stream when `chunk.last` is true, even if the final chunk does not contain a `resumeToken` (bypassing the wait for the gRPC stream to close).
- **Clean stream lifecycle**: Properly finishes the stream's writable side so that `'finish'`, `'close'`, and `stream.finished()` resolve naturally.
- **Graceful drain**: Detaches the underlying gRPC stream on the final chunk so trailers drain cleanly in the background.


handwritten/spanner/src/partial-result-stream.ts
```diff
@@ -713,11 +730,22 @@ export function partialResultStream(
...
+      if (errorListener) {
+        streamToClean.removeListener('error', errorListener);
+      }
+      streamToClean.on('error', () => {});
+      streamToClean.unpipe(requestsStream);
+      if (receivedLast) {
+        // Query completed successfully. Do not cancel the gRPC call; allow it
+        // to drain remaining trailers/EOF in the background so it is not marked
+        // CANCELLED by Spanner or Cloud Monitoring.
+        streamToClean.resume();
```
sakthivelmanii, line 745: Should we add some drain timeout here incase if trailers are not received for any reasons?

```js
  const drainTimeout = setTimeout(() => streamToClean.destroy(), 5000);
  streamToClean.once('end', () => clearTimeout(drainTimeout));
  streamToClean.once('close', () => clearTimeout(drainTimeout));
```

olavloite: No, I don't think we should:
1. Calling `setTimeout(..)` for each query introduces additional event loop congestion, which is exactly the kind of thing that we are trying to get rid of with this optimization (and other optimzations)
2. The probability that the stream gets stuck __after__ having sent a `PartialResultSet` with `last=true`, but __before__ receiving the trailers, is extremely unlikely.
3. In the extremely unlikely case that this happens, then the stream will eventually be cleaned up once the query timeout is reached (but that is 1hr by default)
4. We do not add an extra timeout for the same optimization in Java and Go.
5. The Node Firestore client in this monorepo also contains a similar optimization, and does also not add an additional timeout for receiving the trailers.

sakthivelmanii: My concern was if it stuck for some reasons and if we are not cleaning up the stream for next 1hr then in channel capacity will be reduced due to MAX_CONCURRENT_STREAMS(max concurrent streams is 100).

Since this is unlikely high, we can skip.

sakthivelmanii (approved):

PR #9260, opened 2026-09-08
perf(spanner): eliminate throwaway transactions

olavloite:
- Reuse a singleton NOOP_SPAN instead of allocating on every untraced span lookup
- Dispatch cached multiplexed sessions via process.nextTick to avoid Promise churn
- Stop instantiating unused Transaction objects on read queries in MultiplexedSession.getSession
- Defer Transaction creation to SessionFactory.getSessionForReadWrite for read-write operations

### DIFF
--- a/handwritten/spanner/src/database.ts
+++ b/handwritten/spanner/src/database.ts
@@ -25,6 +25,7 @@ import {
 const common = require('./common-grpc/service-object');
 import {promisify, promisifyAll, callbackifyAll} from '@google-cloud/promisify';
 import * as extend from 'extend';
+// eslint-disable-next-line import/namespace
 import * as r from 'teeny-request';
 import * as streamEvents from 'stream-events';
 import * as through from 'through2';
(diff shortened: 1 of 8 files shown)


handwritten/spanner/src/database.ts
```diff
@@ -3341,69 +3342,96 @@ class Database extends common.GrpcServiceObject {
...
+          } else {
+            setImmediate(runFn, err);
+            release();
+          }
+          return null;
+        },
+      )
+      .catch(err => {
+        process.nextTick(() => {
+          throw err;
```
sakthivelmanii: Should we throw an error here? or can we just emit an error? throwing an error may break the application

sakthivelmanii (changes requested):

handwritten/spanner/src/database.ts
olavloite: Yes, good catch. Fixed that, and added a test for it.

sakthivelmanii (approved):

PR #9266, opened 2026-09-08
feat(gax): attach internalMethodName to otherArgs in constructSettings

shivanee-p:
Update constructSettings() to pass through the method name to `google-gax`


danieljbruce (approved):
Thanks for breaking this up into isolated changes. This looks great.

PR #9267, opened 2026-09-08
feat: add o11y gRPC tracing

shivanee-p:
Call TracerHelper within createApiCall() and update unit tests accordingly
- build StaticTraceContext from the internal telemetry information from CallSettings
- determine if calls are streaming or not
- add unit tests for fallback detection


quirogas:
## Nit: optional chaining on `internalTelemetryInfo`

[`createApiCall.ts#L184-L187`](https://github.com/googleapis/google-cloud-node/blob/shivaneep-o11y-grpc-tracing/core/packages/gax/src/createApiCall.ts#L184-L187)

Nit: `settings.otherArgs.internalTelemetryInfo?.…` starts the optional chain one level too late. Line 194 just below uses `settings.otherArgs?.internalMethodName`, so this one looks out of step with it.

To be clear, this isn't currently reachable, because `checkTelemetryEnabled(settings)` already guarantees `otherArgs` is defined. Flagging it purely for consistency and defensive hardening, since it becomes a live `TypeError` the moment the gating is refactored.

```ts
settings.otherArgs?.internalTelemetryInfo?.gcpClientService,
```

---

## Cross-reference

The callback-style span issue I raised on #9269 (spans ending in tick 0) manifests through this PR's wiring, though the mechanism lives in `traceAttempt` in #9269. Noting it here so the two don't get looked at in isolation. Whichever PR takes the fix, the other should reference it.

core/packages/gax/src/createApiCall.ts
```diff
@@ -168,5 +228,45 @@ export function createApiCall(
     // or to cancel the ongoing call.
     return currentApiCaller.result(ongoingCall);
   };
-  return invokeCall;
+
+  if (tracingEnabled) {
+    const staticArgs: StaticTraceContext = {
+      gcpClientService:
+        settings.otherArgs.internalTelemetryInfo?.gcpClientService,
```
quirogas: Nit: `settings.otherArgs.internalTelemetryInfo?.…` starts the optional chain one level too late. Line 245 just below uses `settings.otherArgs?.internalMethodName`, so this one looks out of step with it.

To be clear, this isn't currently reachable, because `checkTelemetryEnabled(settings)` already guarantees `otherArgs` is defined. Flagging it purely for consistency and defensive hardening, since it becomes a live `TypeError` the moment the gating is refactored.

```ts
settings.otherArgs?.internalTelemetryInfo?.gcpClientService,
```

quirogas (commented):
Thanks for wiring `traceCall`, the callback wrapper, and `getTracingBackstopMs` through `createApiCall`, along with the new unit tests in `apiCallable.ts`. One small nit inline.

shivanee-p:
> Thanks for wiring `traceCall`, the callback wrapper, and `getTracingBackstopMs` through `createApiCall`, along with the new unit tests in `apiCallable.ts`. One small nit inline.

Ty for the comment! I updated the logic in the stack, but I removed the backstop parts. [PR#9334](https://github.com/googleapis/google-cloud-node/pull/9334) has the fix instead. The TLDR is that we never actually verified that REST calls were complete.

quirogas (approved):

PR #9269, opened 2026-09-08
feat(gax): o11y tracer helper updates

shivanee-p:
Updates `traceCall` in `TracerHelper.ts` to manage span lifecycles across Promises or streaming RPCs, and callbacks without closing spans prematurely.
- Add overloads of `traceAttempt()` function call to account for Promises or Streams returned from GAX
- Add helper functions to close spans if they are Promises or Streams
- Make sure spans only close once
- Account for users making calls with callback functions: passes the callback as a parameter
- set span status within traceCall

Note: confirmed that the `google-gax` test suite passes with `npm run test-application`


core/packages/gax/src/observability/TracerHelper.ts
```diff
@@ -67,23 +69,165 @@ export function getGaxTracer(): Tracer {
...
+  stream: EventEmitter,
+  recordError: (err: unknown) => void,
+  endSpan: () => void,
+): void {
+  let spanEnded = false;
+
+  const cleanup = () => {
+    stream.removeListener('error', onError);
+    stream.removeListener('end', onEnd);
+    stream.removeListener('close', onClose);
```
danieljbruce, line 163: I don't know too much about how this code fits in with the broader architecture, but one thought I had is if we remove the listeners here which I see intends to remove the listeners applied on the stream.on('error', onError) lines of code then it is going to remove other listeners as well that we need?

I remember google-gax applies a lot of error listeners in its middleware to handle retries and other such tasks.

core/packages/gax/test/unit/tracerHelper.ts
```diff
@@ -163,5 +172,572 @@ describe('TracerHelper', () => {
       assert.strictEqual(spans.length, 1);
       assert.strictEqual(spans[0].attributes['gcp.method.type'], 'http');
     });
```
danieljbruce, line 174: Could it be in scope to test retries here? I know they get complex with the streaming calls.

danieljbruce (commented):
Left a couple comments. Overall, I think I need to understand the problem and architecture more broadly to understand how traceAttempt and traceHelper.ts are used.

core/packages/gax/src/observability/TracerHelper.ts
shivanee-p: I double checked, and I confirmed taht `removeListener()` removes only the reference for that specific callback that's used. In contrast, `removeAllListeners()` cleans up *all* of the listeners and would affect the middleware listeners.

core/packages/gax/test/unit/tracerHelper.ts
shivanee-p: Good point, thank you! I've added tests for retries in streams

danieljbruce (commented):
Ok. The code LGTM. I see how the traceAttempt is used as a wrapper with invokeCall. Maybe consider inlining traceAttempt into createApiCall and then refactoring pieces of it out to eliminate the need to pass a `fn: () => T` parameter, but that shouldn't block. Maybe just consider it as a backlog item.

The broader thing that might just be good to check is to run the google-gax tests. After the monorepo migration, I don't think they run in the CI pipeline anymore. I'll send a link on chat for how to run the gax tests. Probably would be good to run the gax tests before merging.

quirogas:
Reviewed this against a real in-process `grpc-js` server with spans exported to live Cloud Trace, rather than by reading alone. Architecture looks right; the findings below are all in async lifecycle handling.

---

## Client-streaming spans are never ended

[`TracerHelper.ts#L189-L191`](https://github.com/googleapis/google-cloud-node/blob/shivaneep-o11y-tracer-helper-updates/core/packages/gax/src/observability/TracerHelper.ts#L189-L191)

`handleStream()` subscribes to `'error'`, `'end'`, and `'close'`. For a client-streaming call none of those can fire, so the span is never ended.

For example, writing a couple of messages up to the server and closing the stream:

```js
const call = createApiCall(
  client.collect.bind(client),
  settings('Collect'),
  new StreamDescriptor(StreamType.CLIENT_STREAMING, true),
);

const stream = call(null, undefined, (err, resp) => {
  // fires normally: resp.content === 'collected 2'
});

stream.write({content: 'a'});
stream.write({content: 'b'});
stream.end();
```

The RPC itself works fine. It's only the span that goes missing.

`StreamProxy` is constructed with `readable: type !== StreamType.CLIENT_STREAMING` ([`streaming.ts#L130`](https://github.com/googleapis/google-cloud-node/blob/shivaneep-o11y-tracer-helper-updates/core/packages/gax/src/streamingCalls/streaming.ts#L130)). A non-readable duplex never emits `'end'`, a writable side emits `'finish'`. And since nothing destroys the proxy on the success path, `'close'` never fires either.

Verified against a real in-process `grpc-js` 1.14.4 server: a client-streaming RPC that succeeds (`resp: "collected 2"`) exports zero spans. Observed emissions: `finish: true, end: false, close: false`. The span is also absent from the live Cloud Trace readback.

Two consequences:
1. Successful client-streaming RPCs are invisible in tracing.
2. `cleanup()` never runs, so 3 listeners are retained per call, which adds up on a long-lived client.

I think adding `'finish'` to the listener list would cover it? `endSpanOnce()` is already idempotent, so it should be safe to just hook it alongside the others:

```ts
const onFinish = () => { endSpanOnce(); };

stream.on('error', onError);
stream.on('end', onEnd);
stream.on('close', onClose);
stream.on('finish', onFinish);   // client-streaming: writable side
```

Couple of things worth checking if you go that route: `cleanup()` would need to remove `'finish'` as well, and it'd be good to confirm bidi still ends exactly once, since bidi emits both `'end'` and `'finish'`. I'd expect `endSpanOnce()` to handle it, but I haven't tested that combination.

---

## Callback-style invocations end the span in tick 0

[`TracerHelper.ts#L272-L279`](https://github.com/googleapis/google-cloud-node/blob/shivaneep-o11y-tracer-helper-updates/core/packages/gax/src/observability/TracerHelper.ts#L272-L279)

When an RPC is invoked callback-style, the API caller returns an `OngoingCall` with no `.promise`, so `fn()` returns `undefined`. `promiseTarget` is then `null` and `isStreamCall` is false, so control reaches `else { endSpan(); }` and the span closes synchronously, before the RPC has even been sent.

For example, a real in-process `grpc-js` server with a 150 ms delay, real `createApiCall`, same method invoked both ways:

```js
const unary = () => createApiCall(client.echo.bind(client), settings('Echo'));

// R1 — promise style (control)
await unary()({content: 'hi'}, undefined);

// R2 — callback style, same method
let spansAtTick0 = null;
await new Promise((resolve, reject) => {
  unary()({content: 'hi'}, undefined, (err, resp) =>
    err ? reject(err) : resolve(resp),
  );
  spansAtTick0 = gaxSpans().length - before;  // span already exported here
});

// R4 — callback style against a failing RPC
unary()({content: 'hi', fail: true}, undefined, (err, resp) => { ... });
```

`spansAtTick0 === 1` shows the span is finished synchronously, before the promise executor even returns. Results:

| Case | Duration | Covers RPC? |
|---|---|---|
| unary, promise | 177.55 ms | ✅ |
| unary, callback | 0.15 ms | ❌ |
| unary, callback, failing | 0.06 ms | ❌ `exceptionEvents: 0` |

That last row is the one to look at. The callback gets gRPC code 5 like you'd expect, but the span doesn't record any of it, so a failing callback style RPC doesn't show up in tracing at all.

This isn't just a `NormalApiCaller` problem. All four callers do the same thing, they hand back an `OngoingCall` built from the callback, and then ask it for a `.promise` that was never set:

- `NormalApiCaller`
- `BundleApiCaller`
- `LongrunningApiCaller` (measured: 0.10 ms)
- `PagedApiCaller`

So it hits all of them, not just unary calls.

This also explains page-streaming, so we don't need to track that separately. `PageDescriptor.createStream` calls `apiCall(request, options, callback)`, it goes through a callback, so `listItemsStream()` breaks for exactly the same reason. Fix this and page-streaming is fixed too.

One way to approach the fix: when a callback is supplied, wrap it so `recordError(err)` and `endSpan()` run when it fires, and skip the synchronous `endSpan()` in the `else` branch. The wrapper would need to preserve `this` and the full argument list.

---

## Is the span status ever set?

[`TracerHelper.ts#L248-L265`](https://github.com/googleapis/google-cloud-node/blob/shivaneep-o11y-tracer-helper-updates/core/packages/gax/src/observability/TracerHelper.ts#L248-L265)

`recordError` sets `error.message` / `error.type` and calls `span.recordException(e)`, but I don't see a call to `span.setStatus(...)` anywhere in here. Am I missing it, or is the status genuinely never set?

The reason it matters: [Record Exception](https://opentelemetry.io/docs/specs/otel/trace/api/#record-exception) is just a specialized form of `AddEvent`, so it attaches an event and nothing more. [Set Status](https://opentelemetry.io/docs/specs/otel/trace/api/#set-status) is the separate call that moves a span off the default, and the spec is explicit that the default is `Unset`. On a real failing gRPC call the span came back with the error attributes and one exception event, but `statusCodeRaw: 0` (UNSET) even though the RPC returned `NOT_FOUND`.

Backends key their error surfacing (dashboards, error rates, alerting) off `status.code`, so as written a failed RPC isn't distinguishable from a successful one by status.

```ts
import {SpanStatusCode} from '@opentelemetry/api';

span.setStatus({
  code: SpanStatusCode.ERROR,
  message: e instanceof Error ? e.message : String(e),
});
```

---

## Nit: `error.type` / `exception.type` disagree

[`TracerHelper.ts#L252-L257`](https://github.com/googleapis/google-cloud-node/blob/shivaneep-o11y-tracer-helper-updates/core/packages/gax/src/observability/TracerHelper.ts#L252-L257)

Nit: worth settling before GA. On the real gRPC path `e.constructor?.name` resolves to `"Error"`, not `"GoogleError"`, so `error.type` is less useful than it looks. Separately, `exception.type` is set as a span attribute here, but in OTel semconv it belongs on the exception event, and `recordException()` already populates it there. So we end up emitting it twice, and for `GoogleError` the two values disagree.

Suggest dropping the manual `exception.type` attribute and deciding explicitly what `error.type` should carry (gRPC status name? `e.name`?).

---

## Nit: `traceAttempt` is per-call, not per-attempt

Nit: the name reads as per-attempt, but the span wraps the whole call including retries. 3 attempts produced 1 span of 376 ms. That's a defensible design, it just means retry behavior is invisible in traces, and the name may throw off future readers.

Not asking to change it in this PR. It's probably worth an explicit decision on per-attempt vs per-call semantics before this ships, since changing it later would be a breaking change to trace shape.

---

## Test coverage

The suite is green (79 passing) but the o11y tests assert on `traceAttempt`'s arguments only, never on span lifecycle. All three issues above pass CI today.

Suggest adding lifecycle assertions to `otelHarness`:
- exported span count per call (catches the client-streaming leak)
- span duration > 0 / covers the RPC (catches the callback case)
- status code on failure
- listener count returns to baseline after completion

Happy to share the repro scripts and raw evidence if useful.

core/packages/gax/src/observability/TracerHelper.ts
```diff
@@ -67,23 +69,226 @@ export function getGaxTracer(): Tracer {
...
+
+  const onFinish = () => {
+    endSpanOnce();
+  };
+
+  stream.on('error', onError);
+  stream.on('end', onEnd);
+  stream.on('close', onClose);
+  if (isWriteOnly) {
+    stream.on('finish', onFinish);
```
quirogas, line 251: Nice catch gating `'finish'` on `isWriteOnly` so bidi streams aren't truncated when the client finishes writing while the server is still streaming responses.

core/packages/gax/src/observability/TracerHelper.ts
```diff
@@ -93,22 +298,115 @@ export async function traceAttempt<T = unknown>(
...
+        }
+        span.end();
+      }
+    };
+
+    const recordError = (e: unknown) => {
+      if (e instanceof Error) {
+        span.setAttributes({
+          'error.message': e.message,
+          'error.type': e.constructor?.name ?? e.name,
```
quirogas: Nit: worth settling before GA. On the real gRPC path `e.constructor?.name` resolves to `"Error"`, not `"GoogleError"`, so `error.type` is less useful than it looks. Separately, `exception.type` is set as a span attribute on line 337, but in OTel semconv it belongs on the exception event, and `recordException()` already populates it there. So we end up emitting it twice, and for `GoogleError` the two values disagree.

Suggest dropping the manual `exception.type` attribute and deciding explicitly what `error.type` should carry (gRPC status name? `e.name`?).

quirogas (commented):
Thanks for the quick turnaround in `48c56bbc00` and `804e233c72`! Verified that all three lifecycle issues (client-streaming `'finish'` gated on `isWriteOnly`, callback wrapping + `maxDurationMs` backstop, and `SpanStatusCode.ERROR` / `OK` status resolution) are addressed, along with the rename to `traceCall` and the expanded test suite in `tracerHelper.ts`. Left a couple small inline notes below.

core/packages/gax/src/observability/TracerHelper.ts
```diff
@@ -93,22 +298,115 @@ export async function traceAttempt<T = unknown>(
...
+    // For callback-style invocations the span's lifetime is bound to the
+    // callback rather than to a promise or stream. Declared as a `function` and
+    // forwarding `arguments` via `apply` so the caller's `this` binding and the
+    // full argument list (err, response, next, rawResponse) are preserved.
+    //
+    // The span is ended *before* the user callback runs so the span measures the
+    // RPC itself, and so a throwing user callback cannot leak the span.
+    const tracedCallback: APICallback | undefined =
+      !isStreamCall && callback
+        ? function (this: unknown, ...args: Parameters<APICallback>) {
```
quirogas, line 369: Clean fix wrapping the callback with `.apply(this, args)` and ending the span before invoking the user callback so a throwing callback cannot leak the span.

core/packages/gax/src/observability/TracerHelper.ts
```diff
@@ -67,23 +70,239 @@ export function getGaxTracer(): Tracer {
...
+  };
+
+  const onFinish = () => {
+    endSpanOnce();
+  };
+
+  stream.on('error', onError);
+  stream.on('end', onEnd);
+  stream.on('close', onClose);
+  if (isWriteOnly) {
```
quirogas:
```suggestion
  if (isWriteOnly && !hasCallback) {
```

add a callback check to make this termination more robust.

core/packages/gax/src/observability/TracerHelper.ts
```diff
@@ -93,22 +312,88 @@ export async function traceAttempt<T = unknown>(
...
+
+    // For callback-style invocations the span's lifetime is bound to the
+    // callback rather than to a promise or stream. Declared as a `function` and
+    // forwarding `arguments` via `apply` so the caller's `this` binding and the
+    // full argument list (err, response, next, rawResponse) are preserved.
+    //
+    // The span is ended *before* the user callback runs so the span measures the
+    // RPC itself, and so a throwing user callback cannot leak the span.
+    const tracedCallback: APICallback | undefined =
+      !isStreamCall && callback
```
quirogas:
```suggestion
      callback
```

Only check for callback to allow client streaming callbacks to be wrapped and later handled by the onClose function.

quirogas (approved):
I think this las two suggestions might break helpers and other.

PR #9281, opened 2026-09-10
chore(gax): resolves linter warnings in gax

shivanee-p:


quirogas:
For your awareness, not a merge gate.

I can reproduce the failure (`GIT_DIFF_ARG="HEAD^1...HEAD" node ./bin/linter.mjs --strict` gives exit 1, 7 errors / 17 warnings).

That said, it isn't blocking anything. The `lint` job in `presubmit.yaml` is `continue-on-error: true` (L41), so the red X is informational.

Root cause looks like commit `087aea5` ran Prettier across `gax` and reformatted imports in legacy files (`fallbackServiceStub.ts`, `pagedApiCaller.ts`, `test/unit/pagedIteration.ts`), which pulled pre-existing `promise/*` violations into the `HEAD^1...HEAD` diff window. They're inherited debt rather than regressions from this stack.

One option is reverting the formatting-only churn on those three legacy files and keeping the targeted `apiCallable.ts` fixes. I'd avoid rewriting promise chains in the pagination and fallback stubs here, since that's a lot of blast radius for a cleanup PR.

core/packages/gax/src/fallbackServiceStub.ts
```diff
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
-import type {Response as NodeFetchResponse} from 'node-fetch' with {'resolution-mode': 'import'};
+import type {Response as NodeFetchResponse} from 'node-fetch' with {
```
quirogas, line 17: Nit: commit `087aea5` ran Prettier across `gax` and reformatted imports in legacy files (`fallbackServiceStub.ts`, `pagedApiCaller.ts`, `test/unit/pagedIteration.ts`), which pulled pre-existing `promise/*` violations into the `HEAD^1...HEAD` diff window checked by `linter.mjs --strict`. One option is reverting the formatting-only churn on those three legacy files and keeping the targeted `apiCallable.ts` fixes so we avoid rewriting promise chains in the pagination and fallback stubs.

quirogas (commented):
For your awareness, not a merge gate. The `lint` job in `presubmit.yaml` is `continue-on-error: true` (L41), so the red X is informational. Note that since #9277 has new commits merged from #9269/#9267, this branch will need a merge/rebase.

PR #9286, opened 2026-09-10
chore(generator): regenerate protos

feywind:
Refreshes protos/protos.json and protos/index.js from the checked-in protos with the current protobufjs toolchain. This is generated-file churn only and is independent of the resumable upload work.


feywind:
Header check is for a generated file that's always lacked a header.

feywind:
I think I'm just going to close this one.

PR #9287, opened 2026-09-10
feat(gax): support resumable uploads

feywind:
Adds the client-side implementation of the resumable upload protocol to `google-gax`.

### What's here

- `ResumableUploadDescriptor` / `ResumableUploadSession` and the `resumableUploadStub` that generated clients wire into `createApiCall`
- `resumableSourceFromFile`, a seekable upload source backed by a local file
- `CallOptions.resumableUpload`, the transport context generated clients pass to the stub
- exports from `index`, `fallback` and `descriptor`, plus user documentation in `client-libraries.md`
- unit tests and a hermetic system test (real HTTP server, no credentials) covering the state machine, transient retries, recovery from state mismatches and resume from a saved session URL

`onProgress` callbacks may return `void`; the previous signature rejected the documented usage, which only logs progress.

### Verification

- `npx tsc -p .` — no errors in the touched files
- `npx mocha build/test/unit` — 406 passing
- `npx mocha build/test/system-test/resumableUpload.js` — 2 passing

Nothing in existing behaviour changes: the new session is only reachable through the new descriptor and stub.

### DIFF
--- a/core/packages/gax/client-libraries.md
+++ b/core/packages/gax/client-libraries.md
@@ -149,6 +149,48 @@ in the second parameter:
 const [response] = await client.sampleMethod(request, options);
 ```
 
+### Resumable uploads
+
+Some APIs expose methods that upload large payloads through the resumable
+upload protocol. For these methods, the client method no
+longer returns the response directly; it returns a
+[`ResumableUpload`](https://googleapis.dev/nodejs/google-gax/latest/classes/ResumableUpload.html)
+helper. Call `start()` with a `NodeJS.ReadableStream` and await `finished()`
+for the final response:
+
+```ts
+const helper = await client.createResumableUpload(request);
+await helper.start({
+  uploadStream: dataStream,
+  chunkSize: 8 * 1024 * 1024, // 8MB chunks
+  onProgress: status => {
+    console.log(`Committed ${status.bytesUploaded} bytes to ${status.uploadUrl}`);
+  },
+});
+const response = await helper.finished();
+```
+
+The session URL is available as `helper.uploadUrl` once the upload has
+started. Save it if you need to resume the upload later — for example after a
+process crash or network drop. To resume, pass the saved URL to `start()` on a
+new helper, along with a fresh stream of the same payload:
+
+```ts
+const helper = await client.createResumableUpload();
+await helper.start({
+  uploadStream: dataStream,
+  resumeUrl: savedUploadUrl,
+});
+const response = await helper.finished();
+```
+
+The current implementation requires a seekable stream (for example, a file
+stream). Errors fall into three categories: transient errors (retried with
+exponential backoff), state mismatches (recovered by querying the server for
+the committed byte offset), and fatal errors (propagated to the caller).
+The whole session is bounded by a global deadline (10 minutes by default,
+scaled up for large payloads and overridable via `globalDeadlineMs`).
+
 ### Long-running operations
 
 Some methods are expected to run longer. They return an object of type
--- a/core/packages/gax/client-libraries.md
+++ b/core/packages/gax/client-libraries.md
@@ -154,14 +154,14 @@ const [response] = await client.sampleMethod(request, options);
 Some APIs expose methods that upload large payloads through the resumable
 upload protocol. For these methods, the client method no
 longer returns the response directly; it returns a
-[`ResumableUpload`](https://googleapis.dev/nodejs/google-gax/latest/classes/ResumableUpload.html)
-helper. Call `start()` with a `NodeJS.ReadableStream` and await `finished()`
-for the final response:
+[`ResumableUploadSession`](https://googleapis.dev/nodejs/google-gax/latest/classes/ResumableUploadSession.html)
+helper. Call `start()` with a `ResumableSource` (see `getResumableSource()`
+below) and await `finished()` for the final response:
 
 ```ts
 const helper = await client.createResumableUpload(request);
 await helper.start({
-  uploadStream: dataStream,
+  uploadSource: client.getResumableSource(filePath),
   chunkSize: 8 * 1024 * 1024, // 8MB chunks
   onProgress: status => {
     console.log(`Committed ${status.bytesUploaded} bytes to ${status.uploadUrl}`);
@@ -173,12 +173,12 @@ const response = await helper.finished();
 The session URL is available as `helper.uploadUrl` once the upload has
 started. Save it if you need to resume the upload later — for example after a
 process crash or network drop. To resume, pass the saved URL to `start()` on a
-new helper, along with a fresh stream of the same payload:
+new helper, along with a new source for the same payload:
 
 ```ts
 const helper = await client.createResumableUpload();
 await helper.start({
-  uploadStream: dataStream,
+  uploadSource: client.getResumableSource(filePath),
   resumeUrl: savedUploadUrl,
 });
 const response = await helper.finished();
--- a/core/packages/gax/src/fallback.ts
+++ b/core/packages/gax/src/fallback.ts
@@ -55,6 +55,18 @@ export {
   PageDescriptor,
   StreamDescriptor,
 } from './descriptor';
+export {
+  ResumableUploadDescriptor,
+  ResumableUploadSession,
+  ResumableUploadState,
+  resumableUploadStub,
+} from './resumableUpload';
+export type {
+  ResumableUploadContext,
+  ResumableUploadProgress,
+  ResumableUploadStartParams,
+  ResumableSource,
+} from './resumableUpload';
 
 export {StreamType} from './streamingCalls/streaming';
 
--- a/core/packages/gax/src/resumableUpload.ts
+++ b/core/packages/gax/src/resumableUpload.ts
@@ -725,7 +725,10 @@ export class ResumableUploadSession {
 
   private async runTransmission(sessionUrl: string): Promise<void> {
     try {
-      let buffer = Buffer.alloc(0);
+      // Annotated rather than inferred: Buffer.alloc() narrows to
+      // Buffer<ArrayBuffer>, while the remainder returned by readNextChunk()
+      // and skipBytes() is the wider Buffer<ArrayBufferLike>.
+      let buffer: Buffer = Buffer.alloc(0);
       let offset = this.committedBytes_;
       let previousChunk: Buffer | null = null;
       let response: {} | null = null;
--- a/core/packages/gax/src/resumableUpload.ts
+++ b/core/packages/gax/src/resumableUpload.ts
@@ -1,0 +1,1459 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Resumable upload protocol support.
+ *
+ * This module implements the client-side state machine for the resumable
+ * upload protocol used by Google APIs to transfer large payloads over
+ * HTTP(S). The upload session is managed by a {@link ResumableUploadSession}
+ * object, which is returned by GAPIC-generated client methods for resumable
+ * upload RPCs.
+ *
+ * The protocol commands are sent through the `X-Goog-Upload-Command` header
+ * and the payload is transferred in discrete chunks. Transient errors
+ * (Category 1) are retried with exponential backoff, state mismatches
+ * (Category 2) trigger a recovery phase that queries the server for the
+ * committed byte offset, and everything else (Category 3) is fatal.
+ */
+
+import type {AuthClient, GoogleAuth} from 'google-auth-library';
+import * as protobuf from 'protobufjs';
+import * as serializer from 'proto3-json-serializer';
+
+import {APICaller} from './apiCaller';
+import {APICallback, GRPCCall, SimpleCallbackFunction} from './apitypes';
+import {OngoingCall, OngoingCallPromise} from './call';
+import {Descriptor} from './descriptor';
+import {decodeResponse} from './fallbackRest';
+import {CallSettings, createDefaultBackoffSettings, RetryOptions} from './gax';
+import {GoogleError} from './googleError';
+import {Status, rpcCodeFromHttpStatusCode} from './status';
+import {transcode} from './transcoding';
+
+export const DEFAULT_CHUNK_SIZE = 8 * 1024 * 1024;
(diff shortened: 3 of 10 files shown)


core/packages/gax/src/resumableUpload.ts
```diff
@@ -0,0 +1,1459 @@
...
+            ? undefined
+            : typeof body === 'string'
+              ? body
+              : Buffer.from(body),
+        signal: controller.signal,
+        responseType: 'stream',
+        timeout: this.params?.timeout ?? DEFAULT_PER_REQUEST_TIMEOUT_MS,
+        validateStatus: () => true,
+      })) as unknown as Response;
+      const responseBody = Buffer.from(await response.arrayBuffer());
```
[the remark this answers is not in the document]
feywind: Reply from other-Gemini (tl;dr - this isn't a real issue)

In old gaxios v6, GaxiosResponse was a plain wrapper object ({ config, data, status, headers }) without .arrayBuffer().
However, google-gax 6.x uses "google-auth-library": "^11.0.0", which uses gaxios v7.

core/packages/gax/test/unit/resumableUpload.ts
```diff
@@ -0,0 +1,1245 @@
...
+    headers: {
+      get(name: string): string | null {
+        return normalized[name.toLowerCase()] ?? null;
+      },
+    },
+    async arrayBuffer(): Promise<ArrayBuffer> {
+      return Buffer.from(body) as unknown as ArrayBuffer;
+    },
+  };
+}
```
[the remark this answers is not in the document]
feywind: Same response.

core/packages/gax/test/system-test/resumableUpload.ts
```diff
@@ -0,0 +1,286 @@
...
+    signal?: AbortSignal;
+  }) {
+    return fetch(opts.url!, {
+      method: opts.method,
+      headers: opts.headers,
+      body: opts.body,
+      signal: opts.signal,
+    });
+  },
+};
```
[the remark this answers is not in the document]
feywind: Same response.

feywind:
Santi was going to look at this one, I think, but I'll bug you if that changes.

core/packages/gax/src/fallback.ts
```diff
@@ -55,6 +55,18 @@ export {
   PageDescriptor,
   StreamDescriptor,
 } from './descriptor';
+export {
+  ResumableUploadDescriptor,
+  ResumableUploadSession,
+  ResumableUploadState,
+  resumableUploadStub,
+} from './resumableUpload';
```
quirogas, line 63: Please also export resumableSourceFromFile from src/fallback.ts so generated clients work when initialized with fallback: true or fallback: 'rest'.

Generated clients implement getResumableSource(filePath) by calling this._gaxModule.resumableSourceFromFile(filePath). When a client is created with {fallback: true}, this._gaxModule is gax.fallback (src/fallback.ts), where resumableSourceFromFile is currently undefined:

```ts
const client = new ResumableUploadServiceClient({fallback: true});
client.getResumableSource('/tmp/video.mp4');
// Throws TypeError: this._gaxModule.resumableSourceFromFile is not a function
```

To keep browser builds free of top-level fs imports, resumableSourceFromFile in src/resumableSourceFromFile.ts can check isNodeJS() and lazy-load require('fs') (similar to line 454 in src/fallback.ts), and then src/fallback.ts can re-export resumableSourceFromFile.

core/packages/gax/src/resumableUpload.ts
```diff
@@ -0,0 +1,1462 @@
...
+    }
+    if (this.context.minifyJson) {
+      queryString = `${queryString ? `${queryString}&` : ''}$prettyPrint=0`;
+    }
+
+    const uploadPrefix = this.context.uploadPrefix ?? '/resumable/upload';
+    const url = `${this.getEndpointBase()}${uploadPrefix}${
+      queryString ? `?${queryString}` : ''
+    }`;
+    const body = JSON.stringify(json);
```
quirogas, line 644: In sendStart(), transcode(json, rpc.parsedOptions) is called at line 627, but only transcoded?.queryString is used while transcoded.url and transcoded.data are ignored. Also, startHeaders does not include x-goog-upload-header-content-length when the upload size is known.

For RPCs with a google.api.http rule (for example, post: "/v23/customers/{customer_id=*}/youTubeVideoUploads:create" and body: "you_tube_video_upload"), transcode() puts the resolved URL path into transcoded.url and extracts the inner body field into transcoded.data. Ignoring them sends POST /resumable/upload without the service path and serializes the outer request wrapper:

```ts
// Request: { customerId: '123', youTubeVideoUpload: { videoTitle: 'Ad' } }
// Actual URL sent:   https://googleads.googleapis.com:443/resumable/upload
// Expected URL sent: https://googleads.googleapis.com:443/resumable/upload/v23/customers/123/youTubeVideoUploads:create
// Actual body sent:   {"customerId":"123","youTubeVideoUpload":{"videoTitle":"Ad"}}
// Expected body sent: {"videoTitle":"Ad"}
```

Appending transcoded.url to uploadPrefix, serializing transcoded.data (falling back to {} if transcoded.data is ''), and setting x-goog-upload-header-content-length when uploadSizeForDeadline(this.params) > 0 will fix both.

core/packages/gax/src/resumableUpload.ts
```diff
@@ -0,0 +1,1462 @@
...
+          );
+        }
+        throw new Category3Error(
+          `The resumable upload ${command} response did not include a ` +
+            `${UPLOAD_STATUS_HEADER} header.`,
+        );
+      }
+      return;
+    }
+    if (CATEGORY_2_RETRY_CODES.has(response.status)) {
```
quirogas, line 1376: In throwOnNonTransientStatus(), non-2xx responses do not check response.headers.get(UPLOAD_STATUS_HEADER) before checking CATEGORY_2_RETRY_CODES.has(response.status), and Category3Error discards response.body instead of decoding the server error with decodeResponse().

This causes three issues:
1. If the server rejects an upload during 'upload' or 'upload, finalize' with HTTP 400 and x-goog-upload-status: final, line 1376 throws Category2Error, causing transmitChunk() to loop on queryOffset() instead of aborting immediately.
2. If 'start' receives HTTP 400 without x-goog-upload-status, line 1376 throws Category2Error (failing immediately) instead of retrying as TransientError.
3. On fatal errors (like 400 with x-goog-upload-status: final or 403), throwing Category3Error('Resumable upload failed: HTTP ...') drops the server error message, gRPC status code, and statusDetails.

To fix this:
- Check uploadStatus === 'final' first on non-2xx responses and treat it as a fatal error.
- Only throw Category2Error for CATEGORY_2_RETRY_CODES when command is 'upload', 'upload, finalize', or 'finalize' (and throw TransientError on 'start' when uploadStatus is null).
- Decode fatal error bodies with decodeResponse(this.context.rpc, false, response.body, response.status) and rethrow GoogleError in fetchCommand() and sendCommandWithRetry().

core/packages/gax/src/resumableUpload.ts
```diff
@@ -0,0 +1,1462 @@
...
+        throw new Category3Error(
+          `Resumable upload recovery query failed: HTTP ${response.status}.`,
+          response.status,
+        );
+      }
+      const size = parseHeaderInt(
+        response.headers.get(UPLOAD_SIZE_RECEIVED_HEADER),
+      );
+      if (size === null) {
+        throw new Category3Error(
```
quirogas, line 715: In queryOffset(), the client only reads x-goog-upload-size-received and does not check if response.headers.get(UPLOAD_STATUS_HEADER) === 'final'.

When resuming a session that already completed on the server (or querying after the final chunk succeeded), the server replies to 'query' with HTTP 200, x-goog-upload-status: final, and the final JSON response body:
- If x-goog-upload-size-received is absent, queryOffset() throws Category3Error here.
- If x-goog-upload-size-received is present, queryOffset() discards response.body and causes start() or transmitChunk() to send a duplicate finalize request.

Checking if (response.headers.get(UPLOAD_STATUS_HEADER) === 'final') in queryOffset() so the session decodes response.body via decodeFinalResponse(response) and finishes immediately will handle this cleanly. Similarly, in transmitChunk() (line 983) and transmitFinalize() (line 1119), please check that uploadStatus === 'final' before calling decodeFinalResponse(response).

core/packages/gax/src/resumableUpload.ts
```diff
@@ -0,0 +1,1462 @@
...
+    } catch (err) {
+      if (err instanceof RestartUploadError) {
+        // The server reported an offset that is not covered by the local
+        // buffer. Re-open the source at the committed offset and continue.
+        try {
+          this.committedBytes_ = err.offset;
+          this.reportProgress();
+          await this.openSourceAt(err.offset);
+          // Restart the transmission loop from the new offset.
+          await this.runTransmission(sessionUrl);
```
quirogas, line 802: In runTransmission(), openSourceAt(offset) opens an fs.ReadStream at line 735, but releaseActiveStream() is not called in a finally block or in catch (err) when an upload fails.

Because readNextChunk() pulls chunks manually via activeIterator_.next(), Node does not auto-close the paused fs.ReadStream when an HTTP error (like 403) stops the upload mid-file, leaving activeStream_ open (`destroyed === false`).

Also, when RestartUploadError is caught at line 794, line 800 calls await this.openSourceAt(err.offset) and line 802 recursively calls await this.runTransmission(sessionUrl), which immediately calls await this.openSourceAt(offset) a second time at line 735.

Calling this.releaseActiveStream() in a finally block inside runTransmission() and removing the duplicate openSourceAt(err.offset) at line 800 fixes both.

core/packages/gax/src/resumableUpload.ts
```diff
@@ -0,0 +1,1462 @@
...
+        }
+
+        // The server is behind the local state: data from the previous chunk
+        // must be re-transmitted from the in-memory buffer.
+        const bufferedStart = currentPrevious
+          ? currentOffset - currentPrevious.length
+          : currentOffset;
+        if (currentPrevious && serverOffset >= bufferedStart) {
+          const rebuilt = Buffer.concat([currentPrevious, currentChunk]);
+          currentChunk = rebuilt.subarray(serverOffset - bufferedStart);
```
quirogas, line 1081: Two small recovery issues in transmitChunk():

1. Here at lines 1080-1084, when serverOffset rewinds into currentPrevious, Buffer.concat([currentPrevious, currentChunk]).subarray(...) combines both chunks and sends them in a single 'upload' request. For example, with a 256 KiB chunkSize/granularity and serverOffset = 100 KiB, it sends 412 KiB in one request (exceeding chunkSize and not a multiple of 256 KiB). Throwing new RestartUploadError(serverOffset) (or sending at most effectiveChunkSize_ bytes per request) keeps chunks aligned.

2. At lines 1016-1019, when serverOffset === currentOffset, the loop runs continue with 0ms delay and no retry limit. If the server repeatedly returns 409 at the same offset, it spins in a tight loop without backoff until globalDeadlineMs expires.

quirogas (commented):
Thanks for putting this together! Here are a few findings and nit suggestions from testing the state machine and generated client integration.

core/packages/gax/src/resumableUpload.ts
```diff
@@ -0,0 +1,1462 @@
...
+      canceller.callback!(
+        createGoogleError(
+          'The resumable upload transport context was not provided. ' +
+            'This is a bug in the generated client library.',
+        ),
+      );
+      return;
+    }
+    const session = new ResumableUploadSession({
+      ...context,
```
quirogas, line 370: In ResumableUploadApiCaller.call(), settings (of type CallSettings) holds the merged client defaults and per-call CallOptions (such as settings.timeout, settings.retry, and settings.otherArgs?.headers), but lines 359-375 only read settings.resumableUpload.

Passing settings.timeout, settings.retry, and settings.otherArgs?.headers into ResumableUploadSession as defaults (when not overridden in session.start(params)) ensures per-call options and headers are respected.

core/packages/gax/src/resumableUpload.ts
```diff
@@ -0,0 +1,1462 @@
...
+  }
+  return Buffer.from(String(chunk));
+}
+
+function parseHeaderInt(value: string | null): number | null {
+  if (value === null || value === '') {
+    return null;
+  }
+  const parsed = parseInt(value, 10);
+  return Number.isNaN(parsed) ? null : parsed;
```
quirogas, line 313: Nit: parseHeaderInt(value) allows negative numbers (for example, parseInt('-5', 10) returns -5). Returning null when parsed < 0 avoids passing a negative offset to openSourceAt().

core/packages/gax/src/resumableUpload.ts
```diff
@@ -0,0 +1,1462 @@
...
+          err instanceof Category2Error || err instanceof StallError;
+        if (!recoverable) {
+          throw err;
+        }
+        const serverOffset = await this.queryOffset(sessionUrl);
+        this.reportProgress();
+        if (err instanceof StallError) {
+          throw new RestartUploadError(serverOffset);
+        }
+        if (chunk) {
```
quirogas, line 1131: Nit: A few pieces of unreachable or unused code can be removed to simplify the class:
- In transmitFinalize(sessionUrl, finalChunk, offset), all three callers (lines 749, 1025, and 1050) pass null for finalChunk because final chunks with data go through transmitChunk(..., isFinal = true). Removing finalChunk lets you delete this 25-line if (chunk) block (lines 1131-1155).
- private response_ (line 404) is written at line 789 but never read.
- if (response !== null) { break; } (lines 776-778) is unreachable because lines 750 and 774 already break whenever response is set.

core/packages/gax/src/resumableUpload.ts
```diff
@@ -0,0 +1,1462 @@
...
+      );
+      if (size === null) {
+        throw new Category3Error(
+          'The resumable upload recovery query did not include a ' +
+            `${UPLOAD_SIZE_RECEIVED_HEADER} header.`,
+        );
+      }
+      return size;
+    } finally {
+      this.state_ = ResumableUploadState.TRANSMISSION;
```
quirogas, line 722: Nit: In queryOffset(), this.state_ is set to RECOVERY at line 696 and reset to TRANSMISSION here in the finally block. Because reportProgress() is called after queryOffset() returns (at lines 526 and 1009), session.state is already back to TRANSMISSION when onProgress runs, so onProgress callbacks never observe the RECOVERY state.

core/packages/gax/client-libraries.md
```diff
@@ -149,6 +149,48 @@ in the second parameter:
...
+started. Save it if you need to resume the upload later — for example after a
+process crash or network drop. To resume, pass the saved URL to `start()` on a
+new helper, along with a new source for the same payload:
+
+```ts
+const helper = await client.createResumableUpload();
+await helper.start({
+  uploadSource: client.getResumableSource(filePath),
+  resumeUrl: savedUploadUrl,
+});
```
quirogas, line 183: Nit: When resuming with resumeUrl, the start command is skipped so the resumed session does not receive x-goog-upload-chunk-granularity from the server. Showing how to also save helper.chunkSize and pass chunkSize: savedChunkSize alongside resumeUrl: savedUploadUrl will keep resumed uploads aligned to the server granularity.

quirogas (approved):

PR #9288, opened 2026-09-10
test(gax): add showcase harness for resumable uploads

feywind:
Adds an end-to-end harness for the gapic-showcase `ResumableUploadService`, so the resumable upload client path can be exercised against a real server. It uses the resumable upload support from the `google-gax` PR in this series, so merge that one first.

- checked-in generated client for `google.showcase.v1beta1.ResumableUploadService`
- `sample.js`, which uploads a local file through a resumable session
- `run.sh`, which downloads gapic-showcase, compiles the client against the local `google-gax` checkout and runs the sample
- harness README plus a pointer from `test/README.md`, and a `.gitignore` entry for the generated protos

This is test tooling only; nothing in the published package changes.

Merge first: https://github.com/googleapis/google-cloud-node/pull/9287


feywind:
Closing to re-open on main repo as a stack.

feywind:
Never mind on the stack.

core/packages/gax/test/showcase-resumable-upload/sample.ts
```diff
@@ -0,0 +1,1156 @@
...
+
+async function main() {
+  const filePath = process.env.UPLOAD_FILE;
+  const port = Number(process.env.SHOWCASE_PORT || 7469);
+  if (!filePath) {
+    throw new Error('Set UPLOAD_FILE to the path of the file to upload.');
+  }
+  if (!fs.existsSync(filePath)) {
+    throw new Error(`Upload file does not exist: ${filePath}`);
+  }
```
quirogas, line 41: It looks like a rebase or merge conflict in sample.ts left a few syntax and wiring issues that currently break npm run compile, the linter, and the instrumented auth checks at runtime:

1. Lines 33-41 have an unclosed async function main() { stub left over from the earlier JS version (the real async function main(): Promise<void> is at line 1133).
2. const GRANULARITY = 256 * 1024; was dropped, causing 38 TS2304: Cannot find name 'GRANULARITY' errors when compiling core/packages/gax.
3. In createClient() (lines 132-148), the closing } brace is missing, and lines 145-147 pass a fresh new GoogleAuth(...) instead of the auth variable computed on lines 136-140. Passing auth is needed so createInstrumentedAuth() in Tests 3, 4, 5, 6, and 8 records the request log and runs the stallAuth hook.
4. Lines 172 and 192 in runBaselineUpload() need trailing commas to satisfy Prettier.

Replacing lines 33-41 with const GRANULARITY = 256 * 1024; and updating createClient() as follows fixes compilation and makes the instrumented auth log assertions work:

```ts
function createClient(
  port: number,
  customAuth?: GoogleAuth,
): ShowcaseResumableUploadClient {
  const auth =
    customAuth ??
    new GoogleAuth({
      authClient: new googleAuthLibrary.PassThroughClient(),
    });
  return new ResumableUploadServiceClient({
    servicePath: '127.0.0.1',
    port,
    protocol: 'http',
    auth,
  });
}
```

core/packages/gax/test/showcase-resumable-upload/sample.ts
```diff
@@ -0,0 +1,1156 @@
...
+        uploadSource: clientB.getResumableSource(payload.filePath),
+        chunkSize: GRANULARITY,
+        startHeaders: {
+          'X-Goog-Test-Scenario': 'fatal_error_on_start',
+          'X-Goog-Test-Scenario-Config': JSON.stringify({
+            error_code: 403,
+          }),
+        },
+      }),
+      /HTTP 403/,
```
quirogas, line 430: In Test 3 Part B (fatal_error_on_start, lines 430 and 433), assert.rejects(..., /HTTP 403/) fails at runtime against the updated ResumableUploadSession from PR #9287.

Because throwFatalHttpResponse() now decodes fatal HTTP responses through decodeResponse(), the thrown GoogleError has httpStatusCode: 403 and message:
"Error response from RPC UploadMedia with status 403 could not be decoded: the response body is not JSON: Injected fatal start error"

Updating the matcher on lines 430 and 433 from /HTTP 403/ to /(?:HTTP|status) 403/ (or checking err.httpStatusCode === 403) makes Test 3 Part B pass against gapic-showcase 0.44.0.

core/packages/gax/test/showcase-resumable-upload/README.md
```diff
@@ -0,0 +1,58 @@
+# Showcase resumable upload harness
+
+This directory contains a small end-to-end example of the resumable upload
+support added in the `scotty-1` work:
```
quirogas, line 4: On line 4, please replace the internal codename (scotty-1) with generic wording so we keep internal project names out of public repo files:

support added in google-gax:

core/packages/gax/tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "tools/src/prepublish.ts",
     "src/*/*.ts",
     "test/system-test/*.ts",
+    "test/showcase-resumable-upload/*.ts",
```
quirogas, line 21: Right now tsconfig.json compiles test/showcase-resumable-upload/*.ts during npm run compile, but nothing in core/packages/gax/package.json or .github/workflows/ executes test/showcase-resumable-upload/run.sh during CI.

Also note that ci/run_conditional_tests.sh (lines 223-224) currently hits "skipping core package core/packages/gax/ in non-core trigger" during TEST_TYPE=units whenever GIT_DIFF_ARG is set. To make sure the Showcase test runs during gax CI on Linux/macOS, please either wire core/packages/gax/test/showcase-resumable-upload/run.sh into .github/workflows/presubmit.yaml or enable core/packages/* unit test execution in ci/run_conditional_tests.sh.

core/packages/gax/test/showcase-resumable-upload/run.sh
```diff
@@ -0,0 +1,113 @@
...
+    tarball="$DOWNLOAD_DIR/gapic-showcase-$SHOWCASE_VERSION-$os-$arch.tar.gz"
+    echo "Downloading gapic-showcase $SHOWCASE_VERSION from GitHub releases"
+    curl -fsSL \
+      "https://github.com/googleapis/gapic-showcase/releases/download/v$SHOWCASE_VERSION/gapic-showcase-$SHOWCASE_VERSION-$os-$arch.tar.gz" \
+      -o "$tarball"
+    tar -xzf "$tarball" -C "$DOWNLOAD_DIR"
+  fi
+fi
+
+LOG_FILE="$DOWNLOAD_DIR/showcase-$SHOWCASE_PORT.log"
```
quirogas, line 80: Nit: On line 80, LOG_FILE="$DOWNLOAD_DIR/showcase-$SHOWCASE_PORT.log" writes to $DOWNLOAD_DIR, but mkdir -p "$DOWNLOAD_DIR" (line 70) only runs inside if [[ -z "$SHOWCASE_BIN" ]]. If someone sets SHOWCASE_BIN to an existing binary before $DOWNLOAD_DIR exists, line 82 (>"$LOG_FILE") fails with No such file or directory.

Moving mkdir -p "$DOWNLOAD_DIR" right after line 54 avoids that error.

core/packages/gax/test/README.md
```diff
@@ -67,6 +67,19 @@ The following steps will regenerate new Echo and Sequence clients from the lates
 1. Once you have verified that nothing is broken, commit these changes and make a PR from the `regenerate-showcase-client` branch in your fork to the main gax-nodejs repo.
 
 
+## [showcase-resumable-upload](./showcase-resumable-upload/)
+### About
+A small end-to-end harness for the resumable upload API. It downloads the
+[gapic-showcase](https://github.com/googleapis/gapic-showcase) server, compiles
+the checked-in generated client for the showcase `ResumableUploadService`, and runs
+[`sample.js`](./showcase-resumable-upload/sample.js) against it:
```
quirogas, line 75: Nit: Since sample.js was converted to TypeScript (sample.ts), please update the link on line 75 from [sample.js](./showcase-resumable-upload/sample.js) to [sample.ts](./showcase-resumable-upload/sample.ts).

core/packages/gax/test/showcase-resumable-upload/sample.ts
```diff
@@ -0,0 +1,1156 @@
...
+      authClient: new googleAuthLibrary.PassThroughClient(),
+    });
+  return new ResumableUploadServiceClient({
+    servicePath: '127.0.0.1',
+    port,
+    protocol: 'http',
+    auth: new GoogleAuth({
+      authClient: new googleAuthLibrary.PassThroughClient(),
+    }),
+  });
```
quirogas, line 148: Nit: Currently createClient() always initializes ResumableUploadServiceClient with default options (fallback: false). Adding a small check that new ResumableUploadServiceClient({fallback: true, ...}) can also upload a file and that passing sslCreds without fallback: true rejects uploadMedia() with a GoogleError would give end-to-end coverage of both branches in the generated client.

quirogas (approved):
Approving so this is not blocked! Left a few inline comments on sample.ts (fixing a small rebase splice so compile and the showcase run pass) plus a couple of quick nits.

core/packages/gax/tsconfig.json
feywind: I decided not to include it in `.github/workflows`, because it seems a bit too high level, but it's been added to the npm scripts.

Issue #8295, opened 2026-05-18
Update request: @google-cloud/storage is stuck at 7.19.0 and incompatible with google-auth-library@10.x


kackyt:
## Summary
It appears that updates for `@google-cloud/storage` have stalled since version `7.19.0`, and newer changes have not been pulled into this repository. As a result, the package is currently incompatible with `google-auth-library` 10.x. 

## Impact
Because `@google-cloud/storage@7.19.0` relies on an older major version of the auth library, developers are unable to upgrade to `google-auth-library@10.x`. This causes dependency resolution conflicts (e.g., in `npm`, `yarn`, or `pnpm`) and prevents us from utilizing the latest features, bug fixes, and potential security patches provided in the v10 auth library.

## Suggested Fix
Could you please look into updating the `@google-cloud/storage` dependencies to support `google-auth-library@^10.0.0` and cut a new release?

## Context
We are currently developing a GenAI project where we use Google Cloud Storage to store files and send requests. However, due to the `google-auth-library` version incompatibility, we are encountering a header compatibility error when making these requests. This effectively blocks our development workflow, as we cannot communicate properly with the storage service.

thiyaguk09:
> Hi @kackyt, 
>
> Thank you for raising this! We hear you loud and clear.
>
> We have been actively working on a major upgrade under the hood to address this and modernize our stack. **`@google-cloud/storage` v8** will resolve this issue by officially supporting **`google-auth-library@10.x`**.
>
> **What's happening:**
> *   We are transitioning the core transport client to **`gaxios@7`**, which resolves the dependency conflicts you are seeing with `google-auth-library@10.x`.
> *   Due to these updates, **`@google-cloud/storage` v8 will require Node.js 18+** as the minimum supported version. This allows us to leverage standard native global web APIs (like `Headers` and the standard `URL` constructor).
>
> **Current Status:**
> The migration is currently in the final stages of review and performance verification. We are running comprehensive benchmarks to ensure no performance regressions. 
>
> We expect v8 to be released soon. Once published, you will be able to cleanly resolve dependencies with the v10 auth library. We appreciate your patience as we finalize this major release!

danez:
I also see that 7.20.0 was tagged, but not released or published.

markuso:
[...]

PR #9289, opened 2026-09-11
fix(storage): upgrade google-auth-library to v11, gaxios to v7 and align auth client types

### LINKED ISSUE #7341: DeprecationWarning for punycode module

When using the latest version@google-cloud/storage@7.17.1, a runtime warning appears:

```
(node:99526) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

The fundamental problem is that [google-cloud/storage](https://github.com/googleapis/nodejs-storage/blob/main/package.json#L82) latest version is pinned to gaxios 6.x which has a dependency on punycode.

[gaxios 7.x](https://www.npmjs.com/package/gaxios?activeTab=dependencies) does not have a dependency on punycode anymore.

Would it be possible to migrate to gaxios 7.x to eliminate this dependency?

Users have reported this in the gemini-cli repository, which has a dependency on this library: https://github.com/google-gemini/gemini-cli/issues/8405

https://github.com/google-gemini/gemini-cli

Launch any application with a dependency on the latest version of this library in a terminal such as just running gemini-cli --version

A red DeprecationWarning appears on the screen about the punycode module being deprecated

Latest version of googleapis/nodejs-storage dependes on gaxios 6.x. This bug has been fixed in gaxios 7.x.

We expect the latest version to not have any deprecation warnings.

thiyaguk09:
- Upgrade `google-auth-library` to `^11.0.0` and `gaxios` to `^7.1.4` in `@google-cloud/storage`.

Fixes #7341, #7342, #7349, #8295

### DIFF
--- a/handwritten/storage/src/nodejs-common/util.ts
+++ b/handwritten/storage/src/nodejs-common/util.ts
@@ -144,7 +144,7 @@ export interface MakeAuthenticatedRequest {
   getCredentials: (
     callback: (err?: Error | null, credentials?: CredentialBody) => void
   ) => void;
-  authClient: GoogleAuth<AuthClient>;
+  authClient: GoogleAuth;
 }
 
 export interface Abortable {
@@ -643,7 +643,7 @@ export class Util {
       delete googleAutoAuthConfig.projectId;
     }
 
-    let authClient: GoogleAuth<AuthClient>;
+    let authClient: GoogleAuth;
 
     if (googleAutoAuthConfig.authClient instanceof GoogleAuth) {
       // Use an existing `GoogleAuth`
@@ -652,7 +652,8 @@ export class Util {
       // Pass an `AuthClient` & `clientOptions` to `GoogleAuth`, if available
       authClient = new GoogleAuth({
         ...googleAutoAuthConfig,
-        authClient: googleAutoAuthConfig.authClient,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        authClient: googleAutoAuthConfig.authClient as any,
         clientOptions: googleAutoAuthConfig.clientOptions,
       });
     }
--- a/handwritten/storage/src/resumable-upload.ts
+++ b/handwritten/storage/src/resumable-upload.ts
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import AbortController from 'abort-controller';
 import {createHash} from 'crypto';
 import {
   GaxiosOptions,
@@ -99,9 +98,8 @@ export interface UploadConfig extends Pick<WritableOptions, 'highWaterMark'> {
    * emulator context is detected.
    */
   authClient?: {
-    request: <T>(
-      opts: GaxiosOptions
-    ) => Promise<GaxiosResponse<T>> | GaxiosPromise<T>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    request(opts: any): Promise<any>;
   };
 
   /**
@@ -301,9 +299,8 @@ export class Upload extends Writable {
    * emulator context is detected.
    */
   authClient: {
-    request: <T>(
-      opts: GaxiosOptions
-    ) => Promise<GaxiosResponse<T>> | GaxiosPromise<T>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    request(opts: any): Promise<any>;
   };
   cacheKey: string;
   chunkSize?: number;
@@ -1335,9 +1332,9 @@ export class Upload extends Writable {
       }
     }
 
-    const res = await this.authClient.request<{error?: object}>(
+    const res = (await this.authClient.request(
       combinedReqOpts
-    );
+    )) as GaxiosResponse<{error?: object}>;
     if (res.data && res.data.error) {
       throw res.data.error;
     }
@@ -1378,7 +1375,9 @@ export class Upload extends Writable {
       }
     }
 
-    const res = await this.authClient.request(combinedReqOpts);
+    const res = (await this.authClient.request(
+      combinedReqOpts
+    )) as GaxiosResponse;
     const successfulRequest = this.onResponse(res);
     this.removeListener('error', errorCallback);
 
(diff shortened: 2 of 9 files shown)


handwritten/storage/src/transfer-manager.ts
```diff
@@ -220,7 +224,7 @@ class XMLMultiPartUploadHelper implements MultiPartUploadHelper {
     };
   }
 
-  #setGoogApiClientHeaders(headers: Headers = {}): Headers {
+  #setGoogApiClientHeaders(headers = new Headers()): Headers {
```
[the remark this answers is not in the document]
thiyaguk09: Switched from Object.entries(headers) to headers.forEach(...) to properly iterate over the Headers instance.

handwritten/storage/src/resumable-upload.ts
```diff
@@ -1295,10 +1302,14 @@ export class Upload extends Writable {
   private async makeRequest(reqOpts: GaxiosOptions): GaxiosPromise {
     if (this.encryption) {
       reqOpts.headers = reqOpts.headers || {};
-      reqOpts.headers['x-goog-encryption-algorithm'] = 'AES256';
-      reqOpts.headers['x-goog-encryption-key'] = this.encryption.key.toString();
-      reqOpts.headers['x-goog-encryption-key-sha256'] =
-        this.encryption.hash.toString();
+      (reqOpts.headers as Record<string, string>)[
```
kalragauri: In `google-auth-library` v11 (`gaxios` v7), `reqOpts.headers` may be an instance of the global `Headers` class. Setting bracket properties like `(reqOpts.headers as Record<string, string>)['x-goog-encryption-algorithm'] = 'AES256'` on a `Headers` instance does not add the header to the internal `Headers` map. 

This will silently omit all three CSEK encryption headers over the wire.
Please handle `Headers` instances here just as you did in `#addLocalChecksumValidationHeaders` (perhaps using a small helper method?).

handwritten/storage/src/transfer-manager.ts
```diff
@@ -263,21 +270,26 @@ class XMLMultiPartUploadHelper implements MultiPartUploadHelper {
...
           method: 'POST',
           url,
         });
 
-        if (res.data && (res.data as any).error) {
-          throw (res.data as any).error;
+        if ((res?.data as MultiPartUploadErrorResponse)?.error) {
+          throw (res.data as MultiPartUploadErrorResponse).error;
+        }
+        if (typeof res.data === 'string') {
```
kalragauri: This check causes `initiateUpload` to silently skip setting `this.uploadId` whenever `res.data` is a `Buffer`.  When that happens, `initiateUpload()` resolves without error while `this.uploadId` is `undefined`, causing subsequent `uploadPart` requests to hit `?partNumber=1&uploadId=undefined`.

`fast-xml-parser` natively supports both `string` and `Buffer`. Please accept `Buffer.isBuffer(res.data)` as well, and throw an error if `UploadId` cannot be parsed.

handwritten/storage/src/transfer-manager.ts
```diff
@@ -224,35 +227,39 @@ class XMLMultiPartUploadHelper implements MultiPartUploadHelper {
     };
   }
 
-  #setGoogApiClientHeaders(headers: Headers = {}): Headers {
+  #setGoogApiClientHeaders(headers = new Headers()): Headers {
     let headerFound = false;
     let userAgentFound = false;
 
-    for (const [key, value] of Object.entries(headers)) {
+    headers.forEach((value, key) => {
```
kalragauri: Since `headers` is now a Web `Headers` instance, `.get()` and `.has()` are already case-insensitive per the Fetch spec. We can simplify this method and avoid mutating `headers` while iterating over it.

handwritten/storage/src/nodejs-common/util.ts
```diff
@@ -652,7 +652,8 @@ export class Util {
       // Pass an `AuthClient` & `clientOptions` to `GoogleAuth`, if available
       authClient = new GoogleAuth({
         ...googleAutoAuthConfig,
-        authClient: googleAutoAuthConfig.authClient,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        authClient: googleAutoAuthConfig.authClient as any,
```
kalragauri: Avoid suppressing the type checker with `as any`. Instead, preserve explicit generic typing (GoogleAuth<AuthClient>) on authClient to match @google-cloud/common and widen the input config type (AuthClient | GoogleAuth<any>) so all AuthClient subclasses (such as OAuth2Client and Impersonated) type-check cleanly.

handwritten/storage/src/resumable-upload.ts
thiyaguk09: Added `#setHeader` and `#headersToObject` helper methods to safely set and merge headers across Headers instances, tuple arrays, and plain objects, and added unit tests covering CSEK headers with Headers instances.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: Updated `initiateUpload` to accept `Buffer.isBuffer(res.data)` in addition to string, and added an explicit error if `this.uploadId` is missing after parsing. Added unit test coverage for this case.

handwritten/storage/src/transfer-manager.ts
thiyaguk09: Simplified `#setGoogApiClientHeaders` to use `headers.get('x-goog-api-client')` and `headers.has('user-agent')` directly instead of iterating over the `Headers` instance.

handwritten/storage/src/nodejs-common/util.ts
thiyaguk09: Removed the `as any` cast and eslint-disable comment. 

Updated `authClient` to use explicit generic typing `(GoogleAuth<AuthClient>)` across `util.ts` and `service.ts`, allowing all `AuthClient` subclasses to type-check cleanly.

thiyaguk09:
Also upgraded `"gaxios"` from `"^6.0.2"` to `"^7.1.4"` in `package.json` to match the base version used by `google-auth-library` v11, and updated the header types and test mocks (`HeadersInit`, `GaxiosOptionsPrepared`) for `gaxios` v7 compatibility.

Fixes #7341, #7342, #7349, #8295

PR #9310, opened 2026-09-13
perf(spanner): carry MetricsTracer on call context instead of global registry

olavloite:
Pass the MetricsTracer instance directly through gRPC call options rather than registering and retrieving it via a global map keyed by Spanner request ID.

- Attach MetricsTracer to call options in request() and requestStream(), resolving it directly in MetricInterceptor.
- Eliminate the global _currentOperationTracers map and the background setInterval cleanup timer from MetricsTracerFactory.
- Remove per-attempt metadata regex parsing for project ID and request ID in MetricInterceptor.
- Record attempt completion before forwarding status to downstream listeners to prevent race conditions during operation completion.
- Guard synchronous requestFn invocations to guarantee operation completion metrics on early failures.
- Retain getCurrentTracer, clearCurrentTracer, and cleanup interval constants as @deprecated stubs for backwards compatibility.

### DIFF
--- a/handwritten/spanner/src/index.ts
+++ b/handwritten/spanner/src/index.ts
@@ -534,7 +534,7 @@ class Spanner extends GrpcService {
     if (!this.clients_.has(clientName)) {
       this.clients_.set(
         clientName,
-        new v1[clientName](this.options as ClientOptions),
+        new v1.InstanceAdminClient(this.options as ClientOptions),
       );
     }
     return this.clients_.get(clientName)! as v1.InstanceAdminClient;
@@ -558,7 +558,7 @@ class Spanner extends GrpcService {
     if (!this.clients_.has(clientName)) {
       this.clients_.set(
         clientName,
-        new v1[clientName](this.options as ClientOptions),
+        new v1.DatabaseAdminClient(this.options as ClientOptions),
       );
     }
     return this.clients_.get(clientName)! as v1.DatabaseAdminClient;
@@ -615,6 +615,7 @@ class Spanner extends GrpcService {
 
     if (callback) {
       // process.nextTick prevents Unhandled Promise Rejections if callback throws
+      // eslint-disable-next-line promise/catch-or-return
       res.then(
         () => process.nextTick(() => callback(null)),
         err => process.nextTick(() => callback(err)),
@@ -1727,6 +1728,7 @@ class Spanner extends GrpcService {
       const clientName = config.client;
       try {
         if (!this.clients_.has(clientName)) {
+          // eslint-disable-next-line import/namespace
           this.clients_.set(clientName, new v1[clientName](this.options));
         }
       } catch (err) {
@@ -1759,6 +1761,7 @@ class Spanner extends GrpcService {
         });
         this.projectIdReplaced_ = true;
       }
+      config.headers = extend(true, {}, config.headers);
       config.headers[CLOUD_RESOURCE_HEADER] = replaceProjectIdToken(
         config.headers[CLOUD_RESOURCE_HEADER],
         projectId!,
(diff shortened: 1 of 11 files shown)


handwritten/spanner/src/metrics/metrics-tracer-factory.ts
```diff
@@ -101,17 +88,21 @@ export class MetricsTracerFactory {
...
 
     // Create a singleton instance, enabling/disabling metrics can only be done on the initial call
     if (MetricsTracerFactory._instance === null) {
-      MetricsTracerFactory._instance = new MetricsTracerFactory(projectId);
+      MetricsTracerFactory._instance = new MetricsTracerFactory(
+        projectId || '',
+      );
+    } else if (projectId && !MetricsTracerFactory._instance._projectId) {
+      MetricsTracerFactory._instance._projectId = projectId;
     }
```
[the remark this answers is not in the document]
olavloite: Existing issue and out of scope for this fix

handwritten/spanner/src/index.ts
```diff
@@ -2081,30 +2135,76 @@ class Spanner extends GrpcService {
...
+      if (cleanedUp) {
+        return;
+      }
+      cleanedUp = true;
+      if (
+        callStream &&
+        typeof callStream.destroy === 'function' &&
+        !callStream.destroyed
+      ) {
+        callStream.destroy();
```
alkatrivedi: I think we might also need to call `callStream.cancel()` here (before `callStream.destroy()`), as `callStream.destroy()` alone doesn't seem to cancel the underlying gRPC call in `@grpc/grpc-js`.

Because `callStream` is a GAX `StreamProxy`, `.destroy()` only destroys the outer `PassThrough` stream. Hence, we need to make a call to [cancel](https://github.com/googleapis/google-cloud-node/blob/main/core/packages/gax/src/streamingCalls/streaming.ts#L143) to cancel the gRPC call.

handwritten/spanner/src/index.ts
```diff
@@ -2081,30 +2135,76 @@ class Spanner extends GrpcService {
...
+      }
+      metricsTracer?.recordOperationCompletion();
+    };
+
     const stream = streamEvents(through.obj());
+    const origDestroy = stream._destroy;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    stream._destroy = function (err: any, cb: any) {
+      cleanup();
```
alkatrivedi: this could break one test case:

when the `stream` gets destroyed on the client side while the gRPC stream is still in-flight (e.g. breaking out of `for await (const row of db.runStream(...)`) early), `stream._destroy()` calls `cleanup()` synchronously, where `metricsTracer.recordOperationCompletion()` gets called before `MetricInterceptor.onReceiveStatus` is fired.

Since `recordAttemptCompletion()` hasn't run yet, this causes incorrect operation status and attempt metrics:

1. [`recordOperationCompletion()`](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/spanner/src/metrics/metrics-tracer.ts#L362) will record the operation status as `UNKNOWN` and will set the `this.currentOperation` to null.
2. When `MetricInterceptor.onReceiveStatus` fires afterward, [`recordAttemptCompletion()`](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/spanner/src/metrics/metrics-tracer.ts#L329) hits if `(!this.enabled || !this.currentOperation?.currentAttempt) return;` and drops `attempt_count` and `attempt_latencies` (recording 1 `UNKNOWN` operation and 0 `attempts`, instead of 1 `CANCELLED` operation and 1 `CANCELLED` attempt).

handwritten/spanner/src/index.ts
olavloite: Yes, good point. Done. (Note: This was actually an existing bug in the client, and not something introduced in this change.)

handwritten/spanner/src/index.ts
olavloite: Yes, good point. Done. (Note: This was actually an existing bug in the client, and not something introduced in this change.)

alkatrivedi: Got it. Thanks!

handwritten/spanner/src/index.ts
alkatrivedi: Got it. Thanks for fixing this!

alkatrivedi (approved):

PR #9330, opened 2026-09-14
fix(ci): install pnpm 7 via npm and include workflow in paths filter

### LINKED ISSUE #9328: Librarian generate diff check failed on main branch
The librarian generate diff check failed on main branch.

To keep the `main` branch healthy, please consider **reverting the triggering change** first. Once the revert is merged, you can investigate the failure and submit a new PR with the fix.

Please check the logs: https://github.com/googleapis/google-cloud-node/actions/runs/34887863926

zhumin8:
1. Replace pnpm/action-setup with 'npm install -g pnpm@7' in generation_check.yaml. pnpm/action-setup throws 'Multiple versions of pnpm specified' when package.json has 'packageManager: pnpm@10.0.0' and the action step specifies 'version: ^7.0.0'. Installing pnpm 7 directly via npm avoids this action-setup conflict while providing the required pnpm 7 environment for 'librarian install'.

2. Add .github/workflows/generation_check.yaml to dorny/paths-filter so PRs modifying this workflow actually trigger and validate the check on presubmit.

Fix https://github.com/googleapis/google-cloud-node/issues/9328


zhumin8:
Now the `generation-check` run successfully, all the way to generate --all with no diff: https://screenshot-v2.corp.google.com/5mddut7jha9vg

feywind (approved):

.github/workflows/generation_check.yaml
```diff
@@ -42,11 +43,9 @@ jobs:
         uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6.5.0
         with:
           node-version: 22
-      - name: Setup pnpm
+      - name: Setup pnpm 7
         if: steps.changes.outputs.librarian == 'true' || github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && github.ref == 'refs/heads/main')
-        uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
-        with:
-          version: ^7.0.0
+        run: npm install -g pnpm@7
```
danieljbruce, line 48: It's worth a try to do it like this, but the other actions have the `uses` pattern so we may want to figure out how to make it work that way next.

zhumin8: We also have an https://github.com/googleapis/librarian/issues/6889, I suppose librarian wants to support pnpm 8+ soon. 
As far as I recall, the whole reason librarian on 7 is some blocker from google-cloud-node, I suspect now we can work on bringing support to later versions.

PR #9334, opened 2026-09-15
fix(gax): enforce the call deadline on the REST transport

shivanee-p:
Update bug where REST calls didn't enforce deadlines. Since oTel spans were bound to the lifetime of calls, it's discovered that when the fallback service stub is used, it's not verified that the call is completed

Fixes:
- enforce deadline from callOptions.deadline
- record why the request was aborted
- prevent `metadata` from being dropped from wrong type definition
- add `CallOutcome` as a parameter for `_toGoogleError`
- cancel guard wraps the responseStream branch
- cancelling a server stream now reads from cancelSignal


danieljbruce (approved):
Looks good to me. We should really add the gax tests back to the CI pipeline because it will make it a lot easier to catch.

core/packages/gax/src/fallbackServiceStub.ts
```diff
@@ -213,14 +306,30 @@ export function generateServiceStub(
...
+                // inherited 'Error', so the check never matched and cancelled
+                // calls still reported an error. Use the state we recorded.
+                if (timedOut || !cancelRequested) {
                   if (rpc.responseStream) {
                     if (callback) {
-                      callback(err);
+                      callback(callErr);
                     }
-                    streamArrayParser.emit('error', err);
+                    streamArrayParser.emit('error', callErr);
```
quirogas: When `cancelRequested` is `true` here, skipping `callback!(callErr)` for unary calls (`!rpc.responseStream`) causes `OngoingCallPromise` and the active OpenTelemetry span in `traceCall` to hang forever if the caller calls `promise.cancel()` while the response body (`response.arrayBuffer()`) is being read.

In `core/packages/gax/src/call.ts` (`OngoingCall.cancel()`, lines 54 to 66), once `this.cancelFunc` is set, `OngoingCall.cancel()` only calls `this.cancelFunc()` (`canceller.cancel()`) and relies on the transport invoking `callback(err)` to reject `OngoingCallPromise.promise` and run `tracedCallback` (`endSpan()`):

```ts
// core/packages/gax/src/call.ts
cancel(): void {
  if (this.completed) {
    return;
  }
  this.completed = true;
  if (this.cancelFunc) {
    this.cancelFunc();
  } else {
    const error = new GoogleError('cancelled');
    error.code = Status.CANCELLED;
    this.callback!(error);
  }
}
```

For example, here is a test snippet that shows cancelling a traced fallback call after the response headers arrive while the body is still streaming:

```ts
const apiCall = fallback.createApiCall(
  func,
  new gax.CallSettings({
    otherArgs: {
      internalMethodName: 'Echo',
      internalTelemetryInfo: {gcpClientService: 'echo.googleapis.com'},
    },
  }),
  undefined,
  true,
);

const callPromise = apiCall({content: 'hello'});
// Trigger cancel() after response headers arrive (inside response.arrayBuffer())
callPromise.cancel();
await callPromise; // hangs forever, and harness.getSpans('google-gax') stays empty (spanEnded: false)
```

To keep `streamArrayParser` from emitting an unhandled error event on cancelled server streams (`rpc.responseStream`) while still settling unary promises and closing OTel spans, we could only gate the `rpc.responseStream` branch on `if (timedOut || !cancelRequested)` and always invoke `callback!(callErr)` for unary calls (mapping `cancelRequested && !timedOut` to `Status.CANCELLED` to match `call.ts` and PR #9346):

```ts
if (rpc.responseStream) {
  if (timedOut || !cancelRequested) {
    if (callback) {
      callback(callErr);
    }
    streamArrayParser.emit('error', callErr);
  }
} else {
  if (!response204Ok) {
    if (cancelRequested && !timedOut && !(callErr instanceof GoogleError)) {
      const cancelErr = new GoogleError('Cancelled', {cause: callErr});
      cancelErr.code = Status.CANCELLED;
      callback!(cancelErr);
    } else {
      callback!(callErr);
    }
  } else {
    const resp = _formatEmptyResponse(rpc);
    callback!(null, resp);
  }
}
```

core/packages/gax/test/unit/grpc-fallback.ts
```diff
@@ -539,4 +549,407 @@ describe('grpc-fallback', () => {
...
+      // back to it. The guard here used to test `err.name !== 'AbortError'`,
+      // but gaxios wraps node-fetch's AbortError and never sets its own
+      // `name`, leaving the inherited 'Error', so the check never matched and
+      // cancelled calls reported an error anyway.
+      assert.strictEqual(
+        callbackCalled,
+        false,
+        `callback was invoked with ${callbackErr}`,
+      );
+    });
```
quirogas: Related to the comment in `fallbackServiceStub.ts`, for a unary call (`echoStub.echo`), `OngoingCall` needs `callback` to be called with `Status.CANCELLED` when `cancel()` is invoked during body read so `OngoingCallPromise` rejects and `traceCall` ends the span. For server streaming (`echoStub.expand`), `streamArrayParser` should not emit an `'error'` event when cancelled by the caller.

core/packages/gax/src/fallbackServiceStub.ts
```diff
@@ -112,13 +150,38 @@ export function generateServiceStub(
...
+      // flowing, so forwarding the deadline would abort a healthy stream
+      // mid-read. Bounding those is a separate, user-visible change.
+      let timeoutMs: number | undefined;
+      if (callOptions?.deadline && !rpc.responseStream) {
+        // `AbortSignal.timeout` rejects a negative delay with a RangeError, so
+        // an already-expired deadline is clamped. Zero is a fine value here: it
+        // aborts on the next tick, which is the right answer for a deadline
+        // that has already passed.
+        timeoutMs = Math.max(0, callOptions.deadline.getTime() - Date.now());
+      }
```
quirogas, line 272: Nit: If a caller configures `new CallSettings({timeout: Infinity})`, `addTimeoutArg` computes `new Date(now.getTime() + Infinity)`, which creates an `Invalid Date` where `callOptions.deadline.getTime()` is `NaN`. Passing `Math.max(0, NaN)` (`NaN`) to `AbortSignal.timeout(timeoutMs)` on line 223 throws `RangeError [ERR_OUT_OF_RANGE]: The value of "delay" is out of range. It must be an integer. Received NaN`.

We can guard this with `Number.isFinite` before setting `timeoutMs`:

```ts
let timeoutMs: number | undefined;
if (callOptions?.deadline && !rpc.responseStream) {
  const deadlineMs =
    callOptions.deadline instanceof Date
      ? callOptions.deadline.getTime()
      : Number(callOptions.deadline);
  if (Number.isFinite(deadlineMs)) {
    timeoutMs = Math.max(0, Math.trunc(deadlineMs - Date.now()));
  }
}
```

core/packages/gax/src/fallbackServiceStub.ts
shivanee-p: Thank you! Now, the cancel guard only wraps the `responseSteam` branch instead of wrapping both branches.

core/packages/gax/src/fallbackServiceStub.ts
shivanee-p: ACK

shivanee-p: I also added a corresponding unit test.

quirogas (approved):

shivanee-p:
Note: linter errors will be resolved in PR #9281

PR #9344, opened 2026-09-15
feat(gax): report response status codes on traced calls

shivanee-p:
Pass through the status code to spans. `rpc.response.status_code` holds the gRPC status name on both transports, since gax resolves it everywhere. `grpc.response.status_code` mirrors it on gRPC spans, and `http.response.status_code` carries the received HTTP status on fallback spans.

The HTTP status is plumbed from the fetch response through decodeResponse onto `GoogleError.httpStatusCode`, and is absent when no response arrived at all, such as an expired deadline.

Success is inferred rather than observed. The unary path never surfaces a status object to gax, so a call that reported no error is recorded as OK, and 200 on the fallback.

Adds `OtelHarness.assertResponseStatus`, which derives the expected attribute shape from the span's own `gcp.method.type` and asserts both the presence of the attribute that applies and the absence of the one that does not, so a test cannot assert a combination the tracer should never produce.


core/packages/gax/src/fallbackServiceStub.ts
```diff
@@ -426,12 +428,20 @@ export function generateServiceStub(
...
+            const httpStatusCode = response.status;
             return Promise.all([
               Promise.resolve(response.ok),
               response.arrayBuffer(),
             ])
               .then(([ok, buffer]: [boolean, Buffer | ArrayBuffer]) => {
-                const response = responseDecoder(rpc, ok, buffer);
+                const response = responseDecoder(
+                  rpc,
+                  ok,
```
quirogas, line 445: Passing `httpStatusCode` into `responseDecoder` works for most status codes, but `auth.fetch` is configured at line 384 with `validateStatus: (status: number) => status !== 401 && status !== 403`.

Because of that, `401` and `403` responses reject and go straight to `_toGoogleError()` (`fallbackServiceStub.ts:154-161`), which sets `error.code` from `httpStatus` but never sets `error.httpStatusCode`. This leaves `http.response.status_code` as `undefined` on `401` and `403` spans.

In `_toGoogleError()` (`fallbackServiceStub.ts:158-161`), please also set `error.httpStatusCode` to make sure `401` and `403` responses report `http.response.status_code`:

```ts
  if (typeof httpStatus === 'number') {
    error.code = rpcCodeFromHttpStatusCode(httpStatus);
    error.httpStatusCode = httpStatus;
    return error;
  }
```

core/packages/gax/src/fallbackRest.ts
```diff
@@ -99,6 +100,17 @@ export function decodeResponse(
...
+    // missing entirely. Record the received one when the caller knows it.
+    //
+    // Optional because `decodeResponse` is also called from
+    // `streamArrayParser`, which only ever decodes an already-successful body
+    // and so has no status to pass.
+    if (httpStatusCode !== undefined) {
+      error.httpStatusCode = httpStatusCode;
+    }
     throw error;
   }
```
quirogas: `JSON.parse(decodedString)` on line 100 runs before `if (!ok)`. When a REST call receives a non-JSON error body (such as a GFE or proxy HTML `404` or `502` page), `JSON.parse()` throws a `SyntaxError` before `error.httpStatusCode = httpStatusCode` is set. `_toGoogleError()` then catches the `SyntaxError` without an HTTP status, leaving `http.response.status_code` as `undefined` and defaulting `error.code` to `Status.UNAVAILABLE` (`14`).

Guarding `JSON.parse` when `!ok` (and importing `rpcCodeFromHttpStatusCode` from `./status`) will make sure non-JSON error responses keep `httpStatusCode` and the mapped `code`:

```suggestion
  if (!ok) {
    let error: GoogleError;
    try {
      const json = JSON.parse(decodedString);
      error = GoogleError.parseHttpError(json);
    } catch {
      error = new GoogleError(decodedString);
    }
    if (httpStatusCode !== undefined) {
      error.httpStatusCode = httpStatusCode;
      if (error.code === undefined) {
        error.code = rpcCodeFromHttpStatusCode(httpStatusCode);
      }
    }
    throw error;
  }
  const json = JSON.parse(decodedString);
```

quirogas (changes requested):
Good stuff!

Did some testing against the REST fallback path, and found two cases where `http.response.status_code` is dropped:

1. HTTP `401` and `403` responses skip `responseDecoder` and drop `httpStatusCode`, leaving `http.response.status_code` as `undefined` on `401` and `403` spans.
2. Non-JSON HTTP error responses (such as GFE or proxy HTML `404` / `502` pages) throw a `SyntaxError` at `JSON.parse()` in `decodeResponse` before `error.httpStatusCode` is set, dropping `http.response.status_code` and misclassifying a `404` HTML error as `Status.UNAVAILABLE`.

I left comments and suggested fixes on `fallbackServiceStub.ts` and `fallbackRest.ts`.

quirogas (approved):

PR #9347, opened 2026-09-15
fix(gax): update error handling on traced calls

shivanee-p:
- Left span status unset on success (removed explicit OK).
- Removed deprecated error.message span attribute in favor of the span status description.
- Set error.type based on transport protocol
- created recordException() to allow exception names to be passed onto spans


core/packages/gax/src/observability/TracerHelper.ts
```diff
@@ -79,46 +79,48 @@ export function getGaxTracer(): Tracer {
...
+  } else {
+    const httpStatusCode = resolveHttpStatusCode(e);
+    if (httpStatusCode !== undefined) {
+      return String(httpStatusCode);
+    }
   }
+  const code = (e as {code?: unknown}).code;
   if (typeof code === 'string' && code.length > 0) {
     return code;
   }
```
quirogas: On the REST fallback path, _toGoogleError wraps the fetch error in a GoogleError with a numeric e.code and puts the raw error on e.cause. Checking e.cause?.code will make sure system error codes like ECONNREFUSED are preserved on HTTP spans:

```suggestion
  const rawCode = (e as {code?: unknown}).code;
  const cause = (e as {cause?: {code?: unknown}}).cause;
  const code =
    typeof rawCode === 'string'
      ? rawCode
      : cause && typeof cause === 'object' && typeof cause.code === 'string'
        ? cause.code
        : undefined;
  if (code && code.length > 0) {
    return code;
  }
```

quirogas (changes requested):
Did some testing against the REST fallback path, and found one case where system error codes (like ECONNREFUSED) are dropped from error.type:

- REST fallback errors are wrapped in a GoogleError with a numeric e.code and the underlying error stored on e.cause, so resolveErrorType(e, 'http') reports GoogleError instead of the system error code on e.cause.code.

core/packages/gax/src/observability/TracerHelper.ts
shivanee-p: Ty! I've broken out the system error logic into helper functions to make the code easier to read.

quirogas (approved):

PR #9353, opened 2026-09-16
fix: move to pnpm in generator-tests and update-api-lists

bshaffer:
- Remove orphaned `package-lock.json` in `gapic-generator-typescript`
- Switch `generator-tests.yaml` and `update-api-list.yaml` from **npm** to **pnpm** so root only-allow pnpm succeeds.


.github/workflows/generator-tests.yaml
```diff
@@ -41,9 +41,8 @@ jobs:
       with:
         node-version: 24.x
 
-    - name: Install Node dependencies with npm
-      run: npm install
-      working-directory: core/generator/gapic-generator-typescript
+    - name: Install global CLI tools
+      run: npm install -g @bazel/bazelisk pnpm@10.0.0 --silent
```
quirogas, line 45: Nit: could you use pnpm to install the generator.

quirogas: I believe it would look similar to what librarian does to install it. 

https://github.com/googleapis/google-cloud-node/blob/main/librarian.yaml#L28-L30

quirogas (approved):

bshaffer:
Closing in favor of #9413 (generator) and #9412 (`update-api-list.yaml`), which together are a strict superset of this PR.

Two reasons this one can't land as-is:

1. `presubmit` is red. `pnpm install --ignore-workspace --frozen-lockfile` under pnpm 10 cannot read the committed `pnpm-lock.yaml`, which is `lockfileVersion: '6.0'` — pnpm 10 errors with `ERR_PNPM_LOCKFILE_BREAKING_CHANGE`. #9413 regenerates that lockfile at `9.0`, and its `generator-tests` run is green (188 passing).
2. It installs `@bazel/bazelisk`, which #9413 removes along with the rest of the Bazel build.

Thanks for the approval — it carried over as the design for #9413.

PR #9377, opened 2026-09-18
feat(o11y): let callers enable tracing with env var

shivanee-p:
Tracing could only be switched on from code, by passing enableTelemetryTracing in clientOptions, so collecting traces meant editing and redeploying the application.

Read GOOGLE_SDK_NODE_ENABLE_TRACING in checkTelemetryEnabled and let it win over the client option whenever it is set, so tracing can be turned on or off per process. An absent or empty value leaves the client option in charge, and GOOGLE_SDK_NODE_EXPERIMENTAL_O11Y_ENABLED is still required while the feature is experimental.


core/packages/gax/src/util.ts
```diff
@@ -23,16 +23,45 @@ const randomUUID = () =>
...
+  // separates it from an explicit `false`: only an explicit value overrides
+  // the opt-in the caller passed in `clientOptions`.
+  const envOptIn = env.GOOGLE_SDK_NODE_ENABLE_TRACING?.trim();
+  const tracingRequested = envOptIn
+    ? envOptIn.toLowerCase() === 'true'
+    : Boolean(settings?.enableTelemetryTracing);
+
+  return (
+    tracingRequested &&
+    env.GOOGLE_SDK_NODE_EXPERIMENTAL_O11Y_ENABLED === 'true' &&
```
quirogas: Nit: Consider applying the same `.trim().toLowerCase() === 'true'` normalization to `GOOGLE_SDK_NODE_EXPERIMENTAL_O11Y_ENABLED` to make sure both flags behave consistently if a caller sets `GOOGLE_SDK_NODE_EXPERIMENTAL_O11Y_ENABLED=TRUE`:

```suggestion
    env.GOOGLE_SDK_NODE_EXPERIMENTAL_O11Y_ENABLED?.trim().toLowerCase() ===
      'true' &&
```

quirogas (approved):
LGTM

PR #9381, opened 2026-09-18
chore(generator): move gapic-generator-typescript off Bazel into the root pnpm workspace

bshaffer:
Split into reviewable pieces. Land these first, then merge `main` here and this PR shrinks to just the generator change:

- #9410 — drop dead nested `pnpm-lock.yaml` files
- #9411 — run `update-api-list` on pnpm
- #9412 — move librarian to pnpm 10

### What's left here

The Bazel `WORKSPACE` contained `npm_translate_lock(pnpm_lock = "//:pnpm-lock.yaml")`, hard-wiring Bazel to the generator's nested lockfile. So removing Bazel and moving the generator into the root workspace cannot be separated — they're one change.

Hand-written surface is ~150 lines:

| File | What |
|---|---|
| `pnpm-workspace.yaml` | add the generator as a member |
| `typescript/src/generator.ts` | protobufjs v7/v8 cast bridge + rationale |
| `package.json` | drop `@bazel/bazelisk` and a dead npm-style `overrides` |
| `.github/workflows/generator-tests.yaml` | bazelisk → pnpm + protoc |

Everything else is generated-file deletion (`WORKSPACE`, `BUILD.bazel`, `MODULE.bazel`, `.bazelrc`, `package-lock.json`, the nested `pnpm-lock.yaml`) plus the root lockfile update. **Reviewers: filter out `*lock*` and this is 4 files.**

### On the protobufjs casts

`proto3-json-serializer@4` (latest) hard-depends on `protobufjs@^7`; the generator wants `^8`. Hoisting into the root workspace loses the nested `pnpm.overrides` that used to unify them, so TypeScript now sees two protobufjs type trees.

Fixing this with a root `pnpm.overrides` entry **does not work** — it was tried in 2cc19a6 and broke all 18 `units` shards, because it forces protobufjs@8 onto google-gax, which needs v7. The casts are the correct fix until upstream supports v8. Documented inline so nobody retries the override.

### Known follow-ups (not addressed here)

1. The Bazel removal is incomplete: `.bazeliskrc`, `repositories.bzl` (loaded by the deleted `WORKSPACE`), and `rules_typescript_gapic/` survive unreferenced.
2. Four steps in `generator-tests.yaml` (lines ~159-180) are gated on `steps.generator-changes.outputs.changes`, but no step defines `id: generator-changes` — they have never run. They also use `generator/...` paths missing the `core/` prefix. These are the combined-library Speech/Tasks tests, and the only consumer of `rules_typescript_gapic/` from (1).

Needs a call on whether those tests should be revived or deleted.


librarian.yaml
```diff
@@ -25,7 +25,7 @@ tools:
       package: https://github.com/googleapis/google-cloud-node/archive/gapic-generator-v5.2.0.tar.gz
       checksum: a2afaa498c7938837e65a1eecb2041bda5b0bda232e6bdbd5117e012cdc7e75b
       build:
-        - pnpm install --frozen-lockfile --config.auto-install-peers=true
+        - pnpm install --no-frozen-lockfile --force --config.auto-install-peers=true
```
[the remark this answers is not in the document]
bshaffer: Removed `--force`.

**Note**: `--ignore-workspace --no-frozen-lockfile` is temporarily required here because `librarian install` downloads the pre-consolidation `gapic-generator-v5.3.0` tarball (where `gapic-generator-typescript` was not yet in `pnpm-workspace.yaml` or the root `pnpm-lock.yaml`). Once this PR merges and the next `gapic-generator` tag is released, we will update this to `pnpm install --frozen-lockfile --ignore-scripts` in follow-up PR https://github.com/googleapis/google-cloud-node/pull/9407.

pnpm-lock.yaml
```diff
@@ -298,7 +298,7 @@ importers:
         version: 7.0.3
       gapic-tools:
         specifier: ^2.0.0
-        version: 2.0.1(protobufjs@7.6.5)
+        version: 2.0.1(protobufjs@8.8.0)
```
dashpole: Since `4fadb437` removed `core/generator/gapic-generator-typescript` from `pnpm-workspace.yaml` and kept its standalone lockfile, no workspace `package.json` files changed relative to `main`, but this root `pnpm-lock.yaml` still has leftover diff shifting `gapic-tools` and `protobufjs-cli` from `protobufjs@7.6.5` to `protobufjs@8.8.0` across 249 workspace packages. Consider restoring `pnpm-lock.yaml` from `main` (unless you intend to re-apply `40c9c303` to keep the generator in `pnpm-workspace.yaml`).

.github/workflows/generator-tests.yaml
```diff
@@ -29,50 +29,51 @@ jobs:
...
       run: |
         mkdir -p ~/artifacts
-        cp bazel-testlogs/unit_tests/test.outputs/outputs.zip ~/artifacts/
-        bazelisk run -- @pnpm//:pnpm --dir $PWD install --ignore-workspace
-        tar cfz ~/artifacts/node_modules.tar.gz node_modules
+        zip -r ~/artifacts/outputs.zip .test-out-*
+        # -h dereferences symlinks. In a pnpm workspace this directory is a
+        # symlink farm pointing into the root node_modules/.pnpm store, so
+        # without -h the uploaded artifact is a pile of dangling links.
+        tar czhf ~/artifacts/node_modules.tar.gz node_modules
```
dashpole: With `pnpm install --ignore-workspace` on line 50, `node_modules/.pnpm` is self-contained inside `core/generator/gapic-generator-typescript/node_modules` rather than pointing to the root store, so `tar -h` duplicates every package inside `node_modules/.pnpm` alongside the top-level symlinks:
```suggestion
        tar czf ~/artifacts/node_modules.tar.gz node_modules
```

bshaffer:
Closing. This PR has been split into two reviewable pieces, and its one remaining idea was dropped on purpose.

| | |
|---|---|
| #9413 | generator: remove Bazel, build with pnpm (**generator stays standalone**) |
| #9412 | CI: finish the move to pnpm 10 (`generation_check.yaml`, `update-api-list.yaml`, `librarian.yaml`, dead lockfiles) |

**Dropped: moving the generator into the root pnpm workspace.**

1. Librarian builds the *released* generator from a pinned tarball with `--ignore-workspace` (`src_dir` is a path inside the fetched archive). Workspace membership buys the production path nothing.
2. Workspace membership is the sole cause of the `--no-frozen-lockfile` reproducibility regression. Standalone + a `9.0` lockfile means `--frozen-lockfile` keeps working.
3. It costs type safety. In the root workspace, `proto3-json-serializer@4` pins `protobufjs@^7` while the generator needs `^8`, and there is no per-project override in pnpm 10 — so `generator.ts` needs casts. A **root** `pnpm.overrides.protobufjs` is not an option: 2cc19a6 tried it and broke all 18 `units` shards (it forces protobufjs@8 onto google-gax, which needs v7); reverted in 05c2a62.
4. It silently changes CI to test the generator against local sources instead of the published packages librarian actually installs.

PR #9393, opened 2026-09-21
feat(tools): add runtime-agnostic test runner and Bun mocha hooks

quirogas:
Currently, packages run tests using `c8 mocha build/test`. When invoked via `pnpm test` (even with `TEST_CMD="bun --bun run"`), two problems prevent tests from actually running on the Bun runtime:
- `c8` relies on Node-only V8 coverage internals and has a `#!/usr/bin/env node` shebang, which causes the OS to silently respawn tests under Node.js instead of Bun.
- Mocha's `.mocharc.cjs` enables `parallel: true` (which relies on Node `worker_threads` serialization that crashes in Bun) and spawns child Mocha processes when flags like `--enable-source-maps` and `--throw-deprecation` are present. Additionally, test-only mocking tools like `proxyquire` and `nock` behave differently under Bun.

This PR introduces a shared test runner (`bin/run-test.cjs`) and a Bun test compatibility preload (`bin/proxyquire-bun-shim.cjs`) so packages can execute their unit tests on either runtime without changing how Node.js tests work. When tests run on Node.js, `bin/run-test.cjs` still delegates directly to `c8 mocha` with full V8 coverage and worker-thread parallelism. When tests run on Bun, it skips `c8`, runs Mocha in single-process mode inside the active Bun runtime, and preloads the Bun test shim so existing `proxyquire` and `nock` unit tests work out of the box.

BEGIN_COMMIT_OVERRIDE
chore(tools): add runtime-agnostic test runner and Bun mocha hooks
END_COMMIT_OVERRIDE


danieljbruce (approved):
Looks fine to me. Do we know the reason proxyquire isn't co-operating with bun?

bin/run-test.cjs
```diff
@@ -0,0 +1,107 @@
...
+//
+// - Under Node.js (`typeof Bun === 'undefined'` and `JS_RUNTIME !== 'bun'`):
+//   Executes `c8 mocha <args>` (unless `--no-c8` is passed), preserving V8 code
+//   coverage and Mocha worker-thread parallelism (`parallel: true`).
+// - Under Bun (`typeof Bun !== 'undefined'` or `JS_RUNTIME === 'bun'`):
+//   Skips `c8` (which relies on Node's `NODE_V8_COVERAGE`), disables Mocha
+//   worker-thread parallelism (`--no-parallel`), preloads the Bun `proxyquire`
+//   shim (`bin/proxyquire-bun-shim.cjs`), and loads `mocha/bin/mocha.js`
+//   directly in-process so `#!/usr/bin/env node` shebangs never switch back to
+//   Node.js.
```
danieljbruce, line 26: Just one more question I had. Still LGTM, but does this file run the tests for exactly one library when it is invoked? I just want to be sure the target is correct when `bun run test` is executed. Maybe a comment to show where the scope is limited would be good.

danieljbruce:
Also, one more comment about c8. It looks like to me that c8 is only used for collecting metrics so I think we might be able to remove it altogether if that makes the code cleaner.

danieljbruce:
FYI: The test runner right now could make it difficult to test in the future with .only. See b/565023469.

PR #9394, opened 2026-09-21
feat(generator): use runtime-agnostic test runner in GAPIC package templates

quirogas:
Currently, the GAPIC TypeScript generator templates hardcode `c8 mocha build/test` as the `"test"` script in every generated `package.json`. Whenever libraries are generated or regenerated, their test scripts default back to `c8`, which forces `#!/usr/bin/env node` and prevents tests from running on the Bun runtime.

This PR updates the CommonJS and ESM GAPIC `package.json` templates in `core/generator/gapic-generator-typescript` (along with the generator test baselines) to use `node ../../bin/run-test.cjs build/test` instead. When developers or CI run `pnpm test` on Node.js, `bin/run-test.cjs` continues to run `c8 mocha` with coverage as usual. When invoked with `bun --bun run test`, Bun intercepts the `node` call and executes `bin/run-test.cjs` directly inside the Bun runtime.


danieljbruce (approved):
This looks fine for the sake of getting the bun testing changes in, but we really should try to understand the full impact of removing c8 and we should try to understand the reason it was there in the first place.

quirogas:
> This looks fine for the sake of getting the bun testing changes in, but we really should try to understand the full impact of removing c8 and we should try to understand the reason it was there in the first place.

We use c8 for test coverage and should still aim to have test coverage; however, keeping it within our test execution causes issues when running on different runtimes. The solution is to remove C8 from individual per-package execution and implement a dedicated CI check for test coverage, allowing for a more runtime-agnostic test configuration.

PR #9395, opened 2026-09-21
chore(packages): migrate generated packages to runtime-agnostic test runner

quirogas:
Even with the GAPIC generator templates updated in the previous PR, existing packages in the monorepo still have `c8 mocha` in their `package.json` files until the next generator release rolls out. Because `c8` uses a `#!/usr/bin/env node` shebang, running tests in `packages/*` and `core/packages/*` today still spawns Node.js instead of Bun.

This PR updates the `"test"` script across `packages/*/package.json` and `core/packages/*/package.json` to call `bin/run-test.cjs` directly so we can test and validate all generated and core libraries on Bun right away without waiting for a generator release. Node.js test execution and `c8` coverage remain unchanged.


core/packages/gaxios/package.json
```diff
@@ -26,7 +26,7 @@
   },
   "scripts": {
     "lint": "gts check --no-inline-config",
-    "test": "c8 mocha build/esm/test",
+    "test": "node ../../../bin/run-test.cjs build/esm/test",
```
danieljbruce, line 29: For the sake of getting bun support in I can approve this, but we really should try to understand the impact of c8 and why it was there in the first place.

danieljbruce (approved):

core/packages/gaxios/package.json
danieljbruce: Oops. I clicked the wrong one. I guess this is not ready yet.

danieljbruce: Actually, I think the answer is that c8 is just used to report analytics so I think it is fine to remove.

PR #9401, opened 2026-09-21
feat(gax): add resend_count for o11y tracing

shivanee-p:
Adds a resend_count span attribute reporting how many times a request was resent after a retryable failure. The initial send is recorded with a value of 0, and then 1 on the first retry. It is written on every traced span, including the 0 case.

- traceCall hands the traced operation a ResendRecorder and counts the invocations
- passing unit test suite with both HTTP and gRPC calls
- Legacy retry-request streaming path when gaxStreamingRetries is off is not recorded. Retries happen inside the third-party module with no hook to observe them.


core/packages/gax/src/normalCalls/retries.ts
```diff
@@ -170,6 +176,9 @@ export function retryable(
             const rpcTimeout = maxTimeout ? maxTimeout : 0;
             const newDeadline = deadline ? deadline - now.getTime() : Infinity;
             timeout = Math.min(timeoutCal, rpcTimeout, newDeadline);
+            // Every repeat scheduled here is a resend of the request, so the
+            // tracer is told about each one as it happens.
+            recordResend?.();
             repeat(lastError);
```
quirogas: Calling recordResend?.() before repeat(lastError) causes resend_count to be off by +1 when repeat() immediately exits at the deadline or maxRetries check without sending another request. Guarding recordResend?.() here (or moving it inside repeat() before retries++) will make sure it only increments when a retry is actually sent:

```suggestion
            if (
              (!deadline || now.getTime() < deadline) &&
              (!maxRetries || retries < maxRetries)
            ) {
              recordResend?.();
            }
            repeat(lastError);
```

core/packages/gax/src/observability/TracerHelper.ts
```diff
@@ -414,6 +460,7 @@ export function traceCall(
           span.setStatus({code: SpanStatusCode.OK});
         }
         setStatusAttributes();
+        span.setAttribute(resendCountAttribute, resendCount);
```
quirogas: Per the OpenTelemetry spec and the Golden Signals test plan, the resend count attribute should be omitted when no retries occurred and start at 1 on the first retry:

```suggestion
        if (resendCount > 0) {
          span.setAttribute(resendCountAttribute, resendCount);
        }
```

core/packages/gax/src/streamingCalls/streaming.ts
```diff
@@ -547,6 +554,11 @@ export class StreamProxy extends duplexify implements GRPCCallResult {
                   }
 
                   retries++;
+                  // Reported separately from `retries`, which is the retry
+                  // budget and is reset to 0 every time data arrives. The
+                  // tracer wants the total for the call, so it counts the
+                  // calls it receives instead of reading that counter.
+                  this.recordResend?.();
                   let retryArgument = this.argument! as RequestType;
```
quirogas: Two small things on streaming retries:

1. Moving `this.recordResend?.()` inside `newRequest` (after `if (this._isCancelCalled) return;`) will make sure a retry cancelled during the backoff sleep is not counted as a resend.
2. In `setStream` (lines 333-357), the default `retryRequest` branch never calls `this.recordResend?.()`, so standard server streams don't record resends when `gaxStreamingRetries` is unset. Calling `this.recordResend?.()` on the 2nd+ `request()` call in that branch will make sure both paths record resends.

```suggestion
                  let retryArgument = this.argument! as RequestType;
```

quirogas (changes requested):
Did some testing across unary and streaming retries, and found three issues with resend_count:

- Failed unary retries that hit maxRetries or totalTimeoutMillis report resend_count off by +1 because recordResend() runs before repeat() checks whether the retry limit was reached.
- Setting resend_count to 0 on unretried calls conflicts with the OpenTelemetry spec and the Golden Signals test plan, which expect the attribute to be omitted when no retries occurred.
- Default server-streaming retries (using retry-request) never call recordResend(), and cancelled gaxServerStreamingRetries calls increment recordResend() before checking _isCancelCalled.

quirogas (approved):

PR #9407, opened 2026-09-22
chore(librarian): use frozen lockfile and workspace for generator build

bshaffer:
Use `pnpm install --frozen-lockfile --ignore-scripts` in `librarian.yaml` once the next `gapic-generator` tag is released after #9381 and the `package` and `checksum` in `librarian.yaml` are updated to match


bshaffer:
Keeping this open as the tracker for restoring a reproducible generator build, but the target line has changed.

The generator is staying **standalone**, not joining the root pnpm workspace (rationale in #9381). So `--ignore-workspace` needs to stay; the only thing blocking `--frozen-lockfile` is the pinned `gapic-generator-v5.3.0` tarball, whose nested lockfile is `lockfileVersion: '6.0'` and unreadable by pnpm 10.

Sequence to close this out:

1. Merge #9413 — regenerates the nested lockfile at `9.0`
2. Cut a new `gapic-generator-*` tag and repin `package:` / `checksum:` here
3. Flip to `pnpm install --ignore-workspace --frozen-lockfile --config.auto-install-peers=true`

Step 3 before step 2 will fail with `ERR_PNPM_LOCKFILE_BREAKING_CHANGE`.

Also note `--ignore-scripts` isn't safe here as written — `pnpm run compile` on the next line needs the install to have completed normally.

PR #9410, opened 2026-09-22
chore: drop dead nested pnpm-lock.yaml files

bshaffer:
Split out of #9381.

`core/packages/gapic-node-processing` and `core/packages/tools` are already root workspace members via the `core/packages/*` glob in `pnpm-workspace.yaml`, so pnpm has never read these nested lockfiles. Pure deletion of generated content, no behavior change.


bshaffer:
Superseded by #9412, which now carries these two deletions.

PR #9411, opened 2026-09-22
chore(ci): run update-api-list on pnpm

bshaffer:
Split out of #9381.

The repo is pnpm-only (the root `preinstall` runs `only-allow pnpm`), but this nightly job still ran `npm install`. Scoped with `--filter google-cloud-node` because it only needs the root package to run `bin/generate-readme.mjs`.


bshaffer:
Superseded by #9412, which now carries the `update-api-list.yaml` change.

PR #9416, opened 2026-09-22
test(gax): add showcase harness for resumable uploads

feywind:
Re-creating this so I can stack it.

Adds an end-to-end harness for the gapic-showcase ResumableUploadService, so the resumable upload client path can be exercised against a real server. It uses the resumable upload support from the google-gax PR in this series, so merge that one first.

checked-in generated client for google.showcase.v1beta1.ResumableUploadService
sample.js, which uploads a local file through a resumable session
run.sh, which downloads gapic-showcase, compiles the client against the local google-gax checkout and runs the sample
harness README plus a pointer from test/README.md, and a .gitignore entry for the generated protos
This is test tooling only; nothing in the published package changes.

Merge first: https://github.com/googleapis/google-cloud-node/pull/9287


feywind:
Switching back to the previous PR. Sorry.

PR #9427, opened 2026-09-23
feat: update API sources and regenerate

sofisl:
Update API sources and regenerate.

### System Details
- Node.js version: v22.23.1
- npm version: 10.9.8
- OS: Linux 7.1.6-1rodete1-amd64 x86_64
- Librarian version: v0.45.0

### Commands Executed
- `go run github.com/googleapis/librarian/cmd/librarian@v0.45.0 update sources.googleapis`
rm -rf /usr/local/google/home/sofialeon/.cache/librarian/bin/nodejs_tools
- `go run github.com/googleapis/librarian/cmd/librarian@v0.45.0 install`
- `go run github.com/googleapis/librarian/cmd/librarian@v0.45.0 generate --all`
- `git add .`
- `git commit -m "feat: update API sources and regenerate"`
### Included googleapis CLs

**google/bigtable/admin/v2**
- http://cl/985340194

**google/chat/v1**
- http://cl/985471432
- http://cl/984009633

**google/cloud/agentregistry/v1**
- http://cl/984147399

**google/cloud/geminidataanalytics/v1**
- http://cl/985350099

**google/cloud/geminidataanalytics/v1alpha**
- http://cl/985426965

**google/cloud/geminidataanalytics/v1beta**
- http://cl/985427080

**google/cloud/lustre/v1**
- http://cl/985572102

**google/cloud/sql/v1**
- http://cl/985143287

**google/cloud/sql/v1beta4**
- http://cl/985077866

**google/cloud/tasks/v2**
- http://cl/984260262

**google/cloud/vectorsearch/v1**
- http://cl/984037515

**google/cloud/vectorsearch/v1beta**
- http://cl/984035901

**google/marketingplatform/admin/v1alpha**
- http://cl/983821889

**google/monitoring/dashboard/v1**
- http://cl/986124202

PR #8266, opened 2026-05-13
chore: clean up per-package eslintrc configurations

shivanee-p:
Remove the per-package generation logic for the .eslintignore and .eslintrc.json files within the GAPIC libraries

This PR cleans up legacy per-package linter configurations across the generated GAPIC libraries to prepare for linter implementation under ESLint v10. The individual per-package configurations are being consolidated into a unified, top-level flat configuration architecture to reduce merge conflicts and maintenance overhead.

Fixes [#4585](https://github.com/googleapis/librarian/issues/4585) 🦕

BEGIN_COMMIT_OVERRIDE
feat: clean up per-package eslintrc configurations
END_COMMIT_OVERRIDE

PR #9429, opened 2026-09-23
chore(lint): exempt generated GAPIC files from promise/prefer-const rules

bshaffer:
Generated GAPIC clients (`packages/*/src/v*/`) and their `gapic_*.ts` tests lost their per-package `root: true` eslintrc in #8266, so the root config's `plugin:promise/recommended` + gts rules now fail every regeneration PR (e.g. #9427, 59 errors).

Adds an override turning off `promise/always-return`, `promise/catch-or-return`, `@typescript-eslint/no-floating-promises`, and `prefer-const` for those paths — mirroring the existing `handwritten/firestore` GAPIC override.


quirogas (approved):
Thank you!!

